PR #8, opened 2021-01-14
Basic DB setup

ttung:
### Description
It is an empty schema with the most rudimentary schema migration support.

Depends on #7 

#### Issue

### Test plan

1. `make init-local-db` and `make drop-local-db` both work.
2. Made some dummy changes to the model code and ran `ENV=local alembic revision --autogenerate -m "SOME DESCRIPTIVE MESSAGE" --rev-id $(date +%Y%M%d_%H%M%S)` to generate a migration.  Hand checked it to see that it made sense.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,62 @@
 SHELL := /bin/bash
 PYTHON_CODE_DIRECTORIES = src/py
 
+
+##########
+# DATABASE
+##########
+
+LOCAL_DB_CONTAINER_NAME = covidr-local
+LOCAL_DB_CONTAINER_ID = $(shell docker ps -a | grep $(LOCAL_DB_CONTAINER_NAME) | awk '{print $$1}')
+LOCAL_DB_CONTAINER_RUNNING_ID = $(shell docker ps | grep $(LOCAL_DB_CONTAINER_NAME) | awk '{print $$1}')
+LOCAL_DB_NAME = covidr_db
+LOCAL_DB_ADMIN_USERNAME = postgres  # This has to be "postgres" to ease moving snapshots from RDS.
+LOCAL_DB_ADMIN_PASSWORD = admin
+LOCAL_DB_RW_USERNAME = user_rw
+LOCAL_DB_RW_PASSWORD = password_rw
+LOCAL_DB_RO_USERNAME = user_ro
+LOCAL_DB_RO_PASSWORD = password_ro
+DOCKER_IMAGE = czbiohub/covidhub-postgres:11.5-alpine
+
+start-local-db:
+	@if [ "$(LOCAL_DB_CONTAINER_ID)" == "" ]; then \
+		docker create --name $(LOCAL_DB_CONTAINER_NAME) -p 5432:5432 \
+		-e POSTGRES_USER=$(LOCAL_DB_ADMIN_USERNAME) \
+		-e POSTGRES_PASSWORD=$(LOCAL_DB_ADMIN_PASSWORD) \
+		-e POSTGRES_DB=$(LOCAL_DB_NAME) \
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -1,6 +1,62 @@
...
+	ENV=local alembic stamp head
+
+stop-local-db:
+	@if [ "$(LOCAL_DB_CONTAINER_RUNNING_ID)" != "" ]; then \
+		docker stop $(LOCAL_DB_CONTAINER_NAME); \
+	fi
+
+drop-local-db:
+	@$(MAKE) stop-local-db
+	@( read -p "Delete local database container? [y/N]: " sure && case "$$sure" in [yY]) true;; *) false;; esac )
```
phoenixAja, line 51: it's throwing me off that this @ sign is red, do you know why it's doing that?

phoenixAja: i looked at the same line in the covidhub Makefile and it looks the same so i think this is correct, i was jw

src/py/covidr/cli/db.py
```diff
@@ -0,0 +1,19 @@
+import click
+from covidr.database.connection import get_db_uri, init_db
+from covidr.database.schema import create_tables_and_schema
+from covidr.sandbox.runtimeenvironment import RuntimeEnvironment
+
+from .toplevel import cli
+
+
+@cli.group()
```
phoenixAja, line 9: interesting, going with click instead of argparse?

src/py/covidr/database/modelmixins.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+# for things that have names
+class NameMixin:
+    """Mixin that adds a required unique name string field"""
+
+    name = Column(String, unique=True, nullable=False)
+
+    def __repr__(self):
+        return self.name
```
phoenixAja, line 20: i'm confused when we would need this over the `BaseMixin`? Like the primary_key will always be unique so why would we need this?

phoenixAja: oh nevermind i get it the comment was throwing me off

src/py/covidr/database/models.py
```diff
@@ -0,0 +1,17 @@
+# the main models in the database
+from sqlalchemy import MetaData
+from sqlalchemy.ext.declarative import declarative_base
+
+from . import modelmixins as mx
+
+meta = MetaData(
+    schema="covidr",
```
phoenixAja, line 8: do we need to create this schema with a migration?

src/py/covidr/database/schema.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def create_tables_and_schema(interface: SqlAlchemyInterface):
+    """Creates the schema and all tables"""
+    with session_scope(interface) as session:
+        sqlalchemy_event.listen(
+            meta,
+            "before_create",
+            DDL(f"CREATE SCHEMA IF NOT EXISTS {meta.schema}"),
+        )
```
phoenixAja, line 15: oh ok interesting, why not just create the schema in the initial migration?

phoenixAja (approved):
looks good, just had a couple questions

Makefile
ttung: the `@` silences the echoing of the command when it's run.  it's mostly for aesthetic reasons.  Not sure why it's highlighted red.

src/py/covidr/cli/db.py
ttung: I go back and forth. :)

src/py/covidr/database/schema.py
ttung: The reason is that I want `make init-local-db` actually produce a functional db.

PR #9, opened 2021-01-14
Merge RuntimeEnvironment and Config

ttung:
### Description
- Switch up config to use properties.  This will allow implementations to override more easily.
- Add a mechanism to track all the subclasses of config and mandate a descriptive name for each config.  That descriptive name is used to map from environment variable to a runtime config (done in `Config.by_descriptive_name`, used in `database_migrations/env.py`).

#### Issue

### Test plan
make drop-local-db && make init-local-db


ttung:
I changed the db to be "dev" instead of "local" even though "local" is more accurate because it merges better with `Config`.

lvreynoso (approved):

phoenixAja (approved):
coolio

PR #11, opened 2021-01-15
Top level user/group objects

ttung:
### Description
These are the basic objects we need to log someone in.

Also enabled the plugin for mypy+sqlalchemy to be nice to each other.

#### Issue
[ch65314](https://app.clubhouse.io/genepi/story/65314/)

### Test plan
created the tables.


phoenixAja (approved):
awesome!

PR #12, opened 2021-01-15
Improve the management of requirements files.

ttung:
### Description
1. Separate the requirements needed to run the application from the requirements to develop the requirements.
2. Separate the documentation for the backend.
3. Add makefile rules to generate pinned requirements.
4. Snuck in a fix for phony makefile targets.

### Test plan
1. `make pin-all-requirements`
2. deployed the app.


lvreynoso (approved):
LGTM. Looks like I'll have to switch from `python3 -m venv $(DIR)` to using `virtualenv`.

ttung:
This doesn't use virtualenv except for travis.  Actually, that's old, and I can remove it.

PR #13, opened 2021-01-15
Configure Elasticbeanstalk deployment and build

lvreynoso:
### Description

Adds Elasticbeanstalk-specific configuration files to properly build the frontend and deploy the application.

#### Issue
[ch65516](https://app.clubhouse.io/genepi/stories/space/65516)

### Test plan

Smoke test at http://covidr-staging.us-west-2.elasticbeanstalk.com/


requirements.txt
```diff
@@ -0,0 +1 @@
+./src/py
```
ttung, line 1: TIL this.

src/py/covidr/app.py
```diff
@@ -1,10 +1,12 @@
 import os
 
 from covidr.config import DevelopmentConfig
-from flask import Flask
+from flask import Flask, send_from_directory
+
+dir_path = os.path.dirname(os.path.realpath(__file__))
```
ttung: Would prefer using pathlib over os.path. Also, is the realpath part necessary?

src/py/covidr/app.py
```diff
@@ -13,6 +15,10 @@
...
-@application.route("/")
-def root():
-    return "Hello world"
+@application.route("/", defaults={"path": ""})
+@application.route("/<path:path>")
+def serve(path):
+    if path != "" and os.path.exists(application.static_folder + "/" + path):
+        return send_from_directory(application.static_folder, path)
+    else:
+        return send_from_directory(application.static_folder, "index.html")
```
ttung: ditto re pathlib.

ttung (approved):
looks awesome!  See comments pls.

.gitignore
```diff
@@ -299,3 +299,8 @@ sketch
 static
 
 # End of https://www.toptal.com/developers/gitignore/api/flask,react,macos,node
+
+# Elastic Beanstalk Files
```
phoenixAja, line 303: we want these config files to be ignored? I thought it would be helpful to version these.

.gitignore
```diff
@@ -299,3 +299,8 @@ sketch
...
+.elasticbeanstalk/*
+!.elasticbeanstalk/*.cfg.yml
+!.elasticbeanstalk/*.global.yml
```
phoenixAja, line 306: we want these config files to be ignored? I thought it would be helpful to version these.

.platform/hooks/prebuild/02_build_frontend.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+source ~/.nvm/nvm.sh
+npm --prefix src/ts run build
```
phoenixAja, line 8: doesn't webpack do something like this automatically? Or is this something different?

.platform/hooks/prebuild/01_software_dependencies.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Install Node 12.16.1
+curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.37.2/install.sh | bash
+source ~/.nvm/nvm.sh
+nvm install 12.16.1
+
+# Install Frontend dependencies
+npm --prefix src/ts ci
```
phoenixAja: are we calling this file whenever we're deploying?

.gitignore
lvreynoso: This is `.gitignore` speak for "Ignore everything in the `.elasticbeanstalk` folder, *except* files that end with `.cfg.yml` or `.global.yml`." So the config files will be brought under version control. These lines were added by the `eb` command itself :)

I know EB uses the folder for temp files during deploy, for example.

.platform/hooks/prebuild/02_build_frontend.sh
lvreynoso: This invokes webpack through a script, and webpack places the build in the `static` folder of our flask app. We have to invoke it though. Webpack and Node do not need to run in production.

.platform/hooks/prebuild/01_software_dependencies.sh
lvreynoso: Yes, I should edit this script a bit so we aren't doing anything redundant on the instances.

lvreynoso: changed!

src/py/covidr/app.py
lvreynoso: Change to using Path. And a relative path does work, I was using absolute paths when debugging and got into the habit.

.gitignore
phoenixAja: oh haha you're right! It's been a long day and i misread the `.gitignore `

PR #14, opened 2021-01-16
Suggest that deployments go to separate environments

ttung:
### Description
No reason we should all be stomping over each other with -staging.

### Test plan
it's documentation <shrug>


phoenixAja:
for this to work do we have to create the eb environment before hand or will it automatically create it for us?

ttung:
> for this to work do we have to create the eb environment before hand or will it automatically create it for us?

Good point, let me fix it.  It's an extra step but pretty straightforward.

ttung:
Ok fixed!

lvreynoso (approved):

phoenixAja (approved):
sweet

PR #27, opened 2021-01-22
Add navbar and browser routing

lvreynoso:
### Description

Adds a navbar to the top of the app. Replaces the example React app page with an app component. Adds browser routing using React Router.

#### Issue
[ch65516](https://app.clubhouse.io/genepi/stories/space/65516)

### Test plan

Test for proper rendering in browser.

Screenshot from 2021-01-22 14-24-19


src/py/covidr/config/development.py
```diff
@@ -15,3 +15,8 @@ class DevelopmentDatabaseConfig(DatabaseConfig):
     @property
     def URI(self):
         return "postgresql://user_rw:password_rw@localhost:5432/covidr_db"
+
+    # ensures that latest static assets are read during frontend dev work
+    @property
+    def SEND_FILE_MAX_AGE_DEFAULT(self):
+        return 0
```
ttung: python's doc blocking style is _weird_

```suggestion

    @property
    """Ensures that latest static assets are read during frontend dev work."""
        def SEND_FILE_MAX_AGE_DEFAULT(self):
        return 0
```

src/ts/config/webpack.config.js
```diff
@@ -574,8 +577,8 @@ module.exports = function (webpackEnv) {
       // to restart the development server for webpack to discover it. This plugin
       // makes the discovery automatic so you don't have to restart.
       // See https://github.com/facebook/create-react-app/issues/186
-      isEnvDevelopment &&
-        new WatchMissingNodeModulesPlugin(paths.appNodeModules),
+      // isEnvDevelopment &&
+      //   new WatchMissingNodeModulesPlugin(paths.appNodeModules),
```
ttung: Did you intend to comment this out?

src/ts/src/App.scss
```diff
@@ -1,38 +1,50 @@
...
+// }
+
+// @keyframes App-logo-spin {
+//     from {
+//         transform: rotate(0deg);
+//     }
+//     to {
+//         transform: rotate(360deg);
+//     }
+// }
```
ttung: this intentional?

src/ts/config/webpack.config.js
lvreynoso: Yes, it was crashing watching my `node_modules` folder. I'm not sure if I should remove it entirely

src/py/covidr/config/development.py
lvreynoso: Got it, thanks

src/ts/src/App.scss
lvreynoso: Yes, I meant to just delete it outright! Will fix that.

src/ts/config/webpack.config.js
ttung: I don't see this crash.  Do you have a trace?

Also, it looks like you added the `watchOptions` directive above.  Does that avoid the problem?

src/ts/src/Landing.tsx
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+// import cx from "classnames";
```
ttung: intended?

src/py/covidr/config/development.py
```diff
@@ -15,3 +15,8 @@ class DevelopmentDatabaseConfig(DatabaseConfig):
     @property
     def URI(self):
         return "postgresql://user_rw:password_rw@localhost:5432/covidr_db"
+
+    @property
+    """Ensures that latest static assets are read during frontend dev work."""
+        def SEND_FILE_MAX_AGE_DEFAULT(self):
+        return 0
```
ttung:
```suggestion
    @property
    def SEND_FILE_MAX_AGE_DEFAULT(self):
        """Ensures that latest static assets are read during frontend dev work."""
        return 0
```

ttung (changes requested):

src/ts/src/Landing.tsx
lvreynoso: No, not yet anyways. As for `watchOptions`, that does not solve the issue since those two lines specifically tell the plugin to watch `node_modules`.

It seems to be behaving today so I can't reproduce the stack trace :/ perhaps all the heavy dev I was doing last week confused it? I'm not sure. But I can undo these changes.

ttung (approved):

src/py/covidr/config/development.py
```diff
@@ -15,3 +15,8 @@ class DevelopmentDatabaseConfig(DatabaseConfig):
     @property
     def URI(self):
         return "postgresql://user_rw:password_rw@localhost:5432/covidr_db"
+
+    @property
+    def SEND_FILE_MAX_AGE_DEFAULT(self):
+        """Ensures that latest static assets are read during frontend dev work."""
+        return 0
```
phoenixAja, line 22: interesting, i thought the static assets would be overwritten when recompiled.

src/ts/src/NavBar.module.scss
```diff
@@ -0,0 +1,15 @@
...
+
+.navbar {
+    background-color: var(--primary);
+    min-height: 50px;
+}
+
+.titleLink {
+    color: var(--white);
+
+    &:hover {
```
phoenixAja, line 11: can you remind me what the `&` does in scss?

phoenixAja (approved):
sorry for my late review, looks good!

src/py/covidr/config/development.py
lvreynoso: They are, but that doesn't necessarily mean Flask will send them instead of telling the browser to use it's cache.

phoenixAja: oh cool makes sense, thanks for explaining

src/ts/src/NavBar.module.scss
lvreynoso: The `&` means this applies when an element has both the scope class and this class. So in this case, the stuff in the `&:hover { }` block will only apply to elements that have the `titleLink` *and* the `hover` class.

PR #28, opened 2021-01-23
adding auth0 authentication with callback, login, logout views

phoenixAja:
### Description

adding in auth0 config

#### Issue
[ch64740](https://app.clubhouse.io/genepi/stories/space/64740)

### Test plan
hit login/logout endpoints try to hit a protected view when authenticated/ not authenticated, also deployed to eb-env `covidr-staging`

### DIFF
--- a/docs/backend/README.md
+++ b/docs/backend/README.md
@@ -1,3 +1,13 @@
+## Setup Auth0:
+
+you will need to have a specific `.env` file to read in `AUTH0` variables (place this file in `src/py/covidr`), ask project leads for this file, there will be different `.env` files for local development, staging, and production. 
+
+if deploying to a custom staging environment change this variable in the .env file:
+
+`AUTH0_CALLBACK_URL="http://<custom-staging-env-name>.us-west-2.elasticbeanstalk.com/callback"`
+
+if you are new to working on covidr ask a project lead to ask your eb staging env to the list of allowed callbacks in Auth0. 
+
 ## Deployment of the python web services
 
 Because the [AWS Elastic Beanstalk CLI](https://github.com/aws/aws-elastic-beanstalk-cli) depends on libraries that conflict with the libraries required by our development workflow, it is recommended to install the AWS EB CLI tools in a separate virtual environment.
--- a/src/py/covidr/app.py
+++ b/src/py/covidr/app.py
@@ -1,25 +1,91 @@
 import os
+from functools import wraps
 from pathlib import Path
+from urllib.parse import urlencode
 
+from authlib.integrations.flask_client import OAuth
 from covidr.config import DevelopmentConfig
-from flask import Flask, send_from_directory
+from covidr.config.config import Auth0Config
+from flask import Flask, redirect, send_from_directory, session, url_for
 
 static_folder = Path("static")
 
 # EB looks for an 'application' callable by default.
 application = Flask(__name__, static_folder=str(static_folder))
 
-
 if os.environ.get("FLASK_ENV") == "development":
     application.config.from_object(DevelopmentConfig())
+
+
+auth0_config = Auth0Config()
+application.secret_key = auth0_config.SECRET_KEY
+
+
+oauth = OAuth(application)
+
+auth0 = oauth.register(
+    "auth0",
+    client_id=auth0_config.AUTH0_CLIENT_ID,
+    client_secret=auth0_config.AUTH0_CLIENT_SECRET,
+    api_base_url=auth0_config.AUTH0_BASE_URL,
+    access_token_url=auth0_config.ACCESS_TOKEN_URL,
+    authorize_url=auth0_config.AUTHORIZE_URL,
+    client_kwargs=auth0_config.CLIENT_KWARGS,
+)
+
+
+# use this to wrap protected views
+def requires_auth(f):
+    @wraps(f)
+    def decorated(*args, **kwargs):
+        if "profile" not in session:
+            # Redirect to Login page
+            return redirect("/login")
+        return f(*args, **kwargs)
+
+    return decorated
 
 
 # Catch all routes. If path is a file, send the file;
 # else send index.html. Allows reloading React app from any route.
 @application.route("/", defaults={"path": ""})
 @application.route("/<path:path>")
+@requires_auth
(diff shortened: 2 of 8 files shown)


src/py/covidr/app.py
```diff
@@ -1,25 +1,104 @@
 import os
 from pathlib import Path
+from functools import wraps
+
+from dotenv import find_dotenv, load_dotenv
+from flask import Flask, send_from_directory, redirect, session, url_for
+from authlib.integrations.flask_client import OAuth
+from six.moves.urllib.parse import urlencode
```
ttung: We don't support python 2, so we can just import urllib.parse's urlencode.

src/py/requirements.txt
```diff
@@ -1,8 +1,12 @@
...
-itsdangerous==1.1.0
 Jinja2==2.11.2
 MarkupSafe==1.1.1
 SQLAlchemy==1.3.22
 Werkzeug==1.0.1
+click==7.1.2
+itsdangerous==1.1.0
+python-dotenv==0.15.0
+requests==2.24.0
+six==1.15.0
```
ttung: You should edit `src/py/requirements.txt.in`, add in python-dotenv and Authlib, and run `make -j pin-requirements`

ttung (changes requested):

src/py/requirements.txt
phoenixAja: oh ok so in the future we add new requirements to requirements.txt.ini and make -j pin-requirements will add them to requirements.txt? Can I ask what the benefit is of structuring it this way?

src/py/covidr/app.py
phoenixAja: good point

src/py/requirements.txt
ttung: Yeah, we are tracking the requirements we actually need.  The transitive dependencies are pinned but kept separate.

src/py/covidr/app.py
```diff
@@ -1,25 +1,104 @@
 import os
 from pathlib import Path
+from functools import wraps
+
+from dotenv import find_dotenv, load_dotenv
+from flask import Flask, send_from_directory, redirect, session, url_for
+from authlib.integrations.flask_client import OAuth
+from urllib.parse import urlencode
```
ttung:
```suggestion
import os
from functools import wraps
from pathlib import Path
from urllib.parse import urlencode

from dotenv import find_dotenv, load_dotenv
from flask import Flask, send_from_directory, redirect, session, url_for
from authlib.integrations.flask_client import OAuth
```

src/py/covidr/app.py
```diff
@@ -1,25 +1,104 @@
...
 
 from covidr.config import DevelopmentConfig
-from flask import Flask, send_from_directory
+
+
+ENV_FILE = find_dotenv()
+if ENV_FILE:
+    load_dotenv(ENV_FILE)
```
ttung: Could you add documentation for what needs to be in the `.env` file and where it's supposed to be located?

src/py/covidr/app.py
```diff
@@ -1,25 +1,104 @@
...
+
+AUTH0_CALLBACK_URL = os.environ.get("AUTH0_CALLBACK_URL")
+AUTH0_CLIENT_ID = os.environ.get("AUTH0_CLIENT_ID")
+AUTH0_CLIENT_SECRET = os.environ.get("AUTH0_CLIENT_SECRET")
+AUTH0_DOMAIN = os.environ.get("AUTH0_DOMAIN")
+AUTH0_BASE_URL = "https://" + AUTH0_DOMAIN
+SECRET_KEY = os.environ.get("SECRET_KEY")
```
ttung: I suspect many of these could live in the Config object.

ttung (changes requested):

src/py/covidr/app.py
phoenixAja: yes that is a very good idea

phoenixAja: do you think this would be better in the README or in the backend docs?

ttung: I imagine that README is a top-level directory plus anything that caters to a broad audience.  So in this case, the backend docs makes more sense.

phoenixAja:
hmm for some reason this is breaking when deploying to ebs but was working fine when running locally, in the process of debugging

src/py/covidr/app.py
lvreynoso: +1

phoenixAja:
> hmm for some reason this is breaking when deploying to ebs but was working fine when running locally, in the process of debugging

fixed now, thanks @ttung for the help!

src/py/covidr/app.py
```diff
@@ -1,25 +1,91 @@
...
+# use this to wrap protected views
+def requires_auth(f):
+    @wraps(f)
+    def decorated(*args, **kwargs):
+        if "profile" not in session:
+            # Redirect to Login page
+            return redirect("/login")
+        return f(*args, **kwargs)
+
+    return decorated
```
ttung, line 46: Not necessary to do in this PR, but putting this here could potentially cause problems with recursive imports if we break up `app.py`.

docs/backend/README.md
```diff
@@ -1,3 +1,13 @@
+## Setup Auth0:
+
+you will need to have a specific `.env` file to read in `AUTH0` variables (place this file in `src/py/covidr`), ask project leads for this file, there will be different `.env` files for local development, staging, and production. 
+
+if deploying to a custom staging environment change this variable in the .env file:
+
+`AUTH0_CALLBACK_URL="http://<custom-staging-env-name>.us-west-2.elasticbeanstalk.com/callback"`
+
+if you are new to working on covidr ask a project lead to ask your eb staging env to the list of allowed callbacks in Auth0. 
+
```
ttung, line 10: I created a couple follow-on stories:

https://app.clubhouse.io/genepi/story/68840/fold-authconfig-into-main-config-object
https://app.clubhouse.io/genepi/story/68841/figure-out-how-to-set-env-variables-as-part-of-eb-deploy

ttung (approved):

PR #29, opened 2021-01-26
Update docs for deployment

ttung:
### Description
1. The eb script is now run at the top level directory.
2. Add --cname option when creating the environment.

### Test plan
Tried creating covidr-ttung


phoenixAja (approved):
nice thanks for adding this

PR #30, opened 2021-01-26
Schema for database

ttung:
#### Issue
[ch68091](https://app.clubhouse.io/genepi/stories/space/68091)

### Test plan
it's a doc. :)

### DIFF
--- a/docs/backend/schema.ddl
+++ b/docs/backend/schema.ddl
@@ -1,0 +1,312 @@
+///////////////////////////////////////////////////
+// Groups and users
+
+Table Group {
+  id INT [pk, increment]
+  name VARCHAR [not null, unique]
+  email VARCHAR [not null, unique]
+  address VARCHAR [not null, unique]
+}
+
+Table User {
+  id INT [pk, increment]
+  name VARCHAR [not null, unique]
+  email VARCHAR [not null, unique]
+  auth0_user_id VARCHAR [not null]
+  group INT [not null, ref: > Group.id]
+  group_admin bool [not null]
+  system_admin bool [not null]
+}
+
+
+///////////////////////////////////////////////////
+// Cross-group viewing
+
+Table TypeOfData {
+  // Types of data people can view.  This table should be pre-populated with at least the following entries:
+  //  1. Trees
+  //  2. Sequences
+  //  3. Metadata
+  id INT [pk, increment]
+  name VARCHAR [not null]
+}
+
+// members of group A (viewer_group) can see data uploaded by group B (owner_group)
+Table GroupCanSee {
+  viewer_group_id INT [not null, ref: > Group.id]
+  owner_group_id INT [not null, ref: > Group.id]
+  type_of_data_id INT [not null, ref: > TypeOfData.id]
+
+  Indexes {
+    (viewer_group_id, owner_group_id, type_of_data_id) [unique]
+  }
+}
+
+
+///////////////////////////////////////////////////
+// Physical Samples
+
+Table PhysicalSample {
+  id INT [pk, increment]
+  submitting_group_id INT [not null, ref: > Group.id]
+  private_identifier VARCHAR [not null]
+
+  original_submission JSONB [not null]
+
+  // This is the public identifier we assign to this sample.  The identifier assigned by the public repositories is
+  // stored on the related Accession objects.
+  public_identifer VARCHAR [not null, unique]              // maps to specimen_collector_sample_id and isolate
+
+  // TODO: (sidneymbell) going to see if these are necessary.
+  // collector VARCHAR [not null]                             // maps to sample_collected_by
+  // collector_email VARCHAR                                  // maps to sample_collector_contact_email
+  // collector_address VARCHAR                                // maps to sample_collector_contact_address
+  // sequencer VARCHAR                                        // maps to sequence_submitted_by
+  // sequencer_email VARCHAR                                  // maps to sequence_submitter_contact_email
+  // sequencer_address VARCHAR                                // maps to sequence_submitter_contact_address
+
+  collection_date DATETIME [not null]                      // maps to sample_collection_date
+  location VARCHAR [not null]
+  division VARCHAR [not null]                              // maps to geo_loc_name_state_province_region
+  country VARCHAR [not null]                               // maps to geo_loc_name_country
+  country_residence VARCHAR                                // maps to host_origin_geo_loc_country
+  country_exposure VARCHAR                                 // maps to location_of_exposure_geo_loc_name_country
+  travel_history VARCHAR                                   // maps to travel_history
+
+  organism VARCHAR [not null]                              // maps to organism
+  purpose_of_sampling VARCHAR                              // maps to purpose_of_sampling
+  anatomical_material VARCHAR                              // maps to anatomical_material
+  anatomical_part VARCHAR                                  // maps to anatomical_part
+  body_product VARCHAR                                     // maps to body_product
+  environmental_material VARCHAR                           // maps to environmental_material
+  environmental_site VARCHAR                               // maps to environmental_site
+  collection_device VARCHAR                                // maps to collection_device
+  collection_method VARCHAR                                // maps to collection_method
+  collection_protocol VARCHAR                              // maps to collection_protocol
+  specimen_processing VARCHAR                              // maps to specimen_processing
+  lab_host VARCHAR                                         // maps to lab_host
+  passage_number VARCHAR                                   // maps to passage_number
+  passage_method VARCHAR                                   // maps to passage_method
+  biomaterial_extracted VARCHAR                            // maps to biomaterial_extracted
+  host_common_name VARCHAR                                 // maps to host_common_name
+  host_scientific_name VARCHAR                             // maps to host_scientific_name
+  host_health_state VARCHAR                                // maps to host_health_state
+  host_health_status_details VARCHAR                       // maps to host_health_status_details
+  host_disease VARCHAR                                     // maps to host_disease
+
+  symptom_onset_date DATETIME                              // maps to symptom_onset_date
+  signs_and_symptoms VARCHAR                               // maps to signs_and_symptoms
+  exposure_event VARCHAR                                   // maps to exposure_event
+
+  Indexes {
+    (submitting_group_id, private_identifier) [unique]
+  }
+}
+
+
+///////////////////////////////////////////////////
+// Entities and Workflows
+
+Table Entity {
+  id INT [pk, increment]
+}
+
+Table WorkflowType {
+  id INT [pk, increment]
+  name VARCHAR [not null, unique]
+}
+
+Table Workflow {
+  id INT [pk, increment]
+  workflow_type_id INT [not null, ref: > WorkflowType.id]
+
+  run_date DATETIME [not null]
+  pipeline_versions JSONB [not null]
+}
+
+Table WorkflowInputs {
+  entity_id INT [not null, ref: > Entity.id]
+  workflow_id INT [not null, ref: > Workflow.id]
+
+  Indexes {
+    (entity_id, workflow_id) [unique]
+  }
+}
+
+Table WorkflowOutputs {
+  entity_id INT [not null, ref: > Entity.id]
+  workflow_id INT [not null, ref: > Workflow.id]
+
+  Indexes {
+    (entity_id, workflow_id) [unique]
+  }
+}
+
+
+///////////////////////////////////////////////////
+// Public accession ids
+
+Table PublicRepository {
+  id INT [pk, increment]
+  name VARCHAR [not null, unique]
+  website VARCHAR [unique]
+}
+
+Table Accession {
+  entity_id INT [not null, ref: > Entity.id]
+  public_repository_id INT [not null, ref: > PublicRepository.id]
+  public_identifier VARCHAR [not null]
+
+  Indexes {
+    (entity_id, public_repository_id) [unique]
+    (public_repository_id, public_identifier) [unique]
+  }
+}
+
+
+///////////////////////////////////////////////////
+// Sequences
+
+// describes a sequencer machine  (TODO: what's a good way of canonicalizing this?)
+Table SequencingInstrument {
+  id INT [pk, increment]
+  name VARCHAR [not null, unique]
+}
+
+// describes a sequencing protocol  (TODO: what's a good way of canonicalizing this?)
+Table SequencingProtocol {
+  id INT [pk, increment]
+  name VARCHAR [not null, unique]
+}
+
+// describes a single set of sequence data from the sequencer. (FASTQ)
+Table SequencingReads {
+  id INT [pk, increment]
+  entity_id INT [not null, unique, ref: - Entity.id]
+
+  physical_sample_id INT [not null, unique, ref: - PhysicalSample.id]
+  s3_bucket VARCHAR [not null]
+  s3_key VARCHAR [not null, unique]
+
+  sequencing_instrument_id INT [not null, ref: > SequencingInstrument.id]
+  sequencing_protocol_id INT [not null, ref: > SequencingProtocol.id]
+
+  // sequencing date is optional, but sample collection date should not be.
+  sequencing_date FLOAT
+
+  Indexes {
+    (s3_bucket, s3_key) [unique]
+  }
+}
+
+// describes a single pathogen sequence
+Table PathogenGenome {
+  id INT [pk, increment]
+  entity_id INT [not null, unique, ref: > Entity.id]
+
+  sequence VARCHAR [not null]
+
+  // sequence statistics we can calculate from the fasta file
+  num_unambiguous_sites INT [not null]
+  num_n INT [not null]
+  num_mixed INT [not null]
+
+}
+
+// describes a single sequence that was uploaded to the system directly.
+Table UploadedPathogenGenome {
+  pathogen_genome_id INT [not null, unique, ref: - PathogenGenome.id]
+  physical_sample_id INT [not null, unique, ref: - PhysicalSample.id]
+
+  // optional field for gisaid submission, we'd like to get users to provide this.
+  sequencing_depth FLOAT
+}
+
+// describes a single sequence that was called from a consensus genome run
+Table CalledPathogenGenome {
+  pathogen_genome_id INT [not null, unique, ref: - PathogenGenome.id]
+}
+
+///////////////////////////////////////////////////
+// Consensus genome pipeline
+
+// TODO: @jackkamm do the BAM/SRA records need to store any metrics?
+
+Table HostFilteredSequence {
+  id INT [pk, increment]
+  entity_id INT [not null, unique, ref: - Entity.id]
+
+  // For the MVP, we will just store them, in the future, we would ship them to archives and store just a pointer.
+  // These should be Entities.
+
+  s3_bucket VARCHAR [not null]
+  s3_key VARCHAR [not null, unique]
+
+  Indexes {
+    (s3_bucket, s3_key) [unique]
+  }
+}
+
+Table Bam {
+  id INT [pk, increment]
+  entity_id INT [not null, unique, ref: - Entity.id]
+
+  // For the MVP, we will just store them, in the future, we would ship them to archives and store just a pointer.
+  // These should be Entities.
+
+  // optional field for gisaid submission, we'd like to get users to provide this.
+  sequencing_depth FLOAT
+
+  s3_bucket VARCHAR [not null]
+  s3_key VARCHAR [not null, unique]
+
+  Indexes {
+    (s3_bucket, s3_key) [unique]
+  }
+}
+
+
+///////////////////////////////////////////////////
+// GISAID dumps
+
+Table RawGisaidDump {
+  id INT [pk, increment]
+  entity_id INT [not null, unique, ref: - Entity.id]
+
+  download_date datetime [not null]
+
+  s3_bucket VARCHAR [not null]
+  s3_key VARCHAR [not null, unique]
+
+  Indexes {
+    (s3_bucket, s3_key) [unique]
+  }
+}
+
+Table ProcessedGisaidDump {
+  id INT [pk, increment]
+  entity_id INT [not null, unique, ref: - Entity.id]
+
+  s3_bucket VARCHAR [not null]
+  s3_key VARCHAR [not null, unique]
+
+  Indexes {
+    (s3_bucket, s3_key) [unique]
+  }
+}
+
+
+///////////////////////////////////////////////////
+// Phylo pipeline
+
+Table PhyloTree {
+  id INT [pk, increment]
+  entity_id INT [not null, unique, ref: - Entity.id]
+
+  s3_bucket VARCHAR [not null]
+  s3_key VARCHAR [not null, unique]
+
+  Indexes {
+    (s3_bucket, s3_key) [unique]
+  }
+}


ttung:
You can paste the text into https://dbdiagram.io/d to visualize it.

docs/backend/schema.ddl
```diff
@@ -0,0 +1,115 @@
...
+  auth0_user_id varchar [not null]
+  dph int [not null, ref: > Dph.id]
+  dph_admin bool [not null]
+}
+
+// members of dph can see data uploaded by owner_dph
+Table DphCanSee {
+  dph int [not null, unique, ref: > Dph.id]
+  owner_dph int [not null, unique, ref: > Dph.id]
+}
```
ttung: @sidneymbell  could you comment on this?

docs/backend/schema.ddl
```diff
@@ -0,0 +1,115 @@
...
+
+Table PcrCtValue {
+  physical_sample_id int [not null, ref: > PhysicalSample.id]
+  pcr_primer_id int [not null, ref: > PcrPrimer.id]
+  value float [not null]
+  
+  Indexes {
+    (physical_sample_id, pcr_primer_id) [unique]
+  }
+}
```
ttung: @jackkamm would love your feedback on if this is adequate to capture CT values but is flexible enough for changes in the PCR assay.

jackkamm:
Random thoughts/questions:

Is `RawSequence` the fastqs?

Should store info about what sequencer was used. In particular, it will be important to distinguish illumina vs oxford nanopore data. Also, what protocol was used -- typically this will be ARTICv3 but this could always change in future

I recommend storing the BAM file intermediates between the fastqs and the fastas as well

I recommend storing some sequence quality statistics -- main ones are depth of coverage, number of sites recovered, number of Ns, and number of mixed/ambiguous sites

I'll be out tomorrow but can give more feedback later in week on Thursday or Friday

docs/backend/schema.ddl
jackkamm: Honestly I'm a little hesitant to store all these CT values. There are so many assays, they are so heterogeneous, some of them don't even give CT values but give other readouts. While they have been generally useful for us in sanity checking our sequencing runs, for covidtracker v2 I think we will devolve this responsibility to the submitting DPHs

jackkamm:
Another comment: would be good to store NCBI names and accession IDs for each of the uploaded fastqs, bams, and fastas.

Speaking of which -- we'd need to upload host-filtered fastqs to SRA, so perhaps there should be an additional object for the host-filtered fastqs.

ttung:
> Is RawSequence the fastqs?

yes.

> Should store info about what sequencer was used. In particular, it will be important to distinguish illumina vs oxford nanopore data. Also, what protocol was used -- typically this will be ARTICv3 but this could always change in future

what fields do we need?  should sequencer be free text or constrained?  protocol?

> I recommend storing the BAM file intermediates between the fastqs and the fastas as well

Do the BAM files have human read data?

> I recommend storing some sequence quality statistics -- main ones are depth of coverage, number of sites recovered, number of Ns, and number of mixed/ambiguous sites

these are all ints, and per Sequence, right?  Do we expect users to upload this?

lvreynoso:
Perhaps we could filter out human reads before doing any analysis? And we could treat the filtered fastq as the "input" file, and only store that. Then human read data should be excluded from any other output files, right?

lvreynoso:
And the sequence quality stats would be generated by us, not the users. Depth and coverage are usually decimals/fractional.

ttung:
> And the sequence quality stats would be generated by us, not the users. Depth and coverage are usually decimals/fractional.

Wouldn't that be generated by users if they're uploading fastas?

lvreynoso:
Fastas/fastqs come with *read quality* information for individual reads, but not higher order quality analysis. Those statistics are only obtainable by doing further data analysis (e.g. running STAR, PriceSeq, etc.) which presumably DPHs wouldn't be doing since they would just be sequencing samples and sending us the fastqs.

docs/backend/schema.ddl
ttung: @sidneymbell if you're on board with this, i can drop this.

sidneymbell:
Haven't had a chance to look (OOO today) but will in the morning

jackkamm:
> Wouldn't that be generated by users if they're uploading fastas?

Are we allowing people to upload their own fastas if they ran their own bioinformatics? Or are we only allowing them to upload fastqs of reads?

In the former case, we would want them to upload some of these quality statistics; otherwise we will generate ourselves when running the bioinformatics pipelines.

docs/backend/schema.ddl
phoenixAja: what is the difference between owner_dph and dph_admin?

ttung: I should rename that as:

viewing_dph
owner_dph

Meaning viewing_dph can see the data for owner_dph.

phoenixAja: I'm also not sure if we would need the `DphCanSee` table since we could determine whether a user was in a dph or an admin by looking at the `User` table directly.

ttung: `DphCanSee` is for people to cross that boundary.

phoenixAja: sorry i'm not really sure what you mean by that. Why not just query the User and see what DPH they belong to (and if they are an admin) directly in the User table? If they are a part of the DPH then they can see the data?

ttung: So for instance, we might have an entry for people in CADPH to view SCCDPH's data.

phoenixAja: ok got it thanks

ttung: I'm not fully sold this is the right model, because it doesn't fully capture the nuances.  For instance, SCCDPH might want to let CADPH view its data but not its metadata.  But making it a full cross product might be a bit too heavyweight.

docs/backend/schema.ddl
```diff
@@ -0,0 +1,115 @@
...
+  
+  // TODO: metedata about the consensus genome run
+}
+
+///////////////////////////////////////////////////
+// Phylo Runs
+
+Table PhyloRun {
+  id int [pk, increment]
+}
```
phoenixAja: should we also have a field here that points to the location of the output JSONs?

ttung: LOL right.  :)

docs/backend/schema.ddl
```diff
@@ -0,0 +1,115 @@
...
+  gisaid_accession varchar [unique]
+
+  // TODO: Do we store the genome directly on this object?
+}
+
+// describes a single sequence that was uploaded to the system directly.
+Table UploadedSequence {
+  sequence_id int [not null, unique, ref: > Sequence.id]
+  physical_sample_id int [not null, ref: > PhysicalSample.id]
+}
```
phoenixAja: why do we need separate tables for `Sequence` and `UploadedSequence`?

ttung: `Sequence` is like the parent object for `UploadedSequence` and `ConsensusGenomeSequence`.  There needs to be two ways for a sequence to be connected to a physical sample.  If the user is uploading a fasta, then the metadata is immediately connected to the fasta.  If the user is uploading a fastq, then the metadata is connected to the fastq.  After the consensus genome is called, then the consensus genome is connected to the fastq (plausibly more than once if the pipeline is called multiple times).

docs/backend/schema.ddl
```diff
@@ -0,0 +1,142 @@
+///////////////////////////////////////////////////
+// DPHs and users
```
sidneymbell: Not a blocker, but eventually want to generalize from "DPH" --> "Group" or something like this.

docs/backend/schema.ddl
```diff
@@ -0,0 +1,142 @@
...
+///////////////////////////////////////////////////
+// Physical Samples
+
+Table PhysicalSample {
+  id int [pk, increment]
+  submitting_dph int [not null, ref: > Dph.id]
+  date_taken datetime [not null]
+  external_accession varchar [not null]
+  location varchar [not null]
+  extraction_method varchar [not null]
```
sidneymbell: This seems LIMS-y to me. @shanaxel42?

docs/backend/schema.ddl
```diff
@@ -0,0 +1,142 @@
...
+
+Table PcrPrimer {
+  id int [pk, increment]
+  gene_id int [not null, ref: > Gene.id]
+  name varchar [not null]
+
+  Indexes {
+    (gene_id, name) [unique]
+  }
+}
```
sidneymbell: I think of PCR primer as an attribute of `protocol`, relevant only for consensus genome generation

docs/backend/schema.ddl
```diff
@@ -0,0 +1,142 @@
...
+
+Table PcrCtValue {
+  physical_sample_id int [not null, ref: > PhysicalSample.id]
+  pcr_primer_id int [not null, ref: > PcrPrimer.id]
+  value float [not null]
```
sidneymbell: This also seems LIMS-y to me.

docs/backend/schema.ddl
```diff
@@ -0,0 +1,142 @@
...
+// describes a single set of sequence data from the sequencer. (FASTQ)
+Table RawSequence {
+  id int [pk, increment]
+  physical_sample_id int [not null, ref: > PhysicalSample.id]
+  s3_bucket str [not null]
+  s3_key str [not null, unique]
+
+  // TODO: @jackkamm suggested storing the sequencer information.  "In particular, it will be important to distinguish
+  // illumina vs oxford nanopore data. Also, what protocol was used -- typically this will be ARTICv3 but this could
+  // always change in future"
```
sidneymbell: Yes (see note on `primers` above)

docs/backend/schema.ddl
```diff
@@ -0,0 +1,142 @@
...
+    (s3_bucket, s3_key) [unique]
+  }
+}
+
+// describes a single pathogen sequence
+Table PathogenSequence {
+  id int [pk, increment]
+  gisaid_name varchar [unique]
+  gisaid_accession varchar [unique]
+  sequence varchar [not null]
```
sidneymbell: plus`external_accession`?

As a comp bio, when working with this data, I'd want the following information easily available as a "record":
* Public ID (currently called `gisaid_name`)
* Private ID (currently called (`external_accession`)
* Consensus genome
* Pointer to metadata (location + collection date)

docs/backend/schema.ddl
```diff
@@ -0,0 +1,142 @@
...
+
+  Indexes {
+    (s3_bucket, s3_key) [unique]
+  }
+}
+
+// describes a single pathogen sequence
+Table PathogenSequence {
+  id int [pk, increment]
+  gisaid_name varchar [unique]
```
sidneymbell: `gisaid_name` -> `public_id`

docs/backend/schema.ddl
```diff
@@ -0,0 +1,142 @@
...
+  gisaid_accession varchar [unique]
+  sequence varchar [not null]
+}
+
+// describes a single sequence that was uploaded to the system directly.
+Table UploadedPathogenSequence {
+  pathogen_sequence_id int [not null, unique, ref: > PathogenSequence.id]
+  physical_sample_id int [not null, ref: > PhysicalSample.id]
```
sidneymbell: The vast majority of users won't have this

docs/backend/schema.ddl
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// describes a single sequence that was called from a consensus genome run
+Table CalledPathogenSequence {
+  pathogen_sequence_id int [not null, unique, ref: > PathogenSequence.id]
+  raw_genome_id int [not null, ref: > RawSequence.id]
+
+  // TODO: @jackkamm suggested storing the intermediate files (BAM file?, host-filtered short reads).  Need more
```
sidneymbell: I'd put this in the same bucket as storing the fastq files -- good to keep on hand in case something needs to be reprocessed, but we don't need to store indefinitely

docs/backend/schema.ddl
```diff
@@ -0,0 +1,142 @@
...
+  // information how to work that into the schema.
+
+  // TODO: @jackkamm suggested storing quality statistics.  "main ones are depth of coverage, number of sites recovered,
+  // number of Ns, and number of mixed/ambiguous sites"
```
sidneymbell: Yes.

docs/backend/schema.ddl
```diff
@@ -0,0 +1,142 @@
...
+// Phylo Runs
+
+Table PhyloRun {
+  id int [pk, increment]
+  s3_bucket str [not null]
+  s3_key str [not null, unique]
+
+  // metadata about the phylo run
+  run_date datetime [not null]
+  // TODO: additional metadata?  sw version?  information about the gisaid data dump.
```
sidneymbell: Yeah. I'd potentially even suggest having a more generic `analysis` table that just lists out
1. type of analysis (consensus or phylo)
2. pipeline and version
3. date of run
4. any errors that were raised
5. pointers for where to find the output files (including raw results files like `tree.newick` and `aligned.fasta`)

docs/backend/schema.ddl
```diff
@@ -0,0 +1,142 @@
...
+
+  Indexes {
+    (s3_bucket, s3_key) [unique]
+  }
+}
+
+// this enables a many:many relationship between Sequence and PhyloRun.
```
sidneymbell: Yussssss

docs/backend/schema.ddl
sidneymbell: Idea being that eventually the api for a generic pipeline module looks something or other like:

input: 
`records` as described above
` parameters` (out of scope for now, but eventually)
`output` (where output files should be stored

output:
the above ^ 
`intermediate results` --> `tree.newick`, etc. to be stored
`end product` --> `build.json`, `consensus genome`, etc.

jackkamm:
I suggest referring to these for designing the schema:
- https://github.com/pha4ge/SARS-CoV-2-Contextual-Data-Specification/blob/master/PHA4GE%20SARS-CoV-2%20Contextual%20Data%20Template.xlsx
- https://github.com/seattleflu/documentation/blob/master/id3c-warehouse-schema.pdf

jackkamm:
Another good reference is COG-UK's system

https://www.biorxiv.org/content/10.1101/2020.10.06.328328v1

image

ttung:
I made a bunch of changes to this.  Please review when you have time. :)

docs/backend/schema.ddl
```diff
@@ -0,0 +1,280 @@
...
+
+// describes a single sequence that was called from a consensus genome run
+Table CalledPathogenSequence {
+  pathogen_sequence_id INT [not null, unique, ref: > PathogenSequence.id]
+}
+
+///////////////////////////////////////////////////
+// Consensus genome pipeline
+
+// TODO: @jackkamm do the BAM/SRA records need to store any metrics?
```
ttung, line 233: @jackkamm

erinstargazer:
I am just getting up to speed on this so don't have many comments since I'm mainly still learning it.

I was wondering about the situation with `Entity` though. It looks like it's being used as an abstraction for specific relationships (I might just be way off the mark here though). For instance it seems like `ProcessedGisaidDump` would have an associated `RawGisaidDump` but the only way I could see that happening is through the `Entity` relationships they both have. At some point `Entity` becomes so generalized and it feels like almost everything could be an `Entity`. This can lead to less information being conveyed. For instance, if I have a car with 4 tires, having a relationship that each tire can go to 1 car is clearer than having each tire having an `Entity` and each car also having an `Entity` and then needing to infer the rest of the info. Sorry if this is nonsense but I wanted to find out more of the motivation and guiding principles around the `Entity` abstraction.

ttung:
So entity being the base unit is motivated by two things:
1. A lot of things may be submitted to public repositories.
2. It's a way to glue workflows from beginning to end.

The relationship between entities is a bit more complicated mostly because of the nature of workflows.  You'd have to find the workflows and find the corresponding outputs.  There isn't a table that directly maps input to output.

I think the solution is to provide helper methods in each of the specific entities that simplify finding the relationships.  For instance, `RawGisaidDump` might have a `processed_dumps()` method that abstracts that.

jackkamm:
Getting the following error on most recent commit in dbdiagram.io

```
 (64:5) Column "submitting_group" do not exist in table "PhysicalSample" 
```

image

jackkamm:
It's fixed by the following patch
```
   // TODO: the other fields from PHA4GE
 
   Indexes {
-    (submitting_group, private_identifier) [unique]
+    (submitting_group_id, private_identifier) [unique]
   }
 }
```

docs/backend/schema.ddl
```diff
@@ -0,0 +1,281 @@
...
+///////////////////////////////////////////////////
+// Public accession ids
+
+Table PublicRepository {
+  id INT [pk, increment]
+  name VARCHAR [not null, unique]
+  website VARCHAR [unique]
+}
+
+Table Accession {
```
jackkamm, line 155: I think this table needs a column for the actual accession (which should be a varchar)

docs/backend/schema.ddl
```diff
@@ -0,0 +1,281 @@
...
+}
+
+// describes a sequencing protocol  (TODO: what's a good way of canonicalizing this?)
+Table SequencingProtocol {
+  id INT [pk, increment]
+  name VARCHAR [not null, unique]
+}
+
+// describes a single set of sequence data from the sequencer. (FASTQ)
+Table RawSequence {
```
jackkamm: Is this the raw fastq or the host-filtered fastq?

If the former, we need another object for the host-filtered fastq.

If we host-filter the fastq upfront then we could just store that, and not store the raw fastq. (Similar to how idseq does it)

Also, this needs to be linked to the BAM

docs/backend/schema.ddl
```diff
@@ -0,0 +1,281 @@
...
+  s3_key VARCHAR [not null, unique]
+
+  sequencer_id INT [not null, ref: > Sequencer.id]
+  sequencing_protocol_id INT [not null, ref: > SequencingProtocol.id]
+
+  // sequencing date is optional, but sample collection date should not be.
+  sequencing_date FLOAT
+
+  // optional field for gisaid submission, we'd like to get users to provide this.
+  sequencing_depth FLOAT
```
jackkamm: This should be a column of the BAM not the fastq

docs/backend/schema.ddl
```diff
@@ -0,0 +1,281 @@
+///////////////////////////////////////////////////
+// Groups and users
+
+Table Group {
+  id INT [pk, increment]
+  name VARCHAR [not null, unique]
+  gisaid_division_name VARCHAR [unique]
```
jackkamm: I don't think it is a safe assumption that each Group will be 1-1 with a location. For example, there is a single public lab for Napa, Sonoma, Yolo, Marin, which ideally should be separate locations in GISAID.

Also, I think this might be redundant with the PhysicalSample.location column for location of physical samples.

On the other hand, we do need to provide an address for the group, for submission to GISAID

docs/backend/schema.ddl
```diff
@@ -0,0 +1,281 @@
...
+Table PhysicalSample {
+  id INT [pk, increment]
+  submitting_group_id INT [not null, ref: > Group.id]
+  private_identifier VARCHAR [not null]
+
+  original_submission JSONB [not null]
+
+  // TODO: dedup these fields with PHA4GE
+  date_taken DATETIME [not null]
+  location VARCHAR [not null]
```
jackkamm, line 69: Is this always going to be a county?

Also, will we ever include samples from other states/countries? If so we should include columns for these.

Perhaps instead of a single VARCHAR, we could link to a separate table that would store county, state, country and perhaps other info (lat/lon?)

docs/backend/schema.ddl
```diff
@@ -0,0 +1,281 @@
...
+}
+
+
+///////////////////////////////////////////////////
+// Sequences
+
+// describes a sequencer machine  (TODO: what's a good way of canonicalizing this?)
+Table Sequencer {
+  id INT [pk, increment]
+  name VARCHAR [not null, unique]
```
jackkamm, line 173: Here are the valid Sequencer names that NCBI will accept. I am not sure where/how we should encode them specifically

- Illumina HiSeq 1000
- Illumina HiSeq 1500
- Illumina HiSeq 2000
- Illumina HiSeq 2500
- Illumina HiSeq 3000
- Illumina HiSeq 4000
- Illumina iSeq 100
- Illumina NovaSeq 6000
- Illumina MiniSeq
- Illumina MiSeq
- NextSeq 500
- NextSeq 550

docs/backend/schema.ddl
```diff
@@ -0,0 +1,281 @@
...
+
+}
+
+// describes a single sequence that was uploaded to the system directly.
+Table UploadedPathogenSequence {
+  pathogen_sequence_id INT [not null, unique, ref: > PathogenSequence.id]
+  physical_sample_id INT [not null, ref: > PhysicalSample.id]
+
+  // optional field for gisaid submission, we'd like to get users to provide this.
+  sequencing_depth FLOAT
```
jackkamm, line 222: I think this should be mandatory not optional

docs/backend/schema.ddl
```diff
@@ -0,0 +1,281 @@
...
+Table CalledPathogenSequence {
+  pathogen_sequence_id INT [not null, unique, ref: > PathogenSequence.id]
+}
+
+///////////////////////////////////////////////////
+// Consensus genome pipeline
+
+// TODO: @jackkamm do the BAM/SRA records need to store any metrics?
+
+Table HostFilteredBam {
```
jackkamm: `HostFiltered` is redundant since it's aligned to SARS-CoV-2. Just call it `Bam` or `AlignedReads`.

Also, this needs to be linked to the `RawSequence` and the `CalledPathogenSequence` somehow

docs/backend/schema.ddl
```diff
@@ -0,0 +1,281 @@
...
+  pathogen_sequence_id INT [not null, unique, ref: > PathogenSequence.id]
+  physical_sample_id INT [not null, ref: > PhysicalSample.id]
+
+  // optional field for gisaid submission, we'd like to get users to provide this.
+  sequencing_depth FLOAT
+}
+
+// describes a single sequence that was called from a consensus genome run
+Table CalledPathogenSequence {
+  pathogen_sequence_id INT [not null, unique, ref: > PathogenSequence.id]
```
jackkamm: This needs to be linked to the Bam

docs/backend/schema.ddl
```diff
@@ -0,0 +1,281 @@
...
+  Indexes {
+    (viewer_group_id, owner_group_id, type_of_data_id) [unique]
+  }
+}
+
+
+///////////////////////////////////////////////////
+// Physical Samples
+
+Table PhysicalSample {
```
jackkamm, line 49: This needs a column for the `isolate` (the name used for submission to GISAID/Genbank/etc)

jackkamm: We should also add a column for `purpose of sampling` as recommended by PH4AGE. Important to distinguish between samples collected for surveillance, for outbreak investigation, and for targeted searches e.g. S-gene dropout

ttung:
> It's fixed by the following patch

thanks! patched.

docs/backend/schema.ddl
```diff
@@ -0,0 +1,281 @@
...
+
+  // sequence statistics we can calculate from the fasta file
+  num_unambiguous_sites INT [not null]
+  num_n INT [not null]
+  num_mixed INT [not null]
+
+}
+
+// describes a single sequence that was uploaded to the system directly.
+Table UploadedPathogenSequence {
```
jackkamm: Add column for the software used to generate the sequence. Needed for submission to GISAID/NCBI

PH4AGE actually recommends separating this out into separate fields for the software used for raw sequence data processing, assembly method, and consensus sequence method. I think the first 2 may only be relevant for long-read (Oxford Nanopore) data but I'm not sure -- should check in with someone who knows more about the long-read genome pipelines.

docs/backend/schema.ddl
jackkamm: Add column for the software used to generate the sequence. Needed for submission to GISAID/NCBI

PH4AGE actually recommends separating this out into separate fields for the software used for raw sequence data processing, assembly method, and consensus sequence method. I think the first 2 may only be relevant for long-read (Oxford Nanopore) data but I'm not sure -- should check in with someone who knows more about the long-read genome pipelines.

docs/backend/schema.ddl
jackkamm: Add a column for the reference genome

docs/backend/schema.ddl
jackkamm: Add a column for the reference genome accession

docs/backend/schema.ddl
jackkamm: These are recommended by PHA4GE, and nice to have, but I think not strictly necessary:
- library insert size (int)
- number base pairs (int)
- mean contig length (float)
- N50 (int)

jackkamm (changes requested):
I've cross-checked with PHA4GE. Please see the attached spreadsheet. I've also made inline comments in the relevant locations.

[PHA4GE-aspen-crosscheck.xlsx](https://github.com/chanzuckerberg/aspen/files/5907741/PHA4GE-aspen-crosscheck.xlsx)

docs/backend/schema.ddl
ttung: Will fix.

docs/backend/schema.ddl
ttung: Sounds good.

docs/backend/schema.ddl
ttung: HAHA yeah you're totally right.

docs/backend/schema.ddl
ttung: I did not realize there were two. :)

ttung: IDseq doesn't store the raw upload?  Did not know that.

docs/backend/schema.ddl
ttung: I thought this was user-provided data.  In that case, it would need to be attached to the fastq object.

docs/backend/schema.ddl
ttung: > Add column for the software used to generate the sequence. Needed for submission to GISAID/NCBI

> PH4AGE actually recommends separating this out into separate fields for the software used for raw sequence data processing, assembly method, and consensus sequence method. I think the first 2 may only be relevant for long-read (Oxford Nanopore) data but I'm not sure -- should check in with someone who knows more about the long-read genome pipelines.

are these mandatory fields?

> Add a column for the reference genome

is this mandatory? Also, is this a choice of a few acceptable reference genomes?  or is it freeform text?

docs/backend/schema.ddl
ttung: I thought we looked it up and it was an optional field for gisaid.

docs/backend/schema.ddl
ttung: > This needs to be linked to the Bam

It is.  This should is the output of a workflow, so you can trace it back from the Workflow to the BAM.

> Add column for the software used to generate the sequence. Needed for submission to GISAID/NCBI

> PH4AGE actually recommends separating this out into separate fields for the software used for raw sequence data processing, assembly method, and consensus sequence method. I think the first 2 may only be relevant for long-read (Oxford Nanopore) data but I'm not sure -- should check in with someone who knows more about the long-read genome pipelines.

The workflow object contains the software information.

docs/backend/schema.ddl
ttung: > `HostFiltered` is redundant since it's aligned to SARS-CoV-2. Just call it `Bam` or `AlignedReads`.

👍 

> Also, this needs to be linked to the `RawSequence` and the `CalledPathogenSequence` somehow

It's done through the workflow objects.

> Add a column for the reference genome accession

Anything that is an entity can have an accession (which you noted was missing).

docs/backend/schema.ddl
ttung: `isolate` would all be SARS-CoV-2, right?  this is futureproofing, I assume?

docs/backend/schema.ddl
ttung: What about nanopore sequencing?

docs/backend/schema.ddl
jackkamm: Not 100% sure about how IDseq stores the fastqs, I might be wrong about it

docs/backend/schema.ddl
jackkamm: No it's the average number of reads covering each position in the reference and is computed by aligning reads to reference. The user needs to supply it if they upload a fasta, but if they upload a fastq we compute it ourselves

docs/backend/schema.ddl
jackkamm: You're right sorry, here are the valid values (taken from the PH4AGE spreadsheet I attached above)

```
sequencing instrument
Illumina
    HiSeq X
    HiSeq X Five
    HiSeq X Ten
    Illumina Genome Analyzer
    Illumina Genome Analyzer II
    Illumina Genome Analyzer IIx
    Illumina HiScanSQ
    Illumina HiSeq 1000
    Illumina HiSeq 1500
    Illumina HiSeq 2000
    Illumina HiSeq 2500
    Illumina HiSeq 3000
    Illumina HiSeq 4000
    Illumina iSeq 100
    Illumina NovaSeq 6000
    Illumina MiniSeq
    Illumina MiSeq
    NextSeq 500
    NextSeq 550
Pacific Biosciences
    PacBio RS
    PacBio RS II
    PacBio Sequel
    PacBio Sequel II
Ion Torrent
    Ion Torrent PGM
    Ion Torrent Proton
    Ion Torrent S5 XL
    Ion Torrent S5
Oxford Nanopore Technologies
    GridION
    MinION
    PromethION
BGI Genomics
    BGISEQ-500
MGI
    DNBSEQ-T7
    DNBSEQ-G400
    DNBSEQ-G400 FAST
    DNBSEQ-G50
Not Applicable
Not Collected
Not Provided
Missing
Restricted Access
```

docs/backend/schema.ddl
jackkamm: We should test uploading a sequence to GISAID without it, but it's the most basic sequence quality statistic so I think we should require it

docs/backend/schema.ddl
jackkamm: No, `isolate` is what we've been calling `gisaid_name` -- basically, it's the human-readable public name of the sample, and is usually something like `USA/CA-CZB-1234/2021`

docs/backend/schema.ddl
jackkamm: I think consensus sequence method is mandatory. In any case it's really important information to provide publicly. For us, a minimal version of it is `ivar=1.2`

docs/backend/schema.ddl
jackkamm: Or if not require, then strongly strongly encourage -- perhaps by making it mandatory to start, and then only consider loosening this if a submitter can make a compelling case why this would be too onerous for them to collect and submit

docs/backend/schema.ddl
```diff
@@ -0,0 +1,342 @@
+///////////////////////////////////////////////////
+// Groups and users
+
+Table Group {
+  id INT [pk, increment]
+  name VARCHAR [not null, unique]
+  email VARCHAR [not null, unique]
+  address VARCHAR [not null, unique]
```
sidneymbell, line 8: Referring to group owner?

docs/backend/schema.ddl
```diff
@@ -0,0 +1,342 @@
...
+}
+
+
+///////////////////////////////////////////////////
+// Cross-group viewing
+
+Table TypeOfData {
+  // Types of data people can view.  This table should be pre-populated with at least the following entries:
+  //  1. Trees
+  //  2. Sequences
```
sidneymbell, line 28: 3. metadata (per column)

docs/backend/schema.ddl
```diff
@@ -0,0 +1,342 @@
...
+Table PhysicalSample {
+  id INT [pk, increment]
+  submitting_group_id INT [not null, ref: > Group.id]
+  private_identifier VARCHAR [not null]
+
+  original_submission JSONB [not null]
+
+  // This is the public identifier we assign to this sample.  The identifier assigned by the public repositories is
+  // stored on the related Accession objects.
+  public_identifer VARCHAR [not null, unique]              // maps to specimen_collector_sample_id and isolate
```
sidneymbell, line 58: Re: comment on `isolate` above... are these the same things? (I personally find it more intuitive to think of this as `public_identifier`, although agree we should take the ph4ge terminology for the sake of being good citizens

docs/backend/schema.ddl
```diff
@@ -0,0 +1,342 @@
...
+  collector VARCHAR [not null]                             // maps to sample_collected_by
+  collector_email VARCHAR                                  // maps to sample_collector_contact_email
+  collector_address VARCHAR                                // maps to sample_collector_contact_address
+  sequencer VARCHAR                                        // maps to sequence_submitted_by
```
sidneymbell: I first read this as what instrument model was used

docs/backend/schema.ddl
```diff
@@ -0,0 +1,342 @@
...
+  sequencer_email VARCHAR                                  // maps to sequence_submitter_contact_email
+  sequencer_address VARCHAR                                // maps to sequence_submitter_contact_address
+
+  collection_date DATETIME [not null]                      // maps to sample_collection_date
+  received_date DATETIME                                   // maps to sample_received_date
```
sidneymbell: If we're not doing the sequencing anymore, we no longer have a `received_date` for physical samples?

docs/backend/schema.ddl
sidneymbell: IDSeq has a very nice auto-suggest feature in their upload process where the user starts typing a location (address, city, county), and it provides discrete (read: standardized) options and "rounds" these values to the appropriate level of detail for confidentiality (i.e., county or less resolution)

docs/backend/schema.ddl
```diff
@@ -0,0 +1,342 @@
...
+  location VARCHAR [not null]
+  state VARCHAR [not null]                                 // maps to geo_loc_name_state_province_region
```
sidneymbell:
```suggestion
  division VARCHAR [not null]                                 // maps to geo_loc_name_state_province_region
```

sidneymbell: this is what nextstrain uses as a catchall for `state` `province` etc

docs/backend/schema.ddl
```diff
@@ -0,0 +1,342 @@
...
+  country VARCHAR [not null]                               // maps to geo_loc_name_country
+  country_residence VARCHAR                                // maps to host_origin_geo_loc_country
+  country_exposure VARCHAR                                 // maps to location_of_exposure_geo_loc_name_country
+  travel_history VARCHAR                                   // maps to travel_history
```
sidneymbell, line 74: potential pii

docs/backend/schema.ddl
```diff
@@ -0,0 +1,342 @@
...
+  host_scientific_name VARCHAR                             // maps to host_scientific_name
+  host_health_state VARCHAR                                // maps to host_health_state
+  host_health_status_details VARCHAR                       // maps to host_health_status_details
+  host_disease VARCHAR                                     // maps to host_disease
+
+  host_age INT                                             // maps to host_age
+  host_sex ENUM                                            // maps to host_gender
+  symptom_onset_date DATETIME                              // maps to symptom_onset_date
+  signs_and_symptoms VARCHAR                               // maps to signs_and_symptoms
+  exposure_event VARCHAR                                   // maps to exposure_event
```
sidneymbell: potential phi

docs/backend/schema.ddl
```diff
@@ -0,0 +1,342 @@
...
+
+
+///////////////////////////////////////////////////
+// Physical Samples
+
+Enum Sex {
+  Female
+  Male
+  Unknown
+}
```
sidneymbell: potential pii

docs/backend/schema.ddl
```diff
@@ -0,0 +1,342 @@
...
+  workflow_type_id INT [not null, ref: > WorkflowType.id]
+
+  run_date DATETIME [not null]
+  pipeline_versions JSONB [not null]
+}
+
+Table WorkflowInputs {
+  entity_id INT [not null, ref: > Entity.id]
+  workflow_id INT [not null, ref: > Workflow.id]
+
```
sidneymbell, line 130: are we logging the many-to-many relationship of samples <-> workflow_runs?

docs/backend/schema.ddl
```diff
@@ -0,0 +1,342 @@
...
+Table Entity {
+  id INT [pk, increment]
+}
+
+Table WorkflowType {
+  id INT [pk, increment]
+  name VARCHAR [not null, unique]
+}
+
+Table Workflow {
```
sidneymbell, line 119: `workflow_run`?

docs/backend/schema.ddl
```diff
@@ -0,0 +1,342 @@
...
+
+  Indexes {
+    (entity_id, workflow_id) [unique]
+  }
+}
+
+Table WorkflowIntermediateEntities {
+  entity_id INT [not null, ref: > Entity.id]
+  workflow_id INT [not null, ref: > Workflow.id]
+
```
sidneymbell: does this then get intermediate output files attached to it?

docs/backend/schema.ddl
```diff
@@ -0,0 +1,342 @@
...
+Table WorkflowInputs {
+  entity_id INT [not null, ref: > Entity.id]
+  workflow_id INT [not null, ref: > Workflow.id]
+
+  Indexes {
+    (entity_id, workflow_id) [unique]
+  }
+}
+
+Table WorkflowIntermediateEntities {
```
sidneymbell:
```suggestion
Table WorkflowIntermediateOutputs {
```

docs/backend/schema.ddl
```diff
@@ -0,0 +1,342 @@
...
+  Indexes {
+    (submitting_group_id, private_identifier) [unique]
+  }
+}
+
+
+///////////////////////////////////////////////////
+// Entities and Workflows
+
+Table Entity {
```
sidneymbell, line 110:
```suggestion
Table Workflow {
```

docs/backend/schema.ddl
```diff
@@ -0,0 +1,342 @@
...
+///////////////////////////////////////////////////
+// Public accession ids
+
+Table PublicRepository {
+  id INT [pk, increment]
+  name VARCHAR [not null, unique]
+  website VARCHAR [unique]
+}
+
+Table Accession {
```
sidneymbell, line 155: how does this link back to the `private_id`, etc?

docs/backend/schema.ddl
```diff
@@ -0,0 +1,342 @@
...
+    (public_repository_id, public_identifier) [unique]
+  }
+}
+
+
+///////////////////////////////////////////////////
+// Sequences
+
+// describes a sequencer machine  (TODO: what's a good way of canonicalizing this?)
+Table Sequencer {
```
sidneymbell: `Instrument`...? Idk

docs/backend/schema.ddl
sidneymbell: we should 100% be filtering out host reads and discarding them; these should never be stored

docs/backend/schema.ddl
sidneymbell: While this _can_ be user provided if they're uploading a fasta, the vast majority will not have this information. 

If they upload a fastq, we'll compute it ourselves.

docs/backend/schema.ddl
ttung: Done.

docs/backend/schema.ddl
ttung: Dropped.

docs/backend/schema.ddl
ttung: Removed.

docs/backend/schema.ddl
ttung: Commented out. :)

docs/backend/schema.ddl
ttung: Dropped.

docs/backend/schema.ddl
ttung: Clarified in the meeting.

docs/backend/schema.ddl
ttung: Clarified in the meeting.

docs/backend/schema.ddl
ttung: Yes.

jackkamm (approved):

docs/backend/schema.ddl
ttung: Added, thanks!!

PR #32, opened 2021-01-27
Update .gitignore to ignore the .venv-ebcli virtual environment

ttung:
### Description
The changes in #12 suggested a separate virtual environment to install the ebcli because of a conflict between ebcli's requirements and the app's requirements.  This avoids installing the second virtualenv when deploying.

Also, to ensure that we are all using the same [ignore set while deploying](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/eb-cli3-configuration.html#eb-cli3-ebignore), we explicitly do not allow the .ebignore file to be committed.  This means we will all use .gitignore.

### Test plan
deployed successfully with a .venv-ebcli virtualenv in my setup.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -113,8 +113,9 @@ celerybeat.pid
 *.sage.py
 
 # Environments
-.env
-.venv
+.env/
+.venv/
+.venv-ebcli/
 env/
 venv/
 ENV/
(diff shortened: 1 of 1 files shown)


phoenixAja (approved):
thank you!

PR #37, opened 2021-02-01
Finish renaming covidr → aspen

ttung:
### Description
covidr is a dumb name.

### Test plan
pip install -e src/py


phoenixAja (approved):
looks good, i pulled locally and it ran fine for me with these changes, you missed one covidr replacement though [here](https://github.com/chanzuckerberg/aspen/blob/21be8ce477fd67242010e32253ecb19623ad59ea/src/ts/src/NavBar.tsx#L25)

ttung:
I missed two!

PR #38, opened 2021-02-01
[easy] Fix import ordering

ttung:
### Description
Did not configure .isort.cfg correctly.  This now sorts first-party correctly.

### Test plan
Github actions


phoenixAja (approved):
cool, looks good

PR #40, opened 2021-02-01
Update user and group models.

ttung:
### Description
This brings the user and group models in compliance with the schema in #30.

Depends on #30, #42

#### Issue

### Test plan
run migration.

### DIFF
--- a/src/py/aspen/database/models.py
+++ b/src/py/aspen/database/models.py
@@ -20,12 +20,18 @@ meta = MetaData(
 base: DeclarativeMeta = declarative_base(cls=mx.BaseMixin, metadata=meta)
 
 
+########################################################################################
+# groups and users
+
+
 class Group(base):
     """A group of users, generally a department of public health."""
 
     __tablename__ = "groups"
 
     name = Column(String, unique=True, nullable=False)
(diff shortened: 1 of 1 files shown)


src/py/aspen/database/models.py
```diff
@@ -26,6 +26,8 @@ class Group(base):
     __tablename__ = "groups"
 
     name = Column(String, unique=True, nullable=False)
+    gisaid_division_name = Column(String, unique=True)
+    email = Column(String, unique=True, nullable=False)
```
phoenixAja, line 33: wouldn't a better way to do this be to query all users in a group and get the emails from each user record?

phoenixAja: or i guess i don't see the use case of having a group email and user emails.

ttung: So the reason is that public archives seem to want an email address.  In that case, whose email address should we use?  It would seem that a group email address is appropriate here.

phoenixAja: ok makes sense.

phoenixAja (approved):
lg!

PR #42, opened 2021-02-01
Fix alembic tooling

ttung:
### Description
Autogenerate requires the `include_schemas=True` to work.

#### Issue

### Test plan
Generated some migrations.


database_migrations/env.py
```diff
@@ -1,6 +1,8 @@
 import os
 from logging.config import fileConfig
 
+from sqlalchemy import create_engine
```
lvreynoso: looks like this appears twice now?

ttung: Fixed.

lvreynoso (approved):

PR #43, opened 2021-02-01
Models for the physical sample

ttung:
### Description
It's still missing the fields from PHA4GE but I think it's easier to throw them in after people agree to the general concepts.

Depends on #30, #42

#### Issue

### Test plan
Ran migration.

### DIFF
--- a/src/py/aspen/database/models.py
+++ b/src/py/aspen/database/models.py
@@ -1,5 +1,15 @@
 # the main models in the database
-from sqlalchemy import Boolean, Column, ForeignKey, Integer, MetaData, String
+from sqlalchemy import (
+    Boolean,
+    Column,
+    DateTime,
+    ForeignKey,
+    Integer,
+    JSON,
+    MetaData,
+    String,
+    UniqueConstraint,
+)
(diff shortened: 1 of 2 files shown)


src/py/aspen/database/models.py
```diff
@@ -46,3 +55,21 @@ class User(base):
...
+    __table_args__ = (UniqueConstraint("submitting_group_id", "private_identifier"),)
+
+    submitting_group_id = Column(
+        Integer, ForeignKey(f"{Group.__tablename__}.id"), nullable=False
+    )
+    submitting_group = relationship(
+        Group, backref=backref("physical_samples", uselist=True)
+    )
+    private_identifier = Column(String, nullable=False)
+    original_submission = Column(JSON, nullable=False)
```
phoenixAja: oh cool a json field

phoenixAja (approved):
this looks good, i think it captures all the metadata we would need to collect when storing information on physical samples

ttung:
I updated this with the latest round of revisions, along with some annotations on the db columns.

PR #47, opened 2021-02-02
Make enumtables schema aware

ttung:
### Description
This library was unaware of schemas and as such choked whenever it came across a schema.

### Test plan
used in later PR.

PR #46, opened 2021-02-02
Import sqlalchemy-enum-tables

ttung:
### Description
Importing this repo because we need to make changes to it.  I am working to clear this with the open source team @ CZI.

We are importing from https://git.heptacle.fr/neshiraini/sqlalchemy-enum-tables @ f2d832ad78aeac2acd6c0a6e22a6d43049a2c409

### Test plan
none

PR #49, opened 2021-02-02
Support for DataTypes/CanSee

ttung:
### Description
This adds the part of the schema for DataTypes and CanSee.

Depends on #30, #40, #46, #47, #48, #50, #51, #52, #55

#### Issue

### Test plan
1. Ran tests!
2. Created a db from scratch and observed that the "enum"s were created.
3. Created a db from the prior revision, applied the autogenerated script, and observed that the "enum"s were created.

### DIFF
--- a/src/py/aspen/database/models.py
+++ b/src/py/aspen/database/models.py
@@ -1,13 +1,21 @@
+from __future__ import annotations
+
 # the main models in the database
+import enum
+from typing import Type, TYPE_CHECKING, TypeVar
+
+import enumtables
 from sqlalchemy import (
     Boolean,
     Column,
     DateTime,
+    event,
     ForeignKey,
     Integer,
     JSON,
     MetaData,
     String,
+    text,
     UniqueConstraint,
 )
 from sqlalchemy.ext.declarative import declarative_base, DeclarativeMeta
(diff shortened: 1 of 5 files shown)


ttung:
This basically works but python packaging is thwarting me!

src/py/aspen/database/models.py
```diff
@@ -27,14 +35,32 @@
 )
 # typing "base" as DeclarativeMeta is the ugly workaround for
 # https://github.com/python/mypy/issues/2477
-base: DeclarativeMeta = declarative_base(cls=mx.BaseMixin, metadata=meta)
+base: DeclarativeMeta = declarative_base(metadata=meta)
+idbase: DeclarativeMeta = declarative_base(cls=mx.BaseMixin, metadata=meta)
+
+# Mypy gets confused about whether sqlalchemy enum columns are strings or enums, see here:
+# https://github.com/dropbox/sqlalchemy-stubs/issues/114
+# This is the (gross) workaround. Keep an eye on the issue and get rid of it once it's fixed.
```
phoenixAja, line 43: lol

src/py/aspen/database/models.py
```diff
@@ -65,11 +91,70 @@ def __repr__(self):
...
+def _after_datatypes_create(target, connection, **kw):
+    for datatype in DataType:
+        connection.execute(
+            text(
+                f"INSERT INTO {target.schema}.{target.name} (item_id) VALUES (:name)"
+            ).params(name=datatype.value)
+        )
+
+
+event.listen(_DataTypeTable.__table__, "after_create", _after_datatypes_create)
```
phoenixAja, line 121: i'm a little confused on how this is going to be used

ttung: For enum tables, we need to pre-create the enum types in the table.  This is a hook that allows some code to be called after the table is created, and we shove in the enum types into the table at that time.

.hgtags
```diff
@@ -0,0 +1,5 @@
+624b2a967271f0cc24b0f6af19ae73fc4bdbd634 t1
+3bf4e2032f47137600539a08db24b9341c0dc838 t1
+624b2a967271f0cc24b0f6af19ae73fc4bdbd634 t1
+624b2a967271f0cc24b0f6af19ae73fc4bdbd634 t1
+c32aa987429ba54e4c7be9dd679a8a9c76fde661 t1
```
phoenixAja: what is this file for?

ttung: It should be gone now.

src/py/aspen/database/models.py
phoenixAja: so we'll need to create similar listeners for any enum we create in the future?

phoenixAja (approved):

src/py/aspen/database/models.py
ttung: Yes, that's correct.

Actually, you gave me a great idea.  We should fold that logic into the `EnumTable` function.

phoenixAja: yeah that's actually why i was a little confused, i was thinking that we should fold that logic in there instead. No rush though!

ttung: Done in #61

PR #50, opened 2021-02-03
Switch to pipenv for dependency management

ttung:
### Description
Standard requirements.txt files have a few issues with a third party library installed from source:
1. pip does not handle [relative paths](https://github.com/pypa/pip/issues/8765) in a predictable way.  It is relative to current working directory instead of relative to the path of the requirements.txt.  As we cannot constrain that, it leads to unpredictable behavior.
2. `pip freeze` does not generate the correct output when there is a local library.

### Test plan
1. tried to deploying to elastic beanstalk, and it seems to work.
2. ran `make update-deps`
3. ran `make update-deps` with a local library and `Pipfile.lock` references that instead of just a version number.


ttung:
I think `pyproject.toml` is the [future of python packaging](https://www.python.org/dev/peps/pep-0518/), and plausibly can solve these problems, but as [elastic beanstalk](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/python-configuration-requirements.html) wants either requirements.txt or Pipenv.

phoenixAja:
> I think `pyproject.toml` is the [future of python packaging](https://www.python.org/dev/peps/pep-0518/), and plausibly can solve these problems, but as [elastic beanstalk](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/python-configuration-requirements.html) wants either requirements.txt or Pipenv.

interesting, i would personally like to move towards this approach if we end up doing a deploy solution that's not ebs.

phoenixAja (approved):

PR #51, opened 2021-02-03
Fix the test db setup code

ttung:
### Description
This was ported from covidhub, but the constants were not updated.

#### Issue

### Test plan
Used in a separate PR that uses the session object.


phoenixAja (approved):
funny we didn't catch this earlier, we're all probably so used to seeing cliahub stuff it didn't even register

PR #53, opened 2021-02-04
Statistics for consensus genome pipelines

ttung:
### Description
This is attached to the BAM object, as discussed.

#### Issue
[ch68091](https://app.clubhouse.io/genepi/stories/space/68091)

### Test plan
it's a doc. :)


lvreynoso (approved):
Checked in dbdiagram, looked good to me db-wise. I'll leave it to Jack or Sidney let you know if it's what we want

PR #58, opened 2021-02-05
entity and workflows

ttung:
### Description
Created the base models for entities and workflows.  Added the many-to-many tables to connect the entities and workflows.

Depends on #57 

#### Issue

### Test plan
Used `aspen-cli db interact` to make some models and save them to the db.  It seems to produce reasonable results.


erinstargazer (approved):
My only question here is one I asked on another PR about naming conventions but it looks pretty consistent on this one that we are not using underscores for enum tables.

PR #59, opened 2021-02-05
Fix enumtables' table detection

ttung:
### Description
enumtables uses an empty select to detect the presence of tables.  Unfortunately, as postgres wraps this whole process in a transaction, this can lead to situations where the tables are not detected properly.

Here's how it goes wrong:
1. It gathers all the tables that are enums.
2. It tests for the existence of the tables.  The first table check that fails will cause all subsequent operations in the transaction to fail, whether or not the operation is actually doable.
3. A lot of spurious table changes are "detected".

This PR uses the table detection logic alembic internally uses (refactored for our needs), and this produces a changeset that makes sense.

Depends on #47, #48

### Test plan
Autogenerated a schema migration with multiple enumtables.  It produced the migration script.

### DIFF
--- a/third-party/sqlalchemy-enum-tables/enumtables/alembic_autogen.py
+++ b/third-party/sqlalchemy-enum-tables/enumtables/alembic_autogen.py
@@ -1,5 +1,6 @@
 
 import alembic.autogenerate
+from alembic.autogenerate.render import _ident
 from sqlalchemy import MetaData
 
 from . import enum_column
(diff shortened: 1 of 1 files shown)


third-party/sqlalchemy-enum-tables/enumtables/alembic_autogen.py
```diff
@@ -36,12 +30,12 @@ def compare_enums(autogen_context, upgrade_ops, schema_names):
...
 		for table, values in enums.items():
-			fully_qualified_table_name = "%s.%s" % (schema, table) if schema else table
-			if is_table_present(fully_qualified_table_name, autogen_context.connection):
+			if (schema, table) in known_tables:
 				items = {
 					r[0]
 					for r in autogen_context.connection.execute(
-						"SELECT item_id FROM {}".format(fully_qualified_table_name)
+						"SELECT item_id FROM {}".format(
+							"%s.%s" % (schema, table) if schema else table)
```
erinstargazer: It looks like `get_connection_tables` always has schema table. It uses the default schema if one isn't provided. Is there a case where the `if schema else table` would happen where there isn't a schema?

ttung: Yes, you're right.

ttung: Fixed, and used [alembic.autogenerate.render._ident](https://github.com/sqlalchemy/alembic/blob/92b4ed39147aca1d102a2f92bb8dd78580c92c1f/alembic/autogenerate/render.py#L477) to escape identifiers.  cc: @phoenixAja

phoenixAja (approved):

phoenixAja:
it will be really nice to be able to catch enum changes with auto generated migrations 💯

PR #60, opened 2021-02-05
Models for PublicRepository and Accessions

ttung:
### Description
Pretty boiler plate stuff. :)

Depends on #58, #59

#### Issue

### Test plan
Ran the migration.

### DIFF
--- a/src/py/aspen/database/models/accessions.py
+++ b/src/py/aspen/database/models/accessions.py
@@ -1,0 +1,87 @@
+import enum
+
+import enumtables
+from sqlalchemy import (
+    Column,
+    event,
+    ForeignKey,
+    Integer,
+    String,
+    text,
+    UniqueConstraint,
+)
+from sqlalchemy.orm import backref, relationship
+
+from .base import base, idbase
+from .entity import Entity
+from .enum import Enum
+
+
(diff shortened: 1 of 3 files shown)


src/py/aspen/database/models/accessions.py
```diff
@@ -0,0 +1,87 @@
...
+    GISAID = "GISAID"
+    NCBI_SRA = "NCBI_SRA"
+
+
+# Create the enumeration table
+# Pass your enum class and the SQLAlchemy declarative base to enumtables.EnumTable
+_PublicRepositoryTypeTable = enumtables.EnumTable(
+    PublicRepositoryType,
+    base,
+    tablename="publicrepositorytypes",
```
erinstargazer, line 30: What's the naming convention? I would have guessed this would be `public_repository_types`. Are we doing something special for the enum tables with the names?

erinstargazer (approved):
I have a question about the naming conventions but it looks consistent in the code base so I'm approving.

src/py/aspen/database/models/accessions.py
ttung: I can change it.

phoenixAja (approved):

PR #62, opened 2021-02-08
Simplify db metadata schema

sidneymbell:


docs/backend/schema.ddl
```diff
@@ -247,23 +223,25 @@ Table HostFilteredSequence {
...
+--   // optional field for gisaid submission, we'd like to get users to provide this.
+--   sequencing_depth FLOAT
+--
+--   s3_bucket VARCHAR [not null]
+--   s3_key VARCHAR [not null, unique]
+--
+--   Indexes {
+--     (s3_bucket, s3_key) [unique]
+--   }
+-- }
```
ttung: where will the sequencing depth be stored?

sidneymbell:
I'd imagine it fitting neatly as an attribute of the BAM file `output`
entity? Actually, this is a good point -- each pipeline (and therefore each
`output` entity) will have unique quality metrics. Perhaps a flexible JSONB
/ dictionary here? (Can hard-code for now for simplicity)

ttung (changes requested):
I think we need to have the BAM object, even if it's nearly empty.  The changes to PhysicalSample, however, look good to me.

sidneymbell:
Ok, fine by me. At some point I'd like to understand why BAM isn't treated
like any other 'output' entity, but I don't feel particularly
strongly about this one.

docs/backend/schema.ddl
```diff
@@ -5,7 +5,7 @@ Table Group {
   id INT [pk, increment]
   name VARCHAR [not null, unique]
   email VARCHAR [not null, unique]
-  address VARCHAR [not null, unique]
+  address VARCHAR [not null, unique] // required only for repository submission
```
jackkamm: If it's not strictly required, maybe it should be nullable?

docs/backend/schema.ddl
```diff
@@ -44,59 +45,33 @@ Table GroupCanSee {
...
   submitting_group_id INT [not null, ref: > Group.id]
-  private_identifier VARCHAR [not null]
+  private_identifier VARCHAR [not null]  // ideally this is the same as specimen_collector_sample_id noted below, but many DPHs will not be comfortable sharing these
 
   original_submission JSONB [not null]
 
   // This is the public identifier we assign to this sample.  The identifier assigned by the public repositories is
   // stored on the related Accession objects.
-  public_identifer VARCHAR [not null, unique]              // maps to specimen_collector_sample_id and isolate
+  public_identifer VARCHAR [not null, unique]              // maps to isolate, force format to pathogen/country/specimen_collector_sample_id/year, e.g., hCoV-19/USA/CZB-1234/2020
```
jackkamm: Should maybe just be `USA/CZB-1234/2020` (without the `hCov-19` in front) since the prefix differs between GISAID and NCBI -- NCBI uses `SARS-CoV-2/human` as the prefix instead

docs/backend/schema.ddl
```diff
@@ -180,6 +155,7 @@ Table SequencingProtocol {
 }
 
 // describes a single set of sequence data from the sequencer. (FASTQ)
+// these should be replaced by the `HostFilteredSequence`as soon as it is available so we are not storing any host reads
```
jackkamm: For consistency, I would also rename `HostFilteredSequence` to `HostFilteredSequencingReads`

docs/backend/schema.ddl
jackkamm: Agree we should keep this

docs/backend/schema.ddl
```diff
@@ -44,59 +45,33 @@ Table GroupCanSee {
...
-  biomaterial_extracted VARCHAR                            // maps to biomaterial_extracted
-  host_common_name VARCHAR                                 // maps to host_common_name
-  host_scientific_name VARCHAR                             // maps to host_scientific_name
-  host_health_state VARCHAR                                // maps to host_health_state
-  host_health_status_details VARCHAR                       // maps to host_health_status_details
-  host_disease VARCHAR                                     // maps to host_disease
-
-  symptom_onset_date DATETIME                              // maps to symptom_onset_date
-  signs_and_symptoms VARCHAR                               // maps to signs_and_symptoms
-  exposure_event VARCHAR                                   // maps to exposure_event
```
jackkamm, line 99: I support deleting all this stuff :+1:

jackkamm (changes requested):

jackkamm:
> Ok, fine by me. At some point I'd like to understand why BAM isn't treated like any other 'output' entity, but I don't feel particularly strongly about this one.

I may be wrong, but I don't think there is such as a thing as a generic 'output' entity. The workflow output entity is just the superclass, but every output needs to have a more specific type (as far as I understand).

jackkamm:
> Perhaps a flexible JSONB / dictionary here? (Can hard-code for now for simplicity)

I think this is a good idea, actually the stats on the Bam object on current trunk don't totally make sense, which is my fault for being slow to review and providing bad advice -- going to submit a PR to fix it.

ttung:
> I may be wrong, but I don't think there is such as a thing as a generic 'output' entity. The workflow output entity is just the superclass, but every output needs to have a more specific type (as far as I understand).

This is correct.

docs/backend/schema.ddl
ttung: I have done this rename in https://github.com/chanzuckerberg/aspen/pull/79/

docs/backend/schema.ddl
sidneymbell:
```suggestion
  address VARCHAR // required only for repository submission. feasible that there are multiple labs at the same address.
```

src/py/aspen/database/models/sample.py
```diff
@@ -48,6 +50,43 @@ class PhysicalSample(idbase):
         comment="This is the public identifier we assign to this sample.",
     )
 
+    sample_collected_by = Column(
```
phoenixAja, line 51: i feel like since this is already in the Sample table that we don't need these `sample_` prefixes to these columns?

phoenixAja (approved):

docs/backend/schema.ddl
ttung: Added this (and the related migration).

src/py/aspen/database/models/sample.py
ttung: That's the way I had it originally.  @sidneymbell changed it, so maybe she can speak to her thinking on this matter?

ttung:
I'm going to land this for now.  Hopefully @sidneymbell can comment on [this thread](https://github.com/chanzuckerberg/aspen/pull/62#discussion_r578692509), which we can address after merging.

PR #63, opened 2021-02-08
Fix EnumColumn for use in polymorphic types

ttung:
### Description
For polymorphic types, instances are instantiated using the same parameters as `sa.Column`.  Because the pre-existing `EnumColumn` redefines the constructor with different arguments, it does not work.

This PR removes `EnumColumn`.  Users of the enumtables should just call `sa.Column` with the type set to `EnumType(my_python_enum_type)`.  A bonus side effect is that we can remove some of the type checking hacks we introduced earlier in other PRs.

#### Issue

### Test plan
Used in subsequent PRs.


phoenixAja (approved):
interesting, lgtm

PR #71, opened 2021-02-09
Rename enum tables

ttung:
### Description
@neevor suggested we do this in some PR but I forgot to do it before I committed it.

Depends on #60. #72 is needed to autogen the migration script.

#### Issue

### Test plan
upgraded and downgraded from this schema

### DIFF
--- a/database_migrations/versions/20210209_134758_rename_enumtables.py
+++ b/database_migrations/versions/20210209_134758_rename_enumtables.py
@@ -1,0 +1,264 @@
+"""rename enumtables
+
+Revision ID: 20210209_134758
+Revises: 20210205_105007
+Create Date: 2021-02-09 13:47:59.276384
+
+"""
+import enumtables  # noqa: F401
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "20210209_134758"
+down_revision = "20210205_105007"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    ####################################################################################
+    # Create the new types tables.
+    op.create_table(
+        "data_types",
+        sa.Column("item_id", sa.String(), nullable=False),
+        sa.PrimaryKeyConstraint("item_id", name=op.f("pk_data_types")),
+        schema="aspen",
+    )
+    op.create_table(
+        "entity_types",
+        sa.Column("item_id", sa.String(), nullable=False),
+        sa.PrimaryKeyConstraint("item_id", name=op.f("pk_entity_types")),
+        schema="aspen",
+    )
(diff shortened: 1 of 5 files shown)


database_migrations/versions/20210209_134758_rename_enumtables.py
```diff
@@ -0,0 +1,264 @@
...
+Revision ID: 20210209_134758
+Revises: 20210205_105007
+Create Date: 2021-02-09 13:47:59.276384
+
+"""
+import enumtables  # noqa: F401
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
```
phoenixAja, line 12: idt we need this comment

database_migrations/versions/20210209_134758_rename_enumtables.py
```diff
@@ -0,0 +1,264 @@
...
+        referent_schema="aspen",
+    )
+    op.drop_constraint(
+        "fk_public_repository_entity_type_publicrepositorytypes",
+        "public_repository",
+        schema="aspen",
+        type_="foreignkey",
+    )
+    op.create_foreign_key(
+        op.f("fk_public_repository_entity_type_public_repository_types"),
```
phoenixAja, line 111: these are really long names! would be nice to simplify some of these in the future if possible.

database_migrations/versions/20210209_134758_rename_enumtables.py
```diff
@@ -0,0 +1,264 @@
...
+        referent_schema="aspen",
+    )
+    op.drop_constraint(
+        "fk_workflows_workflow_type_workflowtypes",
+        "workflows",
+        schema="aspen",
+        type_="foreignkey",
+    )
+    op.create_foreign_key(
+        op.f("fk_workflows_workflow_type_workflow_types"),
```
phoenixAja, line 126: this is a really confusing/ redundant name as well

phoenixAja: is it common practice to name foreign keys like this? Maybe I'm unfamiliar with how these are usually named.

phoenixAja (approved):

database_migrations/versions/20210209_134758_rename_enumtables.py
ttung: It's part of the template. 🤷

database_migrations/versions/20210209_134758_rename_enumtables.py
ttung: It's the default scheme used by sqlalchemy.

PR #72, opened 2021-02-09
Fix logic for getting the enum table name

ttung:
### Description
Previously, we were declaring enum columns as Enum(Table), but to make it compatible with SqlAlchemy's Enum, we [changed it](https://github.com/chanzuckerberg/aspen/pull/69) to Enum(python_enum_class).  This broke autogenerate for enum.

Hopefully this is the last big change to enumtables for a while, as the API should be stable now and most of the kinks have been worked out.

#### Issue

### Test plan
Autogenerated the migration in #71


phoenixAja (approved):
cool looks good!

PR #77, opened 2021-02-10
Support the Column model for specifying the FK for an enum

ttung:
### Description
It is code-wise saner to specify the FK column with the python object rather than a string name.  When that is done, colspec is a `schemaname.tablename.columnname`.  In that case, we want the second-to-last split.

### Test plan
Used in subsequent PR.

### DIFF
--- a/third-party/sqlalchemy-enum-tables/enumtables/alembic_autogen.py
+++ b/third-party/sqlalchemy-enum-tables/enumtables/alembic_autogen.py
@@ -9,7 +9,11 @@ from . import alembic_ops
 def _get_fk_table_name(column):
 	for fk in column.foreign_keys:
 		colspec = fk._get_colspec()
-		tablename, _ = colspec.split(".")
+		# the colspec is either in the form tablename.columnname or
+		# schemaname.tablename.columnname.
+		splitted = colspec.split(".")
+		assert len(splitted) >= 2
+		tablename = splitted[-2]
 		return tablename
 
 


third-party/sqlalchemy-enum-tables/enumtables/alembic_autogen.py
```diff
@@ -9,8 +9,10 @@
 def _get_fk_table_name(column):
 	for fk in column.foreign_keys:
 		colspec = fk._get_colspec()
-		tablename, _ = colspec.split(".")
-		return tablename
+		splitted = colspec.split(".")
+		assert len(splitted) >= 2
+		# second to the last is the table name.
+		return splitted[-2]
```
phoenixAja: i think remove comment and do this instead
```suggestion
		tablename = splitted[-2]
		return tablename
```

third-party/sqlalchemy-enum-tables/enumtables/alembic_autogen.py
```diff
@@ -9,8 +9,10 @@
 def _get_fk_table_name(column):
 	for fk in column.foreign_keys:
 		colspec = fk._get_colspec()
-		tablename, _ = colspec.split(".")
-		return tablename
+		splitted = colspec.split(".")
+		assert len(splitted) >= 2
```
phoenixAja, line 15: will the length ever be > 2?

phoenixAja (approved):
small questions/comments

third-party/sqlalchemy-enum-tables/enumtables/alembic_autogen.py
ttung: Yes, it can be `schemaname`.`tablename`.`columnname`.

third-party/sqlalchemy-enum-tables/enumtables/alembic_autogen.py
ttung: I have updated this, incorporating some of your suggestion.  I will land this now, but if you still have concerns, I'm happy to revise.

PR #78, opened 2021-02-10
Specify the ForeignKeys not as a string

ttung:
### Description
Strings are fickle and can easily be broken.  TableObject.ColumnObject is more durable.

Depends on #77 

### Test plan
Autogenerated a migration and verified that the migration was empty (this is essentially a no-op).


phoenixAja (approved):
cool this is much nicer

PR #81, opened 2021-02-11
Rework the relationship between workflows and entities.

ttung:
### Description
The key changes is that only one workflow is considered to ever be responsible for creating an entity.  If more than one workflow contributes to an entity, we should have separate entities to represent them.

Added a method to easily traverse between entities and its parents (entities that feed into the workflow that creates the entity in question) and children (entities that result from workflows that consume the entity in question).

### Test plan
used in next PR

### DIFF
--- a/src/py/aspen/database/models/entity.py
+++ b/src/py/aspen/database/models/entity.py
@@ -1,11 +1,30 @@
+from __future__ import annotations
+
 import enum
-from typing import Mapping, Union
+from typing import (
+    Mapping,
+    MutableSequence,
+    Optional,
+    Sequence,
+    Tuple,
+    Type,
+    TYPE_CHECKING,
+    TypeVar,
+    Union,
+)
 
 import enumtables
-from sqlalchemy import Column, ForeignKey
+from sqlalchemy import Column, ForeignKey, Integer
+from sqlalchemy.orm import backref, relationship
 
 from .base import base, idbase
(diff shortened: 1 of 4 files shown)


src/py/aspen/database/models/entity.py
```diff
@@ -1,12 +1,31 @@
...
 import enumtables
-from sqlalchemy import Column, ForeignKey
+from sqlalchemy import Column, ForeignKey, Integer
+from sqlalchemy.orm import backref, relationship
 
 from .base import base, idbase
 from .enum import Enum
 
+if TYPE_CHECKING:
+    from .workflow import Workflow
```
phoenixAja, line 24: why do you need this part?

phoenixAja (approved):

src/py/aspen/database/models/entity.py
ttung: I use it in type annotations, for example [here](https://github.com/chanzuckerberg/aspen/pull/81/files#diff-c25c7ebf669612dcdc0aa054e26ff3d103f4b125ab2b05a48a63a0691357e552R67).  If I don't gate it with `if TYPE_CHECKING`, however, I get recursive imports.

ttung: to be specific, mypy is two-pass, normal python execution is one-pass.

PR #83, opened 2021-02-11
Auth0 secret store

phoenixAja:
### Description

instead of a .env file use the aws secret store instead

#### Issue
[ch68841](https://app.clubhouse.io/genepi/stories/space/68841)

### Test plan

deployed app with new changes and hit authentication endpoints

### DIFF
--- a/src/py/aspen/config/config.py
+++ b/src/py/aspen/config/config.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import json
 import os
 from collections import MutableMapping
+from functools import lru_cache
 from typing import Type
 
 import boto3
@@ -45,6 +46,7 @@ class Config:
 
 class Auth0Config:
     @property
+    @lru_cache()
     def AWS_SECRET(self):
 
         secret_name = os.environ.get("SECRET_NAME")
--- a/src/py/aspen/config/config.py
+++ b/src/py/aspen/config/config.py
@@ -45,7 +45,7 @@ class Config:
 
 
 class Auth0Config:
-    @property
+    @property  # type: ignore
     @lru_cache()
     def AWS_SECRET(self):
 
--- a/src/py/aspen/config/config.py
+++ b/src/py/aspen/config/config.py
@@ -4,7 +4,7 @@ import json
 import os
 from collections import MutableMapping
 from functools import lru_cache
-from typing import Type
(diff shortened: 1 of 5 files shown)


src/py/aspen/config/config.py
```diff
@@ -42,26 +44,38 @@ def AUTH0_CONFIG(self):
...
+        session = boto3.session.Session()
+        client = session.client(service_name="secretsmanager", region_name=region_name)
+
+        try:
+            get_secret_value_response = client.get_secret_value(SecretId=secret_name)
+        except ClientError as e:
+            raise e
+        else:
+            secret = get_secret_value_response["SecretString"]
+            return json.loads(secret)
```
ttung: My guess is that you'll want to cache this value, otherwise you'll keep hitting `secretsmanager`

ttung (changes requested):

src/py/aspen/config/config.py
ttung: The other thing to be aware of is that these values are slurped into the flask app at startup, and we can't really change them.

phoenixAja: good points... I'll have to think about how to address the second point

phoenixAja: @ttung what do you think we should do about the second point? I think this stuff will stay fairly static or the most part, but do you think there's some stuff we should move out of the config?

ttung: For the second point, I think one possible strategy is just to require that we restart the web services whenever a secret changes.

ttung: A better strategy might be to fetch the secrets once per web request, and only cache it for the duration of the request.

phoenixAja: hmmm i'm a little confused, so you think we should decouple auth0 from the config? otherwise these values will be stored in the flask app indefinitely  like what you were saying before, and we won't be able to store these values for only the length of the request (which i think is fine here since these values are more about being able to connect to the auth0 backend instead of holding auth data for a logged on user)

phoenixAja: i added `lru_cache` decorator to AWS_SECRET which will prevent us from hitting the secret store repeatedly

ttung (approved):

PR #85, opened 2021-02-12
Normalize the singular/plural

ttung:
### Description
1. `processed_gisaid_dump` should be `processed_gisaid_dumps` because it returns a sequence.
2. Make all the python entities singular (i.e., `SequencingRead`).

#### Issue

### Test plan
GA

### DIFF
--- a/database_migrations/versions/20210222_220412_rename_sequencing_reads_tables.py
+++ b/database_migrations/versions/20210222_220412_rename_sequencing_reads_tables.py
@@ -1,0 +1,89 @@
+"""rename *sequencing_reads tables
+
+Revision ID: 20210222_220412
+Revises: 20210218_164249
+Create Date: 2021-02-22 22:04:14.173048
+
+"""
+import enumtables  # noqa: F401
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "20210222_220412"
+down_revision = "20210218_164249"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.rename_table(
+        "sequencing_reads",
+        "sequencing_read_collections",
+        schema="aspen",
+    )
+    op.rename_table(
+        "host_filtered_sequencing_reads",
+        "host_filtered_sequencing_read_collections",
+        schema="aspen",
+    )
+    op.drop_constraint(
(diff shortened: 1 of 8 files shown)


ttung:
The only part of this that is potentially questionable is `SequencingRead`.  I know it's really multiple reads, but keeping a consistent pluralization is helpful for devs.  @jackkamm  is this a showstopper? :)

erinstargazer (commented):
I think `SequencingRead` also needs to be updated in the `backend/schema.ddl` doc.

jackkamm:
`SequencingRead` does read really strange to me. A better choice might be one of `SequencingReadsCollection`, `SequencingReadsPool`, or `SequencingReadsLibrary`. I'm not 100% confident of the terminology, someone with a stronger molecular background should comment, maybe @danrlu ?

erinstargazer:
> `SequencingRead` does read really strange to me. A better choice might be one of `SequencingReadsCollection`, `SequencingReadsPool`, or `SequencingReadsLibrary`. I'm not 100% confident of the terminology, someone with a stronger molecular background should comment, maybe @danrlu ?

That's a good point. The current suggestions do sounds a bit odd as well (although no as much). It's really more along the lines of a sequencing run produces a fasta that has the reads. So something along the lines of `SequencingRun` (although that doesn't really capture that this is specific to reads) or `SequencingFasta` or even `ReadsFromSequeneRun` (very verbose) seems to match reality more without adding a modifier like `Pool` (I personally don't feel that adds clarification), `Library` (you can have multiple runs with the same library so this also seems a bit confusing) or `Collection` (this is the clearest of these from my perspective).

danrlu:
I agree that `Pool` and `Library` means specific things in sequencing. `SequencingRead` sounds good to me, `SequencingFastq` or the good old `SampleFastq` too (me non-native speaker alert!). The actual table name is `sequencing_reads` + consistent singular in py code should be least confusing to ppl.

ttung:
I renamed it to `SequencingReadCollection` and `HostFilteredSequencingReadCollection`.  It's a mouthful, but hopefully we're not mucking around there too often?  :)

Added a migration, and updated `schema.ddl`.

phoenixAja (approved):

PR #86, opened 2021-02-12
Accessors to go between SequencingRead <-> HostFilteredSequencingRead

ttung:
### Description
This makes it easier to walk between objects.

Depends on #85 

#### Issue

### Test plan
None.

### DIFF
--- a/src/py/aspen/database/models/host_filtering.py
+++ b/src/py/aspen/database/models/host_filtering.py
@@ -1,7 +1,14 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
 from sqlalchemy import Column, ForeignKey, Integer, String, UniqueConstraint
 
 from .entity import Entity, EntityType
 from .workflow import Workflow, WorkflowType
+
+if TYPE_CHECKING:
+    from .sequences import SequencingReadCollection
 
 
 class HostFilteredSequencingReadCollection(Entity):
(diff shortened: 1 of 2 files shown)


phoenixAja (approved):

src/py/aspen/database/models/host_filtering.py
```diff
@@ -14,6 +21,16 @@ class HostFilteredSequencingRead(Entity):
     s3_bucket = Column(String, nullable=False)
     s3_key = Column(String, nullable=False)
 
+    @property
+    def sequencing_read(self) -> SequencingRead:
+        """The raw gisaid dump this processed gisaid dump was generated from."""
+        # this import has to be here for circular dependencies reasons. :(
+        from .sequences import SequencingRead
```
erinstargazer: Is this expensive to import or does it cause a loop (and if it does create a loop why isn't that a problem when `TYPE_CHECKING` is set)? I'm curious about the choice to only import at the top of the file if `TYPE_CHECKING` is true, but then import it here since it's used by this property as opposed to just importing that at the top level as usual.

erinstargazer (approved):
Just had a clarification question but otherwise looks good.

src/py/aspen/database/models/host_filtering.py
ttung: It does cause a loop.  mypy is a two-pass compiler, so it doesn't have loop issues, but cpython is one-pass.

PR #76, opened 2021-02-10
Add models for SequencingReads and *PathogenGenomes

ttung:
### Description
Enums for sequencing instrument and sequencing protocol.  Entities for the sequencing reads and pathogen genomes.

Added a test.

Depends on #74, #75

#### Issue
[ch64743](https://app.clubhouse.io/genepi/stories/space/64743)

### Test plan
ran test!

PR #88, opened 2021-02-12
Add documentation for pathogen genome statistics

jackkamm:
At request of https://github.com/chanzuckerberg/aspen/pull/76#discussion_r574857286, here is documentation for fields in PathogenGenome, in the form of comments.

As an aside, I think `num_n` might be clearer as `num_missing_alleles`, but don't have the time to run a proper migration now.


ttung:
I folded the comments into the sql comment field so any layman checking out the db will understand.  I also renamed `num_n` to `num_missing_alleles` and created a migration script.

ttung (approved):

ttung:
For testing I ran the migration script forwards and backwards.

PR #90, opened 2021-02-18
adding usergroup view

phoenixAja:
### Description

adding view to grab user info from session and return group data

#### Issue
[ch118684](https://app.clubhouse.io/genepi/stories/space/118684)

### Test plan
hit endpoint url (http://localhost:3000/api/usergroup) and verified json was returned correctly, i added a test group and a test user (with `aspen-cli db interact`)
also added some unit tests for this endpoint


src/py/aspen/app.py
```diff
@@ -85,3 +90,19 @@ def logout():
...
+
+@requires_auth
+@application.route("/api/usergroup", methods=["GET"])
+def usergroup():
+    with session_scope(db_interface) as db_session:
+        # since authentication is required to access view this information is guaranteed to be in session
+        profile = session["profile"]
+        user = (
+            db_session.query(User)
+            .options(joinedload(User.group))
```
ttung, line 107: yay :)

src/py/aspen/app.py
```diff
@@ -85,3 +90,19 @@ def logout():
...
+            .filter(User.auth0_user_id == profile["user_id"])
+            .one_or_none()
```
ttung: why is this `one_or_none()`?

src/py/aspen/app.py
phoenixAja: learned it from reviewing your code!

src/py/aspen/app.py
phoenixAja: good point, it will always be one otherwise this would not work at all

phoenixAja: fixed

phoenixAja:
weird `test_alembic_head.py` test is failing when running `DB=dev alembic upgrade HEAD` locally i get this: 

> ERROR [alembic.util.messaging] Can't locate revision identified by '20214716_114729'
>   FAILED: Can't locate revision identified by '20214716_114729'

@ttung am i doing something wrong or did we miss a migration?

ttung:
Where are you seeing this error?

phoenixAja:
@ttung [here](https://github.com/chanzuckerberg/aspen/runs/1938061503) and when running the tests locally 😕

phoenixAja:
i got that other error from running `DB=dev alembic upgrade HEAD`

ttung:
The error you see in github actions is fixed with https://github.com/chanzuckerberg/aspen/pull/97.

phoenixAja:
> The error you see in github actions is fixed with #97.

ok cool, just approved thanks!!

ttung:
The other error is because I messed up the migration's name, which was fixed in #91.  If you know what version of the database you have (`alembic current`), you can probably stamp it with the corrected name (`alembic stamp ...`).

.github/workflows/py-tests.yml
```diff
@@ -74,5 +74,7 @@ jobs:
         run: |
           pip install -e src/py
       - name: Run with pytest
+        env:
+          SECRET_KEY: secret_key
```
ttung: I would assume that you wouldn't want this to be defined in the github workflow, because that implies that someone running the test standalone won't have this.

Makefile
```diff
@@ -88,3 +88,12 @@ update-deps :
...
 ##############################################################
+
+### TESTTING #############################################
+#
+
+unit-tests:
+	pytest src/py/aspen
+
+#
+##############################################################
```
ttung, line 99: 👍 

might even just make the target `test`.

src/py/aspen/test_models/fixtures.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+
+from ..database.models.usergroup import Group, User
```
ttung:
```suggestion
from aspen.database.models.usergroup import Group, User
```

src/py/aspen/test_app/fixtures.py
```diff
@@ -0,0 +1,19 @@
+import pytest
+
+from ..app import application
+from ..config.testing import TestingConfig
```
ttung:
```suggestion
from aspen.app import application
from aspen.config.testing import TestingConfig
```

src/py/aspen/database/models/mixins.py
```diff
@@ -8,3 +10,18 @@ class BaseMixin:
     """Base model: all models have integer primary keys"""
 
     id = Column(Integer, primary_key=True, autoincrement=True)
+
+
+class DictMixin:
+    def to_dict(self):
+        d = {}
+        for column in self.__table__.columns:
+            if isinstance(
```
ttung: i would hoist `getattr(self, column.name)` out of this if-else block, i.e.,

```
column_value = getattr(self, column.name)
if ...:
   ...
else:
   ...
```

src/py/aspen/database/connection.py
```diff
@@ -38,7 +36,8 @@ def session_scope(interface: SqlAlchemyInterface):
         session.close()
 
 
-def get_db_uri(runtime_config: Config, readonly: bool = False) -> str:
+# TODO: add back in Config type hint and resolve circular dependency
+def get_db_uri(runtime_config, readonly: bool = False) -> str:
```
ttung: I'm guessing you can do:

```
from __future__ import annotations

if TYPE_CHECKING:
    from aspen.config.config import Config
```

and put back the type annotation.

src/py/aspen/test_models/fixtures.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+@pytest.fixture(scope="function")
+def group(session):
+    group = Group(name="groupname", email="groupemail", address="123 Main St")
+    session.add(group)
+    session.commit()
```
ttung, line 10: Strictly speaking, this can be a `session.flush()`.

src/py/aspen/test_models/fixtures.py
```diff
@@ -0,0 +1,26 @@
...
+    user = User(
+        name="test",
+        auth0_user_id="test_auth0_id",
+        email="test_user@dph.org",
+        group_admin=True,
+        system_admin=True,
+        group=group,
+    )
+    session.add(user)
+    session.commit()
```
ttung, line 25: ditto.

src/py/aspen/test_models/fixtures.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+
+from ..database.models.usergroup import Group, User
+
+
+@pytest.fixture(scope="function")
+def group(session):
```
ttung:
```suggestion
def group(session) -> Group:
```

src/py/aspen/test_models/fixtures.py
```diff
@@ -0,0 +1,26 @@
...
+    group = Group(name="groupname", email="groupemail", address="123 Main St")
+    session.add(group)
+    session.commit()
+    return group
+
+
+@pytest.fixture(scope="function")
+def user(session, group):
```
ttung:
```suggestion
def user(session, group) -> User:
```

src/py/aspen/tests/test_views.py
```diff
@@ -0,0 +1,16 @@
+import json
+
+
+def test_usergroup_with_session_profile(session, app, client, user, group):
+    with client.session_transaction() as sess:
+        sess["profile"] = {"name": "test", "user_id": "test_auth0_id"}
+    res = client.get("/api/usergroup")
+    expected = {"user": user.to_dict(), "group": group.to_dict()}
+    assert expected == json.loads(res.get_data(as_text=True))
```
ttung, line 9: interesting that this doesn't have to be in the `session_transaction` scope.

src/py/aspen/app.py
```diff
@@ -4,32 +4,47 @@
...
+flask_env = os.environ.get("FLASK_ENV")
+if flask_env == "production":
+    application.config.from_object(ProductionConfig())
+elif flask_env == "development":
     application.config.from_object(DevelopmentConfig())
-if os.environ.get("FLASK_ENV") == "staging":
+elif flask_env == "staging":
     application.config.from_object(StagingConfig())
+else:
+    application.config.from_object(TestingConfig())
```
ttung: Is this still necessary?  L36 should prevent the auth0 from initializing, and you are setting up the TestingConfig object in the app fixture.

src/py/aspen/config/config.py
```diff
@@ -111,3 +111,7 @@ def URI(self):
     @property
     def READONLY_URI(self):
         raise NotImplementedError()
+
+    @property
+    def INTERFACE(self):
+        raise NotImplementedError()
```
ttung, line 117: What's the rationale behind making the Config object produce the `SqlAlchemyInterface` object?

ttung (changes requested):

.github/workflows/py-tests.yml
phoenixAja: we needed this previously but i removed this since we're not getting the secret_key from the enironment anymore (needed previously to establish test client)

Makefile
phoenixAja: true, i was used to using `unit-tests` target with covidhub

src/py/aspen/database/connection.py
phoenixAja: oh that's right!! Forgot about that work around

src/py/aspen/test_models/fixtures.py
phoenixAja: why is that better than doing `session.commit()`?

src/py/aspen/app.py
phoenixAja: yeah probs not actually

src/py/aspen/test_models/fixtures.py
ttung: As a general practice, I just use `flush()` unless I actually want it persisted to the database.

Makefile
ttung: Ah ok, either is fine w/ me.

src/py/aspen/config/config.py
phoenixAja: i thought this was a cleaner way to have an interface rather than declaring a global from inside app.py. Especially because the interface depends on attributes already declared within the config. I thought it was also nice with testing that you could create the application from the config directly without also creating an interface fixture object that would have to be shared across fixtures. I'm open to changing it though.

src/py/aspen/tests/test_views.py
phoenixAja: yeah it is interesting, this actually won't work if you add it to the session_transaction scope

ttung: That is even more bizarre!

phoenixAja: yeah.. the flask session/ request scope is V interesting, especially when comparing to other frameworks

src/py/aspen/database/models/mixins.py
phoenixAja: yeah that's much cleaner

phoenixAja: done

src/py/aspen/database/connection.py
phoenixAja: added

src/py/aspen/test_models/fixtures.py
phoenixAja: so when i do flush the tests break,  so i added the commit back in. i think since the test client is establishing a new session scope in the usergroup view that this needs to be committed?

src/py/aspen/app.py
phoenixAja: taken out

src/py/aspen/config/config.py
ttung: I have a preference for keeping the config object to just config things.  I don't mind a new global.  We may have to do more work in terms of using scoped sessions, so I filed [a ticket](https://app.clubhouse.io/genepi/story/123072/use-scoped-sessions-for-db-connections-in-the-app) for that.

src/py/aspen/test_models/fixtures.py
ttung: Ah! Yes, that is a great reason to do the commit.

src/py/aspen/config/config.py
phoenixAja: i think i'm already using a scoped session in the usergroup view though

phoenixAja: oh so also i think that creating a global would be problematic for testing because we're not creating the application from the testing config in app.py, so if we initialize the interface globally in app.py then we would need different logic for also first instantiating the application from the TestConfig (which is problematic becuase the DB_URI isn't stable in the testing environment).

ttung: Go ahead and merge this as is then.  We can think of a better way to handle this in a future PR.

ttung (approved):

src/py/aspen/config/config.py
phoenixAja: yayyy!!

PR #92, opened 2021-02-18
Add more types of data that access can be granted to.

ttung:
### Description
This is originally from #62.

### Test plan
YOLO

### DIFF
--- a/docs/backend/schema.ddl
+++ b/docs/backend/schema.ddl
@@ -26,7 +26,8 @@ Table TypeOfData {
   // Types of data people can view.  This table should be pre-populated with at least the following entries:
   //  1. Trees
   //  2. Sequences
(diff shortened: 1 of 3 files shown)


sidneymbell (approved):
Thank you, Tony! LGTM.

docs/backend/schema.ddl
```diff
@@ -26,7 +26,8 @@ Table TypeOfData {
   // Types of data people can view.  This table should be pre-populated with at least the following entries:
   //  1. Trees
   //  2. Sequences
-  //  3. Metadata
+  //  3. Other metadata
```
phoenixAja: seems like this is wrong based on the enum values

ttung: Clarified.

PR #93, opened 2021-02-18
Update stylesheets to follow design system

lvreynoso:
### Description

Adds typography, colors, and spacing SASS (not CSS) variables to match the design system.

#### Issue

### Test plan

They're SASS variables, so testing can only be done by building on top of them.


src/ts/src/common/styles/colors.scss
```diff
@@ -0,0 +1,61 @@
...
+// A scale with these colors
+// #482778
+// #1F968B
+// #55C567
+// #BF1464
+// #E58740
+// #AB4ECC
+
+// Sequential Colors; refer to design system
+// or to D3's "YlOrRd" scale
```
phoenixAja, line 61: are we eventually going to need to implement this?

src/ts/src/common/styles/typography.scss
```diff
@@ -0,0 +1,147 @@
+@import url("https://fonts.googleapis.com/css2?family=Open+Sans:ital,wght@0,300;0,400;0,600;0,700;0,800;1,300;1,400;1,600;1,700;1,800&display=swap");
+
+// font families
+$font-family-primary: "Open Sans", "Helvetica Neue", Arial, Helvetica,
```
phoenixAja, line 4: is it ok that only some of these in quotes?

lvreynoso: Yeah, they're in quotes because there is a space in their name. The font and variable should look fine on your browser, let me know if this isn't the case!

src/ts/src/common/styles/colors.scss
lvreynoso: I'm not sure. I don't know if it would be better to implement these in CSS or React since they are supposed to be continuous scales. I think we can wait until they are needed

phoenixAja (approved):

PR #95, opened 2021-02-19
Use date instead of datetime

ttung:
### Description
datetime is probably too much precision, and introduces wonky timezone issues.  Simplify by changing them to `DATE`.

Depends on #62 

### Test plan
Run migration, check the db.


phoenixAja (approved):
i agree datetime is too intense haha

PR #99, opened 2021-02-19
Fix accessions, sequencing_reads, and uploaded_pathogen_genomes

ttung:
### Description
Added a migration to handle the change.

1. There should only be one UploadedPathogenGenome _OR_ SequencingReads per Sample.
2. There can be more than one accession per sample id.

Depends on #110 

### Test plan
ran migration forwards and backwards.

### DIFF
--- a/src/py/aspen/database/models/sample.py
+++ b/src/py/aspen/database/models/sample.py
@@ -1,3 +1,7 @@
+from __future__ import annotations
+
+from typing import Optional, TYPE_CHECKING
+
 from sqlalchemy import (
     Column,
     Date,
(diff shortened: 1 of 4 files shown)


src/py/aspen/database/models/sample.py
```diff
@@ -14,6 +18,9 @@
 from .base import idbase
 from .usergroup import Group
 
+if TYPE_CHECKING:
+    from .sequences import SequencingReads, UploadedPathogenGenome
```
erinstargazer: What happens if `TYPE_CHECKING` is not true when `SequencingReads` is used on line 167?

erinstargazer: I'm guessing it just ignores the entire line where it's used as it looks like it's just a type specification, is that right?

ttung: When you do `from __future__ import annotations`, it ignores type annotations when not type checking.

shanaxel42 (approved):
looks good!

PR #100, opened 2021-02-19
script to import from covidhub

ttung:
### Description
This script allows for an individual project to be imported into the aspen database.

### Test plan
Ran `AWS_PROFILE=biohub aspen-cli db import-covidhub-project --rr-project-id RR074e --covidhub-db-secret cliahub/cliahub_rds_read_prod`

### DIFF
--- a/src/py/aspen/covidhub_import/implementation.py
+++ b/src/py/aspen/covidhub_import/implementation.py
@@ -1,0 +1,136 @@
+import warnings
+from typing import Iterable, Mapping
+
+from covid_database import init_db as Cinit_db
+from covid_database import SqlAlchemyInterface as CSqlAlchemyInterface
+from covid_database import util as Cutil
+from covid_database.models.enums import ConsensusGenomeStatus
+from covid_database.models.ngs_sample_tracking import (
+    ConsensusGenome,
+    CZBID,
+    DphCZBID,
+    Project,
+    SampleFastqs,
+)
+from sqlalchemy import and_, or_
+from sqlalchemy.orm import joinedload, Session
+
+from aspen.database.connection import session_scope, SqlAlchemyInterface
+from aspen.database.models import Group, Sample, UploadedPathogenGenome
+
+
+def covidhub_interface_from_secret(secret_id: str) -> CSqlAlchemyInterface:
+    interface = Cinit_db(Cutil.get_db_uri(secret_id))
+    return interface
+
+
+def import_project(
+    interface: SqlAlchemyInterface,
+    covidhub_secret_id: str,
+    rr_project_id: str,
+):
+    covidhub_interface = covidhub_interface_from_secret(covidhub_secret_id)
+    covidhub_session: Session = covidhub_interface.make_session()
+
+    with session_scope(interface) as session:
+        project = (
+            covidhub_session.query(Project)
+            .filter(Project.rr_project_id == rr_project_id)
+            .one()
+        )
+
+        group = (
+            session.query(Group)
+            .filter(Group.name == project.originating_lab)
+            .one_or_none()
+        )
+        if group is None:
+            # copy the project info into the "group"
+            group = Group(
+                name=project.originating_lab,
+                address=project.originating_address,
+                # FIXME: this needs to be updated.
+                email="FAKE_EMAIL@gmail.com",
+            )
+            session.add(group)
+
+        consensus_genomes: Iterable[ConsensusGenome] = (
+            covidhub_session.query(ConsensusGenome)
+            .join(SampleFastqs, CZBID, Project)
+            .distinct(CZBID.id)
+            .filter(Project.rr_project_id == rr_project_id)
+            .filter(
+                or_(
+                    ConsensusGenome.status == ConsensusGenomeStatus.PASSED_AUTO,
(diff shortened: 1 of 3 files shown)


src/py/aspen/covidhub_import/implementation.py
```diff
@@ -0,0 +1,139 @@
...
+                consensus_genome.recovered_sites
+            )
+            sample.uploaded_pathogen_genome.num_mixed = consensus_genome.ambiguous_sites
+            sample.uploaded_pathogen_genome.num_missing_alleles = (
+                consensus_genome.missing_sites
+            )
+            sample.uploaded_pathogen_genome.sequencing_depth = (
+                consensus_genome.avg_depth
+            )
+            sample.uploaded_pathogen_genome.sequence = consensus_genome.fasta
```
ttung: this is where I could really use your help, @jackkamm

src/py/aspen/covidhub_import/implementation.py
```diff
@@ -0,0 +1,137 @@
...
```
phoenixAja, line 136: we don't need to do `session.add(sample)` at the end here?

phoenixAja (approved):
very nice!

src/py/aspen/covidhub_import/implementation.py
ttung: we only need to do that for new objects that are unlinked to any other objects.

src/py/aspen/covidhub_import/implementation.py
```diff
@@ -0,0 +1,136 @@
...
+                    submitting_group=group, private_identifier=external_accession
+                )
+
+            sample.original_submission = {}
+            sample.public_identifier = (
+                f"USA/{czbid.submission_base}/{czbid.collection_date.year}"
+            )
+            sample.sample_collected_by = project.originating_lab
+            sample.sample_collector_contact_address = project.originating_address
+            sample.collection_date = czbid.collection_date
```
jackkamm, line 116: `collection_date` doesn't exist for `InternalCZBID`. I see you're outputting a warning for non-DPH samples, but it isn't clear to me whether you expect internal CZBIDs here. If you really need it, there's some complicated function to grab the date for internal czbids, though note it is bugged for samples that were PCR'd twice::
https://github.com/czbiohub/covidhub/issues/1485

src/py/aspen/covidhub_import/implementation.py
```diff
@@ -0,0 +1,136 @@
...
+            sample.location = project.location
```
jackkamm, line 117: Note, this can be null, for example the recent VRDL samples. However, I see sample.location is `[not null]` in `schema.ddl`.

src/py/aspen/covidhub_import/implementation.py
jackkamm: I made a couple separate comment threads for `collection_date` and `location`. Otherwise it looks good to me.

src/py/aspen/covidhub_import/implementation.py
```diff
@@ -0,0 +1,136 @@
...
+            session.query(Group)
+            .filter(Group.name == project.originating_lab)
+            .one_or_none()
+        )
+        if group is None:
+            # copy the project info into the "group"
+            group = Group(
+                name=project.originating_lab,
+                address=project.originating_address,
+                # FIXME: this needs to be updated.
```
erinstargazer, line 52: Is this going to be fixed later or did you mean to update this in the PR? If this is something we are going to update/fix later it might be good to update the note.

src/py/aspen/covidhub_import/implementation.py
```diff
@@ -0,0 +1,136 @@
...
+                joinedload(ConsensusGenome.sample_fastqs)
+                .joinedload(SampleFastqs.czb_id)
+                .joinedload(CZBID.genome_submission_info),
+            )
+        )
+
+        external_accessions = {
+            consensus_genome.sample_fastqs.czb_id.external_accession
+            for consensus_genome in consensus_genomes
+            if isinstance(consensus_genome.sample_fastqs.czb_id, DphCZBID)
```
erinstargazer, line 83: Maybe this is getting filtered out at a different step but it seems odd to import samples from collaborators but not include the external accession. Is the idea that we just won't specify any project ids that aren't or public DPH projects when running this script?

src/py/aspen/covidhub_import/implementation.py
ttung: I think for Orange County, it's all DphCZBID.  I'll resolve when we get to that.

ttung: https://app.clubhouse.io/genepi/story/123614/importing-covidhub-data-does-not-work-for-non-dphczbid

src/py/aspen/covidhub_import/implementation.py
ttung: https://app.clubhouse.io/genepi/story/123615/how-to-manage-vrdl-samples-in-aspen

src/py/aspen/covidhub_import/implementation.py
ttung: https://app.clubhouse.io/genepi/story/123616/covidhub-importer-does-not-provide-a-sane-email-address-for-dph

PR #102, opened 2021-02-23
Allow for multiple staging environments

ttung:
### Description
Creates an easy-to-setup staging environment.  This allows developers to create arbitrary staging environments without setting up a new secret.  The callback URL is determined by extracting the Elastic Beanstalk environment name.  Note that this is actually dependent on the cname, but if all the instructions are followed, the cname should match the environment name.

Also updated docs to reflect the new setup.

### Test plan
1. Created and deployed to an AWS Elastic Beanstalk environment.
2. Run locally.

### DIFF
--- a/src/py/aspen/config/config.py
+++ b/src/py/aspen/config/config.py
@@ -6,8 +6,9 @@ from collections import MutableMapping
 from functools import lru_cache
 from typing import Any, Mapping, Type
 
-import boto3
 from botocore.exceptions import ClientError
+
+from aspen import aws
 
 
 class Config:
@@ -29,7 +30,7 @@ class Config:
 
     @property
     def SECRET_KEY(self):
-        return os.environ.get("SECRET_KEY")
+        raise NotImplementedError()
 
     @property
     def TESTING(self):
@@ -41,7 +42,7 @@ class Config:
 
     @property
     def AUTH0_CONFIG(self):
-        return Auth0Config()
+        return NotImplementedError()
 
 
 class Auth0Config:
@@ -50,12 +51,10 @@ class Auth0Config:
     # https://github.com/python/mypy/issues/1362
     @lru_cache()
     def _AWS_SECRET(self) -> Mapping[str, Any]:
+        session = aws.session()
 
-        secret_name = os.environ.get("SECRET_NAME")
-        region_name = os.environ.get("AWS_REGION")
-
-        session = boto3.session.Session()
-        client = session.client(service_name="secretsmanager", region_name=region_name)
+        secret_name = os.environ.get("AUTH0_CONFIG_SECRET_NAME", "aspen-auth0")
+        client = session.client(service_name="secretsmanager")
 
         try:
             get_secret_value_response = client.get_secret_value(SecretId=secret_name)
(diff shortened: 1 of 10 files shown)


src/py/aspen/aws/session.py
```diff
@@ -0,0 +1,16 @@
+from typing import Optional
+
+from boto3 import Session
+
+from .region import region
+
+_session: Optional[Session] = None
+
+
+def session() -> Session:
```
phoenixAja: since we have many different sessions (one for db, one for app, and now one for aws) i wonder if we should rename to something more specific? Like `aws_session`? Interested to hear what you think

src/py/aspen/config/config.py
```diff
@@ -29,7 +30,16 @@ def DEBUG(self):
 
     @property
     def SECRET_KEY(self):
-        return os.environ.get("SECRET_KEY")
+        # TODO: This probably needs to be sourced from secrets.  Somehow.  Who knows.
```
phoenixAja: yeah that's a good idea to store in secrets, but then we would need an auth0 secret and an app secret, unless we create a mega secret. this definitely should be more secure for a production deploy

ttung: yeah, let me fix this.  I don't think we should be committing this as is.

phoenixAja: i think we can keep the env variable option for now, or maybe have a randomly generated string?

src/py/aspen/config/config.py
```diff
@@ -29,7 +30,16 @@ def DEBUG(self):
...
+        return "abcabcabc"
+
+    @property
+    def SESSION_TYPE(self):
+        # TODO: This needs to be updated to something more sane to persist sessions
+        # across restarts.
+        # https://flask-session.readthedocs.io/en/latest/#sqlalchemysessioninterface is
+        # probably a sane option.
+        return "filesystem"
```
phoenixAja: this is interesting. Do we need to persist sessions across restarts? I think if something like that happens sqlalchemy would automatically role back the session right?

src/py/aspen/aws/session.py
ttung: I used it as `aws.session()` in all the places to make it explicit.   Google's python style guide has some [recommendations](https://google.github.io/styleguide/pyguide.html#224-decision), but I don't know how to enforce that.  We'll just have to do it by convention...

src/py/aspen/config/config.py
ttung: No, it wouldn't.  It'd force people to hit auth0 again, I believe.

src/py/aspen/config/config.py
ttung: ok i'm setting it to a random value for now.  I'll clean this up in a subsequent PR.

src/py/aspen/config/config.py
ttung: Ok so it looks like we don't need this.  From my testing, having a random value for `SECRET_KEY` is not great because it forces people to return to auth0, so we will want some persistent value for that.

phoenixAja (approved):

PR #104, opened 2021-02-24
Fix installation for aspen on elastic beanstalk

ttung:
### Description
1. Don't need `requirements.txt`, `Pipfile.lock` is sufficient.
2. Add `sqlalchemy-enum-tables` to the path.  The backstory is that the `.egg-link` is installed, but the link is to the staging directory, and not the deployment directory.  Once the deployment is complete, the staging directory is removed, and as such, the link is to a non-existent directory.

### Test plan
Deployed some code based on these changes to elastic beanstalk, and it seems to work.  Without these changes, `sqlalchemy-enum-tables` is not found.


phoenixAja (approved):
wierd

PR #106, opened 2021-02-24
Make logo size explicit

lvreynoso:
### Description

Fixes a bug where the logo does not appear on some browsers.

#### Issue

### Test plan

### DIFF
--- a/src/ts/src/NavBar.module.scss
+++ b/src/ts/src/NavBar.module.scss
@@ -39,7 +39,10 @@
         }
 
         .logo {
-            height: 100%;
+            svg {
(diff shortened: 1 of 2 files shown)


src/ts/src/NavBar.module.scss
```diff
@@ -39,7 +39,10 @@
         }
 
         .logo {
-            height: 100%;
+            svg {
+                height: 100%;
+                width: 22%;
```
ttung, line 44: do you fully control the height and the width of the enclosing div?  if not, couldn't this stretch in unusual ways?

PR #107, opened 2021-02-24
Rearrange config objects

ttung:
### Description
Move most config objects to the top level.  Maintain two second-level config objects:

1. One for holding and caching AWS secrets.  This is separate because we don't always want to trigger AWS secrets at app startup.
2. `DatabaseConfig`, which remains separate because not all fields are defined for every valid configuration.  Flask will retrieve all fields.

Depends on #104 

### Test plan
Deployed to Elastic Beanstalk, and verified I can log in.

### DIFF
--- a/src/py/aspen/config/config.py
+++ b/src/py/aspen/config/config.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import abc
 import json
 import os
 from collections import MutableMapping
@@ -9,6 +10,7 @@ from typing import Any, Mapping, Type
 from botocore.exceptions import ClientError
 
 from aspen import aws
+from aspen.database.connection import SqlAlchemyInterface
 
 
 class Config:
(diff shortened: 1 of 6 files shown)


src/py/aspen/config/config.py
```diff
@@ -24,94 +26,110 @@ def __init_subclass__(cls: Type[Config], descriptive_name: str, *args, **kwargs)
     def by_descriptive_name(cls, descriptive_name: str) -> Config:
         return Config._subclasses[descriptive_name]()
 
+    ####################################################################################
+    # secondary config objects.
     @property
-    def DEBUG(self):
-        return False
+    @abc.abstractmethod
```
phoenixAja, line 32: this is much better

src/py/aspen/config/production.py
```diff
@@ -1,8 +1,29 @@
-from .config import Config, DatabaseConfig
+import uuid
+from typing import Any, Mapping
+
+from .config import Config, DatabaseConfig, SecretsConfig
```
phoenixAja, line 4: should we change to aspen.config?

ttung: ok you seem to really dislike relative imports, i will dump them all in a fire in a single PR coming up. :)

phoenixAja (approved):

src/py/aspen/config/production.py
phoenixAja: no no no i was just thinking because now we have a smattering of relative imports and aspen imports. But a unifying PR _would_ be nice 😉

ttung: believe it or not, there is actually a pattern and a rationale to when i prefer relative and absolute.

PR #108, opened 2021-02-24
Phoenix/sampledata view

phoenixAja:
### Description

sample data api view to return all samples from a particular group, returning specificially: 

```
Private Id -> Sample.private_identifier
Public Id -> Sample.public_identifier
Upload Date -> SequencingRead.sequencing_date
Collection Date -> Sample.collection_date
Collection Location -> Sample.location
```

#### Issue
[ch120101](https://app.clubhouse.io/genepi/stories/space/120101)

### Test plan

wrote unit test for view


src/py/aspen/app.py
```diff
@@ -99,14 +102,47 @@ def logout():
...
+            {"sequence": seq.to_dict(), "sample": seq.sample.to_dict()}
+            for seq in sequencing_reads
+        ]
+
+        # filter for only information we need in sample table view
+        filtered_sample_data = [
+            {
+                "private_identifier": s["sample"]["private_identifier"],
+                "public_identifier": s["sample"]["public_identifier"],
+                "upload_date": s["sequence"]["sequencing_date"],
```
sidneymbell: assuming the `sequencing_date` field will eventually be updated to say `upload_date` per your issue/story?

src/py/aspen/api_utils.py
```diff
@@ -0,0 +1,15 @@
...
+def get_usergroup(session, user_id: str):
+    # since authentication is required to access view this will always return one
+    user = (
+        session.query(User)
+        .options(joinedload(User.group))
+        .filter(User.auth0_user_id == user_id)
+        .one()
+    )
+
+    return user
```
ttung: You should have this return a semi-composed query, and where you're using this, you can grab the `one()` or join it to other things.

```suggestion
def get_usergroup_query(session, user_id: str):
    # since authentication is required to access view this will always return one
    return (
        session.query(User)
        .options(joinedload(User.group))
        .filter(User.auth0_user_id == user_id)
    )
```

src/py/aspen/app.py
```diff
@@ -99,14 +102,47 @@ def logout():
...
     with session_scope(application.config["DATABASE_CONFIG"].INTERFACE) as db_session:
-        # since authentication is required to access view this information is guaranteed to be in session
         profile = session["profile"]
-        user = (
-            db_session.query(User)
-            .options(joinedload(User.group))
-            .filter(User.auth0_user_id == profile["user_id"])
-            .one()
-        )
+        user = get_usergroup(db_session, profile["user_id"])
```
ttung: here, it would be:

```suggestion
        user = get_usergroup_query(db_session, profile["user_id"]).one()
```

src/py/aspen/app.py
```diff
@@ -99,14 +102,47 @@ def logout():
...
+
+        sequencing_reads = (
+            db_session.query(SequencingReads)
+            .join(Sample, Group, User)
+            .options(
+                joinedload(SequencingReads.sample).joinedload(Sample.submitting_group)
+            )
+            .filter(Sample.submitting_group == user.group)
+            .all()
+        )
```
ttung: You could plausibly just do the following and avoid the extra db roundtrip:

```suggestion
        sequencing_reads = (
            db_session.query(SequencingReads)
            .join(Sample, Group, User)
            .options(
                joinedload(SequencingReads.sample)
            )
            .filter(User.auth0_user_id == profile["user_id"])
        )
```

ttung (changes requested):
generally looks good to me. see comments.

src/py/aspen/app.py
ttung: I don't think you pull the group info, so the joined load to the group is unnecessary.

src/py/aspen/app.py
phoenixAja: @sidneymbell yes we'll update this to be upload date once we make that migration 👍, but for now i'm using as a proxy for `upload_date` since we currently don't have an `upload_date` field

sidneymbell (approved):

phoenixAja:
@ttung is it cool if i merge this?

src/py/aspen/api_utils.py
```diff
@@ -0,0 +1,12 @@
+from sqlalchemy.orm import joinedload
+
+from aspen.database.models.usergroup import User
+
+
+def get_usergroup_query(session, user_id: str):
+    # since authentication is required to access view this will always return one
```
ttung: Not relevant any more.

ttung (approved):

src/py/aspen/app/views/base.py
```diff
@@ -0,0 +1,17 @@
...
+
+from aspen.app.app import application, requires_auth
+
+
+# Catch all routes. If path is a file, send the file;
+# else send index.html. Allows reloading React app from any route.
+@application.route("/", defaults={"path": ""})
+@application.route("/<path:path>")
+@requires_auth
+def serve(path):
```
ttung: I wouldn't call this file `base.py`.  Maybe `static.py`.

ttung (approved):

src/py/aspen/app/views/base.py
phoenixAja: what about `root.py`? Static makes me think of static assets

phoenixAja: or actually `index.py`

phoenixAja: i'll go ahead and change/merge and then we can always rename it in the future.

ttung: > what about root.py? Static makes me think of static assets

These **are** static assets.

PR #111, opened 2021-02-24
[easy] Remove duplicate source of truth for database migrations

ttung:
### Description
The revision id and downward revision is in the python code a few lines down.

### Test plan
These are comments, YOLO.


phoenixAja (commented):
interesting, I'm still a little confused why you want to remove this though

phoenixAja:
just read your description again, so ignore the last comment.

phoenixAja (approved):

PR #112, opened 2021-02-24
Api call navbar

phoenixAja:
### Description

integrate usergroup api with navbar component

#### Issue
[ch119785](https://app.clubhouse.io/genepi/stories/space/119785)

### Test plan

refreshed page when logged in, saw user name and dph name values change

### DIFF
--- a/src/py/aspen/app/views/index.py
+++ b/src/py/aspen/app/views/index.py
@@ -2,14 +2,13 @@ from pathlib import Path
 
 from flask import send_from_directory
 
-from aspen.app.app import application, requires_auth
+from aspen.app.app import application
 
 
 # Catch all routes. If path is a file, send the file;
 # else send index.html. Allows reloading React app from any route.
 @application.route("/", defaults={"path": ""})
 @application.route("/<path:path>")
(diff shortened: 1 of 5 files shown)


src/py/aspen/app.py
```diff
@@ -55,7 +55,7 @@ def decorated(*args, **kwargs):
 # else send index.html. Allows reloading React app from any route.
 @application.route("/", defaults={"path": ""})
 @application.route("/<path:path>")
-@requires_auth
+#@requires_auth
```
ttung: is this change intentional?

phoenixAja: no i meant to remove that, since it doesn't fit in with the v0 user flow

ttung (approved):

lvreynoso (changes requested):
Can we bring the state up to the level of `App`? We don't want state that will be shared by the entire app to be stuck in the `NavBar` component.

phoenixAja:
@lvreynoso moved the state up to App

lvreynoso (approved):
Thanks Phoenix!

PR #113, opened 2021-02-24
refactor flask app

ttung:
### Description
Break up app.py into the flask app initialization and the routes.  In the long term, we will want to split up the routes into separate python modules and this simplifies that future step.

Depends on #107 

### Test plan
1. pytest
1. run locally
1. run on eb

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,1 @@
-web: PYTHONPATH=$(pwd)/src/py/:$(pwd)/third-party/sqlalchemy-enum-tables/ gunicorn --bind localhost:8000 --workers 1 --threads 15 aspen.app
+web: PYTHONPATH=$(pwd)/src/py/:$(pwd)/third-party/sqlalchemy-enum-tables/ gunicorn --bind localhost:8000 --workers 1 --threads 15 aspen.main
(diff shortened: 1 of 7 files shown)


phoenixAja:
oh no you beat me, i wanted to do this after this demo deadline.

src/py/aspen/flask_app/flask_app.py
```diff
@@ -0,0 +1,47 @@
+import os
```
phoenixAja: can we name this src/py/aspen/app/app ?

src/py/aspen/flask_app/routes.py
```diff
@@ -0,0 +1,72 @@
+from pathlib import Path
```
phoenixAja: can we rename this src/py/aspen/app/views ?

src/py/aspen/flask_app/flask_app.py
ttung: probably.  we'd need to change our entry point as well.

phoenixAja (approved):
this looks good to me, just wanted to rename some stuff

src/py/aspen/flask_app/flask_app.py
phoenixAja: i just think the flask part is unnecessary, but if it's too annoying i don't mind waiting until later

ttung: no problem.

ttung:
ok it's all been renamed.

PR #114, opened 2021-02-25
create an aspen app class

ttung:
### Description
This has an explicit `aspen_config` object attached to it that we can reference, instead of reaching through the config.

Other globals such as db handles can also live in this aspen app class.

Depends on #113 

### Test plan
1. pytest
1. run locally
1. run on eb


phoenixAja (approved):
looks good

PR #115, opened 2021-02-25
Add @flaskproperty decorator

ttung:
### Description
Only fields decorated as such are fed into flask's flask.config.from_* method.  This allows us to collapse all the config objects!

Depends on #114 

### Test plan
1. pytest
1. run locally
1. run on eb


shanaxel42 (approved):
nicccccce

PR #116, opened 2021-02-25
merge all the config objects

ttung:
### Description
Now with #115, there's no need to have separate config objects.  Collapse it all into one.

Depends on #115 obviously.

### Test plan
1. pytest
1. flask run


shanaxel42 (approved):
ONE CONFIG TO RULE THEM ALL

PR #117, opened 2021-02-25
Switch to absolute imports

ttung:
### Description
Noop, this is just a refactor to use absolute imports.

Depends on #116 

### Test plan
1. pytest
1. flask run


phoenixAja (approved):
👏 👏 👏

PR #121, opened 2021-02-25
various aws fixes

phoenixAja:
### Description

i had to patch this because i have multiple eb envs tied to my profile and `_get_environment_name` was returning the wrong env

### Test plan

deployed eb env successfully


ttung (changes requested):
The problem is likely deeper, as we discussed offline.  I'll investigate.

phoenixAja:
@ttung ok i'll close this

PR #123, opened 2021-02-25
Begin Terraform tooling for Aspen

kislyuk:
### Description

Begin TF tooling for the Batch compute environment to manage Aspen phylogeny inference jobs.

### Test plan

Deploy and run:
```
aegea batch submit --queue aspen-dev --execute pwd --dry-run
aegea batch submit --queue aspen-dev --execute pwd --watch
```
to test the Batch compute environment works as expected.

@ttung please push your SSH_PUBLIC_KEY changes to the environment and batch_queue.tf files.

The `DEPLOYMENT_ENVIRONMENT` variable controls which environment (staging, prod) the code gets deployed to. The default is `dev`. All asset names are keyed by this variable so they don't interfere between deployment environments, even when deployed to the same AWS account.

Some points for discussion/future PRs:
* Do you want to use step functions to envelope the Batch jobs? We have TF tooling for this as well. I prefer to use step functions because they provide first-class JSON I/O, better error handling, a better state tracking system than Batch jobs, remain longer in the API so you can easily examine them, and provide better event integration with other AWS services as necessary.
* How do you plan to manage the Docker image for these jobs? The IDseq Dockerfiles are here: https://github.com/chanzuckerberg/idseq/tree/main/workflows/docker and they are built and pushed to ECR here: https://github.com/chanzuckerberg/idseq/blob/main/scripts/build_docker_image.sh - I can generate a minimal version of this config if needed.


kislyuk:
Closing this since it was adapted/copied and merged in other PRs.

PR #124, opened 2021-02-26
add workflow status, start/end times for workflows, and fix repository_type column

ttung:
#### Accessions:
- Rename entity_type to repository_type (copy pasta probably)

#### Workflows:
- Add a status field
- Explode run_date to start_datetime and end_datetime

### Test plan
upgraded migration, downgraded, upgraded, and downgraded.


phoenixAja (approved):
yay! 🚀

PR #126, opened 2021-02-26
simplify models for accessions and add genbank as a public repository

ttung:
### Description
Rather than link an accession to a public repository object, just label each accession with the repository enum.

### Test plan
ran migration upgrade and downgrade.  ran import in next pr.


phoenixAja (approved):
i like this a lot more 👍

PR #128, opened 2021-02-26
Do not cache the DB interface without the db uri as a cachekey

ttung:
### Description
`@lru_cache` takes the arguments of the method as a cachekey.  Since DB URI can change during the course of a unit test run (we change the database name so each test has its own clean database to work with), we end up pulling a stale db handle.  By adding the db_uri as a parameter to the `_DATABASE_INTERFACE` method, we guarantee the correct interface.

### Test plan
Successfully passed other tests.

### DIFF
--- a/src/py/aspen/app/aspen_app.py
+++ b/src/py/aspen/app/aspen_app.py
@@ -25,9 +25,9 @@ class AspenApp(Flask):
         return self._aspen_config
 
     @lru_cache()
-    def _DATABASE_INTERFACE(self) -> aspen_database.SqlAlchemyInterface:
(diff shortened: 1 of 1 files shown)


ttung:
This subtle bug was introduced in #119.  Since the tests would swap out the config object, the old way was insulated from this subtle bug.

phoenixAja (approved):
nice catch, i think in the future we can have the test db stay for the duration of the tests and just rollback the session after each test?

ttung:
> nice catch, i think in the future we can have the test db stay for the duration of the tests and just rollback the session after each test?

yeah, we could plausibly do that.  but the view tests start new sessions, so the data has to be committed.

PR #130, opened 2021-02-26
Add a created_date to SequencingReadsCollection and UploadedPathogenGenome

ttung:
### Description
This represents the datetime it was uploaded.  Modified import script to use DphCZBID.date_received as the upload date.

[ch123185](https://app.clubhouse.io/genepi/story/123185/need-an-uploaded-date-for-upload-able-entities)

### Test plan
ran migration.

PR #131, opened 2021-02-26
Retrieve either SequencingReadsCollection or UploadedPathogenGenome for sampleview

ttung:
### Description
Samples can be one or the other.  Retrieve both, and fill in the result with whichever one is appropriate.

Depends on #128, #129, #130

### Test plan
pytest


phoenixAja (approved):
cool cool cool!!! thanks so much for your help on this!!

PR #133, opened 2021-02-26
Change React internal link to hit backend /login route

lvreynoso:
### Description

It makes the sign in button work!

#### Issue
[ch66711](https://app.clubhouse.io/genepi/stories/space/66711)

### Test plan

Load app; sign in!

### DIFF
--- a/src/ts/src/NavBar.tsx
+++ b/src/ts/src/NavBar.tsx
@@ -72,9 +72,9 @@ const NavBar: FunctionComponent<Props> = ({ org, user }: Props) => {
     );
 
     const signInLink = (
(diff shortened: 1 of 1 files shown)


phoenixAja (approved):
so we don't need to add this as a switch in app.tsx?

lvreynoso:
Nope! Switches in React navigate between components, but in this case we want it to hit the backend. Sorry for the confusion!

phoenixAja:
> Nope! Switches in React navigate between components, but in this case we want it to hit the backend. Sorry for the confusion!

no worries i was jw for my own react education 📚

PR #135, opened 2021-02-26
Don't put the postgres data on a separate volume

ttung:
### Description
This allows us to snapshot database state readily.

### Test plan
building the docker image...


phoenixAja:
how does this allows us to snapshot database state readily?

ttung:
You can't snapshot a volume in docker afaik.  This is basically a port of the same functionality I built for covidhub.

phoenixAja (approved):

PR #139, opened 2021-02-26
v0 Design Fixes

lvreynoso:
### Description

Oooo pretty!!

Screenshot from 2021-02-26 11-31-51

#### Issue
[ch66711](https://app.clubhouse.io/genepi/stories/space/66711)

### Test plan

Load app, navigate around, look at the sparklies.


phoenixAja:
BEAUTIFUL!! ✨ I'm going to pull now and give it a look, we should definitely showcase this in the demo today.

src/ts/src/common/api/index.tsx
```diff
@@ -20,8 +20,8 @@ export const fetchUserData = async (): Promise<
 const SAMPLE_MAP = new Map<string, keyof Sample>([
     ["collection_date", "collectionDate"],
     ["collection_location", "collectionLocation"],
-    ["private_identifier", "privateID"],
-    ["public_identifier", "publicID"],
+    ["private_identifier", "privateId"],
+    ["public_identifier", "publicId"],
     ["upload_date", "uploadDate"],
```
phoenixAja, line 25: do we need to add gisaid to this now too?

lvreynoso: No, since the gisaid key is not different between the api and the TS type it will transfer automatically :)

phoenixAja (approved):
design looks good on firefox! The only thing is if we should add the GISAID field now that @ttung added that functionality to the sample data view?

ttung:
> design looks good on firefox! The only thing is if we should add the GISAID field now that @ttung added that functionality to the sample data view?

#138!

lvreynoso:
gisaid field is there :) the `jsonToType<T>` function uses the mapping, but any key in the input object not in the mapping just gets directly transferred. It's already in the `Sample` type definition and the table def hits that key.

PR #140, opened 2021-02-26
Defer the sequence data from being loaded by default

ttung:
### Description
Sequence data is too big to load by default.

Too bad sqlalchemy 1.4 is not yet release, as it has an awsesome [`defer(raiseload=True)`](https://docs.sqlalchemy.org/en/14/orm/loading_columns.html#raiseload-for-deferred-columns) feature.

### Test plan
gha.

### DIFF
--- a/src/py/aspen/database/models/sequences.py
+++ b/src/py/aspen/database/models/sequences.py
@@ -15,7 +15,7 @@ from sqlalchemy import (
     String,
     UniqueConstraint,
 )
-from sqlalchemy.orm import backref, relationship
+from sqlalchemy.orm import backref, deferred, relationship
 
 from aspen.database.models.base import base
 from aspen.database.models.entity import Entity, EntityType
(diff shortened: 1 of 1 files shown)


src/py/aspen/database/models/sequences.py
```diff
@@ -146,7 +146,7 @@ class PathogenGenome(Entity):
     __tablename__ = "pathogen_genomes"
 
     entity_id = Column(Integer, ForeignKey(Entity.id), primary_key=True)
-    sequence = Column(String, nullable=False)
+    sequence = deferred(Column(String, nullable=False))
```
phoenixAja, line 149: so this is storing the actual sequence data?? i thought we would be storing an s3 key or something like that

ttung: no, we're storing the sequence data in the db.

erinstargazer (approved):

PR #143, opened 2021-02-26
Switch to enable profiling for db interact

ttung:
### Description
aspen-cli db interact --profile will print the sql statements being executed and the time spent.

### Test plan
Tried it, it works great.


phoenixAja (approved):
really nice!! would be nice to add a little comment to the documentation to let users know about this capability

ttung:
Added docs.

PR #144, opened 2021-02-27
Complete switch to postgres 13.1

ttung:
### Description
1. Forgot one place where to switch it.
2. Forgot to set a+x for the `docker-entrypoint.sh`.
3. Accidentally downloaded the html page for `docker-entrypoint.sh` instead of the raw content.

### Test plan
1. docker build && docker run
2. GHA

### DIFF
--- a/docker/postgres/docker-entrypoint.sh
+++ b/docker/postgres/docker-entrypoint.sh
@@ -1,2482 +1,327 @@
-
-
-
-
-
-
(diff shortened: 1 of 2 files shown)


ttung:
If you're wondering why tests didn't catch the previous oopsie, that's because the test infra file wasn't updated.

lvreynoso (approved):

PR #146, opened 2021-03-01
Models to represent phylo trees and runs

ttung:
### Description
Self explanatory?

#### Issue
[ch65314](https://app.clubhouse.io/genepi/story/65314/standup-dummy-local-version-of-aspen-for-demo)

### Test plan
Used in subsequent PR.

### DIFF
--- a/src/py/aspen/database/models/phylo_tree.py
+++ b/src/py/aspen/database/models/phylo_tree.py
@@ -1,0 +1,39 @@
+from sqlalchemy import Column, ForeignKey, Integer, String, UniqueConstraint
+from sqlalchemy.orm import backref, relationship
+
+from aspen.database.models.entity import Entity, EntityType
+from aspen.database.models.usergroup import Group
+from aspen.database.models.workflow import Workflow, WorkflowType
+
+
+class PhyloTree(Entity):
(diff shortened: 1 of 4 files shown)


src/py/aspen/database/models/phylo_tree.py
```diff
@@ -0,0 +1,39 @@
...
+    s3_bucket = Column(String, nullable=False)
+    s3_key = Column(String, nullable=False)
+
+
+class PhyloRun(Workflow):
+    __tablename__ = "phylo_runs"
+    __mapper_args__ = {"polymorphic_identity": WorkflowType.PHYLO_RUN}
+
+    workflow_id = Column(Integer, ForeignKey(Workflow.id), primary_key=True)
+    group_id = Column(
```
erinstargazer, line 25: Is the group related to who ran it, or who the tree was generated for/who owns it?

erinstargazer (approved):
Minor clarification question but otherwise looks good.

src/py/aspen/database/models/phylo_tree.py
ttung: The group that ran it when the runs are self-service, and the group that it is generated for before it becomes self-service.

PR #149, opened 2021-03-02
Remove extraneous paths

lvreynoso:
### Description

Gets rid of the two navigation buttons and the dropdown menu.

#### Issue
[ch123793](https://app.clubhouse.io/genepi/stories/space/123793)

### Test plan

Navigate to the app in your browser.

### DIFF
--- a/src/ts/src/NavBar.tsx
+++ b/src/ts/src/NavBar.tsx
@@ -100,7 +100,6 @@ const NavBar: FunctionComponent<Props> = ({ org, user }: Props) => {
                 </div>
                 <div className={style.center}></div>
                 <div className={style.right}>
-                    {/* navigationLinks */}
                     {rightEdge}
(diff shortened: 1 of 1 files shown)


src/ts/src/NavBar.tsx
```diff
@@ -98,7 +100,7 @@ const NavBar: FunctionComponent<Props> = ({ org, user }: Props) => {
                 </div>
                 <div className={style.center}></div>
                 <div className={style.right}>
-                    {navigationLinks}
+                    {/* navigationLinks */}
```
ttung: should this just be removed?

ttung (approved):

src/ts/src/NavBar.tsx
lvreynoso: sure, I can do that

PR #157, opened 2021-03-02
Add cansee support for sample data

ttung:
### Description
1. Sample view can now pull data from other groups.
2. Add tests!

Depends on #156 

### Test plan
All sorts of exciting combinations!


ttung:
Ok I fixed this so we can do this in two DB requests instead of three.

lvreynoso (approved):
Looks great! Good job knocking out those tests

PR #159, opened 2021-03-02
Create generic, smooth API interface

lvreynoso:
### Description

Adds three generic functions that cover a broad variety of API responses:

* `apiSingleResponse<T>`. This is when the JSON response is a flat mapping of keys that corresponds to a single object, and we want to extract the information from the response and map it onto a type we use in the Typescript app.
* `apiSplitResponse<T extends any[]>`. This is when the JSON response is an object with multiple keys, and each key corresponds to a type. The function is used with a tuple type, eg `apiSplitResponse<[Group, User]>(...)`
* `apiCollectionResponse<T>`. This is when the JSON response is an array of objects, all of the same type; the function returns any array of objects of type T.

This makes adding new API endpoints very, very easy; all you have to do is define a type, define a mapping if any of the keys are different cases, and call one of the generics.

#### Issue

### Test plan

Right now it's kind of tested ad hoc, with the existing api interfaces.

### DIFF
--- a/src/ts/src/App.tsx
+++ b/src/ts/src/App.tsx
@@ -15,9 +15,9 @@ const App: FunctionComponent = () => {
 
     useEffect(() => {
         const setUserData = async () => {
-            const response = await fetchUserData();
-            setUser(response.user.name);
-            setOrg(response.group.name);
+            const [groupResponse, userResponse] = await fetchUserData();
+            setUser(userResponse.name);
+            setOrg(groupResponse.name);
         };
         setUserData();
     }, []);
--- a/src/ts/src/common/api/index.tsx
+++ b/src/ts/src/common/api/index.tsx
@@ -45,7 +45,7 @@ const USER_MAP = new Map<string, keyof User>([
     ["group_admin", "groupAdmin"],
     ["system_admin", "systemAdmin"],
 ]);
-export const fetchUserData = async (): [Group, User] =>
+export const fetchUserData = async (): Promise<[Group, User]> =>
     apiSplitResponse<[Group, User]>([new Map([]), USER_MAP], "/api/usergroup");
 
 const SAMPLE_MAP = new Map<string, keyof Sample>([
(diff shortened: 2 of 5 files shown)


ttung:
General questions: 

1. Is `apiSingleResponse` a special case of `apiSplitResponse`, and if so, can they be collapsed?
2. What happens if the response is an array of composite objects a la `apiSplitResponse`?

lvreynoso:
- Hmm, apiSingleResponse` expects a the root JSON object to represent the type, something like
```json
{
   "key1": "value1",
   "key2": "value2"
}
```
whereas `apiSplitResponse` expects the JSON structure to look like 
```json
{
    "type": {
        "key1": "value1",
        "key2": "value2"
    }
}
```
so they don't match up one to one. We could add a branching statement in a collapsed function to use one or the other pattern depending on the number of mappings passed in, but since we return a tuple with a single value in it which then has to be destructured (not bad, but not intuitive); I'm not sure it's possible to return a union type where one is a tuple member of the specified type but that would require casting the return object to the single type anyway. In short, I don't think it would be saving any code and it makes for a nicer dev experience to just be able to call one or the other.

- that is not supported by any of the functions, but definitely doable. I can't think of why we would ever return an array of compose objects though, instead of different arrays under their own keys.

lvreynoso:
Ok, realized we don't have an API endpoint that returns only a single object. So I'm not sure what that should look like :) but I think the point about having more useful experience with them in two different functions stands.

lvreynoso:
You can already call `apiSplitResponse` as it is written with just one type in the tuple and it should work, though.

ttung:
Philosophically, it's probably the right thing to never return an array.  We should always return keyed responses so we can extend the API.  I'm less certain if we should wrap everything in an extra object for consistency.

e.g., 
```
{'response': [my_magic_array]}
```

Now does that mean for single objects, we do this?
```
{'response': {my_object}}
```

sidneymbell:
Would it be possible to show a short example of each? (Like what Tony started above)

Generally agree that we should always return keyed responses for the sake of any bioinformaticians that may want to contribute or adapt down the line.

shanaxel42:
> Philosophically, it's probably the right thing to return an array. We should always return keyed responses so we can extend the API. I'm less certain if we should wrap everything in an extra object for consistency.
> 
> e.g.,
> 
> ```
> {'response': [my_magic_array]}
> ```
> 
> Now does that mean for single objects, we do this?
> 
> ```
> {'response': {my_object}}
> ```

The first is how I'm most used to working with api responses, but I do see the second as having a lot of advantages! My thing with api's is just as long as the format is consistent across all types of requests things are setup well.

PR #163, opened 2021-03-04
Do not return results as top-level constructs

ttung:
### Description
Wrap everything with a top-level object, so we can extend the API more easily.

The exception is retrieving a single phylo_tree, which we return the json directly.

Depends on #157 

### Test plan
pytest

But the front end still needs to be updated, hopefully @lvreynoso can add the necessary changes to this PR.


lvreynoso (approved):
Awesome!!!

PR #165, opened 2021-03-05
Fix table loading bug

lvreynoso:
### Description

Make sample and trees data optional to better represent state and explicitly handle change in state

#### Issue
[ch124687](https://app.clubhouse.io/genepi/stories/space/124687)

### Test plan

Navigate to samples table and test that the table loads properly without any user interaction.

### DIFF
--- a/src/ts/src/data/index.tsx
+++ b/src/ts/src/data/index.tsx
@@ -10,8 +10,8 @@ import Samples from "./samples";
 import style from "./index.module.scss";
 
 const Data: FunctionComponent = () => {
-    const [samples, setSamples] = useState<Array<Sample>>([]);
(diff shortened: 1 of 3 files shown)


ttung:
Why did the old system not work?  If the samples array changes, shouldn't the display automatically update?

lvreynoso:
Since we are using `useReducer`, we have to handle that state explicitly. The function component might be called again but it won't initialize the state with the updated value since `useReducer` will just ignore the subsequent calls.

ttung (approved):
I don't understand the mechanics, but this seems somewhat surprising that this isn't an obviously handled case.  But I will defer to you as you are the expert. :)

PR #166, opened 2021-03-06
Support for a remote db config

ttung:
### Description
For the most part, accessing the DB is a function of AWS_PROFILE or AWS_ACCESS_KEY_ID/AWS_SECRET_KEY_ID.  We want to have an object that encompasses all the remote db config so we can access the DB from the CLI.

If we didn't do this, then we would require the user to specify --staging or --prod (to initialize `StagingConfig` or `ProductionConfig`), when it actually doesn't matter.

### Test plan
flask run

### DIFF
--- a/src/py/aspen/config/config.py
+++ b/src/py/aspen/config/config.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
-import abc
 import json
 import os
+import platform
 from collections import MutableMapping
 from functools import lru_cache
 from typing import Any, Callable, Mapping, Optional, Set, Type, TypeVar, Union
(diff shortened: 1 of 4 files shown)


src/py/aspen/config/config.py
```diff
@@ -58,26 +58,29 @@ class Config:
...
-            raise Exception(
-                f"Attributes ({missing_properties_string}) defined as flask properties"
-                f" in Config but not in {cls.__name__}"
-            )
+        if descriptive_name is not None:
+            Config._subclasses[descriptive_name] = cls
+
+            # get a list of @flaskproperties for Config and the subclass of Config, and
+            # compare to see if there are any attributes that were marked as
+            # @flaskproperty in Config but not in the subclass.
```
phoenixAja, line 72: don't ABC's do something similar to this?

src/py/aspen/config/config.py
```diff
@@ -187,3 +189,11 @@ def DATABASE_URI(self) -> str:
     @property
     def DATABASE_READONLY_URI(self) -> str:
         raise NotImplementedError()
+
+
+class RemoteConfig(Config):
```
phoenixAja: i feel like a more intuitive name might be `RemoteDBConfig`? Unless we're planning on using this config for things other that a DB down the line

src/py/aspen/config/production.py
```diff
@@ -1,12 +1,10 @@
-import uuid
-
 from aspen.config import config
 
 
-class ProductionConfig(config.Config, descriptive_name="prod"):
+class ProductionConfig(config.RemoteConfig, descriptive_name="prod"):
     @config.flaskproperty
-    def SECRET_KEY(self) -> str:
-        return uuid.uuid4().hex
```
phoenixAja, line 9: do we need to add `SECRET_KEY` to the `RemoteConfig` if we're removing it here?

src/py/aspen/config/staging.py
```diff
@@ -1,29 +1,20 @@
...
-class StagingConfig(config.Config, descriptive_name="staging"):
+class StagingConfig(config.RemoteConfig, descriptive_name="staging"):
     @config.flaskproperty
     def DEBUG(self) -> bool:
         return True
 
-    @config.flaskproperty
-    def SECRET_KEY(self) -> str:
-        return uuid.uuid4().hex
-
```
phoenixAja, line 18: samesies

src/py/aspen/config/production.py
phoenixAja: ok i see you added this back in later PRs, nevermind

phoenixAja (approved):

src/py/aspen/config/config.py
ttung: If you tag something as an abstract class (abc.abstractclass), or there are abstract methods (abc.abstractmethod), it cannot be instantiated.  This is similar in concept but we can't quite use those constructs to achieve the same.

src/py/aspen/config/config.py
ttung: It's subclassed by all configs that use a remote db, so I guess you're right.

src/py/aspen/config/production.py
ttung: A terrible implementation is provided by the base class now.

src/py/aspen/config/config.py
ttung: Updated.

PR #167, opened 2021-03-06
Add support for different Data subviews (including phylo trees)

lvreynoso:
### Description

Creates generic components out of the `Samples` and `SamplesTable` components, allowing the resulting `DataSubview` to take in different types of data.

Since searching across keys was already supported, this includes very basic searching capabilities for trees.

#### Issue
[ch120774](https://app.clubhouse.io/genepi/stories/space/120774)
[ch120778](https://app.clubhouse.io/genepi/stories/space/120778)

### Test plan


src/ts/src/common/types/bioinformatics.d.ts
```diff
@@ -4,6 +4,7 @@
 
 interface Sample {
```
ttung: should there be a base interface for anything that can show up in a table view and have the type field live there?

lvreynoso: Yeah that makes more sense than the union type. Good point

ttung (approved):

PR #172, opened 2021-03-09
Clarify AWS profiles

ttung:
### Description
src/dst is confusing.  Make it explicitly covidhub vs aspen.  This makes it easier to manage multiple connections, which we will need to import to a staging db hosted on AWS.

### Test plan
None.

### DIFF
--- a/src/py/aspen/cli/db.py
+++ b/src/py/aspen/cli/db.py
@@ -76,31 +76,28 @@ def interact(ctx, profile):
 
 
 @db.command("import-covidhub-project")
+@click.option("--covidhub-aws-profile", type=str, required=True)
 @click.option("--covidhub-db-secret", default="cliahub/cliahub_test_db")
 @click.option("--rr-project-id", type=str, required=True)
 @click.option("--s3-src-prefix", type=str, required=True)
-@click.option("--s3-src-profile", type=str, required=True)
 @click.option("--s3-dst-prefix", type=str, required=True)
-@click.option("--s3-dst-profile", type=str, required=True)
 @click.pass_context
 def import_covidhub_project(
     ctx,
+    covidhub_aws_profile,
     covidhub_db_secret,
     rr_project_id,
     s3_src_prefix,
-    s3_src_profile,
     s3_dst_prefix,
-    s3_dst_profile,
 ):
     # these are injected into the IPython scope, but they appear to be unused.
     engine = ctx.obj["ENGINE"]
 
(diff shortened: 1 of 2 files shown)


src/py/aspen/cli/db.py
```diff
@@ -37,31 +37,28 @@ def interact(ctx, profile):
 
 
 @db.command("import-covidhub-project")
+@click.option("--covidhub-aws-profile", type=str, required=True)
```
phoenixAja, line 79: wouldn't this always be the biohub aws profile?

src/py/aspen/covidhub_import/implementation.py
```diff
@@ -48,8 +48,10 @@
 logger = logging.getLogger(__name__)
 
 
-def covidhub_interface_from_secret(secret_id: str) -> CSqlAlchemyInterface:
-    interface = Cinit_db(Cutil.get_db_uri(secret_id))
+def covidhub_interface_from_secret(
+    covidhub_aws_profile: str, secret_id: str
+) -> CSqlAlchemyInterface:
+    interface = Cinit_db(Cutil.get_db_uri(secret_id, aws_profile=covidhub_aws_profile))
```
phoenixAja: does the leading C stand for covidhub? lol

phoenixAja: i feel like `CUtil` should be `c_util` and `Cinit_db` should be `c_init_db`, that seems more python-ish to me. That's a nitpick though

src/py/aspen/cli/db.py
phoenixAja: or maybe i'm missing what you mean by `covidhub-aws-profile` exactly, or what the issue is that this PR is fixing?

ttung: It is always the biohub aws profile.  We can't hardcode it though because everyone could name it differently.

src/py/aspen/covidhub_import/implementation.py
ttung: I'll just be less lazy and call it covidhub_*.

phoenixAja: haha even better!

src/py/aspen/cli/db.py
phoenixAja: makes sense

phoenixAja (approved):

PR #175, opened 2021-03-09
Import users from covidhub

ttung:
### Description
Interface with auth0 and covid database to build up users.  In addition to the group associated with the RR project, we also import users associated with CDPH and admin.

Depends on #172, #174, https://github.com/czbiohub/covidhub/pull/1512, https://github.com/czbiohub/covidhub/pull/1513

#### Issue
[ch125195](https://app.clubhouse.io/genepi/stories/space/125195)

### Test plan
imported users successfully and verified manually on db.

### DIFF
--- a/src/py/aspen/covidhub_import/implementation.py
+++ b/src/py/aspen/covidhub_import/implementation.py
@@ -3,9 +3,11 @@ import json
 import logging
 import re
 import warnings
+from dataclasses import dataclass
 from typing import (
     Any,
     Collection,
+    Iterable,
     Iterator,
     List,
     Mapping,
@@ -17,10 +19,13 @@ from typing import (
 import boto3
 import pytz
 import tqdm
+from auth0.v3 import authentication as auth0_authentication
+from auth0.v3 import management as auth0_management
 from sqlalchemy import and_, or_
 from sqlalchemy.orm import joinedload, Session
 
 from aspen.aws.s3 import S3UrlParser
+from aspen.config import config as aspen_config
 from aspen.database.connection import session_scope, SqlAlchemyInterface
 from aspen.database.models import (
     Accession,
@@ -31,11 +36,13 @@ from aspen.database.models import (
     PublicRepositoryType,
     Sample,
     UploadedPathogenGenome,
+    User,
     WorkflowStatusType,
 )
 from covid_database import init_db as covidhub_init_db
 from covid_database import SqlAlchemyInterface as CSqlAlchemyInterface
 from covid_database import util as covidhub_database_util
+from covid_database.models import covidtracker
 from covid_database.models.enums import ConsensusGenomeStatus
 from covid_database.models.ngs_sample_tracking import (
     ConsensusGenome,
@@ -46,6 +53,13 @@ from covid_database.models.ngs_sample_tracking import (
 )
 
 logger = logging.getLogger(__name__)
+
+
+@dataclass
+class Auth0Entry:
+    nickname: str
+    email: str
+    auth0_token: str
 
 
 def covidhub_interface_from_secret(
@@ -83,6 +97,20 @@ def get_names_from_tree(tree) -> Sequence[str]:
     return results
 
 
+def get_or_make_group(session: Session, name: str, address: str, email: str) -> Group:
+    group = session.query(Group).filter(Group.name == name).one_or_none()
+    if group is None:
+        # copy the project info into the "group"
+        group = Group(
+            name=name,
+            address=address,
+            email=email,
+        )
+        session.add(group)
+
+    return group
+
+
 def import_project(
     interface: SqlAlchemyInterface,
     covidhub_aws_profile: str,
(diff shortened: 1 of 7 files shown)


phoenixAja (changes requested):
looks good, but do you also need to add the piplock file if the requirement dependencies have changed?

ttung:
Update `Pipfile` and `Pipfile.lock`

src/py/aspen/covidhub_import/implementation.py
```diff
@@ -48,6 +55,13 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclass
+class Auth0Entry:
+    nickname: str
```
phoenixAja, line 60: we don't need to add username to this? or does nickname mean username?

src/py/aspen/covidhub_import/implementation.py
```diff
@@ -97,24 +126,85 @@ def import_project(
...
-            group = Group(
-                name=project.originating_lab,
-                address=project.originating_address,
+        group, admin_group, cdph_group = (
+            get_or_make_group(
+                session,
+                project.originating_lab,
+                project.originating_address,
+                # FIXME: this needs to be updated.
+                email="fakeemail@dph.gov",
```
phoenixAja, line 146: i thought that this field had to be unique, or did we disable that until we have a workaround?

phoenixAja: will this break if we are moving over multiple groups (besides admin and cdph)?

src/py/aspen/covidhub_import/implementation.py
```diff
@@ -97,24 +126,85 @@ def import_project(
...
+                "CDPH",
+                "3701 N Freeway Blvd, Sacramento, CA 95834",
+                # FIXME: this needs to be updated.
+                email="cdph.internetadmin@cdph.ca.gov",
+            ),
+        )
+        covidhub_group_to_aspen_group = {
+            project.covidtracker_group.group: group,
+            covidhub_session.query(covidtracker.Group)
+            .filter(covidtracker.Group.name == "Admin")
```
phoenixAja, line 166: does this need to be `"ADMIN"`?

src/py/aspen/covidhub_import/implementation.py
```diff
@@ -97,24 +126,85 @@ def import_project(
...
+            .one(): cdph_group,
+        }
+
+        covidhub_users: Iterable[covidtracker.UsersGroups] = covidhub_session.query(
+            covidtracker.UsersGroups
+        ).filter(
+            or_(
+                covidtracker.UsersGroups.group_id.in_(
+                    covidhub_session.query(covidtracker.Group.id).filter(
+                        covidtracker.Group.name.in_(["Admin", "CDPH"])
```
phoenixAja, line 179: same here? Or is this not case sensitive?

src/py/aspen/covidhub_import/implementation.py
```diff
@@ -97,24 +126,85 @@ def import_project(
...
+
+            # try to create this user in the aspen db.
+            user = (
+                session.query(User).filter(User.email == auth0_user.email).one_or_none()
             )
-            session.add(group)
+            if user is None:
+                user = User()
+
+            user.name = auth0_user.nickname
```
phoenixAja, line 204: this is a nitpick, but i feel like we should rename `nickname` to `name`.

phoenixAja (commented):
looks good for the most part just had some comments

src/py/aspen/covidhub_import/implementation.py
ttung: For the covidtracker auth0 account, the usernames were all email addresses.

src/py/aspen/covidhub_import/implementation.py
ttung: Yeah, this will break with multiple groups.

src/py/aspen/covidhub_import/implementation.py
ttung: No, in the old DB, it was "Admin".

src/py/aspen/covidhub_import/implementation.py
ttung: Same here, in the old DB, it was Admin and CDPH.

src/py/aspen/covidhub_import/implementation.py
ttung: We're looking at the auth0 nickname field, so it makes sense.  Had we populated our auth0 db correctly, we would be looking at username and not nickname.

src/py/aspen/covidhub_import/implementation.py
phoenixAja: oh ok then nevermind

src/py/aspen/covidhub_import/implementation.py
phoenixAja: makes sense, keep nickname then 👍

phoenixAja (approved):

PR #179, opened 2021-03-11
Grant access to postgres if an EC2 instance is in the right security group

ttung:
### Description
The security group policy denies public access to the database by default.  This creates a security group that can access the database.  Instructions are provided for how to update an elastic beanstalk environment to be in that security group.

#### Issue
[ch125756](https://app.clubhouse.io/genepi/story/125756/fix-how-elastic-beanstalk-instances-connect-to-the-db)

### Test plan
1. applied these terraform changes
2. verified I can connect to the db.

### DIFF
--- a/terraform/security_group.tf
+++ b/terraform/security_group.tf
@@ -1,3 +1,8 @@
+resource "aws_security_group" "eb-security-group" {
+  name        = "eb-security-group"
+  description = "security group for eb instances"
+}
+
 resource "aws_security_group" "db-security-group" {
   name        = "db"
   description = "security group for database"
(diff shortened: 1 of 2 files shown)


terraform/security_group.tf
```diff
@@ -10,7 +15,21 @@ resource "aws_security_group" "db-security-group" {
...
+  security_group_id        = aws_security_group.db-security-group.id
+  type                     = "ingress"
+  description              = "Access given to the security group applied to all EB instances"
+  from_port                = 5432
+  to_port                  = 5432
+  protocol                 = "tcp"
+  source_security_group_id = aws_security_group.eb-security-group.id
+}
+
+resource "aws_security_group_rule" "remote-db-access-ttung" {
```
lvreynoso: `ttung`?

ttung: yup, this is a rule that allows my ip address at home to access the instance.  feel free to add one for your ip.

lvreynoso: TBH I'm not a fan of this, I think it would be better to set up a bastion host and use an ssh tunnel.

ttung: I'll remove this from this PR.  Honestly, I think it's not a big deal, but I'd rather move forward.

lvreynoso (approved):

PR #182, opened 2021-03-11
Initial commit of local development environments

mbarrien:
### Description

Adds local development environments based on Docker for running the frontend and backend of Aspen. Initial instructions on how to use are in [DEV_ENV.md].

Left to document: How to update Pipfile, Pipfile.lock, requirements.txt, and package.json in the new world.

It currently is able to bring up all the containers, but we are still working through issues initializing the database contents, and some CORS issues.


phoenixAja:
these failing tests _are_ strange.. Not sure why it's looking for a module named `py.aspen`, it should just be `aspen`.  I'll need to do some more investigating

phoenixAja:
i think we need to add an `__init__.py` file somewhere, or maybe we can try removing `src/py/__init__.py`, i see you added that in but i'm wondering if that might be causing these errors?

phoenixAja:
yay now the tests are failing for the right reason!

PR #184, opened 2021-03-12
Allow system admins to view all phylo trees

lvreynoso:
### Description

Added code to allow all phylo trees to be seen by a system admin. Basically copied from #171 

#### Issue

### Test plan

Navigate to `/data/phylo_trees`


ttung (approved):
we should probably add the cansee thing here too.

PR #185, opened 2021-03-12
update readme

phoenixAja:
### Description

as i was setting up the app i noticed there were some steps missing from the docs

#### Issue
no issue 

### Test plan
reading?

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ aspen% . .venv/bin/activate
 2. Install pipenv.
 ```bash
 (.venv) aspen% pip install pipenv
-(.venv) aspen%
+(.venv) aspen% 
 ```
 3. Install the packages required for development.
 ```bash
--- a/Pipfile
+++ b/Pipfile
@@ -29,6 +29,7 @@ pytest-xdist = "*"
 sqlalchemy-stubs = "*"
 tqdm = "*"
 auth0-python = "*"
+aspen = {editable = true, path = "./src/py"}
 
 [requires]
 python_version = "3.7"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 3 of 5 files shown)


README.md
```diff
@@ -18,11 +18,13 @@ aspen% . .venv/bin/activate
 2. Install pipenv.
 ```bash
 (.venv) aspen% pip install pipenv
-(.venv) aspen%
+(.venv) aspen% 
 ```
 3. Install the packages required for development.
 ```bash
 (.venv) aspen% pipenv install --dev
+(.venv) aspen% pipenv shell
```
ttung: this shouldn't be necessary.

ttung: I just tested it and it's not required.

phoenixAja: it was for me..

phoenixAja: because when you do `pipenv install --dev` it installs all of the dev packages into a virtual env that you then can't access until you do `pipenv shell`. I'm using conda as my base environment (where really just `pipenv` is installed..). I wonder if using virtualenv is more seemless.

phoenixAja: @lvreynoso did you have to run pipenv shell to access the project packages? Are you using virtualenv too?

ttung: When I did it, `pipenv` said that it detected an existing virtualenv and will use it.  I'm guessing that's the difference.  Are you creating a virtualenv from conda?

lvreynoso: I had the same experience as @ttung since I am using a `pyenv virtualenv` that I created for aspen using python 3.7.9. So I didn't actually have to do Step 1 of the Readme, or worry about doing `pipenv shell`. 

But if that's what you have to do when following our instructions exactly then it should be included :)

ttung: Ok I verified that doing `pipenv shell` is harmless in the virtualenv way.

ttung (approved):

phoenixAja (commented):
@ttung would you mind approving again? i added aspen to the dev packages installed through pipenv and want to know if it looks ok to you to do this.

ttung (approved):

PR #186, opened 2021-03-15
Add one more level to the initial db credentials json file

ttung:
### Description
We had to add one more level to make it a single variable we could pass around in terraform.  The code was incorrectly not updated.

### Test plan
`aspen-cli db set-passwords-from-secret`


ttung:
The style check failure is unrelated, and fixed by #187 or #188

phoenixAja (approved):

PR #188, opened 2021-03-15
Fix flake8 error

ttung:
### Description
Flake8 3.9.0 does not like this redefinition of a variable name that is only used in type checking.  Remove the conflict.

### Test plan
installed flake8 3.9.0, and verified the error goes away.


phoenixAja (approved):
interesting

PR #189, opened 2021-03-15
repin requirements

ttung:
### Description
Repin because why not.

Depends on #188 

### Test plan
github actions yo.

PR #190, opened 2021-03-15
Raise on load for the sequence data

ttung:
### Description
We want to raise an error if we ever erroneously load the sequence data because it is expensive to load erroneously.

This is a new feature of sqlalchemy 1.4, which we are bringing in with #189 

### Test plan
Add a test to verify that the `raiseload` does the expected thing.


phoenixAja (approved):
this is very nifty! 👍

lvreynoso (approved):

PR #191, opened 2021-03-15
Re-add dev access to the postgres db

ttung:
### Description
Setting up the bastion host via terraform and using the bastion host to access the postgres instance via terraform is considerably more effort than we ought to be expending on this at this time.

I added a couple tickets[[1]](https://app.clubhouse.io/genepi/story/126386/add-ssh-bastion-host-that-is-configured-by-terraform)[[2]](https://app.clubhouse.io/genepi/story/126389/existing-tooling-and-terraform-rules-that-access-the-database-should-go-through-the-ssh-bastion-host) to capture this future work.

### Test plan
terraform apply completes

### DIFF
--- a/terraform/security_group.tf
+++ b/terraform/security_group.tf
@@ -28,3 +28,25 @@ resource "aws_security_group_rule" "eb-db-access" {
   protocol                 = "tcp"
   source_security_group_id = aws_security_group.eb-security-group.id
 }
+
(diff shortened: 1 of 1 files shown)


ttung:
Error here is also due to #187 and #188.

phoenixAja (approved):

PR #192, opened 2021-03-15
terraform for nextstrain

phoenixAja:
### Description

terraform build for nextstrain pipeline port over

#### Issue
[ch119790](https://app.clubhouse.io/genepi/stories/space/119790)

### Test plan

`make deploy-tf`
`nextstrain build --aws-batch zika-tutorial/`


ttung:
Can you update the test plan for posterity?

terraform/iam_templates/nextstrain_jobs_access_to_logs.json
```diff
@@ -0,0 +1,18 @@
...
+            "Sid": "VisualEditor0",
+            "Effect": "Allow",
+            "Action": [
+                "logs:GetLogEvents",
+                "logs:FilterLogEvents",
+                "logs:DeleteLogStream"
+            ],
+            "Resource": [
+                "arn:aws:logs:*:*:log-group:/aws/batch/job",
+                "arn:aws:logs:*:*:log-group:/aws/batch/job:log-stream:*"
```
ttung: are these log names prefixed with the job queue or anything?  can we limit the scope?

terraform/nextstrain_batch.tf
```diff
@@ -0,0 +1,179 @@
+variable "nextstrain-batch-params" {
+  type  = map
+  default = {
+    retry_strategy = 1
```
ttung: i don't know that we should retry at all.  most errors are resource limit issues, and i don't think we're going to resolve that with retries.

terraform/nextstrain_batch.tf
```diff
@@ -0,0 +1,179 @@
...
+}
+
+
+data "aws_availability_zones" "available" {}
+
+
+resource "aws_default_vpc" "default" {
+  tags = {
+    Name = "Default VPC"
+  }
```
ttung: actually, i think this can be removed.

terraform/nextstrain_batch.tf
```diff
@@ -0,0 +1,179 @@
...
+  default_route_table_id = aws_default_vpc.default.default_route_table_id
+
+  route {
+    cidr_block = "0.0.0.0/0"
+    gateway_id = data.aws_internet_gateway.default.id
+  }
+
+  tags = {
+    Name = "Default Route Table"
+  }
```
ttung: probably get rid of this too.

terraform/nextstrain_batch.tf
```diff
@@ -0,0 +1,179 @@
...
+    protocol    = "-1"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+}
+
+
+# resource "aws_cloudwatch_log_group" "nextstrain-log-group" {
+#   name = "/aws/batch/job"
+#   retention_in_days = 60
+# }
```
ttung: this should probably be un-commented.

terraform/nextstrain_batch.tf
```diff
@@ -0,0 +1,179 @@
...
+resource "aws_batch_compute_environment" "nextstrain-compute-environment" {
+  compute_environment_name = "aspen-nextstrain"
+
+  compute_resources {
+    instance_role = aws_iam_instance_profile.nextstrain-ecs-instance-role.arn
+    allocation_strategy = "BEST_FIT"
+    ec2_key_pair = "phoenix"
+
+    instance_type = [
+      "m5.2xlarge",
```
ttung: can this be m5 family?

terraform/nextstrain_batch.tf
```diff
@@ -0,0 +1,179 @@
...
+    ]
+
+    max_vcpus = 128
+    min_vcpus = 8
+    desired_vcpus = 8
```
ttung: these should be set to 0.  we don't need to keep instances running.

terraform/nextstrain_batch.tf
```diff
@@ -0,0 +1,179 @@
...
+  })
+}
+
+
+resource "aws_iam_role_policy_attachment" "nextstrain-batch-service-role" {
+  role       = aws_iam_role.nextstrain-batch-service-role.name
+  policy_arn = "arn:aws:iam::aws:policy/service-role/AWSBatchServiceRole"
+}
+
+
```
ttung: why are these iam roles not in `nextstrain_iam.tf`?

terraform/nextstrain_s3.tf
```diff
@@ -0,0 +1,18 @@
...
+  acl = "private"
+
+  # this is what nextstrain docs reccomend
+  lifecycle_rule {
+    enabled = true
+
+    transition {
+      days = 30
+      storage_class = "GLACIER"
+    }
```
ttung: i'd just punt to ONEZONE_IA.  Glacier is such a pain to retrieve anyway.

terraform/nextstrain_batch.tf
```diff
@@ -0,0 +1,179 @@
...
+  block_device_mappings {
+    device_name = "/dev/xvdcz"
+
+    ebs {
+      volume_size = 200
+      volume_type = "gp2"
+      delete_on_termination = true
+    }
+  }
+  user_data = base64encode(file("${path.module}/data/user_data.txt"))
```
ttung: this file is missing from your PR.

ttung (changes requested):

terraform/iam_templates/nextstrain_jobs_access_to_logs.json
phoenixAja: these were the policies that were in the [nextstrain docs](https://docs.nextstrain.org/projects/cli/en/latest/aws-batch/#nextstrainjobsaccesstologs), I didn't modify them when implementing them here. I can play around with limiting the scope more, the log group _should_ just be `/aws/batch/job`

terraform/nextstrain_batch.tf
phoenixAja: should we add different tags?

terraform/nextstrain_batch.tf
phoenixAja: i think so, i can try changing it

terraform/nextstrain_batch.tf
phoenixAja: should i set both min and desired to 0?

terraform/nextstrain_batch.tf
phoenixAja: i'll move them

terraform/nextstrain_s3.tf
phoenixAja: cool i didn't know about ONEZONE_IA

terraform/nextstrain_batch.tf
phoenixAja: ahhh thanks!!

phoenixAja:
@ttung do you think that the overall organization is ok? Looking at this now i was thinking of making a new directory for policies (instead of adding the policies to templates since they're not really templates). Does it also seem ok to you that i have the three seperate files for nextstrain related things (nextstrain_batch.tf, nextstrain_s3.tf, and nextstrain.tf) or do you think i should move all the resources into a nextstrain.tf file? I'm having a hard time deciding

terraform/nextstrain_batch.tf
phoenixAja: these options were specified from the nextstrain setup docs as well, but i can take it out

terraform/data/user_data.txt
```diff
@@ -0,0 +1,10 @@
+Content-Type: multipart/mixed; boundary="==BOUNDARY=="
+MIME-Version: 1.0
+
+--==BOUNDARY==
+Content-Type: text/cloud-boothook; charset="us-ascii"
+
+# Set Docker daemon option dm.basesize so each container gets up to 50GB
+cloud-init-per once docker_options echo 'OPTIONS="${OPTIONS} --storage-opt dm.basesize=50GB"' >> /etc/sysconfig/docker
```
ttung: If we're provisioning 200GB EBS volumes, then we should probably set this to 200GB.

terraform/iam_templates/nextstrain_jobs_access_to_logs.json
ttung: the log stream name is always `nextstrain-job-definition/default/*`, can we take advantage of that somehow?

terraform/nextstrain_batch.tf
ttung: Probably not necessary at the moment.

ttung:
> @ttung do you think that the overall organization is ok? Looking at this now i was thinking of making a new directory for policies (instead of adding the policies to templates since they're not really templates). Does it also seem ok to you that i have the three seperate files for nextstrain related things (nextstrain_batch.tf, nextstrain_s3.tf, and nextstrain.tf) or do you think i should move all the resources into a nextstrain.tf file? I'm having a hard time deciding

yeah, i think this is okay.  we can always rearrange as needed.

terraform/nextstrain_batch.tf
ttung: I'd consider naming this something more specific.

terraform/nextstrain_batch.tf
```diff
@@ -0,0 +1,133 @@
...
+}
+
+
+resource "aws_batch_compute_environment" "nextstrain-compute-environment" {
+  compute_environment_name = "aspen-nextstrain"
+
+  compute_resources {
+    instance_role = aws_iam_instance_profile.nextstrain-ecs-instance-role.arn
+    allocation_strategy = "BEST_FIT"
+    ec2_key_pair = "phoenix"
```
ttung, line 91: Can you make a TODO or an issue to make a common ec2 key?

terraform/nextstrain_s3.tf
```diff
@@ -0,0 +1,18 @@
+resource "aws_s3_bucket" "nextstrain_jobs" {
+  bucket = "aspen-nextstrain-jobs"
+  acl = "private"
+
+  # this is what nextstrain docs reccomend
```
ttung:
```suggestion
  # this is what nextstrain docs recommend
```

terraform/nextstrain_batch.tf
phoenixAja: yeah! i was going to ask you if you knew how to do that, should i create a new key that we can all just share? or is there a better way, that way i mentioned sounds sketch

phoenixAja: i'll add it as a TODO for now and we can fix later

terraform/nextstrain_batch.tf
```diff
@@ -0,0 +1,130 @@
...
+  block_device_mappings {
+    device_name = "/dev/xvdcz"
+
+    ebs {
+      volume_size = 200
+      volume_type = "gp2"
+      delete_on_termination = true
+    }
+  }
+  user_data = base64encode(file("${path.module}/advanced_options/user_data.txt"))
```
ttung, line 79: ah, this is my bad, i misread and didn't realize the field was literally `user_data`.  bad choice on aws's part!

terraform/nextstrain_batch.tf
```diff
@@ -0,0 +1,130 @@
...
+    launch_template {
+      launch_template_id = aws_launch_template.nextstrain-launch-template.id
+    }
+  }
+
+  service_role = aws_iam_role.nextstrain-batch-service-role.arn
+  type = "MANAGED"
+  depends_on   = [
+    aws_iam_role_policy_attachment.nextstrain-batch-service-role
+  ]
```
ttung, line 118: is this necessary?  since the service_role is already referencing this, it shouldn't be required.

ttung (approved):

terraform/nextstrain_batch.tf
phoenixAja: haha yeah

terraform/nextstrain_batch.tf
phoenixAja: To prevent a race condition during environment deletion, you have to set this, otherwise the policy may be destroyed too soon and the compute environment will then get stuck in the DELETING state (mentioned in the terraform docs)

ttung: Got it!

PR #193, opened 2021-03-17
Forward tree json files to nextstrain with magic urls

lvreynoso:
### Description

Gets a presigned url for an s3 object and returns it as a "magic" url that can be fed into nextstrain. When this url is accessed, the server simply redirects to the presigned url. The presigned url has an expiration of 5 minutes.

Example:
Screenshot from 2021-03-16 22-11-01

#### Issue
[ch119787](https://app.clubhouse.io/genepi/stories/space/119787)

### Test plan

### DIFF
--- a/src/py/aspen/app/views/magic_url.py
+++ b/src/py/aspen/app/views/magic_url.py
@@ -2,7 +2,7 @@ import json
 from urllib.parse import urlparse, parse_qsl
 
 import boto3
-from flask import jsonify, session, request
+from flask import jsonify, session, request, url_for, redirect
 
 from aspen.app.app import application, requires_auth
 from aspen.database.connection import session_scope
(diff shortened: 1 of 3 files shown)


src/py/aspen/app/views/magic_url.py
```diff
@@ -0,0 +1,50 @@
...
+@requires_auth
+def get_magic_url(object_name: str, object_id: int):
+    object_type = MAGIC_TYPES.get(object_name)
+    with session_scope(application.DATABASE_INTERFACE) as db_session:
+        try:
+            target_object = (
+                db_session.query(object_type)
+                .filter(object_type.entity_id == object_id)
+                .one()
+            )
```
ttung, line 31: this does not verify that the logged-in user has permission to see this object.

src/py/aspen/app/views/magic_url.py
```diff
@@ -0,0 +1,50 @@
...
+        except:
+            return jsonify({})
+
+        s3_client = boto3.client("s3")
+
+        presigned_url = (
+            s3_client.generate_presigned_url(
+                "get_object",
+                Params={ "Bucket": target_object.s3_bucket, "Key": target_object.s3_key },
+                ExpiresIn=300 # 5 minutes
```
ttung, line 41: I would make this longer. I don't think it has to be so short.

src/py/aspen/app/views/magic_url.py
```diff
@@ -0,0 +1,50 @@
...
+            )
+        )
+
+        parse_result = urlparse(presigned_url)
+        netloc_path = parse_result.netloc + parse_result.path
+
+        magic_url_string = url_for("magic_url", _external=True, query_string=parse_result.query, location=netloc_path)
+
```
ttung, line 49: Why not just [quote](https://docs.python.org/3.3/library/urllib.parse.html?highlight=urlencode#urllib.parse.quote_plus) the presigned url and add it as a param?

lvreynoso:
Going to take a different approach - closing this for now.

PR #194, opened 2021-03-18
Docker image to run batch jobs

ttung:
### Description
This docker image is based on nextstrain/base and sets up a virtualenv with aspen's dependencies installed.  This can be used to invoke a variety of workflows.

#### Issue
[ch126778](https://app.clubhouse.io/genepi/story/126778)

### Test plan
docker run -it --rm aspen/batch:latest trunk ls


phoenixAja (approved):
so will we be using this container in the nextstrain job definition then?

ttung:
> so will we be using this container in the nextstrain job definition then?

No, this is the aspen batch job image.

PR #197, opened 2021-03-19
Dockerize aspen.

jgadling:
### Description

This PR updates aspen to be able to run locally in a docker-compose controlled environment.

Some important changes to note:
- This upgrades the app to python 3.9
- Everything python-related is now in src/py
- The frontend container looks for a BACKEND_URL env variable to generate API url's, and the backend container uses a FRONTEND_URL to generate a redirect to the frontend after login
- This application uses a locally-hosted OIDC service for auth rather than Auth0

#### Issue

### Test plan


.github/workflows/py-tests.yml
```diff
@@ -23,11 +23,11 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 2
-      - name: Set up Python 3.7
+      - name: Set up Python 3.9
```
ttung, line 26: What's the rationale behind bumping it to 3.9?  Is this PR sufficient to replace our "staging" environment?  If not, we do not have the ability to stand this version up in Elastic Beanstalk.

Makefile
```diff
@@ -97,4 +105,96 @@ unit-tests:
...
+.PHONY: local-init
+local-init: oauth/pkcs12/certificate.pfx .env.ecr local-ecr-login ## Launch a new local dev env and populate it with test data.
+	docker-compose $(COMPOSE_OPTS) up -d
+	# docker-compose exec -T utility pip3 install awscli
+	docker-compose exec -T utility $(BACKEND_APP_ROOT)/scripts/setup_dev_data.sh
+	docker-compose exec -T utility python scripts/setup_localdata.py
+	docker-compose exec -T utility pip install .
+	-@docker-compose exec -e PGPASSWORD=$(LOCAL_DB_RW_PASSWORD) database psql -h localhost -d $(LOCAL_DB_NAME) -U $(LOCAL_DB_RW_USERNAME) -c "CREATE USER $(LOCAL_DB_RO_USERNAME) WITH PASSWORD '$(LOCAL_DB_RO_PASSWORD)';"
+	docker-compose exec -T utility aspen-cli db --docker create
+	docker-compose exec -T utility alembic upgrade head
```
ttung: does this detect if an existing docker composition is running?

DEV_ENV.md
```diff
@@ -0,0 +1,57 @@
...
+### Containers managed by the dev environment
+The Aspen dev environment is a set of containers defined in [docker-compose.yml](docker-compose.yml). The [backend docker image](src/py/Dockerfile) and [frontend docker image](src/ts/Dockerfile) are built locally. Update any of these files as necessary and run `make local-sync` to sync your dev environment with these configs.
+
+### Updating frontend/backend dependencies
+Both the frontend and backend services will automatically reload when their source code is modified, but they won't automatically rebuild when their dependencies (such as npm or pip package lists) change.
+
+To update the dev env to reflect changes to [src/ts/package.json](src/ts/package.json) or [src/py/Pipfile](src/py/Pipfile), run `make local-sync`
+
+### Update Dev Data
+The dev environment is initialized with AWS Secrets/S3 data in the [src/py/scripts/setup_dev_data.sh](src/py/scripts/setup_dev_data.sh) script, as well as DB data from [src/py/database_migrations](src/py/database_migrations). To add more data, modify these scripts and run `make local-init` to reload the dev environment's data stores.
```
ttung: there's not really any data being introduced by `src/py/database_migrations`.  the entries being jammed into the db are just for enums.

there does appear to be data being written by `setup_localdata.py` though.

Makefile
```diff
@@ -97,4 +105,96 @@ unit-tests:
...
+	docker-compose stop
+
+.PHONY: local-clean
+local-clean: ## Remove everything related to the local dev environment (including db data!)
+	-if [ -f ./oauth/pkcs12/server.crt ] ; then \
+	    export CERT=$$(docker run -v $(PWD)/oauth/pkcs12:/tmp/certs --workdir /tmp/certs --rm=true --entrypoint "" soluto/oidc-server-mock:0.3.0 bash -c "openssl x509 -in server.crt -outform DER | sha1sum | cut -d ' ' -f 1"); \
+	    echo ""; \
+	    echo "Removing this certificate requires sudo access"; \
+	    sudo security delete-certificate -Z $${CERT} /Library/Keychains/System.keychain; \
+	fi;
```
ttung: the `oauth/pkcs12/certificate.pfx` rule supports linux but this rule does not.  mildly odd.

docker-compose.yml
```diff
@@ -0,0 +1,136 @@
...
+  database:
+    image: "${DOCKER_REPO}aspen-devdb:latest"
+    ports:
+      - "5432:5432"
+    environment:
+      - POSTGRES_USER=user_rw
+      - POSTGRES_PASSWORD=password_rw
+      - POSTGRES_DB=aspen_db
+    #volumes:
+    #  - database:/var/lib/postgresql/data
```
ttung: why is this here and commented out?

oauth/pkcs12/generate_cert.sh
```diff
@@ -0,0 +1,9 @@
+if [ ! -f server.crt ]; then
+  openssl req -x509 -newkey rsa:4096 -sha256 -days 3560 -nodes -keyout server.key -out server.crt -subj '/CN=localhost' -extensions san -nodes -config <( \
+    echo '[req]'; \
+    echo 'distinguished_name=req'; \
+    echo '[san]'; \
+    echo 'subjectAltName=DNS:localhost')
+else
+    echo "Cert already exists, skipping"
```
ttung, line 8: is it possible to shove something into the cert / key that makes it easy to distinguish from other keys? maybe jam in the word aspen somewhere in there.

self-signed certs make me uncomfortable, and it would be nice to be able to find them if I botch a `local-clean`.

docker-compose.yml
```diff
@@ -0,0 +1,136 @@
...
+          "Discovery": {
+            "ShowKeySet": true
+          }
+        }
+      API_SCOPES_PATH: /tmp/config/api_scopes.json
+      API_RESOURCES_PATH: /tmp/config/api_resources.json
+      USERS_CONFIGURATION_PATH: /tmp/config/users.json
+      CLIENTS_CONFIGURATION_PATH: /tmp/config/clients-config.json
+      ASPNETCORE_URLS: https://+:443;http://+:80
+      # ASPNETCORE_Kestrel__Certificates__Default__Password: <password for pfx file>
```
ttung, line 132: ?

src/py/aspen/app/app.py
```diff
@@ -5,9 +5,16 @@
...
 from authlib.integrations.flask_client import OAuth
 from flask import redirect, session
+from flask_cors import CORS
 
 from aspen.app.aspen_app import AspenApp
-from aspen.config import config, DevelopmentConfig, ProductionConfig, StagingConfig
+from aspen.config import (
+    config,
+    DevelopmentConfig,
+    LocalConfig,
```
ttung, line 14: is there a reason to keep around `DevelopmentConfig`?

src/py/aspen/config/config.py
```diff
@@ -172,21 +174,40 @@ def AUTH0_DOMAIN(self) -> str:
...
+            return {
+                "scope": "openid profile email",
+            }
+
+    @property
+    def AUTH0_USERINFO_URL(self) -> str:
+        try:
+            return self.AWS_SECRET["AUTH0_USERINFO_URL"]
+        except KeyError:
+            return "userinfo"
```
ttung, line 210: Why the fallbacks?  When do we expect the secret not to be present?  It seems you pre-populate the secretsmanager in the mock docker instance.

Makefile
```diff
@@ -97,4 +105,96 @@ unit-tests:
...
+	if PROFILE=$$(aws configure list-profiles | grep $(AWS_DEV_PROFILE)); then \
+		aws ecr get-login-password --region us-west-2 --profile $(AWS_DEV_PROFILE) | docker login --username AWS --password-stdin $$(aws sts get-caller-identity --profile $(AWS_DEV_PROFILE) | jq -r .Account).dkr.ecr.us-west-2.amazonaws.com; \
+	fi
+
+.PHONY: local-init
+local-init: oauth/pkcs12/certificate.pfx .env.ecr local-ecr-login ## Launch a new local dev env and populate it with test data.
+	docker-compose $(COMPOSE_OPTS) up -d
+	# docker-compose exec -T utility pip3 install awscli
+	docker-compose exec -T utility $(BACKEND_APP_ROOT)/scripts/setup_dev_data.sh
+	docker-compose exec -T utility python scripts/setup_localdata.py
```
ttung, line 146: why do this?  also, how does the database get written to before the `db --docker create` command?

Makefile
ttung, line 149: why the schema update?

docker-compose.yml
```diff
@@ -0,0 +1,136 @@
+version: '3.8'
+
+services:
+  database:
+    image: "${DOCKER_REPO}aspen-devdb:latest"
```
ttung, line 5: where does this image come from?  ECR?  how does it get updated?

ttung (commented):
probably a dumb question, but why does the migration files need to go into the docker image?

.github/workflows/py-tests.yml
jgadling: Just upgrading to the latest available, and we're going to be migrating from Elastic Beanstalk to ECS shortly, where we don't have any python version limitations.

Makefile
jgadling: docker-compose has a declarative interface, it'll just update the existing stack if there's already one running.

DEV_ENV.md
jgadling: No data in database_migrations *yet* ... but I updated the docs to talk about migrations instead of data.

Makefile
jgadling: Docker-compose is launching the db container at `docker-compose up -d` and the image we're pulling is your snapshot from S3.

Makefile
jgadling: Just in case the db snapshot is out of date -- any necessary migrations between the DB snapshot version and the current state of the repo will get run.

docker-compose.yml
jgadling: Yep, ECR. At the moment it's updated manually, but we'll need to work out a plan for how/when to refresh it.

src/py/aspen/app/app.py
jgadling: Right now we're setting up an alternate development path -- the old setup should continue to work for the most part, and each engineer can switch over to dockerized local dev at their own pace. Once we're happy with dockerized dev, we can remove the old configs and docs.

Makefile
jgadling: Yeah, basically we never need to run local-clean in the only linux environments we've worked with so far (CI systems) but updated anyway!

src/py/aspen/config/config.py
jgadling: Just for reverse-compatibility with the existing dev/staging environments.

Makefile
ttung: Are all these operations idempotent then?

src/py/aspen/app/app.py
ttung: Got it.

src/py/aspen/config/config.py
ttung: Where do we have environments that don't have secrets populated?

docker-compose.yml
ttung: Do you have a list of future work you're tracking?

.github/workflows/py-tests.yml
ttung: Then you might as well trash `StagingConfig` because it wouldn't work, right?

src/py/aspen/config/config.py
phoenixAja: @ttung the local dev environment uses oidc instead of auth0 so these secrets wouldn't be populated for local

docker-compose.yml
phoenixAja: how often do you think this would need to be updated?

src/py/aspen/config/config.py
ttung: So couldn't the local dev environment define these functions?

Makefile
jgadling: Yep.

phoenixAja (approved):

src/py/aspen/config/config.py
jgadling: It's the current EBS staging env that doesn't.

ttung: The current staging env doesn't define them because it falls back on the methods here, but the secret for the staging environment has these keys.  I'm asking which environment has a secret that doesn't have these keys defined.

Makefile
```diff
@@ -95,6 +104,112 @@ update-deps :
 unit-tests:
 	pytest -n8 src/py/aspen
 
+help: ## display help for this makefile
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+.PHONY: help
+
 #
 ##############################################################
-include terraform.mk
```
ttung, line 100: Why was this removed?

jgadling: Just a typo or merge issue

ttung: ok np, I readded this in #194

PR #199, opened 2021-03-22
Add aspen terraform batch configuration

ttung:
### Description
This batch configuration will be used to run workflows like download from gisaid.

Depends on #198 

### Test plan
Ran a couple small jobs through the pipeline.

### DIFF
--- a/terraform/aspen_batch.tf
+++ b/terraform/aspen_batch.tf
@@ -1,0 +1,117 @@
+variable "aspen-batch-params" {
+  type  = map
+  default = {
+    timeout = 14400
+  }
+}
+
+resource "aws_batch_job_definition" "aspen-batch-job-definition" {
+  name = "aspen-batch-job-definition"
+  type = "container"
+  parameters = var.aspen-batch-params
+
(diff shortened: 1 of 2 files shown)


phoenixAja (approved):

phoenixAja:
this is all really similar to the nextstrain batch stuff besides using a different docker image for the job definition.. when integrating with the shared infra stuff i think we can remove a lot of this duplication by using the batch module

ttung:
> this is all really similar to the nextstrain batch stuff besides using a different docker image for the job definition.. when integrating with the shared infra stuff i think we can remove a lot of this duplication by using the batch module

different disk space requirement as well.  i think it would probably make sense to use R5 instances here as well, since we probably need a lot of memory and not a lot of CPU.

ttung:
but yeah, it would probably be good to figure out how to dedup the code.

phoenixAja:
> but yeah, it would probably be good to figure out how to dedup the code.

once we convert it to use the shared infra recipes i think we can dedup a lot of this (by using the batch modules that they created)

PR #202, opened 2021-03-22
Combine batch service role / ecs instance roles for the pipelines

ttung:
### Description
No need to have two copies.

### Test plan
`make deploy-tf`

### DIFF
--- a/terraform/aspen_batch.tf
+++ b/terraform/aspen_batch.tf
@@ -69,7 +69,7 @@ resource "aws_batch_compute_environment" "aspen-batch-compute-environment" {
   compute_environment_name_prefix = "aspen-batch-"
 
   compute_resources {
-    instance_role = aws_iam_instance_profile.aspen-batch-ecs-instance-role.arn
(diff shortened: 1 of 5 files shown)


terraform/aspen_batch.tf
```diff
@@ -69,7 +69,7 @@ resource "aws_batch_compute_environment" "aspen-batch-compute-environment" {
   compute_environment_name_prefix = "aspen-batch-"
 
   compute_resources {
-    instance_role = aws_iam_instance_profile.aspen-batch-ecs-instance-role.arn
+    instance_role = aws_iam_instance_profile.batch-ecs-instance-role.arn
```
phoenixAja, line 72: did you think adding the aspen prefix is too redundant?

phoenixAja (approved):

terraform/aspen_batch.tf
ttung: uh, I don't have a strong opinion on that.

PR #203, opened 2021-03-23
Grant aspen batch jobs additional permissions

ttung:
### Description
1. This allows aspen batch jobs access to the secret that holds the gisaid credentials.
1. This allows aspen batch jobs access to the aspen-config secret.
1. This allows aspen batch jobs access to create more batch jobs.
1. This allows aspen batch jobs access to S3.

Depends on #202, #204

#### Issue

### Test plan
`make deploy-tf`

### DIFF
--- a/terraform/aspen_batch_iam.tf
+++ b/terraform/aspen_batch_iam.tf
@@ -1,6 +1,65 @@
+data "aws_secretsmanager_secret" "gisaid-download-credentials" {
+  name = "gisaid-download-credentials"
+}
+
 resource "aws_iam_role" "aspen-batch-jobs-role" {
   name = "aspen-batch-jobs-role"
   assume_role_policy = templatefile("${path.module}/iam_templates/trust_policy.json", {
     trust_services = ["ecs-tasks"]
   })
+}
+
+
+resource "aws_iam_policy" "aspen-batch-jobs-policies" {
+  name = "aspen-batch-jobs-policies"
+  policy = <<-POLICY
+  {
+      "Version": "2012-10-17",
+      "Statement": [
+          {
+              "Effect": "Allow",
+              "Action": "secretsmanager:GetSecretValue",
+              "Resource": [
+                  "${data.aws_secretsmanager_secret.aspen_config.arn}",
+                  "${data.aws_secretsmanager_secret.gisaid-download-credentials.arn}"
+              ]
+          },
+          {
+              "Effect": "Allow",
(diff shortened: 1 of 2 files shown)


terraform/aspen_batch_iam.tf
```diff
@@ -1,6 +1,65 @@
...
   name = "aspen-batch-jobs-role"
   assume_role_policy = templatefile("${path.module}/iam_templates/trust_policy.json", {
     trust_services = ["ecs-tasks"]
   })
 }
+
+
+resource "aws_iam_policy" "aspen-batch-jobs-policies" {
+  name = "aspen-batch-jobs-policies"
+  policy = <<-POLICY
```
phoenixAja, line 15: is this syntax correct or is it supposed to be `<<POLICY`? Why use this syntax instead of `<<EOF` (which seems to be what the terraform docs prefer)

terraform/aspen_batch_iam.tf
```diff
@@ -1,6 +1,65 @@
...
+          {
+              "Effect": "Allow",
+              "Action": [
+                  "s3:GetObject",
+                  "s3:PutObject"
+              ],
+              "Resource": [
+                  "arn:aws:s3:::${aws_s3_bucket.aspen-data.bucket}/raw_gisaid_dump/*",
+                  "arn:aws:s3:::${aws_s3_bucket.aspen-data.bucket}/processed_gisaid_dump/*",
+                  "arn:aws:s3:::${aws_s3_bucket.aspen-data.bucket}/aligned_gisaid_dump/*"
```
phoenixAja: would be nice to move these key names to their own local variables eventually, just in case we need to use them in any other resource down the road.

phoenixAja (approved):

terraform/aspen_batch_iam.tf
ttung: <<- allows you to indent the input.  see https://www.terraform.io/docs/language/expressions/strings.html#indented-heredocs

terraform/aspen_batch_iam.tf
ttung: Yeah, I don't know if there's a way to share strings between terraform and python... @jgadling ?

terraform/aspen_batch_iam.tf
phoenixAja: oh sweet didn't know that

PR #204, opened 2021-03-23
Bucket for data

ttung:
### Description
This brings the aspen-data-dev bucket under terraform management.

### Test plan
```
% terraform import module.aspen.aws_s3_bucket.aspen-data aspen-data-dev
% make deploy-tf
```

### DIFF
--- a/terraform/s3.tf
+++ b/terraform/s3.tf
@@ -1,0 +1,4 @@
+resource "aws_s3_bucket" "aspen-data" {
+  bucket = join("", ["aspen-db-data-", var.DEPLOYMENT_ENVIRONMENT])
+  acl = "private"
+}


terraform/s3.tf
```diff
@@ -0,0 +1,4 @@
+resource "aws_s3_bucket" "aspen-data" {
+  bucket = join("", ["aspen-data-", var.DEPLOYMENT_ENVIRONMENT])
```
phoenixAja: is this bucket going to be holding data relating to certain db fields (like `s3_bucket` fields in the db)? if so i think a more specific name like `aspen-db-data` might be better, just in case we have other aspen data stored on s3 later on? `aspen-app-data` might also be a good name.

phoenixAja (approved):

terraform/s3.tf
ttung: Renamed to `aspen-db-data-*`.

I hand-modified the db entries in the staging db, and am copying the data to the new bucket...

PR #205, opened 2021-03-23
Processed gisaid dumps are stored in two separate files (metadata and sequences)

ttung:
### Description
See description.

#### Issue

### Test plan
applied migration to local database


lvreynoso (approved):
LGTM. "Processed gisaid dump" is kind of a funny name but I guess it will do lol

PR #208, opened 2021-03-24
[easy] standardized software names for recording application versions

ttung:
### Description
Workflows record the versions of the tools used in the workflow, but we need a standardized names to reference the tools.  This is the start of this set of standardized names.

### Test plan
used in subsequent PR.


phoenixAja (approved):
i really like this, would also be nice to add this information to the docs somewhere.

PR #209, opened 2021-03-24
[easy] Add missing type annotation

ttung:
### Description
Add one missing type annotation that improves quality of life. :)

### Test plan
definitely YOLO.

### DIFF
--- a/src/py/aspen/database/connection.py
+++ b/src/py/aspen/database/connection.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 import time
 from contextlib import contextmanager
-from typing import TYPE_CHECKING
+from typing import Generator, TYPE_CHECKING
 
 from sqlalchemy import event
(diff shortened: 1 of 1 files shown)


src/py/aspen/database/connection.py
```diff
@@ -32,7 +32,7 @@ def init_db(db_uri: str) -> SqlAlchemyInterface:
 
 
 @contextmanager
-def session_scope(interface: SqlAlchemyInterface):
+def session_scope(interface: SqlAlchemyInterface) -> Iterator[Session]:
```
phoenixAja: when should we use Iterator vs Generator for the type hints? Generator as the type hint here seems a little more specific to me.

phoenixAja (approved):

src/py/aspen/database/connection.py
ttung: Switched to `Generator`; I agree that it makes more sense.

PR #210, opened 2021-03-24
Schema changes to support an explicit alignment step for gisaid data

ttung:
### Description
New workflow and entity to represent aligned data!

#### Issue
[ch126871](https://app.clubhouse.io/genepi/story/126871/schema-changes-to-support-cached-alignments)

### Test plan
Updated local db.


shanaxel42 (approved):
LGTM!

PR #211, opened 2021-03-25
Ingest and transform workflows for gisaid data

ttung:
### Description
This encompasses the workflows for both the ingest and transform workflows.

Some interesting notes:
1. I did not add an object to the db indicating workflow start (followed by an update to reflect workflow completion).  Rationale is that these workflows are so simple and mostly not-user-facing so that it's really not worth tracking failed workflows.
1. To circumvent the issue where the db/codebase may have moved forward from the time we start the workflow and the time we are ready to write to the DB, we update the local checkout.  This implies that we should be running workflows against a git tag that accurately reflects the state of the database.  i.e., if we run a migration on the database, it is important to advance the tag.

Depends on #206, #207, #208

#### Issue

### Test plan
Successfully ran the workflows!

### DIFF
--- a/workflows/ingest_gisaid/ingest.sh
+++ b/workflows/ingest_gisaid/ingest.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+
+set -Eeuo pipefail
+shopt -s inherit_errexit
+
+# install jq
+apt-get install -y jq
+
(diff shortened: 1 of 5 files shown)


phoenixAja (approved):
looks good to me!

PR #212, opened 2021-03-25
Update style checking to handle workflow directories

ttung:
### Description
mypy struggles when dealing with top-level scripts that don't belong in packages, so for each workflow directory needs to be checked in a separate mypy execution.  This adapts our workflow to invoke style checking on each directory that containes unmodularized python code separately.

Depends on #211

### Test plan
make -j style


ttung:
The alternative is to migrate all the scripts inside the aspen package, but it makes it harder to find the relevant code.

phoenixAja (approved):

shanaxel42 (approved):
Gotta have that style checking!! Especially if I ever contribute ; )

PR #213, opened 2021-03-25
fine tune where we ignore missing imports

ttung:
### Description
Rather than ignoring all missing imports, be explicit about which missing imports are okay.  It's discouraged in the [docs](https://mypy.readthedocs.io/en/stable/running_mypy.html), but I'm not sure if it's that valuable.  Open to a debate whether we should do this or not.

Depends on #212

### Test plan
make -j style

### DIFF
--- a/.github/workflows/py-tests.yml
+++ b/.github/workflows/py-tests.yml
@@ -39,7 +39,7 @@ jobs:
         working-directory: src/py
         run: |
           python -m pip install pipenv
-          pip install $(pipenv lock --requirements --dev | egrep '^(flake8|black|isort|mypy|sqlalchemy-stubs)==')
+          pip install $(pipenv lock --requirements --dev | egrep '^(flake8|black|isort|mypy|sqlalchemy-stubs|pytest)==')
       - name: Verify code style
         run: |
           make -k style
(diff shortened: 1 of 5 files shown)


phoenixAja:
did you mean to delete `src/py/aspen/__main__.py`?

ttung:
> did you mean to delete src/py/aspen/__main__.py?

yeah, do we still need it? it refers to modules that don't even exist any more, so I would be surprised if we need it.

phoenixAja:
so whenever we add a new third party library would we have to update` .mypy.ini`? we should add something to the docs about when/how to update that file.

phoenixAja:
> > did you mean to delete src/py/aspen/**main**.py?
> 
> yeah, do we still need it? it refers to modules that don't even exist any more, so I would be surprised if we need it.

i guess you're right i was just checking

phoenixAja (approved):

shanaxel42 (approved):

ttung:
> so whenever we add a new third party library would we have to update .mypy.ini? we should add something to the docs about when/how to update that file.

It's for third-party libraries that do not support typing that we would need to do this.

ttung:
added [docs](https://github.com/chanzuckerberg/aspen/pull/213/files#diff-7a686e36c672a51917b2c3fc53bbf84890ca2ce8383c8caa5395ecf949fbe609).

PR #214, opened 2021-03-25
[easy] Allow both m5 and r5 instances for running jobs on the aspen batch queue

ttung:
### Description
ingest and transform are both very single threaded, and require r5 instances.  align requires both cpu and memory, and m5 instances are more suitable.

### Test plan
test in prod.


ttung:
This may not be necessary, as align with 96 cores is only marginally faster than 16 (1h vs 1h40m).  It may make more sense to stick with r5.12xlarge (the smallest machine that can fit the dataset in memory).

ttung:
Investigation suggests that there is a long single threaded portion with a short multi threaded portion.

PR #218, opened 2021-03-26
align workflow

ttung:
### Description
This is the workflow to align the gisaid data.

Depends on #211, #212

#### Issue
[ch127651](https://app.clubhouse.io/genepi/story/127651/implement-caching-of-alignment-in-nexstrain)

### Test plan
successfully invoked ingest -> transform -> align

align crashes, but that's because I under-provisioned the EC2 instance.  Apparently mafft uses a ton of scratch space.  (I opened #217 to fix that)

EDIT: working [run](https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logsV2:log-groups/log-group/$252Faws$252Fbatch$252Fjob/log-events/aspen-batch-job-definition$252Fdefault$252Ff093afc269174c368b369dbc480793b1)

### DIFF
--- a/workflows/align_gisaid/align.sh
+++ b/workflows/align_gisaid/align.sh
@@ -1,0 +1,73 @@
+#!/bin/bash
+
+set -Eeuo pipefail
+shopt -s inherit_errexit
+
+if [ $# -ne 1 ]; then
+    echo "Usage: $0 <processed_gisaid_object_id>"
+    exit 1
+fi
+
+# install jq
+apt-get install -y jq
+
+# get the bucket/key from the object id
+processed_gisaid_location=$(/aspen/.venv/bin/python workflows/align_gisaid/lookup_processed_gisaid_object.py --processed-gisaid-object-id "${1}")
+processed_gisaid_s3_bucket=$(echo "${processed_gisaid_location}" | jq -r .bucket)
+processed_gisaid_sequences_s3_key=$(echo "${processed_gisaid_location}" | jq -r .sequences_key)
+processed_gisaid_metadata_s3_key=$(echo "${processed_gisaid_location}" | jq -r .metadata_key)
+
+start_time=$(date +%s)
+build_id=$(date +%Y%m%d-%H%M)
+
+# set up ncov
+mkdir -p /ncov/my_profiles/align
+cd /ncov
+git init
+git fetch git://github.com/nextstrain/ncov.git
+git checkout FETCH_HEAD
+ncov_git_rev=$(git rev-parse HEAD)
+cp /aspen/workflows/align_gisaid/config.yaml /ncov/my_profiles/align
+
+# fetch the gisaid dataset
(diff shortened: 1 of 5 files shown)


workflows/align_gisaid/align.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
+
+set -Eeuo pipefail
```
shanaxel42, line 3: This is more just a curiosity question not at all a suggestion, but is there a reason you wrote this workflow in bash instead of python?

workflows/align_gisaid/config.yaml
```diff
@@ -0,0 +1,10 @@
+configfile:
+  - defaults/parameters.yaml
+
+# NOTE: when the number of cpus is modified, it would be prudent to modify workflows/transform_gisaid/transform.sh.
+cores: 48
```
shanaxel42: damnnnnnn that's a lot of cores

workflows/transform_gisaid/transform.sh
```diff
@@ -75,3 +75,16 @@ entity_id=$(/aspen/.venv/bin/python workflows/transform_gisaid/save.py
...
+# invoke the next workflow.
+# NOTE: when the number of cpus is modified, it would be prudent to modify workflows/align_gisaid/config.yaml.
+aws batch submit-job \
+    --job-name "align-gisaid"                \
+    --job-queue aspen-batch                      \
+    --job-definition aspen-batch-job-definition  \
+    --container-overrides "
+      {
+        \"command\": [\"${ASPEN_GIT_REVSPEC}\", \"workflows/align_gisaid/align.sh\", \"${entity_id}\"],
+        \"vcpus\": 48,
```
shanaxel42: Is this a redefinition from the config.yaml above?

shanaxel42 (approved):
lg with some questions

workflows/align_gisaid/align.sh
ttung: A monolithic python script would not work if we want to pick up updates to aspen.  We could do a python script that starts up another python script for the save portion.  Not sure that'd be significantly simpler in python, but I guess it could plausibly be more readable.

workflows/align_gisaid/config.yaml
ttung: Well, we could probably do with fewer cores, but there aren't machines with 384GB of memory with fewer cores.

workflows/transform_gisaid/transform.sh
ttung: No, it's separate.

workflows/align_gisaid/config.yaml
ttung: ok so it seems 384GB of memory isn't enough to consistently finish the alignment.  I increased the memory requirement so now we are using r5.16xlarge (64 cores, 512GB RAM), but decreased the cores because it's not really that helpful beyond a certain point.  hopefully we can schedule other things onto those machines.

PR #221, opened 2021-03-26
[easy] Change bucket name

ttung:
### Description
It was changed in a later revision of #204.

### Test plan
will test a run.


phoenixAja (approved):
aww thanks ❤️

PR #227, opened 2021-03-30
We don't need to create tables if we're starting with a db snapshot

jgadling:
### Description

Local dockerized dev env was running into trouble running migrations since this `aspen-cli db create` step isn't idempotent. Since we're starting the local dev db as a snapshot, we don't need this.

#### Issue

### Test plan

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -154,7 +154,6 @@ local-init: oauth/pkcs12/certificate.pfx .env.ecr local-ecr-login ## Launch a ne
 	docker-compose exec -T utility $(BACKEND_APP_ROOT)/scripts/setup_dev_data.sh
 	docker-compose exec -T utility python scripts/setup_localdata.py
 	docker-compose exec -T utility pip install .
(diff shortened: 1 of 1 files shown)


ttung (approved):
seems reasonable.

longer term question: is there a path to bootstrapping things without a snapshot?

jgadling:
> longer term question: is there a path to bootstrapping things without a snapshot?

I'd like to be able to bootstrap purely from migrations if that's possible, and then run scripts that import (or better yet, auto-generate) a reasonable set of test data. Did you have something in mind?

ttung:
> I'd like to be able to bootstrap purely from migrations if that's possible, and then run scripts that import (or better yet, auto-generate) a reasonable set of test data. Did you have something in mind?

I don't really have something in mind.  I just think that we need a strategy that allows us to get to a database that's not reliant on a snapshot.  It's not a bad idea to bootstrap purely from migrations IMO because at the very least it exercises the migrations code.

PR #229, opened 2021-03-30
Event trigger for ingest-gisaid

ttung:
### Description
event and target for triggering the ingest-gisaid workflow.

Depends on #228 

#### Issue
[ch126779](https://app.clubhouse.io/genepi/story/126779/cloudwatch-trigger-to-invoke-gisaid-workflow)

### Test plan
Changed it to every 15 minutes, and watched it kick off a build.

### DIFF
--- a/terraform/events.tf
+++ b/terraform/events.tf
@@ -1,0 +1,46 @@
+resource "aws_iam_role" "cloudwatch-batch-trigger-role" {
+  name = "cloudwatch-batch-trigger-role"
+  assume_role_policy = templatefile("${path.module}/iam_templates/trust_policy.json", {
+    trust_services = ["events"]
+  })
+}
+
+
+resource "aws_iam_role_policy_attachment" "cloudwatch-can-trigger-batch-jobs" {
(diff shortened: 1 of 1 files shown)


terraform/events.tf
```diff
@@ -0,0 +1,43 @@
...
+  rule      = aws_cloudwatch_event_rule.ingest-gisaid-rule.name
+  arn       = aws_batch_job_queue.aspen-batch-job-queue.arn
+  role_arn  = aws_iam_role.cloudwatch-batch-trigger-role.arn
+
+  batch_target {
+    job_name       = "ingest-gisaid"
+    job_definition = aws_batch_job_definition.aspen-batch-job-definition.name
+  }
+
+  input = <<-CONTAINER_OVERRIDES
```
phoenixAja, line 36: can you add a comment for what this is doing? I'm not familiar with `CONTAINER_OVERRIDES`

phoenixAja (approved):
exciting looks good!

terraform/events.tf
ttung: done.

PR #230, opened 2021-03-31
Convert accessions to use a workflow model

ttung:
### Description
This turned out to be a lot more complicated than I expected.  Most of the complexity lies in the conversion logic (which we are **not** doing) and the logic for sample view.  The latter is required and now adds another roundtrip to the server.  I suspect it should be possible to do what we want in one query, but I can't seem to massage sqlalchemy to do what I want.

Rest of this is pretty straightforward.

#### Issue
[ch123630](https://app.clubhouse.io/genepi/story/123630/move-accessions-to-a-workflow-model)

### Test plan
make -j style unit-tests

### DIFF
--- a/src/py/aspen/app/views/sample.py
+++ b/src/py/aspen/app/views/sample.py
@@ -1,19 +1,14 @@
-from typing import Any, Iterable, Mapping, MutableSequence, Set
+from typing import Any, Mapping, MutableMapping, MutableSequence, Sequence, Set
 
 from flask import jsonify, session
-from sqlalchemy import or_
-from sqlalchemy.orm import joinedload
+from sqlalchemy import and_, or_
+from sqlalchemy.orm import aliased, joinedload
 
 from aspen.app.app import application, requires_auth
 from aspen.app.views import api_utils
 from aspen.app.views.api_utils import get_usergroup_query
 from aspen.database.connection import session_scope
-from aspen.database.models import (
-    DataType,
-    PublicRepositoryType,
-    SequencingReadsCollection,
-    UploadedPathogenGenome,
-)
+from aspen.database.models import Accession, DataType, Entity, Workflow, WorkflowType
 from aspen.database.models.sample import Sample
 from aspen.database.models.usergroup import Group, User
 
@@ -29,15 +24,21 @@ def _format_created_date(sample: Sample) -> str:
         return "not yet uploaded"
 
 
-def _format_gisaid_accession(sample: Sample) -> str:
+def _format_gisaid_accession(
+    sample: Sample, entity_id_to_accession_map: Mapping[int, Accession]
+) -> str:
     if sample.uploaded_pathogen_genome is not None:
-        for accession in sample.uploaded_pathogen_genome.accessions:
-            if accession.repository_type == PublicRepositoryType.GISAID:
-                return accession.public_identifier
+        accession = entity_id_to_accession_map.get(
+            sample.uploaded_pathogen_genome.entity_id, None
+        )
+        if accession is not None:
+            return accession.public_identifier
     if sample.sequencing_reads_collection is not None:
-        for accession in sample.sequencing_reads_collection.accessions:
-            if accession.repository_type == PublicRepositoryType.GISAID:
-                return accession.public_identifier
+        accession = entity_id_to_accession_map.get(
+            sample.sequencing_reads_collection.entity_id, None
+        )
+        if accession is not None:
+            return accession.public_identifier
     return "NOT SUBMITTED"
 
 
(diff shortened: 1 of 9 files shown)


src/py/aspen/database/models/entity.py
```diff
@@ -123,3 +123,29 @@ def get_children(
                 results.append(tup)
 
         return results
+
+    def accessions(self) -> Sequence[Accession]:
+        from .accessions import Accession
```
phoenixAja, line 128: why add this import in the methods like this?

ttung: Accessions imports Entity, so if I import Accessions from Entity, it'll be a recursive import.

src/py/database_migrations/versions/20210330_120956_accessions_to_workflow_model.py
```diff
@@ -0,0 +1,46 @@
...
+def upgrade():
+    op.execute("DELETE FROM aspen.accessions")
+
+    op.create_table(
+        "accession_workflows",
+        sa.Column("workflow_id", sa.Integer(), nullable=False),
+        sa.ForeignKeyConstraint(
+            ["workflow_id"],
+            ["aspen.workflows.id"],
+            name=op.f("fk_accession_workflows_workflow_id_workflows"),
```
phoenixAja, line 27: so many workflow lol

src/py/aspen/app/views/sample.py
```diff
@@ -29,15 +24,21 @@ def _format_created_date(sample: Sample) -> str:
...
+            return accession.public_identifier
     if sample.sequencing_reads_collection is not None:
-        for accession in sample.sequencing_reads_collection.accessions:
-            if accession.repository_type == PublicRepositoryType.GISAID:
-                return accession.public_identifier
+        accession = entity_id_to_accession_map.get(
+            sample.sequencing_reads_collection.entity_id, None
+        )
+        if accession is not None:
+            return accession.public_identifier
```
phoenixAja, line 41: this code is really similar to the above block, could probably create another function that returns the accession public_identifier?

src/py/aspen/covidhub_import/implementation.py
```diff
@@ -301,19 +302,19 @@ def import_project(
             sample.uploaded_pathogen_genome.sequence = consensus_genome.fasta
 
             if czbid.genome_submission_info is not None:
+                repository_type: Optional[PublicRepositoryType] = None
+                public_identifier: str
```
phoenixAja: why not just define this type hint below on line 309 when you set the `public_identifier` variable?

src/py/aspen/database/models/accessions.py
```diff
@@ -24,23 +26,52 @@ class PublicRepositoryType(enum.Enum):
...
 
-    entity_id = Column(
-        Integer,
-        ForeignKey(Entity.id),
-        nullable=False,
-    )
-    entity = relationship(Entity, backref=backref("accessions", uselist=True))
+    entity_id = Column(Integer, ForeignKey(Entity.id), primary_key=True)
 
+    # statistics for the pathogen genome
```
phoenixAja: is this comment correct? `repository_type` doesn't seem like it would refer to statistics

src/py/aspen/covidhub_import/implementation.py
ttung: i can omit this line if you don't like it.  it's not really that important, as the line on 309 will infer type correctly.

src/py/aspen/database/models/accessions.py
ttung: nope, this is wrong.

src/py/aspen/app/views/sample.py
ttung: https://github.com/chanzuckerberg/aspen/pull/232

src/py/aspen/database/models/accessions.py
ttung: removed.

src/py/aspen/covidhub_import/implementation.py
ttung: Removed.

src/py/aspen/database/models/entity.py
phoenixAja: do you think there's a way to reorganize the files so we don't have to do this? I find these workarounds for circular dependencies a little confusing.. if not maybe we can add this to our agenda to resolve later on?

src/py/aspen/database/models/entity.py
```diff
@@ -123,3 +123,29 @@ def get_children(
...
+        """Adds an accession to this object."""
+        from .accessions import Accession
+
+        Accession.add_to_entity(
+            self,
+            repository_type,
+            public_identifier,
+            workflow_start_datetime,
+            workflow_end_datetime,
+        )
```
phoenixAja: i also find this a little confusing (passing self as the entity for add_to_entity), and the whole entity/ accession relationship, because i read this like add an accession for this entity and then add the accession to this entity. Maybe it's the end of the day and i'm tired though lol

src/py/aspen/database/models/entity.py
ttung: Not really. :(

src/py/aspen/database/models/entity.py
ttung: It's add an accession for this entity.  How about "create_for_entity"?  It's not exactly accurate either because it's not just creating it, it's attaching it to the entity.

phoenixAja: what about `attach_to_entity`? for some reason that seems more clear to me

phoenixAja (approved):

src/py/aspen/database/models/entity.py
ttung: Updated.

PR #233, opened 2021-04-01
Add support for custom cell renderers in Table

lvreynoso:
### Description

Allows custom cell renderers to be declared for a specific key on an item, which means we can easily write custom JSX for a specific column. This is to enable a modal to launch when users click on a phylo tree name.

#### Issue
[ch120777](https://app.clubhouse.io/genepi/story/120777)

### Test plan

From a UI perspective, everything should look the same and render properly. Load app, go to data tables


src/ts/src/common/components/library/data_table/index.tsx
```diff
@@ -27,29 +34,30 @@ const DataTable: FunctionComponent<Props> = ({ data = [], headers }: Props) => {
...
-            }
-            if (index === 0) {
-                icon = ICONS[item.type];
+    const sampleRow = (
+        item: Record<string | number, JSONPrimitive>
+    ): Array<JSX.Element> => {
+        return headers.map((header, index) => {
+            const value = item[header.key];
+            if (renderer !== undefined) {
+                console.log("used created renderer");
```
ttung: intentional?

src/ts/src/common/components/library/data_table/index.tsx
```diff
@@ -27,29 +34,30 @@ const DataTable: FunctionComponent<Props> = ({ data = [], headers }: Props) => {
...
-                displayData = UNDEFINED_TEXT;
-            }
-            if (index === 0) {
-                icon = ICONS[item.type];
+    const sampleRow = (
+        item: Record<string | number, JSONPrimitive>
+    ): Array<JSX.Element> => {
+        return headers.map((header, index) => {
+            const value = item[header.key];
+            if (renderer !== undefined) {
```
ttung: suggestion:

```
if (renderer === undefined) {
  renderer = defaultCellRenderer;
}
return (
    <Table.Cell key={`${item[indexingKey]}-${header.key}`}>
        {renderer({value, header, index})}
    </Table.Cell>
);

File: src/ts/src/common/components/library/data_subview/index.tsx [hunk 3]
```diff
@@ -7,8 +7,9 @@ import { DataTable } from "common/components";
 import style from "./index.module.scss";
 
 interface Props {
-    data?: BioinformaticsData[];
+    data?: Record<string | number, JSONPrimitive>[];
```
ttung: Why is this moving from BioinformaticsData to this new construct?

lvreynoso: To make it more generic, if we want to reuse it.

lvreynoso: `Record` is a built in Typescript type.

phoenixAja:
we should add @tihuan as a reviewer for this PR too

src/ts/src/common/types/table.d.ts
```diff
@@ -1,5 +1,18 @@
 interface Header {
     [index: string]: string;
     text: string;
-    key: keyof BioinformaticsType;
+    key: string | number;
```
phoenixAja: shouldn't this just always be a number?

lvreynoso: `key` here refers to the keys of a JS object, like on a sample or tree, which will almost always be strings.

src/ts/src/common/components/library/data_subview/index.tsx
tihuan: SGTM!

To keep it DRY, should we declare a `Data` type and use it throughout the file?

E.g.,

```ts
// <DataTable /> src/ts/src/common/components/library/data_table/index.tsx
export type Data = Record<string | number, JSONPrimitive>;

// <DataSubview /> src/ts/src/common/components/library/data_subview/index.tsx
import { DataTable, Data } from 'src/common/components/library/data_table'

interface SearchState {
    searching?: boolean;
    results?: Data[]; // <-- this!
}
...

```

src/ts/src/common/components/library/data_table/index.tsx
```diff
@@ -1,66 +1,73 @@
...
+    return <div className={style.cell}>{displayData}</div>;
+}
 
-  function tableRows(data: BioinformaticsData[]): Array<JSX.Element> {
-    return data.map((item) => (
-      <Table.Row key={`${item[indexingKey]}`}>{sampleRow(item)}</Table.Row>
+const DataTable: FunctionComponent<Props> = ({
+    data = [],
+    headers,
+    renderer,
```
tihuan: we can add default value to renderer here, so we don't need to reassign it later!

E.g.,

```tsx
const DataTable: FunctionComponent<Props> = ({
    data = [],
    headers,
    renderer = defaultCellRenderer, // <-- this!
}: Props) => {
```

src/ts/src/common/components/library/data_table/index.tsx
```diff
@@ -1,66 +1,73 @@
...
+            <div className={style.headerCell}>{column.text}</div>
+        </Table.HeaderCell>
     ));
-  }
+    const sampleRow = (
+        item: Record<string | number, JSONPrimitive>
+    ): Array<JSX.Element> => {
+        return headers.map((header, index) => {
+            const value = item[header.key];
+            if (renderer === undefined) {
```
tihuan: we can remove this after adding default value

tihuan (commented):
PR looks great! Just some super minor nits 🤩 👍 Thank you!!

src/ts/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "allowJs": true,
     "allowSyntheticDefaultImports": true,
-    "baseUrl": ".",
+    "baseUrl": "./src",
```
tihuan: Ah sorry I changed the `baseUrl` to `"."`, so it's easier to use autocomplete in important statement starting with `import from 'src/...'` and you get a dropdown of of options 😆

src/ts/src/common/types/table.d.ts
tihuan: Should we just restrict the type of `key` to be string?

The current use cases seem to be string anyway 😆 

```ts
export const SAMPLE_HEADERS: Header[] = [
  {
    key: "privateId",
    text: "Private ID",
  },
  {
    key: "publicId",
    text: "Public ID",
  },
  {
    key: "uploadDate",
    text: "Upload Date",
  },
  {
    key: "collectionDate",
    text: "Collection Date",
  },
  {
    key: "collectionLocation",
    text: "Collection Location",
  },
  {
    key: "gisaid",
    text: "GISAID",
  },
];

export const TREE_HEADERS: Header[] = [
  {
    key: "name",
    text: "Tree Name",
  },
  {
    key: "creationDate",
    text: "Creation Date",
  },
  {
    key: "pathogenGenomeCount",
    text: "Total Samples",
  },
  {
    key: "downloadLink",
    text: "",
  },
];

```

src/ts/tsconfig.json
lvreynoso: Ah, I use Sublime and my TS plugins are not that extensive; I've tried to organize code so we are mostly importing from the current directory or somewhere in `common` anyways.

tihuan: Got it! What if we add another shortcut to import directly from "common"? Would that help?

```ts
    ...
    "baseUrl": ".",
    ...
    "paths": {
      "src/*": [
        "./src/*"
      ],
      "common/*": [ // <-- this!
        "./src/common/*"
      ]
    },
```

tihuan (approved):
LGTM! I added `src/*` back for now, so the app compiles and we can get the PR in for demo. And we can find another time to figure out the important statement strategy that works for everyone 🙏 Thank you!

PR #236, opened 2021-04-04
Add accessors to pathogen genomes

ttung:
### Description
1. Add accessor to retrieve all the pathogen genomes retrieved from a HostFilteredSequencingReadsCollection.
2. Add accessor to retrieve all the pathogen genomes retrieved from a SequencingReadsCollection.

#### Issue

### Test plan
used in subsequent PR.

### DIFF
--- a/src/py/aspen/database/models/host_filtering.py
+++ b/src/py/aspen/database/models/host_filtering.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
+from typing import MutableSequence, Sequence, TYPE_CHECKING
 
 from sqlalchemy import Column, ForeignKey, Integer, String, UniqueConstraint
 
@@ -8,7 +8,10 @@ from aspen.database.models.entity import Entity, EntityType
 from aspen.database.models.workflow import Workflow, WorkflowType
 
 if TYPE_CHECKING:
-    from aspen.database.models.sequences import SequencingReadsCollection
+    from aspen.database.models.sequences import (
+        CalledPathogenGenome,
+        SequencingReadsCollection,
+    )
 
 
 class HostFilteredSequencingReadsCollection(Entity):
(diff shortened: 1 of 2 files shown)


src/py/aspen/database/models/host_filtering.py
```diff
@@ -23,14 +26,26 @@ class HostFilteredSequencingReadsCollection(Entity):
...
         from aspen.database.models.sequences import SequencingReadsCollection
 
         parents = self.get_parents(SequencingReadsCollection)
         assert len(parents) == 1
         return parents[0]
 
+    @property
+    def pathogen_genomes(self) -> Sequence[CalledPathogenGenome]:
+        # this import has to be here for circular dependencies reasons. :(
+        from aspen.database.models.sequences import CalledPathogenGenome
```
phoenixAja, line 41: i feel like we should reorganize this so that we don't have to do this (if that's possible)

src/py/aspen/database/models/host_filtering.py
```diff
@@ -23,14 +26,26 @@ class HostFilteredSequencingReadsCollection(Entity):
...
     @property
     def sequencing_read(self) -> SequencingReadsCollection:
-        """The raw gisaid dump this processed gisaid dump was generated from."""
+        """The sequencing read collection this host-filtered sequencing read collection
+        was generated from."""
         # this import has to be here for circular dependencies reasons. :(
         from aspen.database.models.sequences import SequencingReadsCollection
 
         parents = self.get_parents(SequencingReadsCollection)
         assert len(parents) == 1
```
phoenixAja, line 35: if this were to fail would that mean that it was due to a data integrity issue?

phoenixAja (approved):

src/py/aspen/database/models/host_filtering.py
ttung: It would imply that a host filtering workflow was created that had more than one `SequencingReadsCollection` input.

PR #237, opened 2021-04-04
script to manually create a phylo run workflow

ttung:
### Description
In the future, this will be created by a web endpoint, but to build out the rest of the backend, I need a script that does this for now.

Depends on #236

#### Issue

### Test plan
created a workflow, manual inspection seems to suggest that it was reasonable.

### DIFF
--- a/src/py/aspen/cli/db.py
+++ b/src/py/aspen/cli/db.py
@@ -1,19 +1,38 @@
+import datetime
 import json
 import os
 from pathlib import Path
-from typing import Type
+from typing import Iterable, MutableSequence, Optional, Sequence, Type
 
 import boto3
 import click
 from IPython.terminal.embed import InteractiveShellEmbed
+from sqlalchemy.orm import joinedload
+from sqlalchemy.sql import expression
 
 from aspen import covidhub_import
 from aspen.cli.toplevel import cli
 from aspen.config.config import Config, RemoteDatabaseConfig
 from aspen.config.development import DevelopmentConfig
 from aspen.config.local import LocalConfig
-from aspen.database.connection import enable_profiling, get_db_uri, init_db
+from aspen.database.connection import (
+    enable_profiling,
+    get_db_uri,
+    init_db,
+    session_scope,
+)
 from aspen.database.models import *  # noqa: F401, F403
+from aspen.database.models import (
+    AlignedGisaidDump,
+    Entity,
+    Group,
+    PathogenGenome,
+    PhyloRun,
+    Sample,
+    SequencingReadsCollection,
+    Workflow,
+    WorkflowStatusType,
+)
 from aspen.database.schema import create_tables_and_schema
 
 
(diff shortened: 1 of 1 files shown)


.hgtags
```diff
@@ -0,0 +1,6 @@
+c0cda8b0ffb4a4c71feabfdbf01c4aeb2af27157 t1
+0f598f1ff40c91da04278989f1e4c68d00bc87e0 t1
```
shanaxel42: rebase badness?

src/py/aspen/cli/db.py
```diff
@@ -108,3 +128,115 @@ def import_covidhub_project(
...
+    with session_scope(engine) as session:
+        group = session.query(Group).filter(Group.name == group_name).one()
+
+        all_samples: Iterable[Sample] = (
+            session.query(Sample)
+            .filter(
+                or_(
+                    Sample.public_identifier.in_(samples),
+                    and_(
+                        expression.true()
```
shanaxel42, line 247: What's getting checked here?

src/py/aspen/cli/db.py
```diff
@@ -108,3 +128,115 @@ def import_covidhub_project(
...
+            aligned_gisaid_dump = (
+                session.query(AlignedGisaidDump)
+                .filter(AlignedGisaidDump.entity_id == aligned_gisaid_dump_id)
+                .one()
+            )
+        else:
+            aligned_gisaid_dump = (
+                session.query(AlignedGisaidDump)
+                .join(AlignedGisaidDump.producing_workflow)
+                .order_by(Workflow.end_datetime.desc())
```
shanaxel42, line 296: oh chill, so it just get's the most recent one?

shanaxel42 (approved):
Exciting! Just a couple questions

src/py/aspen/cli/db.py
```diff
@@ -168,3 +188,115 @@ def import_covidhub_project(
...
+            start_datetime=datetime.datetime.now(),
+            workflow_status=WorkflowStatusType.STARTED,
+            software_versions={},
+            group=group,
+            # TODO: need build template / build args.
+        )
+        workflow.inputs = list(pathogen_genomes)
+        workflow.inputs.append(aligned_gisaid_dump)
+        workflow.template_file_path = (
+            "workflows/setup_nextstrain_run/builds_template.yaml"
```
jgadling: Where is this file?

src/py/aspen/cli/db.py
ttung: A sample is included if one of the following is true:

1. The sample is explicitly requested using `--sample`
2. `--all-group-samples` is specified AND the sample belongs to the group.

src/py/aspen/cli/db.py
ttung: yup.

src/py/aspen/cli/db.py
ttung: The workflow uses this file + the build args to generate the builds file.  It should be present in the target git refspec.  I have been iterating on the workflow in [this branch](https://github.com/chanzuckerberg/aspen/tree/ttung/setup-nextstrain-run).

src/py/aspen/cli/db.py
```diff
@@ -168,3 +187,150 @@ def import_covidhub_project(
...
+        containerOverrides={
+            "command": [
+                git_refspec,
+                "src/py/workflows/nextstrain_run/setup.sh",
+                str(workflow_id),
+            ],
+            "vcpus": 32,
+            "memory": 370000,
+        }
+    )
```
ttung: I added code to launch the batch job here.

src/py/aspen/cli/db.py
ttung: [here](https://github.com/chanzuckerberg/aspen/blob/ttung/setup-nextstrain-run/src/py/workflows/nextstrain_run/builds_template.yaml) in this case.

PR #235, opened 2021-04-02
Update phylo tree/run models

ttung:
### Description
1. Add a relationship to express which samples were actually incorporated into a tree.
2. Add data about how the build is made.

Depends on #230 

#### Issue
[ch129581](https://app.clubhouse.io/genepi/story/129581/schema-changes-to-record-which-samples-are-in-a-tree)
[ch130359](https://app.clubhouse.io/genepi/story/130359/schema-changes-to-record-the-build-configuration)

### Test plan
reran the import and pushed the docker image to ECR.

PR #238, opened 2021-04-04
Add region to sample table

ttung:
### Description
Added a migration that manually writes "North America" to all samples.  I'm not aware of any biohub samples that were from outside of North America, and even if there were, there are certainly no plans to import them into aspen.

Depends on #235 

### Test plan
ran the migration

### DIFF
--- a/src/py/aspen/database/models/sample.py
+++ b/src/py/aspen/database/models/sample.py
@@ -1,7 +1,9 @@
 from __future__ import annotations
 
+import enum
 from typing import Optional, TYPE_CHECKING, Union
 
+import enumtables
 from sqlalchemy import (
     Column,
     Date,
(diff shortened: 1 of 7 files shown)


src/py/aspen/database/models/sample.py
```diff
@@ -123,6 +123,11 @@ class Sample(idbase, DictMixin):
             }
         },
     )
+    region = Column(
+        String,
```
phoenixAja: would it make more sense to make this an enum field instead?

ttung: I looked at the region data for all the gisaid data and it appears very constrained to the continents.  I think it's pretty safe to make it an enum.

src/py/aspen/database/models/accessions.py
```diff
@@ -51,6 +51,7 @@ def add_to_entity(
         workflow_start_datetime: Optional[datetime.datetime] = None,
         workflow_end_datetime: Optional[datetime.datetime] = None,
     ):
+        """"""
```
phoenixAja: rm?

phoenixAja (approved):
looks good!!

PR #240, opened 2021-04-05
[easy] Remove schema.ddl

ttung:
### Description
It's out of date anyway.

### Test plan
YOLO


sidneymbell:
Can we keep (and update) this, actually? It's super helpful to me.

ttung:
> Can we keep (and update) this, actually? It's super helpful to me.

oh man, that will be quite an undertaking...

shanaxel42 (approved):
Love a good tony yolo pr

ttung:
> Can we keep (and update) this, actually? It's super helpful to me.

@sidneymbell  I created #270.  It's a compromise between dumping it and updating it (and continually updating it) by hand, which is a big commitment.  Let me know what you think.

sidneymbell:
That's a great path, thank you Tony!

PR #241, opened 2021-04-06
thuang-fe-config

tihuan:
### Description

1. Update FE's configuration to follow convention over customization 
2. Add more linters and plugins, such as `plugin:sonarjs/recommended`, `plugin:prettier/recommended`, `plugin:jest-playwright/recommended`, and `stylelint`, etc.
3. Add a few new script commands in `package.json` to help with linting, type checking, and linting staged commits
4. Use default `prettier` configs, which already has `arrowParens: "always"` and `trailingComma: "es5"` as defaults. The only change is changing `tabWidth: 4` back to `2`, since most FE projects use `2`
5. Update Github action to use the new lint command: `npm run lint-ci`, which runs `eslint`, `stylelint`, and type check concurrently
6. Update `src/ts/src/App.tsx` routes from `<Route path="/data" render={() => <Data />} />` to `<Route path="/data" component={Data} />`, so the route component gets React Router's props. This means we can access `location` via `props.location` now, instead of `window.location` which is more brittle and will break in Node.js based Server Side Rendering and unit testing in Jest 
7. Add `eslint` rule to sort keys alphabetically for consistency. If you use VSCode, [this extension](https://marketplace.visualstudio.com/items?itemName=zengxingxin.sort-js-object-keys) sorts keys for you automatically!
8. Update import statement to start with `src`, so we can take advantage of import path autocomplete in IDEs
9. Import statements will now be automatically sorted as well via [this plugin](https://github.com/simonhaenisch/prettier-plugin-organize-imports#readme)
10. The rest of the file changes are just `prettier` and `eslint` autofixes ✨ 

PTAL thanks all 😄 !

#### Issue
N/A

### Test plan

1. If you make any FE related file changes, at git commit time, it should now run `prettier`. Note that we only run `npm run lint-ci` on PRs instead of at commit time via lint-staged, since PR is gonna catch lint and type errors for us anyway, so we don't need to run the same thing locally. Only when GH Action fails do we need to run `npm run lint` to auto fix the lint errors where needed!


phoenixAja (approved):
cool this looks good to me, should we be adding any of these new checks to github actions?

tihuan:
Thanks for the quick review, @phoenixAja ! 

Yup, already did 🚀 ✨ 🤩 I replaced the existing `lint-ci` with the following one 👌 , which will be run on Github Action's FE lint:

`"lint-ci": "concurrently \"node_modules/.bin/eslint src/**/*.{ts,tsx}\" \"node_modules/.bin/stylelint '**/*.{js,ts,tsx,css,scss}'\"  \"npm run type-check\"",` 

Proof:
Screen Shot 2021-04-06 at 3 24 31 PM

phoenixAja:
awesome! looks good to me 👍

tihuan:
Perfect! Thanks again 😄 Merging!

PR #249, opened 2021-04-07
Make .env.ecr target fail if we fail to get an account id

jgadling:
### Description
The previous make target would sometimes write an invalid .env.ecr file without an account ID if the AWS CLI command failed. This change ensures that we fail instead of writing an invalid .env.ecr file.

#### Issue

### Test plan

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,12 @@ oauth/pkcs12/certificate.pfx:
 	sudo chown -R $$(id -u):$$(id -g) $(PWD)/oauth/pkcs12
 
 .env.ecr:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -42,7 +42,12 @@ oauth/pkcs12/certificate.pfx:
 	sudo chown -R $$(id -u):$$(id -g) $(PWD)/oauth/pkcs12
 
 .env.ecr:
-	echo DOCKER_REPO=$$(aws sts get-caller-identity --profile $(AWS_DEV_PROFILE) | jq -r .Account).dkr.ecr.us-west-2.amazonaws.com/ > .env.ecr;
+	export DOCKER_REPO=$$(aws sts get-caller-identity --profile $(AWS_DEV_PROFILE) | jq -r .Account); \
```
mbarrien: Probably should rename this variable, since it's just the AWS Account number now, not the full Docker repo

mbarrien (approved):

PR #256, opened 2021-04-10
Workflow for nextstrain run

ttung:
### Description
When invoked with phylo run as its parameter, `build_tree.sh` invokes `export.py`.

`export.py` does the following:
1. exports out the aspen sequences and metadata to run nextstrain.
1. dumps information about the linked aligned gisaid sequences, which is consumed by `build_tree.sh` to download the gisaid dataset.
1. generates the build.yaml file from the template and the template args referenced in the phylo run.

#### Duplicate sequences
There may be sequences duplicated in aspen and gisaid.  In that case, nextstrain resolves metadata differences in the order files are specified.  Since we specify the gisaid data first, aspen data overrides that IFF aspen provides that field.

Differences between Sequence data is normally treated by nextstrain as a failure.  As a result, we patch the script that merges sequences to ensure that we use the sequences in the order that we expect.

#### Issue
[ch129580](https://app.clubhouse.io/genepi/story/129580/post-nextstrain-workflow)
[ch129579](https://app.clubhouse.io/genepi/story/129579/record-which-samples-are-actually-in-a-tree)

### Test plan
`aspen-cli db --remote create-phylo-run --group-name 'Orange County Public Health Lab'  --all-group-sequences --builds-template-file src/py/workflows/nextstrain_run/builds_templates/group.yaml --builds-template-args '{"division": "California", "location": "Orange County"}' --git-refspec ttung/setup-nextstrain-run`
and watched the run.

Added a workflow to ensure that the patch file applies cleanly.


.github/workflows/patch-test.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Check out ncov repo.
+        run: |
+          mkdir ncov
+          cd ncov
+          git init
+          git fetch --depth 1 git://github.com/nextstrain/ncov.git
+          git checkout FETCH_HEAD
+      - name: patch
+        run: |
+          cp src/py/workflows/nextstrain_run/combine-and-dedup-fastas.py.patch ncov
```
phoenixAja, line 30: why did you name `combine-and-dedup-fastas.py.patch` with the `.patch` suffix?

src/py/workflows/nextstrain_run/combine-and-dedup-fastas.py.patch
```diff
@@ -0,0 +1,29 @@
...
++        for record in read_sequences(*reversed(args.input)):
+             counter += 1
+             if counter % 10000 == 0:
+                 print(f"Processed {counter} records")
+@@ -49,10 +51,8 @@ if __name__ == '__main__':
+ 
+     if len(duplicate_strains) > 0:
+         print(
+-            "ERROR: Detected the following duplicate input strains with different sequences:",
++            "WARNING: Detected the following duplicate input strains with different sequences:",
```
phoenixAja, line 23: I'm still a little confused by what "duplicate input strains with different sequences" means, can you give an example of when this might happen?

src/py/workflows/nextstrain_run/export.py
```diff
@@ -0,0 +1,243 @@
...
+    builds_file_fh: io.TextIOBase,
+):
+    interface: SqlAlchemyInterface = init_db(get_db_uri(RemoteDatabaseConfig()))
+
+    with session_scope(interface) as session:
+        phylo_run_inputs = with_polymorphic(
+            Entity,
+            [PathogenGenome, AlignedGisaidDump],
+            flat=True,
+        )
```
phoenixAja, line 83: can you add a comment for what this should be returning? Is it supposed to be returning the pipeline parameters passed into ncov?

src/py/workflows/nextstrain_run/save.py
```diff
@@ -0,0 +1,149 @@
...
+        ]
+
+        uploaded_pathogen_genomes: Set[UploadedPathogenGenome] = {
+            pathogen_genome
+            for pathogen_genome in pathogen_genomes
+            if isinstance(pathogen_genome, UploadedPathogenGenome)
+        }
+        sequencing_reads_collections: Set[SequencingReadsCollection] = {
+            pathogen_genome.get_parents(HostFilteredSequencingReadsCollection)[
+                0
```
phoenixAja, line 109: lol at this formatting

.github/workflows/patch-test.yml
phoenixAja: nevermind i get it

phoenixAja (approved):
cool this looks good, can you add a doc where you specify how to run this workflow? Like a docs/workflows.md?

src/py/workflows/nextstrain_run/combine-and-dedup-fastas.py.patch
ttung: If a sample is submitted to gisaid, and it exists in aspen, there could plausibly be a mismatch between the sequence in gisaid and the sequence in aspen.  The reason is that a subsequent consensus genome run could produce a different sequence.

src/py/workflows/nextstrain_run/export.py
ttung: Are you referencing the `with_polymorphic` call?

src/py/workflows/nextstrain_run/save.py
ttung: Yeah, I very much disagree with black on this.

src/py/workflows/nextstrain_run/export.py
ttung: I updated with an explanation of what the `with_polymorphic` call does.  Let me know if you meant something else.

ttung:
started docs in #269

src/py/workflows/nextstrain_run/export.py
phoenixAja: yeah i was confused about with_polymorphic, thanks for clarifying

src/py/workflows/nextstrain_run/combine-and-dedup-fastas.py.patch
phoenixAja: this makes sense now thanks

PR #257, opened 2021-04-10
Reduce resource requirements for tree building

ttung:
### Description
Seems we don't need quite so many resources for tree building.

### Test plan
<strike>Will test a build.</strike>
Ran a nextstrain build successfully.


phoenixAja (approved):

ttung:
I also ran a county + context build, so this seems to be okay.

PR #260, opened 2021-04-12
Import rdev frontend changes

lvreynoso:
### Description

Imports frontend changes made in the `rdev-auspice` branch.

* Fixes the module background opacity to use the proper type (decimal 0-1)
* simplifies the `createTableCellRenderer` utility function
* Adds the Nextstrain fetch link to the tree table modal
* Adds better type checking to custom table cell renderers

#### Issue
[ch120777](https://app.clubhouse.io/genepi/story/120777)

### Test plan

Navigate to Tree Table, test each of the above changes.

### DIFF
--- a/src/ts/src/data/cellRenderers.tsx
+++ b/src/ts/src/data/cellRenderers.tsx
@@ -17,15 +17,12 @@ const DEFAULT_RENDERER = (value: JSONPrimitive): JSX.Element => {
 };
 
 const SAMPLE_CUSTOM_RENDERERS: Record<string | number, CellRenderer> = {
-  privateId: (value: JSONPrimitive): JSX.Element => {
-    const stringValue = stringGuard(value);
-    return (
-      <div className={style.cell}>
-        {<SampleIcon className={style.icon} />}
-        {stringValue}
-      </div>
-    );
-  },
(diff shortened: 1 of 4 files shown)


src/ts/src/data/cellRenderers.tsx
```diff
@@ -14,10 +18,11 @@ const DEFAULT_RENDERER = (value: JSONPrimitive): JSX.Element => {
 
 const SAMPLE_CUSTOM_RENDERERS: Record<string | number, CellRenderer> = {
   privateId: (value: JSONPrimitive): JSX.Element => {
+    const stringValue = stringGuard(value);
```
tihuan: Did we encounter a bug related to value not casted as a string? Otherwise I'm thinking eventually the value will be a string when it's inserted in the DOM by React?

tihuan (approved):
LGTM!! Thanks for the cleanup, @lvreynoso 🚀 

Just one tiny question, otherwise LGTM!

phoenixAja (approved):
lgtm!

src/ts/src/data/cellRenderers.tsx
lvreynoso: You're right, I got a bit overzealous with adding the type guard

PR #261, opened 2021-04-12
Import rdev  Flask auspice view and explicit download response

lvreynoso:
### Description

Imports the changes in `rdev-auspice` pertaining to tree viewing on the backend.

* Adds the less-than-ideal api route for Nextstrain to fetch trees from the app
* Adds explicit content type and disposition to JSON files so they are downloaded and not opened in the browser

#### Issue
[ch119787](https://app.clubhouse.io/genepi/story/119787)

### Test plan

Requires remote dev testing, since this route is intended to be used by Nextstrain.

### DIFF
--- a/src/py/aspen/app/views/auspice.py
+++ b/src/py/aspen/app/views/auspice.py
@@ -1,23 +1,16 @@
 import logging
 import os
-from typing import Iterable
+from typing import Optional
 
 import boto3
 from flask import make_response
-from sqlalchemy.orm import joinedload
 
 from aspen.app.app import application
 from aspen.database.connection import session_scope
-from aspen.database.models import PhyloRun, PhyloTree, WorkflowStatusType
+from aspen.database.models import PhyloTree
+from aspen.error.recoverable import RecoverableError
 
 logger = logging.getLogger(__name__)
-
-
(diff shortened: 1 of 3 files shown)


src/py/aspen/app/views/auspice.py
```diff
@@ -0,0 +1,48 @@
...
+            .filter(PhyloRun.workflow_status == WorkflowStatusType.COMPLETED)
+        )
+
+        phylo_tree: PhyloTree
+        for phylo_run in phylo_runs:
+            for output in phylo_run.outputs:
+                if isinstance(output, PhyloTree) and output.entity_id == phylo_tree_id:
+                    phylo_tree = output
+                    found = True
+                    break
```
phoenixAja: same thing that tony mentioned before that this will only break out of the first loop

lvreynoso: Right yeah.

lvreynoso:
Revised and simplified the code, since this is...not an ideal endpoint anyway, and we have not come up with a more secure way to expose phylo tree jsons.

phoenixAja (approved):
lgtm!

PR #264, opened 2021-04-12
Strip protocol from process.env.API_URL for nextstrain tree link

lvreynoso:
### Description

Strips `https://` from the `nextstraing.org/fetch` link. Adds a `stripProtocol()` utility function.

#### Issue
[ch120777](https://app.clubhouse.io/genepi/story/12077)

### Test plan

Open tree modal; check the generated link doesn't have the protocol included after `nextstrain.org/fetch`.

### DIFF
--- a/src/ts/src/common/utils/urlUtils.tsx
+++ b/src/ts/src/common/utils/urlUtils.tsx
@@ -1,7 +1,7 @@
 export function stripProtocol(url: string | undefined): string | undefined {
-    if (url === undefined) {
-        return undefined;
-    }
-    const re = /^.*:\/\//i;
-    return url.replace(re, "")
+  if (url === undefined) {
+    return undefined;
+  }
+  const re = /^.*:\/\//i;
+  return url.replace(re, "");
 }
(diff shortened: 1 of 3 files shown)


jgadling (approved):

tihuan (approved):
Awesome! Thanks for the util function 🥳 🎉 LGTM!!

src/ts/src/common/utils/urlUtils.tsx
```diff
@@ -0,0 +1,7 @@
+export function stripProtocol(url: string | undefined): string | undefined {
+  if (url === undefined) {
+    return undefined;
+  }
+  const re = /^.*:\/\//i;
```
phoenixAja, line 5: do we need to strip out other protocols here besides just https?

phoenixAja (approved):

src/ts/src/common/utils/urlUtils.tsx
lvreynoso: This would also work on `http://`. It was pretty quick and easy to generalize

phoenixAja: i was just thinking that we could simplify this re to http?s or something like that to make it more obvious as to what this is stripping out.

PR #266, opened 2021-04-13
Targets for connecting to remote dev dbs.

jgadling:
### Description

This adds some make targets to allow connecting from OSX to the remote dev database, either via psql (installed via homebrew) or the ipython console in a backend container.

#### Issue

### Test plan

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,22 @@ help: ## display help for this makefile
 rm-pycache: ## remove all __pycache__ files (run if encountering issues with pycharm debugger (containers exiting prematurely))
 	find . -name '__pycache__' | xargs rm -rf
 
+### Connecting to remote dbs #########################################
+remotedb: # Get a psql console on a remote db (from OSX only!)
+	export ENV=$${ENV:=rdev}; \
+	export config=$$(aws --profile $(AWS_DEV_PROFILE) secretsmanager get-secret-value --secret-id $${ENV}/aspen-config | jq -r .SecretString ); \
+	export DB_URI=$$(jq -r '"postgresql://\(.DB.admin_username):\(.DB.admin_password)@127.0.0.1:5556/$(STACK)"' <<< $$config); \
+	ssh -f -o ExitOnForwardFailure=yes -L 5556:$$(jq -r .DB.address <<< $$config):5432 $$(jq -r .bastion_host <<< $$config) sleep 10; \
+	psql $${DB_URI}
+
+remoteconsole: .env.ecr # Get a python console on a remote db (from OSX only!)
+	export ENV=$${ENV:=rdev}; \
+	export config=$$(aws --profile $(AWS_DEV_PROFILE) secretsmanager get-secret-value --secret-id $${ENV}/aspen-config | jq -r .SecretString ); \
+	export OSX_IP=$$(ipconfig getifaddr en0 || ipconfig getifaddr en1); \
+	export DB_URI=$$(jq -r '"postgresql://\(.DB.admin_username):\(.DB.admin_password)@'$${OSX_IP}':5555/$(STACK)"' <<< $$config); \
+	ssh -f -o ExitOnForwardFailure=yes -L $${OSX_IP}:5555:$$(jq -r .DB.address <<< $$config):5432 $$(jq -r .bastion_host <<< $$config) sleep 10; \
+	docker-compose $(COMPOSE_OPTS) run -e DB_URI backend aspen-cli db --remote interact --connect
+
 ### DOCKER LOCAL DEV #########################################
 oauth/pkcs12/certificate.pfx:
 	# All calls to the openssl cli happen in the oidc-server-mock container.
--- a/Makefile
+++ b/Makefile
@@ -24,14 +24,14 @@ rm-pycache: ## remove all __pycache__ files (run if encountering issues with pyc
 	find . -name '__pycache__' | xargs rm -rf
 
 ### Connecting to remote dbs #########################################
-remotedb: # Get a psql console on a remote db (from OSX only!)
+remote-pgconsole: # Get a psql console on a remote db (from OSX only!)
 	export ENV=$${ENV:=rdev}; \
 	export config=$$(aws --profile $(AWS_DEV_PROFILE) secretsmanager get-secret-value --secret-id $${ENV}/aspen-config | jq -r .SecretString ); \
 	export DB_URI=$$(jq -r '"postgresql://\(.DB.admin_username):\(.DB.admin_password)@127.0.0.1:5556/$(STACK)"' <<< $$config); \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -23,6 +23,22 @@ help: ## display help for this makefile
 rm-pycache: ## remove all __pycache__ files (run if encountering issues with pycharm debugger (containers exiting prematurely))
 	find . -name '__pycache__' | xargs rm -rf
 
+### Connecting to remote dbs #########################################
+remotedb: # Get a psql console on a remote db (from OSX only!)
+	export ENV=$${ENV:=rdev}; \
+	export config=$$(aws --profile $(AWS_DEV_PROFILE) secretsmanager get-secret-value --secret-id $${ENV}/aspen-config | jq -r .SecretString ); \
```
jgadling, line 29: Read db connection info from secrets manager

Makefile
```diff
@@ -23,6 +23,22 @@ help: ## display help for this makefile
...
+	export DB_URI=$$(jq -r '"postgresql://\(.DB.admin_username):\(.DB.admin_password)@127.0.0.1:5556/$(STACK)"' <<< $$config); \
```
jgadling, line 30: Generate a DB URI

Makefile
```diff
@@ -23,6 +23,22 @@ help: ## display help for this makefile
...
+	ssh -f -o ExitOnForwardFailure=yes -L 5556:$$(jq -r .DB.address <<< $$config):5432 $$(jq -r .bastion_host <<< $$config) sleep 10; \
```
jgadling, line 31: Open an SSH tunnel via our bastion host to the aurora DB in the background. This tunnel will CLOSE in 10 seconds if nothing is using it. Bind the aurora db's postgres port (5432) to the localhost port 5556. We're using 5556 to avoid conflicting with any other local containers or postgres services.

Makefile
```diff
@@ -23,6 +23,22 @@ help: ## display help for this makefile
...
+	psql $${DB_URI}
+
+remoteconsole: .env.ecr # Get a python console on a remote db (from OSX only!)
+	export ENV=$${ENV:=rdev}; \
+	export config=$$(aws --profile $(AWS_DEV_PROFILE) secretsmanager get-secret-value --secret-id $${ENV}/aspen-config | jq -r .SecretString ); \
+	export OSX_IP=$$(ipconfig getifaddr en0 || ipconfig getifaddr en1); \
+	export DB_URI=$$(jq -r '"postgresql://\(.DB.admin_username):\(.DB.admin_password)@'$${OSX_IP}':5555/$(STACK)"' <<< $$config); \
+	ssh -f -o ExitOnForwardFailure=yes -L $${OSX_IP}:5555:$$(jq -r .DB.address <<< $$config):5432 $$(jq -r .bastion_host <<< $$config) sleep 10; \
```
jgadling, line 39: We're binding the remote postgres server's port 5432 to port 5555 on our primary OSX IP address so that Docker can route to it (it can't reach ports bound to 127.0.0.1 on OSX!)

Makefile
```diff
@@ -23,6 +23,22 @@ help: ## display help for this makefile
 rm-pycache: ## remove all __pycache__ files (run if encountering issues with pycharm debugger (containers exiting prematurely))
 	find . -name '__pycache__' | xargs rm -rf
 
+### Connecting to remote dbs #########################################
+remotedb: # Get a psql console on a remote db (from OSX only!)
```
phoenixAja: can we rename this to `create-remotedb-uri`?

src/py/aspen/cli/db.py
```diff
@@ -146,14 +146,19 @@ def import_data(s3_path, db_uri):
...
 @click.pass_context
-def interact(ctx, profile):
+def interact(ctx, profile, connect):
     # these are injected into the IPython scope, but they appear to be unused.
     engine = ctx.obj["ENGINE"]  # noqa: F841
 
     if profile:
         enable_profiling()
 
+    if connect:
```
jgadling, line 162: We're adding a flag to force the CLI to connect to the DB immediately so the SSH tunnel connection doesn't time out on us.

src/py/aspen/config/config.py
```diff
@@ -230,6 +230,9 @@ def SECRET_KEY(self) -> str:
 
     @property
     def DATABASE_URI(self) -> str:
+        # Allow db uri to be overridden by env var.
+        if os.getenv("DB_URI"):
+            return os.getenv("DB_URI")
```
jgadling: I'm a little on the fence about adding yet another way to configure DB URI's here, but open to ideas.

docs/REMOTE_DEV.md
```diff
@@ -31,6 +31,11 @@ The general remote dev workflow is:
 
 If you forget which stacks you've created, just run `./scripts/happy list` at any time to list the current remote dev stacks.
 
+### Connecting to remote dev databases
+NOTE - You'll need to [install and configure blessclient](https://wiki.czi.team/display/SI/Install+BlessClient) first!
+
+From the root of this repo, run `make remotedb STACK=STACK_NAME_HERE` (replace STACK_NAME_HERE with the name of your stack!) to open a psql console on the remote dev db, or `make remoteconsole STACK=STACK_NAME_HERE` to get a python console connected to the remote dev db.
```
phoenixAja: oh or actually could we rename to make `remote-pgconsole` and change `remoteconsole` -> `remote-dbconsole` to mirror those target names we have for local?

Makefile
jgadling: It doesn't just create a URI - it passes the URI to psql and actually connects to the db. Should we include psql in the target name?

phoenixAja: nevermind, see other comment.

phoenixAja (approved):
thanks for adding this!!

docs/REMOTE_DEV.md
jgadling: on it!

Makefile
phoenixAja: 10 seconds is pretty short, can we do a minute?

src/py/aspen/cli/db.py
phoenixAja: can you add a comment here saying what you just said in the comment?

src/py/aspen/config/config.py
phoenixAja: i think it's ok

phoenixAja: we could create like a factory method later on if we wanted to polish this, but i think it's fine to add that in later PRs

Makefile
jgadling: The next command in these make targets (docker-compose run or psql) opens a connection to the db within our 10 second window, which prevents the connection from closing!

PR #268, opened 2021-04-15
Thuang 132238 virtual scrolling

tihuan:
RDEV: https://thuang-virtual-scroll-frontend.dev.genepi.czi.technology/data/samples

### Description

1. Rewrite `DataTable` in order to allow virtual scrolling (turned out that `<table />` related elements don't work for virtual scrolling). Current implementation is hand rolled, but eventually we will use Material UI's table if needed
3. Add `react-window` for virtual scrolling
4. Some visual fixes were pulled in from "Visual Fixes" branch to make styling the table easier, such as:
    1. Updating all the theme related variables to match IDseq design system's values (space, typography, corner, etc.)
    2. Remove global style that makes every `div` a flex-box, since it makes styling very difficult and scatters style definitions across multiple places

#### Issue
https://app.clubhouse.io/genepi/story/132238/look-into-rendering-less-items-for-sample-table-view-pagination-or-another-faster-load-fix

### Test plan

demo

1. Log in
2. Load Samples table, you should now see the new virtual scrolling table
2. Switching between Samples and Trees should see no freezing anymore
3. Search/filter should not freeze anymore too
4. UPDATE: Please check the load state too!

demo

PTAL thank you 🎉 !


ttung:
silly question: why the move from `em` to `px`?  I thought `em` is the preferred way of doing things that aren't necessarily lined up down to the pixel.

tihuan:
Oh that's a really important question! 

I think relative units like `rem` and `em` in theory are better than absolute unit `px`. But with browser zoom these days, my impression is that many projects could get away with just using `px`, since elements scale up/down as expected when zooming. And it's much easier to understand and communicate between eng and design using `px`, and design system also "just works" across projects without worrying about the root font size, etc..

And when a project mixes absolute unit and relative units, due to using third party libraries or different contributor preferences, things begin to get unpredictable and hard to manage. So to keep things simple, I think it's easier to start a project with `px` until there's a hard requirement to go with relative unit. And by then, we'd still need to figure out whether to go with `rem` or `em`, or any other relative unit options 😵

src/ts/src/NavBar.module.scss
```diff
@@ -2,82 +2,99 @@
 @import "src/common/styles/spacing";
 @import "src/common/styles/typography";
 
+@mixin item-common {
+  height: 100%;
+  font-family: $font-family-primary;
+  color: $white;
```
ttung, line 8: why is white sometimes a variable and sometimes a constant?

src/ts/src/NavBar.tsx
```diff
@@ -1,6 +1,5 @@
 import cx from "classnames";
 import React, { FunctionComponent } from "react";
-import { ReactComponent as AspenLogo } from "src/common/styles/logos/AspenLogo.svg";
```
ttung, line 3: should we dump the svg file?

src/ts/src/common/components/library/data_table/index.module.scss
```diff
@@ -3,20 +3,38 @@
...
 @import "src/common/styles/corners";
 
+.container {
+  display: flex;
+  flex-direction: column;
+}
+
 .header {
+  display: flex;
+  // (thuang): Account for Chrome scrollbar width
```
ttung: is this a TODO?

ttung (commented):
Can you add to your test plan an empty view state?

src/ts/src/NavBar.tsx
tihuan: Oh yes thanks for catching this! Done ✅ !

src/ts/src/NavBar.module.scss
tihuan: I think only the logo one is constant `white`, the rest should be variable `$white`. And the reason is that the logo white shouldn't depend on the variable white, in case the design wants to change the white variable to a different shade of white, and we want the logo to stay the same!

src/ts/src/common/components/library/data_table/index.module.scss
tihuan: Oh! Sorry the wording is confusing lol `padding-right: $space-l;` is for offsetting the Chrome scrollbar width, so not a TODO. I just changed it to:

```ts
// (thuang): Offset for Chrome scrollbar width
padding-right: $space-l;
```

ttung: Typically when I see a login name in a comment, it's a TODO.  I'm curious what is the convention being used here.

src/ts/src/NavBar.module.scss
ttung: Should we possibly do `$aspen-background` for the `$white` variable or `$aspen-white`?  Not sure what the best practices are, but I kind of suspect that just `$white` will cause confusion in the long run.

ttung (approved):
generally looks fine to me, see comments pls.

src/ts/src/NavBar.module.scss
tihuan: Oh yes for best practice, IDseq's design system has defined color variables and other semantic variable names already. So all variables we use in our code base follow their design system 😄 !

https://www.figma.com/file/EaRifXLFs54XTjO1Mlszkg/Science-Design-System-Reference?node-id=0%3A3498

So `$white` in Aspen maps to the design system one, like so:

Screen Shot 2021-04-16 at 11 37 17 AM

src/ts/src/common/components/library/data_table/index.module.scss
tihuan: Ooh I see! I typically leave my login name in comments, so it's easier for other readers to know who left the comments and therefore easier to track down the person who has the most context about the comment. I don't think it's a hard requirement, but engineers in the past projects do the same thing!

tihuan: For todo, the format I use in the past is `TODO(login name): TICKET-NUMBER short description` 😄

tihuan:
Merging now! Thanks again for reviewing and all the great comments and questions, @ttung 🤩 🙏

PR #271, opened 2021-04-16
FE render link

tihuan:
LIVE: https://thuang-auspice-secure-2-frontend.dev.genepi.czi.technology/data/phylogenetic_trees

### Description

1. Update FE to fetch presigned URL from the private endpoint, and then encode the presigned URL + construct `nextstrain.org` URL
2. Update BE to return a JSON `{ "url": PRESIGNED_URL }`

#### Issue
N/A

### Test plan

1. Load https://thuang-auspice-secure-2-frontend.dev.genepi.czi.technology/data/phylogenetic_trees
2. Click on any tree and see the modal
3. Click on confirm and see the tree on Nextstrain!


src/py/aspen/app/views/auspice.py
```diff
@@ -37,13 +36,13 @@ def auspice_view(phylo_tree_id: int):
...
+        presigned_url = s3_client.generate_presigned_url(
             "get_object",
             Params={"Bucket": phylo_tree.s3_bucket, "Key": phylo_tree.s3_key},
-            ExpiresIn=3600
-            )
+            ExpiresIn=3600,
         )
 
-        return presigned_url
+        response = make_response(json.dumps({"url": presigned_url}))
```
tihuan, line 44: only real change is here, which returns JSON. Everything else is `prettier` formatting!

src/ts/src/common/api/index.tsx
```diff
@@ -2,6 +2,10 @@
 import axios from "axios";
 import { jsonToType } from "src/common/utils";
 
+export const DEFAULT_FETCH_OPTIONS: RequestInit = {
```
tihuan, line 5: shifting to use browser's native Fetch API instead of axios

src/ts/src/common/components/library/modal/index.tsx
```diff
@@ -35,8 +36,9 @@ const Modal: FunctionComponent<Props> = ({
     }
 
     const buttons = buttonInfo.map((button, index) => {
-      let link: string | undefined = button.link;
-      let onClick: (() => void) | undefined = undefined;
+      let link = button.link;
+      let onClick;
+      const Button = button.Button || DefaultButton;
```
tihuan, line 41: Modal component now accepts an optional `button.Button` component, so we can use that custom component to fetch presigned URL and render the link accordingly

src/ts/src/common/types/ui.ts
```diff
@@ -0,0 +1,16 @@
+import { FC } from "react";
```
tihuan, line 1: changed the file from `ui.d.ts` to `ui.ts`, so it's more explicit where the type is coming from

src/ts/src/common/utils/TreeModal/ConfirmButton/index.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { FC, useEffect, useState } from "react";
```
tihuan, line 1: This is the custom `ConfirmButton` that will fetch presigned URL!

src/ts/src/common/utils/tableUtils.tsx
```diff
@@ -40,5 +44,3 @@ function createTreeModalInfo(link: string): ModalInfo {
       "Please confirm you're ready to send your data to Auspice to see your tree.",
   };
 }
-
-export { createTableCellRenderer, createTreeModalInfo };
```
tihuan, line 44: Make export inline at definition

src/ts/src/common/utils/urlUtils.tsx
```diff
@@ -1,6 +1,6 @@
-export function stripProtocol(url: string | undefined): string | undefined {
-  if (url === undefined) {
-    return undefined;
+export function stripProtocol(url?: string): string {
+  if (!url) {
+    return "";
   }
   const re = /^.*:\/\//i;
```
ttung, line 5: btw, shouldn't this be `^[^:]*:\/\///i;`?

ttung (approved):
rest lgtm.

src/ts/src/common/utils/tableUtils.tsx
```diff
@@ -20,15 +24,15 @@ function createTableCellRenderer(
   };
 }
 
-function createTreeModalInfo(link: string): ModalInfo {
+export const createTreeModalInfo = memoize(createTreeModalInfo_);
```
tihuan, line 27: memoize the created object, so we don't unnecessarily re-render the component. The reason is because React by default does shallow diffing, so if we generate a new object every time, React will re-render!

E.g.,

`createTreeModalInfo_(1) === createTreeModalInfo_(1) // FALSE`
`createTreeModalInfo(1) === createTreeModalInfo(1) // TRUE thanks to memoization`

src/ts/src/data/cellRenderers.tsx
```diff
@@ -44,17 +43,11 @@ const TREE_CUSTOM_RENDERERS: Record<string | number, CellRenderer> = {
...
-    )}/api/auspice/view/${treeID}`;
+
     return (
-      <Modal
-        data={createTreeModalInfo(
-          `https://nextstrain.org/fetch/${treeLocation}`
-        )}
-        className={style.cell}
-      >
+      <Modal data={createTreeModalInfo(treeID)} className={style.cell}>
```
tihuan, line 50: We don't need to construct the URL here, now that it's the new `ConfirmButton`'s job!

tihuan:
@phoenixAja will help with the python lint 😆 

Please help merge the PR once it's resolved 🙏 

Thanks so much!!

phoenixAja:
@tihuan you can delete the file that's throwing the linting error (`src/py/aspen/test_infra/models/phylo_tree.py`), i added that one accidentally 😬

src/ts/src/common/api/index.tsx
phoenixAja: why are we shifting?

tihuan:
> @tihuan you can delete the file that's throwing the linting error (`src/py/aspen/test_infra/models/phylo_tree.py`), i added that one accidentally 😬

Ooh thank you!! That worked 😆

phoenixAja (approved):
thanks timmy!!

src/ts/src/common/utils/urlUtils.tsx
tihuan: Hmm I'm not sure familiar with REGEX 😆 What's the difference here?

It seems like `^.*` is capturing beginning of a string where 0 or more any char before `://`, and `^[^:]*` is capturing beginning of a string where 0 or more non `:` char before `://`?

So:

```ts
':://www.g.com'.replace(/^[^:]*:\/\//i, '') //  --> ":://www.g.com"
':://www.g.com'.replace(/^.*:\/\//i, '') //  --> "www.g.com"
```

src/ts/src/common/api/index.tsx
tihuan: Great question! Axios adds extra weight to the browser unnecessarily when the native Fetch API works just fine 👌

tihuan:
We should continue the REGEX discussion, but since it's technically out of scope of this PR, I'll merge this into @phoenixAja 's branch first, so @phoenixAja can open a PR!

Thanks both 😄 !

src/ts/src/common/utils/urlUtils.tsx
ttung: Technically, the : is not a valid character for any component of the URL, but that does presume that everyone is being a good actor and escaping things correctly.  My proposed regex strips everything up to the first `:`, rather than to the last `:`.

tihuan: ahh I see! What's your thought on stripping everything before `//`? Like: `/^.*\/\//i`

ttung: That strips everything before the last `//`.  I generally prefer less greedy regexes.

tihuan: Ahh got it! That sounds good 👍 Thanks @ttung ! [PR](https://github.com/chanzuckerberg/aspen/pull/279) out 😄 PTAL thank you!

PR #272, opened 2021-04-16
Phoenix/auspice secure

tihuan:
### Description

add back in s3 presigned url along with adding back in `requires_auth` to make the redirection to auspice more secure. 

#### Issue
https://app.clubhouse.io/genepi/story/132851/options-for-securing-auspice-linking

### Test plan

- [tested out new view in a remote dev env](https://plogan-frontend.dev.genepi.czi.technology/data/phylogenetic_trees)

- wrote a new unit test (currently a little hacky of a test, but i think it's better than nothing... input is definitely welcome!)

Note: all tsx code has been previously reviewed 🎉


tihuan:
oh I'll resolve the conflict!

tihuan:
Do we still need to review this PR lol? I guess the diff is the BE part! Thank you!

phoenixAja:
@tihuan yes i still need to add in the requires auth part for this view, then we can do a quick review and then merge it in 👍

ttung:
So we fixed how nextstrain's fetch endpoint interprets the URL?

src/ts/src/common/types/ui.ts
```diff
@@ -0,0 +1,16 @@
...
+  Button?: FC;
+}
+
+export interface ModalInfo {
+  header?: string | JSX.Element;
+  body?: string | JSX.Element;
+  buttons?: ButtonInfo[];
+  width?: number;
+  backgroundColor?: string;
+}
```
lvreynoso, line 16: This should be a `.d.ts` file, in line with the TypeScript convention used in the other files. If you want to limit the type to `React.FunctionComponent`, you can reference React types with the triple-slash import directive. https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html. Or, you can use the simpler `JSX.Element`, which encompasses a React FC.

src/ts/src/common/utils/TreeModal/ConfirmButton/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const result = await fetch(
+    `${process.env.API_URL}/api/auspice/view/${treeId}`,
+    DEFAULT_FETCH_OPTIONS
+  );
+
+  const json = await result.json();
+
+  const encodedJsonUrl = encodeURIComponent(stripProtocol(json.url));
+
+  return "https://nextstrain.org/fetch/" + encodedJsonUrl;
```
lvreynoso, line 49: Have you checked to make sure the presigned urls are compatible with nextstrain?

lvreynoso (commented):
A few frontend suggestions. I would like to know if this approach works before approving it.

src/ts/src/common/utils/TreeModal/ConfirmButton/index.tsx
lvreynoso: Just saw the remote dev link, so cool that it works!

phoenixAja:
> So we fixed how nextstrain's fetch endpoint interprets the URL?

@ttung nope we encoded the presigned url on the frontend with `encodeURIComponent` and that fixed the issue

phoenixAja:
> A few frontend suggestions. I would like to know if this approach works before approving it.

@lvreynoso it works! I included the remote dev link if you wanted to try it out

src/ts/src/common/types/ui.ts
phoenixAja: this was from a previous PR that @tihuan wrote, would it be ok if we patch it in a separate PR?

tihuan: Ah we actually don't need `d.ts`, since our code is written in Typescript instead of typing a non typescript library!

https://react-typescript-cheatsheet.netlify.app/docs/advanced/misc_concerns/#working-with-non-typescript-libraries-writing-your-own-indexdts

phoenixAja:
@lvreynoso and @ttung can we merge this? or was there anything else that was concerning?

src/py/aspen/app/views/tests/test_auspice_redirect.py
```diff
@@ -0,0 +1,58 @@
...
+
+    s3_client.put_object(
+        Bucket=phylo_tree.s3_bucket, Key=phylo_tree.s3_key, Body=test_json
+    )
+
+    with client.session_transaction() as sess:
+        sess["profile"] = {"name": user.name, "user_id": user.auth0_user_id}
+
+    res = client.get(f"/api/auspice/view/{phylo_tree.id}")
+    presigned_url_sections = json.loads(res.data)["url"].split("/")
```
ttung: You could just do `res.json["url"].split("/")`

src/py/aspen/test_infra/aws.py
```diff
@@ -0,0 +1,15 @@
+import boto3
+import pytest
+from moto import mock_s3
+
+
+@pytest.fixture(scope="session")
+def s3_resource():
```
ttung: could you rename this `mock_s3_resource`?

src/py/aspen/app/views/tests/test_auspice_redirect.py
```diff
@@ -0,0 +1,58 @@
...
+    # We need to create the bucket since this is all in Moto's 'virtual' AWS account
+    s3_resource.create_bucket(Bucket=phylo_tree.s3_bucket)
+
+    json_test_file = test_data_dir / "ncov_aspen.json"
+    with json_test_file.open() as fh:
+        test_json = json.dumps(json.load(fh))
+
+    s3_client.put_object(
+        Bucket=phylo_tree.s3_bucket, Key=phylo_tree.s3_key, Body=test_json
+    )
```
ttung: You could just do this using `s3_resource`

src/py/aspen/test_infra/aws.py
```diff
@@ -0,0 +1,15 @@
...
+    with mock_s3():
+        s3_resource = boto3.resource("s3")
+        yield s3_resource
+
+
+@pytest.fixture(scope="session")
+def s3_client(s3_resource):
+    yield s3_resource.meta.client
```
ttung: It doesn't seem this is required, as you can use the high-level s3_resource API to upload.

ttung (approved):
I'm really only qualified to comment on the python parts.

I'm approving this, but pls see comments first.

src/py/aspen/app/views/tests/test_auspice_redirect.py
phoenixAja: unfortunately you cannot do this (`res.json` is `None`)

src/py/aspen/test_infra/aws.py
phoenixAja: if i try to upload with resource directly i get `AttributeError: 's3.ServiceResource' object has no attribute 'put_object'`

src/py/aspen/app/views/tests/test_auspice_redirect.py
ttung: Wait that doesn't make sense.  If it's json it ought to be available as `.json`

src/py/aspen/test_infra/aws.py
phoenixAja: i can call this in the test directly though from the mock_s3_resource, i actually don't think we need to split this into the resource and client

src/py/aspen/app/views/tests/test_auspice_redirect.py
ttung: oh wait, it's [slightly different](https://flask.palletsprojects.com/en/1.1.x/testing/#testing-json-apis) than python requests.  Maybe it's `rv.get_json()`?

src/py/aspen/test_infra/aws.py
ttung: you can do:

`mock_s3_resource.Bucket(s3_bucket).Object(s3_key).put(Body=data)`

src/py/aspen/app/views/tests/test_auspice_redirect.py
phoenixAja: let me try that

phoenixAja: that also does not work :( it's wierd that it's so hard to get the json out of this response object..

ttung: So it seems to work if you tweak the result value in the handler.

src/py/aspen/test_infra/aws.py
phoenixAja: ohhh 🤦‍♀️ lol

phoenixAja: done!

src/py/aspen/app/views/auspice.py
```diff
@@ -27,16 +27,21 @@ def auspice_view(phylo_tree_id: int):
...
+            "get_object",
+            Params={"Bucket": phylo_tree.s3_bucket, "Key": phylo_tree.s3_key},
+            ExpiresIn=3600,
+        )
+
+        response = make_response(json.dumps({"url": presigned_url}))
         response.headers["Content-Type"] = "text/json"
-        response.headers["Content-Disposition"] = "attachment; filename=auspice.json"
+
         return response
```
ttung: do this instead:
```
return jsonify({"url": presigned_url})
```

phoenixAja: ok i added this and it works well (tested on remote dev stack too)

phoenixAja: once i verify that GHA passes i'll merge this in

ttung:
Thanks for humoring my requests. :)

src/ts/src/common/types/ui.ts
lvreynoso: Yeah that's the most common usage of `.d.ts` files, yeah. The reason I used them was because those files are "ambient" - the declarations there are available everywhere in the project without having to import them. If this works and it's the style you want to go with, that's fine. I would like to keep the way we declare global types consistent though, so if you want to change up things for this please change them for all the other type declarations.

tihuan: Ahh got it!

I try to steer clear from putting things in global namespace as much as possible to avoid global namespace collision and I find the import statements make things are more explicit too 😆 In theory we could also use Typescript `namespace` to manage collisions, but naming is hard and that's another layer of work we need to manage and could avoid for free by importing type definitions. For ambient or other commonly used entities, in the past projects we use `src/common/entities/*` to host them, so I'm thinking we could do the same thing here?

Keeping type definitions colocated with their code in the same file also has the added benefit of making it easier for engineers to update things and predictably find and use the definitions as well. So I find it easier to work with a codebase that way!

Love the idea of keeping type declaration consistent 👍 Will update other type definitions! Thank you 😄

PR #273, opened 2021-04-17
[easy] Add back missing comment for Sample.region field

ttung:
### Description
This was in there in a previous version of the PR, but dropped accidentally.  The migration code, however, creates it, and causes a mismatch in the schema depending on if it was created from scratch or through migrations.

### Test plan
subsequent autogenerated migrations are empty.


shanaxel42 (approved):
v chill PR

PR #274, opened 2021-04-17
Add a czb_failed_genome_recovery column

ttung:
### Description
Only set this to true if it's imported from CZB and genome recovery failed.

Depends on #273 

#### Issue
[ch127348](https://app.clubhouse.io/genepi/story/127348/port-over-failed-samples-from-covidtracker-just-metadata-not-fasta-q)

### Test plan
ran migration.


shanaxel42 (approved):
woooo!

PR #275, opened 2021-04-17
[easy] Move profiling to the db command

ttung:
### Description
This allows us to enable profiling for any db command.

### Test plan
tried it


shanaxel42 (approved):
helpful

PR #276, opened 2021-04-17
Refactor the creation of sample from the czbid/consensus genome

ttung:
### Description
Almost all this code is reusable for creating failed samples.

#### Issue
[ch127348](https://app.clubhouse.io/genepi/story/127348/port-over-failed-samples-from-covidtracker-just-metadata-not-fasta-q)

### Test plan
Imported successfully.

PR #278, opened 2021-04-17
Import failed samples

ttung:
### Description
This loads all the DphCZBIDs that are in a group but are not from the earlier query to find the consensus genomes.  By definition, these failed genome recovery.

Depends on #274, #276

#### Issue
[ch127348](https://app.clubhouse.io/genepi/story/127348/port-over-failed-samples-from-covidtracker-just-metadata-not-fasta-q)

### Test plan
imported, and visually inspected the results in the db.


shanaxel42 (approved):
Looks good!

PR #283, opened 2021-04-20
adding pangolin fields to pathogen_genome

phoenixAja:
### Description
add pangolin fields to the pathogen_genome table

#### Issue
[ch133637](https://app.clubhouse.io/genepi/story/133637)

### Test plan

`make utility-alembic-upgrade-head`
`make local-pgconsole`
checked `pathogen_genomes` table to verify that new columns were added


src/py/aspen/database/models/sequences.py
```diff
@@ -180,6 +180,11 @@ class PathogenGenome(Entity):
         ),
     )
 
+    pangolin_lineage = Column(String)
+    pangolin_probability = Column(Integer)
```
ttung: Should this be a float?

src/py/aspen/database/models/sequences.py
```diff
@@ -180,6 +180,11 @@ class PathogenGenome(Entity):
...
+    pangolin_version = Column(String)
+    pangolin_last_updated = Column(DateTime)
+
```
ttung: I've been marking `nullable=True`, even though it is the default, to make it super clear what the schema is.

ttung (changes requested):

src/py/aspen/database/models/sequences.py
phoenixAja: i was looking at the `support` column here where it looked like it was an int, but should we be converting that to a float when we upload the table? 

Screen Shot 2021-04-20 at 5 13 08 PM

ttung: Oh, I think integer is fine then.  I think in the mocks, it was a float.

ttung (approved):

PR #284, opened 2021-04-21
renaming to backend and frontend for src directories

phoenixAja:
### Description
renaming:
src/py -> src/backend
src/ts -> src/frontend

#### Issue
[ch133930](https://app.clubhouse.io/genepi/story/133930)

### Test plan
GHA
-test local dev commands
-test remote dev commands

**NOTE**: a test is currently failing due to the aspen batch dockerfile cloning from trunk (where src/py isn't currently renamed), i think the best course of action is to force merge this in as it will be resolved once trunk is updated


tihuan (approved):
LGTM! Thanks for doing this 🎉 !

phoenixAja:
thanks @tihuan !!

PR #285, opened 2021-04-21
Allow the staging stack to drop the contents of its database

jgadling:
We need to be able to drop the staging DB on demand so we can reimport it from a snapshot on S3 on demand.

### Description
We initially assumed the staging db would be long-lived, and only updated via alembic migrations. We wrote those assumptions into our happy env configuration and terraform code to make it very difficult to delete the staging db. 

In actuality, we need to be able to replace the data/schema in the staging DB on a regular basis. Since this is something we need to do in the remote-dev environment pretty frequently, there's already support for dropping a db and importing it from a snapshot on S3. This PR includes 2 key changes:

- Tell the Happy CLI that it's ok to run db drop operations in staging
- Create an ECS task definition in staging that's capable of dropping the staging db. This task definition will be invoked by the Happy CLI when we need to load a new snapshot.

#### Issue

### Test plan

### DIFF
--- a/.happy/config.json
+++ b/.happy/config.json
@@ -15,7 +15,7 @@
             "aws_profile": "genepi-dev",
             "secret_arn": "happy/env-staging-config",
             "terraform_directory": ".happy/terraform/envs/staging",
-            "delete_protected": true,
+            "delete_protected": false,
             "auto_run_migrations": true,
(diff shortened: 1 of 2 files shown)


phoenixAja (approved):
thanks for adding!

PR #287, opened 2021-04-22
[easy] Fix the backref relationship type hints

ttung:
### Description
The can_see* attributes should be on `Group` and not on `User`.

### Test plan
checked the [`relationship` declarations](https://github.com/chanzuckerberg/aspen/blob/ca6dc1e3a59a410a79c1aec39ea9ae8514f59266/src/py/aspen/database/models/cansee.py#L40-L48)

### DIFF
--- a/src/py/aspen/database/models/usergroup.py
+++ b/src/py/aspen/database/models/usergroup.py
@@ -22,6 +22,9 @@ class Group(idbase, DictMixin):  # type: ignore
     email = Column(String, unique=True, nullable=False)
     address = Column(String, nullable=True)
 
+    can_see: MutableSequence[CanSee]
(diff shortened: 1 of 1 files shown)


phoenixAja (commented):
are you missing a migration? Or how are these associations working?

ttung:
> are you missing a migration? Or how are these associations working?

These associations are created by the `backref` calls (see https://github.com/chanzuckerberg/aspen/blob/ca6dc1e3a59a410a79c1aec39ea9ae8514f59266/src/py/aspen/database/models/cansee.py#L40-L48)

ttung:
These are only type hints.

phoenixAja:
oh i see

phoenixAja (approved):

PR #288, opened 2021-04-22
Rename the tree data with private identifiers

ttung:
### Description
Nodes on trees should be rewritten with the private identifiers that the user is privy to.

#### Issue
[ch133611](https://app.clubhouse.io/genepi/story/133611/rewrite-tree-data-with-private-identifiers)

### Test plan
Added a test with samples from own-group, can-see (private identifiers), can-see (not private identifiers), no can-see, and not even in the system.

Note that I'm stealing the `ncov_aspen.json` file for the `test_auspice_redirect.py` test; fortunately, that test is agnostic of the content of the test.

### DIFF
--- a/src/py/aspen/app/views/tests/test_phylo_tree_rename.py
+++ b/src/py/aspen/app/views/tests/test_phylo_tree_rename.py
@@ -1,0 +1,94 @@
+import json
+
+from aspen.app.views.phylo_trees import _process_phylo_tree
+from aspen.database.models import CanSee, DataType
+from aspen.test_infra.models.phylo_tree import phylotree_factory
+from aspen.test_infra.models.sample import sample_factory
+from aspen.test_infra.models.usergroup import group_factory, user_factory
+
+
+def test_phylo_tree_rename(session, mock_s3_resource, test_data_dir):
+    """Create a set of samples belonging to different groups with different levels of
+    can-see relationships.  Rename the nodes according to the can-see rules, and verify
+    that the nodes are renamed correctly."""
+    viewer_group = group_factory()
+    can_see_group = group_factory("can_see", email="can_see@yahoo.com")
+    wrong_can_see_group = group_factory(
+        "wrong_can_see", email="wrong_can_see@hotmail.com"
+    )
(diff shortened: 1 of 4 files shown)


src/py/aspen/app/views/tests/data/ncov_aspen.json
```diff
@@ -1,5 +1,21 @@
-[
-  {
-    "test_tree": "test_tree"
```
phoenixAja: lol i was being lazy

src/py/aspen/app/views/tests/test_phylo_tree_rename.py
```diff
@@ -0,0 +1,94 @@
...
+def test_phylo_tree_rename(session, mock_s3_resource, test_data_dir):
+    """Create a set of samples belonging to different groups with different levels of
+    can-see relationships.  Rename the nodes according to the can-see rules, and verify
+    that the nodes are renamed correctly."""
+    viewer_group = group_factory()
+    can_see_group = group_factory("can_see", email="can_see@yahoo.com")
+    wrong_can_see_group = group_factory(
+        "wrong_can_see", email="wrong_can_see@hotmail.com"
+    )
+    no_can_see_group = group_factory("no_can_see", email="no_can_see@aol.com")
```
phoenixAja, line 19: these are some retro test email accounts lol

ttung: I was trying to think what else I could think of.  juno.com?  I don't know what else is super old.

phoenixAja (approved):

src/py/aspen/app/views/phylo_trees.py
```diff
@@ -84,33 +86,64 @@ def phylo_trees():
...
-            "Content-Disposition"
-        ] = f"attachment; filename={phylo_tree_id}.json"
+    response = make_response(phylo_tree_data)
+    response.headers["Content-Type"] = "text/json"
+    response.headers[
+        "Content-Disposition"
+    ] = f"attachment; filename={phylo_tree_id}.json"
 
-        return response
+    return response
```
phoenixAja: i vote that we rename `response` -> `treesponse` here (haha jk)

PR #289, opened 2021-04-22
[easy] fix mimetype for json

ttung:
### Description
The correct type is application/json, not text/json.

This is built on top of #288.

### Test plan
YOLO!
Also see https://www.ietf.org/rfc/rfc4627.txt


phoenixAja (approved):
interesting..

PR #298, opened 2021-04-27
Update covidhub import to handle internal CZBID

ttung:
### Description
This adds support for Internal CZBIDs.

The old workflow was:

1. look at all consensus genomes.
2. look at all CZBIDs that are not represented by the consensus genomes.
3. write out entries for the consensus genomes.
4. write out entries for the failed genomes.

This refactors the workflow to:

1. Pull all the CZBIDs associtaed with the project.
2. Load all the consensus genomes.
3. Loop through all the CZBIDs and:
   a. Find the consensus genome
   b. Sort out the metadata, depending on the CZBID type.

Depends on #297 

#### Issue
[ch133318](https://app.clubhouse.io/genepi/story/133318/port-over-internal-samples-from-covidhub-db-into-appropriate-projects-about-500)

### Test plan
import marin's internal czbids (RR089i)

### DIFF
--- a/src/backend/aspen/covidhub_import/implementation.py
+++ b/src/backend/aspen/covidhub_import/implementation.py
@@ -1,3 +1,4 @@
+import collections
 import datetime
 import json
 import logging
@@ -6,19 +7,22 @@ import warnings
 from typing import (
     Any,
     Collection,
-    Iterable,
+    DefaultDict,
     Iterator,
+    List,
     Mapping,
     MutableMapping,
+    MutableSequence,
     Optional,
     Sequence,
+    Tuple,
 )
 
 import boto3
 import pytz
 import tqdm
-from sqlalchemy import and_, or_
-from sqlalchemy.orm import configure_mappers, joinedload, Session
+from sqlalchemy import or_
+from sqlalchemy.orm import configure_mappers, joinedload, selectin_polymorphic, Session
 
 from aspen.aws.s3 import S3UrlParser
 from aspen.database.connection import session_scope, SqlAlchemyInterface
@@ -30,6 +34,7 @@ from aspen.database.models import (
     RegionType,
     Sample,
     UploadedPathogenGenome,
+    Workflow,
     WorkflowStatusType,
 )
 from aspen.phylo_tree.identifiers import get_names_from_tree
@@ -41,9 +46,21 @@ from covid_database.models.enums import ConsensusGenomeStatus
 from covid_database.models.ngs_sample_tracking import (
     ConsensusGenome,
     CZBID,
+    CZBIDRnaPlate,
     DphCZBID,
+    InternalCZBID,
     Project,
     SampleFastqs,
+)
+from covid_database.models.qpcr_processing import (
+    AccessionSample,
+    Aliquoting,
+    Extraction,
+    QPCRPlate,
+    QPCRResult,
+    QPCRRun,
+    RNAPlate,
+    SamplePlate,
 )
 
 logger = logging.getLogger(__name__)
@@ -72,6 +89,18 @@ def list_bucket(s3_resource, bucket: str, key_prefix: str) -> Iterator[str]:
         if not results["IsTruncated"]:
             return
         nexttoken = results["NextContinuationToken"]
+
+
+def _get_external_accession(external_accession: str, czbid: CZBID) -> str:
+    """Get the external accession.  If it is provided, use it as is.  Otherwise, create
+    a string out of the czbid's submission base."""
+    if external_accession is None:
+        external_accession = f"UNKNOWN_{czbid.submission_base}"
+        warnings.warn(
+            f"CZBID {czbid} does not have an external accession.  Assigning as"
+            f" {external_accession}"
+        )
+    return external_accession
 
 
 def import_project(
@@ -103,8 +132,25 @@ def import_project(
         group: Group = session.query(Group).filter(Group.id == aspen_group_id).one()
 
         logger.info("Retrieving from COVIDHUB DB...")
-        consensus_genomes: Collection[ConsensusGenome] = (
-            covidhub_session.query(ConsensusGenome)
+        group_czbids: Collection[CZBID] = (
+            covidhub_session.query(CZBID)
+            .join(Project)
+            .filter(Project.rr_project_id == rr_project_id)
+            .options(
+                selectin_polymorphic(CZBID, [DphCZBID, InternalCZBID]),
+                joinedload(CZBID.genome_submission_info),
+            )
+            .all()
+        )
+
+        # load a map of czbid name to czbid object.
+        czbid_lookup_by_str: Mapping[str, CZBID] = {
+            czbid.czb_id: czbid for czbid in group_czbids
+        }
+
+        czbid_to_consensus_genomes: Mapping[str, ConsensusGenome] = {
+            consensus_genome.sample_fastqs.czb_id.czb_id: consensus_genome
+            for consensus_genome in covidhub_session.query(ConsensusGenome)
             .join(SampleFastqs, CZBID, Project)
             .distinct(CZBID.id)
             .filter(Project.rr_project_id == rr_project_id)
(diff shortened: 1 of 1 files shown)


src/backend/aspen/covidhub_import/implementation.py
```diff
@@ -120,68 +152,196 @@ def import_project(
...
+                    external_accession = f"{czbid.external_accession} ({czbid.czb_id})"
+                else:
+                    external_accession = czbid.external_accession
+                collection_date = czbid.collection_date
+                date_received = czbid.date_received or czbid.collection_date
+            elif isinstance(czbid, InternalCZBID):
+                internal_metadata = internal_czb_ids_metadata[czbid.czb_id]
+                external_accession = internal_metadata[0]
+                collection_date = internal_metadata[1]
+                date_received = internal_metadata[1]
```
phoenixAja, line 273: should this be this? 
```suggestion
                date_received = internal_metadata[2]
```

src/backend/aspen/covidhub_import/implementation.py
```diff
@@ -103,8 +120,23 @@ def import_project(
         group: Group = session.query(Group).filter(Group.id == aspen_group_id).one()
 
         logger.info("Retrieving from COVIDHUB DB...")
-        consensus_genomes: Collection[ConsensusGenome] = (
-            covidhub_session.query(ConsensusGenome)
+        group_czbids: Collection[CZBID] = (
+            covidhub_session.query(CZBID)
+            .join(Project)
+            .filter(Project.rr_project_id == rr_project_id)
```
shanaxel42, line 138: So the rr_project id is different for internal/external samples though, ex. `RR0092i` vs `RR0092e` am I reading this wrong or does it assume they are the same?

src/backend/aspen/covidhub_import/implementation.py
```diff
@@ -103,8 +120,23 @@ def import_project(
...
+            .options(
+                selectin_polymorphic(CZBID, [DphCZBID, InternalCZBID]),
+                joinedload(CZBID.genome_submission_info),
+            )
+            .all()
+        )
+
+        czbid_lookup_by_str: Mapping[str, CZBID] = {
+            czbid.czb_id: czbid for czbid in group_czbids
```
shanaxel42, line 148: Why do we do this str comparison?

src/backend/aspen/covidhub_import/implementation.py
ttung: no, this is intentionally the same.  we don't really know the collection date, so we're setting it to `date_received`.

src/backend/aspen/covidhub_import/implementation.py
```diff
@@ -120,68 +152,196 @@ def import_project(
...
             .join(Project)
-            .filter(
-                DphCZBID.czb_id.notin_(
-                    {
-                        consensus_genome.sample_fastqs.czb_id.czb_id
-                        for consensus_genome in consensus_genomes
-                    }
-                )
+            .join(CZBIDRnaPlate)
+            .join(
```
shanaxel42, line 186: clleeeeeean

src/backend/aspen/covidhub_import/implementation.py
```diff
@@ -120,68 +152,196 @@ def import_project(
...
+        # verify that we have no duplicates for external accessions
+        external_accession_to_czbids: DefaultDict[
+            str, set[str]
+        ] = collections.defaultdict(set)
+        for czbid, (external_accession, _) in internal_czb_ids_metadata.items():
+            external_accession_to_czbids[external_accession].add(czbid)
+        czbids_to_process: MutableSequence[CZBID] = list()
+        for external_accession, czbids in external_accession_to_czbids.items():
+            if len(czbids) > 1:
+                # gather up the information for each czbid representing the external
```
shanaxel42, line 218: So I'm not sure we want this logic. If a county accidentally sent two samples with the same external_accession they may actually be two separate samples. Ideally there wouldn't be any of these, but are there?

src/backend/aspen/covidhub_import/implementation.py
```diff
@@ -120,68 +152,196 @@ def import_project(
...
-            if not isinstance(czbid, DphCZBID):
-                warnings.warn(
-                    "consensus genome has non-dphczbid"
-                    f"{consensus_genome.sample_fastqs.czb_id}"
-                )
+        for czbid in tqdm.tqdm(czbids_to_process):
+            if isinstance(czbid, DphCZBID):
+                if czbid.external_accession in (
+                    "OCPHL2207",
+                    "OCPHL2208",
```
shanaxel42: lolz

shanaxel42 (changes requested):

src/backend/aspen/covidhub_import/implementation.py
ttung: We'd run the import twice -- once for `RR0092e` and once for `RR0092i`

src/backend/aspen/covidhub_import/implementation.py
ttung: There is no string comparison...?
This is to look up a czbid based on the string.

src/backend/aspen/covidhub_import/implementation.py
ttung: this should just be one join.

src/backend/aspen/covidhub_import/implementation.py
ttung: Yes, there are.

These accessions are assigned when czb got the tubes (`AccessionSamples`), and not assigned by the counties afaik.

ttung: Example for Marin:

```
External accession H12313 represented by multiple czbids: {'RR089i_00186', 'RR089i_00181'}.  Choosing RR089i_00181 as the czbid forthis sample.
External accession H11851 represented by multiple czbids: {'RR089i_00182', 'RR089i_00187'}.  Choosing RR089i_00182 as the czbid forthis sample.
```

src/backend/aspen/covidhub_import/implementation.py
ttung: fixed!

src/backend/aspen/covidhub_import/implementation.py
shanaxel42: ahhhhh then all good!

src/backend/aspen/covidhub_import/implementation.py
shanaxel42: Oh interesting these are internal samples that were picked twice, ok yea then this makes sense

shanaxel42 (approved):
Looks good!

src/backend/aspen/covidhub_import/implementation.py
```diff
@@ -120,68 +152,196 @@ def import_project(
...
-                        for consensus_genome in consensus_genomes
-                    }
-                )
+        }
+
+        internal_czb_ids_metadata: Mapping[str, Tuple[str, datetime.datetime]] = {
+            czbid: (external_accession, collection_date)
+            for czbid, external_accession, collection_date in covidhub_session.query(
+                InternalCZBID.czb_id.label("czb_id"),
+                AccessionSample.accession.label("external_accession"),
```
shanaxel42: OK! so yes we want this to be

`AccessionSample.location_submitter_id.label("external_accession"),`

shanaxel42 (changes requested):

src/backend/aspen/covidhub_import/implementation.py
```diff
@@ -120,68 +166,195 @@ def import_project(
...
+        }
+
+        internal_czb_ids_metadata: Mapping[str, Tuple[str, datetime.datetime]] = {
+            czbid: (
+                _get_external_accession(external_accession, czbid_lookup_by_str[czbid]),
+                collection_date,
+            )
+            for czbid, external_accession, collection_date in covidhub_session.query(
+                InternalCZBID.czb_id.label("czb_id"),
+                AccessionSample.location_submitter_id.label("external_accession"),
```
shanaxel42, line 183: nice!

shanaxel42 (approved):
Looks good!

PR #299, opened 2021-04-27
Split tree import out to a separate step

ttung:
### Description
This is necessary because we need to import both the internal and the external samples *before* we import trees.  It is probably even a good idea to import _all_ samples before we import any trees.

Depends on #298 

#### Issue
[ch133318](https://app.clubhouse.io/genepi/story/133318/port-over-internal-samples-from-covidhub-db-into-appropriate-projects-about-500)

### Test plan
Ran through this sequence:
1. import all the marin users.
2. import all the marin samples (internal and external).
3. import all the marin trees.

### DIFF
--- a/src/backend/aspen/covidhub_import/import_trees.py
+++ b/src/backend/aspen/covidhub_import/import_trees.py
@@ -1,0 +1,136 @@
+import datetime
+import json
+import logging
+import re
+from typing import Any, Iterator, Mapping, MutableMapping, Sequence, Tuple
+
+import boto3
+import pytz
+from sqlalchemy.orm import configure_mappers
+
+from aspen.aws.s3 import S3UrlParser
+from aspen.database.connection import session_scope, SqlAlchemyInterface
+from aspen.database.models import (
+    Group,
+    PhyloRun,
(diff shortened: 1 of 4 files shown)


src/backend/aspen/covidhub_import/import_trees.py
```diff
@@ -0,0 +1,136 @@
...
+
+        pacific_time = pytz.timezone("US/Pacific")
+        s3_src = boto3.session.Session(profile_name=covidhub_aws_profile).resource("s3")
+        s3_dst = boto3.session.Session().resource("s3")
+
+        src_prefix_url = S3UrlParser(s3_src_prefix)
+        dst_prefix_url = S3UrlParser(s3_dst_prefix)
+        for key in list_bucket(s3_src, src_prefix_url.bucket, src_prefix_url.key):
+            key_mo = re.match(
+                r".*_(?P<year>\d{2})(?P<month>\d{2})(?P<day>\d{2})\.json",
```
shanaxel42, line 73: This is some serious regex

shanaxel42 (approved):

PR #300, opened 2021-04-27
new sequence test fixtures

phoenixAja:
### Description

Wrote some new test model factories for `CallConsensus` and `UploadedPathogenGenome`, <del>also moved `conftest` up a level so that fixtures could be used by tests in workflows</del>

#### Issue
no issue

### Test plan
used them in downstream tests in other branches


src/backend/aspen/test_infra/models/sequences.py
```diff
@@ -35,3 +36,31 @@ def sequencing_read_factory(
             public_identifier=public_identifier,
         )
     return sequencing_reads
+
+
+def uploaded_pathogen_genome_factory(
+    sample,
+    sequence=">test1 NTCGGCG",
```
ttung: I think the nucleotide data should be after a NL.

```suggestion
    sequence=">test1\nNTCGGCG",
```

src/backend/aspen/test_infra/models/workflow.py
```diff
@@ -0,0 +1,24 @@
+from datetime import datetime, timedelta
+
+from aspen.database.models.sequences import CallConsensus
+from aspen.database.models.workflow import WorkflowStatusType, WorkflowType
+
+
+def call_consensus_factory(
+    workflow_type=WorkflowType.CALL_CONSENSUS,
```
ttung: this should not be necessary.

src/backend/aspen/test_infra/models/workflow.py
```diff
@@ -0,0 +1,24 @@
...
+    start_datetime=datetime.now() - timedelta(days=1),
+    end_datetime=datetime.now(),
+    workflow_status=WorkflowStatusType.COMPLETED,
+    software_versions={"test": "v0.0.1"},
+    inputs=[],
+    outputs=[],
+):
+    return CallConsensus(
+        workflow_type=workflow_type,
```
ttung: ditto.

ttung (changes requested):
since you plan to move the workflows into `src/backend/aspen`, is the `conftest.py` change necessary?

phoenixAja:
> since you plan to move the workflows into `src/backend/aspen`, is the `conftest.py` change necessary?

yeah, this was before i talked with you about my plans, i'll move back

ttung (approved):

PR #302, opened 2021-04-27
Support the happy prod environment.

jgadling:
### Description

This adds terraform code that enables the happy CLI to deploy to the production environment.

There are a few other flags changes as well:
- `delete_protected` prevents the happy CLI from being able to delete stacks in staging and prod
- `wait_for_steady_state` means the happy CLI won't return a success or failure status until the frontend and backend services are 100% upgraded (ECS does a rolling deploy on update). We want this to be enabled for staging and prod since they're usually updated by a CI system that needs to run smoke tests on those environment after a deploy.

#### Issue

### Test plan


.happy/terraform/envs/prod/outputs.tf
```diff
@@ -0,0 +1,19 @@
...
+}
+
+output backend_url {
+  value       = module.stack.backend_url
+  description = "The URL endpoint for the backend website service"
+}
+
+output delete_db_task_definition_arn {
+  value       = module.stack.delete_db_task_definition_arn
+  description = "ARN of the ECS Task Definition that deletes dev databases"
```
phoenixAja: do we need this if the db is deletion protected? if so can we change it to the following:
```suggestion
  description = "ARN of the ECS Task Definition that deletes prod database"
```

phoenixAja (approved):
lgtm!

PR #308, opened 2021-04-28
usergroup PUT 

phoenixAja:
### Description

adding put request to usergroup api endpoint that allows someone to update User entry

#### Issue
[ch134917](https://app.clubhouse.io/genepi/story/134917)

### Test plan

wrote some tests

### DIFF
--- a/src/backend/aspen/app/views/tests/test_views.py
+++ b/src/backend/aspen/app/views/tests/test_views.py
@@ -3,7 +3,7 @@ import json
 from aspen.test_infra.models.usergroup import group_factory, user_factory
 
 
-def test_usergroup_view(session, app, client):
+def test_usergroup_view_get(session, app, client):
     group = group_factory()
     user = user_factory(group)
     session.add(group)
@@ -14,6 +14,20 @@ def test_usergroup_view(session, app, client):
     expected = {"user": user.to_dict(), "group": group.to_dict()}
     assert expected == json.loads(res.get_data(as_text=True))
 
+
+def test_usergroup_view_put(session, app, client):
+    group = group_factory()
+    user = user_factory(group)
+    session.add(group)
+    session.commit()
+    with client.session_transaction() as sess:
+        sess["profile"] = {"name": user.name, "user_id": user.auth0_user_id}
+    data = {"agreed_to_tos": True}
+
+    res = client.put("/api/usergroup", json=json.dumps(data))
+    updated_user = session.query(User).filter(User.auth0_user_id == user.auth0_user_id).one()
+    assert updated_user.agreed_to_tos == True
+    assert res.status
 
 def test_redirect(app, client):
     res = client.get("api/usergroup")
--- a/src/backend/aspen/app/views/tests/test_views.py
+++ b/src/backend/aspen/app/views/tests/test_views.py
@@ -1,5 +1,6 @@
 import json
 
+from aspen.database.models.usergroup import User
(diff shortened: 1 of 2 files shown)


phoenixAja:
@ttung can you help me? my test is failing, i think it has something to do with syncing the sessions? Or do you know what I'm doing wrong? I checked that the view PUT option worked manually with pdb and the tos field gets updated, just in the test the session seems like it's not getting updated?

src/backend/aspen/app/views/usergroup.py
```diff
@@ -1,17 +1,40 @@
...
+                "user": user.to_dict(),
+                "group": user.group.to_dict(),
+            }
+            return jsonify(user_groups)
+
+        if request.method == "PUT":
+            fields_to_update: Dict[str, Union[str, bool]] = json.loads(
+                request.get_json()
+            )
+            for key, value in fields_to_update.items():
```
tihuan, line 33: Should we restrict what fields a user is allowed to update using an allow list? Or are the fields here pretty harmless even if they change things unexpectedly? For example, if they change group id to something else, would they gain access to something they shouldn't? Thank you!

phoenixAja: oh that's a good idea!! thanks!

phoenixAja: @tihuan just added this, let me know if you think that solves it!

tihuan: Awesomeee! Thanks so much for the quick update, @phoenixAja 🎉 ! LGTM!!

tihuan (approved):
LGTM!! 🚀 🚀 🚀

src/backend/aspen/app/views/tests/test_views.py
```diff
@@ -15,6 +16,41 @@ def test_usergroup_view(session, app, client):
...
+    session.commit()
+    with client.session_transaction() as sess:
+        sess["profile"] = {"name": user.name, "user_id": user.auth0_user_id}
+
+    data = {"agreed_to_tos": True}
+    res = client.put("/api/usergroup", json=json.dumps(data))
+
+    # start a new transaction
+    session.close()
+    session.begin()
```
ttung, line 32: I suspect `session.commit()` is enough to force a new transaction to start.

src/backend/aspen/app/views/usergroup.py
```diff
@@ -1,17 +1,44 @@
...
+from flask import jsonify, request, Response, session
 
 from aspen.app.app import application, requires_auth
 from aspen.app.views.api_utils import get_usergroup_query
 from aspen.database.connection import session_scope
+from aspen.database.models.usergroup import User
 
 
-@application.route("/api/usergroup", methods=["GET"])
+@application.route("/api/usergroup", methods=["GET", "PUT"])
```
ttung, line 12: I think it probably makes sense to make PUT a separate route from GET.  They don't have enough in commonality to justify a single route.

phoenixAja: really you think? I thought it was cleaner to have (very basic) get, put, and post in the same view. I don't mind splitting it out though if you think that would be better.

src/backend/aspen/app/views/tests/test_views.py
phoenixAja: interesting, i guess it would

PR #309, opened 2021-04-28
remove linting for workflows

phoenixAja:
### Description

Now that workflows have been moved to the aspen directory we don't need to explicitly call it to type check

#### Issue
no issue

### Test plan
GHA


tihuan (approved):
Sgtm!!

PR #319, opened 2021-04-30
Factory for entities can now produce failed accessions

ttung:
### Description
`sequencing_read_factory` can now produce failed accessions.

### Test plan
pytest


phoenixAja (approved):
cool, do you need to update schema.ddl too or does this not really fit in there?

ttung:
No, this isn't part of the schema.

PR #321, opened 2021-05-03
changing pangolin probability to float

phoenixAja:
### Description

the pangolin documentation wasn't updated to reflect the latest output csv format, updated the DB for pangolin probability to be float instead of int to reflect. 

#### Issue
no issue

### Test plan

GHA

### DIFF
--- a/src/backend/database_migrations/versions/20210503_184213_pangolin_probability_to_float.py
+++ b/src/backend/database_migrations/versions/20210503_184213_pangolin_probability_to_float.py
@@ -25,10 +25,4 @@ def upgrade():
 
 
 def downgrade():
-    op.alter_column(
-        "pathogen_genomes",
-        "pangolin_probability",
-        existing_type=sa.Float(),
-        type_=sa.Integer(),
-        schema="aspen",
-    )
+    raise NotImplementedError("Downgrading the database is not allowed")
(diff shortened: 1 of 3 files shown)


phoenixAja:
oh no sqlalchemy-utils import is failing.. @jgadling any suggestions on how to fix?

jgadling:
This is some sort of pip installation error and it's not obvious exactly what the problem is (package failed to download? version conflict? syntax error?) from the output. A bunch of the pip installation output looks REALLY strange though. Is there any chance we're building our package list or creating our requirements.txt files incorrectly?

I don't think we ever want to write to a file called "-e ."
```
Writing supplied requirement line to temporary file: '-e .'
```

phoenixAja:
oh i wonder if this has to do with the latest version of pip being released. I don't think it's the requirements file as the build passed the last time requirements was updated... Maybe i'll try downgrading pip to see if that solves it

src/backend/database_migrations/versions/20210503_184213_pangolin_probability_to_float.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def downgrade():
+    op.alter_column(
+        "pathogen_genomes",
+        "pangolin_probability",
+        existing_type=sa.Float(),
+        type_=sa.Integer(),
+        schema="aspen",
+    )
```
ttung: we should just prohibit a downgrade, as casting a float from 0-1 to an int is pretty meaningless.

ttung (approved):
rest lgtm.

src/backend/database_migrations/versions/20210503_184213_pangolin_probability_to_float.py
phoenixAja: ok i added a `NotImplementedError` for when downgrading the DB

PR #323, opened 2021-05-03
thuang-tos-privacy-pages-134345

tihuan:
### Description

Privacy and ToS page design matching IDseq as much as possible!

The final copies will be ready this Friday, and I'll update them next Monday 😄 Thank you!

ToS:

Screen Shot 2021-05-03 at 3 46 17 PM
Screen Shot 2021-05-03 at 3 46 29 PM

Privacy:
Screen Shot 2021-05-03 at 3 46 40 PM
Screen Shot 2021-05-03 at 3 46 52 PM

#### Issue
https://app.clubhouse.io/genepi/story/134345/create-tos-and-privacy-pages

### Test plan

Visit ToS page: https://thuang-tos-privacy-frontend.dev.genepi.czi.technology/terms
Visit Privacy page: https://thuang-tos-privacy-frontend.dev.genepi.czi.technology/privacy

The pages should be very similar to IDseq's pages:

ToS: https://idseq.net/terms
Privacy: https://idseq.net/privacy

FAQ page will be implemented in a separate ticket!

Thank you 😄 👍


src/frontend/src/App.module.scss
```diff
@@ -4,4 +4,8 @@
   height: 100vh;
   flex-flow: column wrap;
   justify-content: center;
+  * {
+    font-family: Open Sans, -apple-system, BlinkMacSystemFont, Segoe UI, Roboto,
+      Helvetica Neue, Helvetica, Arial, sans-serif !important;
```
phoenixAja, line 9: what is `!important` used for?

src/frontend/src/common/styles/support/components/List/index.tsx
```diff
@@ -0,0 +1,23 @@
+import { List as RawList, ListItem } from "czifui";
+import React from "react";
+
+interface Props {
+  items: React.ReactNode[];
+  ordered?: boolean;
+}
+
+const List = ({ items, ordered }: Props): JSX.Element => {
```
phoenixAja, line 9: would be nice if this were named something a little more descriptive?

phoenixAja (approved):
looks good!!

src/frontend/src/App.module.scss
tihuan: Great question! So `!important` is to ensure that the rule will take precedence over other stylesheets that are setting the `font-family` rule. This is needed because we still have Semantic UI css stylesheet, which sets `font-family` to a different font.

And we still need Semantic UI, because we have components using Semantic UI. On top of that, when we get desperate we might need to port IDseq's components directly (which are using Semantic UI) instead of rewriting them in Material UI. 

However, all new styling uses Material UI by default now, so we will rely on Semantic UI as minimally as possible!

src/frontend/src/common/styles/support/components/List/index.tsx
tihuan: Thanks for the feedback! Do you mean `List` should be something more descriptive? If so it's actually a standard component name in [Material UI](https://material-ui.com/components/lists/) and IDseq's [Design System](https://chanzuckerberg.invisionapp.com/console/Design-System---Current-ckkj0euz36kkd01w6dimd8saa/ckkj0gn5p5fz401t3q88uldh5/play) 😄 !

tihuan:
Thanks so much for the quick and thorough review, @phoenixAja ! PTAL at my response and I'll only merge after you think my response makes sense 🙏 Thanks again!

src/frontend/src/common/styles/support/components/List/index.tsx
phoenixAja: oh whoops i thought this was a variable that you set! nevermind!

phoenixAja:
> Thanks so much for the quick and thorough review, @phoenixAja ! PTAL at my response and I'll only merge after you think my response makes sense 🙏 Thanks again!

looks good to me! Feel free to merge whenever!

tihuan:
Awesome! Thanks so much, @phoenixAja 🎉 !! Will resolve merge conflict then merge!

PR #325, opened 2021-05-04
Adding gisaid status to sample api return value

shanaxel42:
### Description

Adds logic to the view/samples api to return the on of three GISIAD status:

- "Not Eligible" if sample failed genome recovery at czb
- "Rejected" if sample has been rejected by GISAID (if we don't have an ID and it didn't fail genome recovery)
- "Accepted" if sample has been accepted by GISAID (if we have an ID)
- "Submitted" if a sample has been submitted to GISAID (has a consuming workflow) but no accession and it's been less then 4 days. 
- "no_info" if a sample has not been submitted but is eligible WE SHOULD SHOW Eligible, not Submitted

Returns the status as part of the gisaid field. The new format of the return value is 

```
"gisaid": {"status": "STATUS", "gisaid_id": ID_OR_NONE}
```

Relavent for @tihuan ! 

#### Issue
[ch135473](https://app.clubhouse.io/genepi/story/135473)

### Test plan

Added a few tests

### DIFF
--- a/src/backend/aspen/app/views/sample.py
+++ b/src/backend/aspen/app/views/sample.py
@@ -1,4 +1,13 @@
-from typing import Any, Mapping, MutableMapping, MutableSequence, Sequence, Set
+import datetime
+from typing import (
+    Any,
+    Mapping,
+    MutableMapping,
+    MutableSequence,
+    Optional,
+    Sequence,
+    Set,
+)
 
 from flask import jsonify, session
 from sqlalchemy import and_, or_
@@ -8,11 +17,19 @@ from aspen.app.app import application, requires_auth
 from aspen.app.views import api_utils
 from aspen.app.views.api_utils import get_usergroup_query
 from aspen.database.connection import session_scope
-from aspen.database.models import Accession, DataType, Entity, Workflow, WorkflowType
+from aspen.database.models import (
+    Accession,
+    DataType,
+    Entity,
+    Workflow,
+    WorkflowStatusType,
+    WorkflowType,
+)
 from aspen.database.models.sample import Sample
 from aspen.database.models.usergroup import Group, User
 
 SAMPLE_KEY = "samples"
+GISIAD_REJECTION_TIME = datetime.timedelta(days=4)
 
 
 def _format_created_date(sample: Sample) -> str:
@@ -26,12 +43,28 @@ def _format_created_date(sample: Sample) -> str:
 
 def _format_gisaid_accession(
     sample: Sample, entity_id_to_accession_map: Mapping[int, Accession]
-) -> str:
+) -> Mapping[str, Optional[str]]:
     uploaded_entity = sample.get_uploaded_entity()
+    consuming_workflows = uploaded_entity.consuming_workflows
     accession = entity_id_to_accession_map.get(uploaded_entity.entity_id, None)
+    if sample.czb_failed_genome_recovery:
+        # todo need to add the private option here for v3 a user uploads and flags a private sample
+        return {"status": "not_eligible", "gisaid_id": None}
     if accession is not None:
-        return accession.public_identifier
-    return "NOT SUBMITTED"
+        return {"status": "accepted", "gisaid_id": accession.public_identifier}
+    for workflow in consuming_workflows:
+        if (
+            workflow.workflow_type == WorkflowType.PUBLIC_REPOSITORY_SUBMISSION
+            and workflow.workflow_status == WorkflowStatusType.STARTED
+        ):
+            date_since_submitted = (
+                datetime.date.today() - workflow.start_datetime.date()
+            )
+            if date_since_submitted < GISIAD_REJECTION_TIME:
+                return {"status": "submitted", "gisaid_id": None}
+            else:
+                return {"status": "rejected", "gisaid_id": None}
+    return {"status": "no_info", "gisaid_id": None}
 
 
 @application.route("/api/samples", methods=["GET"])
--- a/src/backend/aspen/app/views/tests/test_samples_view.py
+++ b/src/backend/aspen/app/views/tests/test_samples_view.py
@@ -10,6 +10,9 @@ from aspen.database.models import (
     PublicRepositoryType,
     Sample,
     SequencingReadsCollection,
+    Workflow,
+    WorkflowStatusType,
+    WorkflowType,
 )
 from aspen.test_infra.models.accession_workflow import AccessionWorkflowDirective
 from aspen.test_infra.models.sample import sample_factory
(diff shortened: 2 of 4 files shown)


src/backend/aspen/app/views/sample.py
```diff
@@ -26,12 +26,15 @@ def _format_created_date(sample: Sample) -> str:
 
 def _format_gisaid_accession(
     sample: Sample, entity_id_to_accession_map: Mapping[int, Accession]
-) -> str:
+) -> Mapping[str, Union[str, None]]:
```
ttung:
```suggestion
) -> Mapping[str, Optional[str]]:
```

src/backend/aspen/app/views/sample.py
```diff
@@ -26,12 +26,15 @@ def _format_created_date(sample: Sample) -> str:
...
     uploaded_entity = sample.get_uploaded_entity()
     accession = entity_id_to_accession_map.get(uploaded_entity.entity_id, None)
+    if sample.czb_failed_genome_recovery:
+        # todo need to add the private option here for v3 a user uploads and flags a private sample
+        return {"status": "not_eligible", "gisaid_id": None}
     if accession is not None:
-        return accession.public_identifier
-    return "NOT SUBMITTED"
+        return {"status": "accepted", "gisaid_id": accession.public_identifier}
+    return {"status": "rejected", "gisaid_id": None}
```
shanaxel42: @ttung should a sample with an uploaded_entity (for GISAID) but no accession be considered status PENDING?

shanaxel42: Or should we change this status to be "NO_INFO"

ttung: As per our discussions:

1. A GISAID submission workflow that started less than X time ago but hasn't shown up in the GISAID dump is considered pending.
2. A GISAID submission workflow that started more than X time ago but hasn't shown up in the GISAID dump is considered failled.

shanaxel42: Ah ok so would the logic be:
- query all workflows for the given `uploaded_entity`
- check if any are PUBLIC_REPOSITORY_SUBMISSION and have the WorkflowStatusType STARTED
- do a date check....what should the date check be? A week?

ttung: I think the long-term logic is:

1. Retrieve all the workflows coming off of this entity that are of type `AccessionWorkflow`, as well as the entities that are the outputs from those workflows.
2. If there are no workflows, then it was never submitted.
3. If there are workflows that are still RUNNING, then it should be considered "IN SUBMISSION".
4. If there are failed workflows, then the submission has failed.  There should be a separate mechanism to mark workflows failed if they are too old.

ttung: Shorter term, we may need to think things through...

src/backend/aspen/app/views/sample.py
```diff
@@ -26,12 +43,26 @@ def _format_created_date(sample: Sample) -> str:
...
+    consuming_workflows = uploaded_entity.consuming_workflows
     accession = entity_id_to_accession_map.get(uploaded_entity.entity_id, None)
+    if sample.czb_failed_genome_recovery:
+        # todo need to add the private option here for v3 a user uploads and flags a private sample
+        return {"status": "not_eligible", "gisaid_id": None}
     if accession is not None:
-        return accession.public_identifier
-    return "NOT SUBMITTED"
+        return {"status": "accepted", "gisaid_id": accession.public_identifier}
+    for workflow in consuming_workflows:
```
ttung, line 55: can you just inline `consuming_workflows`?

src/backend/aspen/app/views/tests/test_samples_view.py
```diff
@@ -39,7 +42,107 @@ def test_samples_view(
...
+def test_samples_view_gisaid_rejected(
+    session,
+    app,
+    client,
+):
+    group = group_factory()
+    user = user_factory(group)
+    sample = sample_factory(group)
+    # Test no GISAID accession logic
+    sequencing_read = sequencing_read_factory(sample, accessions={})
```
ttung: doesn't this imply "not_submitted" and not "rejected"?

src/backend/aspen/app/views/sample.py
```diff
@@ -26,12 +43,26 @@ def _format_created_date(sample: Sample) -> str:
...
 def _format_gisaid_accession(
     sample: Sample, entity_id_to_accession_map: Mapping[int, Accession]
-) -> str:
+) -> Mapping[str, Optional[str]]:
     uploaded_entity = sample.get_uploaded_entity()
+    consuming_workflows = uploaded_entity.consuming_workflows
     accession = entity_id_to_accession_map.get(uploaded_entity.entity_id, None)
+    if sample.czb_failed_genome_recovery:
+        # todo need to add the private option here for v3 a user uploads and flags a private sample
+        return {"status": "not_eligible", "gisaid_id": None}
```
ttung, line 52: I wonder if "not eligible" is the best terminology for this.

ttung (changes requested):

src/backend/aspen/app/views/sample.py
shanaxel42: That was the product choice

src/backend/aspen/app/views/sample.py
shanaxel42: what do you mean?

src/backend/aspen/app/views/tests/test_samples_view.py
shanaxel42: Well this hits the final case in the _format_gisaid_accession, where there was no accession, or consuming workflow within a 4 day period. But this comment might imply theres another case of "eligible but not yet submitted"?

src/backend/aspen/app/views/sample.py
```diff
@@ -26,12 +43,28 @@ def _format_created_date(sample: Sample) -> str:
...
+            workflow.workflow_type == WorkflowType.PUBLIC_REPOSITORY_SUBMISSION
+            and workflow.workflow_status == WorkflowStatusType.STARTED
+        ):
+            date_since_submitted = (
+                datetime.date.today() - workflow.start_datetime.date()
+            )
+            if date_since_submitted < GISIAD_REJECTION_TIME:
+                return {"status": "submitted", "gisaid_id": None}
+            else:
+                return {"status": "rejected", "gisaid_id": None}
```
ttung, line 66: I think we need to create some tickets around doing this out of band, and having this key off of `WorkflowStatusType.FAILED` (in the future).

src/backend/aspen/app/views/sample.py
ttung: That variable is used once and nowhere else.

ttung (approved):
see comments.

PR #327, opened 2021-05-04
Fix sample count for phylo trees

ttung:
### Description
Sample counts should reflect what's in the phylo tree.

The previous query was not working because sqlalchemy was creating an implicit join to the PhyloRun table in the inner query.  What we want is a correlated query, where the PhyloRun is coming from the outer query.  Use an alias to explicitly name which PhyloRun we want.

Depends on #326 

#### Issue
[ch135382](https://app.clubhouse.io/genepi/story/135382/number-of-samples-in-tree-table-view-seems-off)

### Test plan
Added a test to test the phylo_tree endpoint, which verifies the sample counts.

### DIFF
--- a/src/backend/aspen/app/views/tests/test_phylo_tree_view.py
+++ b/src/backend/aspen/app/views/tests/test_phylo_tree_view.py
@@ -1,0 +1,53 @@
+import json
+import random
+
+from aspen.app.views.phylo_trees import PHYLO_TREE_KEY
+from aspen.test_infra.models.phylo_tree import phylorun_factory, phylotree_factory
+from aspen.test_infra.models.sample import sample_factory
+from aspen.test_infra.models.sequences import uploaded_pathogen_genome_factory
+from aspen.test_infra.models.usergroup import group_factory, user_factory
+
+
+def test_phylo_tree_view(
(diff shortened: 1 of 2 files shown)


src/backend/aspen/app/views/tests/test_phylo_tree_view.py
```diff
@@ -0,0 +1,53 @@
...
+    n_samples=5,
+    n_trees=3,
+):
+    group = group_factory()
+    user = user_factory(group)
+    samples = [
+        sample_factory(
+            group,
+            public_identifier=f"public_identifier_%{ix}",
+            private_identifier=f"private_identifier_%{ix}",
```
phoenixAja: are you mixing up two different types of string formatting here?

shanaxel42 (approved):

src/backend/aspen/app/views/tests/test_phylo_tree_view.py
ttung: Yes, there should just be no `%`

ttung: good catch!

PR #329, opened 2021-05-04
pango save seqs

phoenixAja:
### Description

once pangolin has been run, take output csv and save results back to the DB

#### Issue
[ch135584](https://app.clubhouse.io/genepi/story/135584)

### Test plan
wrote a test

### DIFF
--- a/src/backend/aspen/workflows/pangolin/save.py
+++ b/src/backend/aspen/workflows/pangolin/save.py
@@ -31,7 +31,7 @@ def cli(pangolin_fh: io.TextIOBase, pangolin_last_updated: datetime):
                 .one()
             )
             pathogen_genome.pangolin_lineage = row["lineage"]
-            pathogen_genome.pangolin_probability = float(row["probability"])
+            pathogen_genome.pangolin_probability = float(row["probability"])  # type: ignore
             # TODO: change pangolin_version to date, looks like that's how pangolearn version is tracked
             pathogen_genome.pangolin_version = row["pangoLEARN_version"]
             pathogen_genome.pangolin_last_updated = pangolin_last_updated
--- a/src/backend/aspen/workflows/pangolin/save.py
+++ b/src/backend/aspen/workflows/pangolin/save.py
@@ -1,7 +1,6 @@
 import csv
 import io
 from datetime import datetime
-from decimal import Decimal
 from typing import Mapping, Union
 
 import click
@@ -16,7 +15,7 @@ from aspen.database.connection import (
 from aspen.database.models import PathogenGenome, UploadedPathogenGenome
 
 
-def get_probability(conflict: float) -> Union[Decimal, None]:
+def get_probability(conflict: float) -> float:
     assert conflict <= 1
     return 1.0 - conflict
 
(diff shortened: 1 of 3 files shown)


src/backend/aspen/workflows/pangolin/save.py
```diff
@@ -0,0 +1,42 @@
...
+    interface: SqlAlchemyInterface = init_db(get_db_uri(RemoteDatabaseConfig()))
+
+    with session_scope(interface) as session:
+        pango_csv: csv.DictReader = csv.DictReader(pangolin_fh)
+        for row in pango_csv:
+            pathogen_genome: UploadedPathogenGenome = (
+                session.query(PathogenGenome)
+                .filter(PathogenGenome.entity_id == int(row["taxon"]))
+                .one()
+            )
```
ttung: this will be a pretty slow query.  might be worth gathering up all the taxon ids, and running a single query.

src/backend/aspen/workflows/pangolin/save.py
```diff
@@ -0,0 +1,42 @@
...
+            pathogen_genome.pangolin_lineage = row["lineage"]
+            pathogen_genome.pangolin_probability = 1.0 - float(row["conflict"])  # type: ignore
```
ttung: worth asserting that `row["conflict"] <= 1.0`?

src/backend/aspen/workflows/pangolin/tests/test_pangolin_workflow.py
```diff
@@ -42,3 +59,35 @@ def test_pangolin_export(mocker, session, postgres_database):
...
+    # start new transaction
+    session.close()
+    session.begin()
+
+    for i in range(1, 3):
+        pathogen_genome: UploadedPathogenGenome = (
+            session.query(UploadedPathogenGenome)
+            .filter(UploadedPathogenGenome.entity_id == i)
+            .one()
+        )
```
ttung: i would not rely on the entity_id being a certain number.  you do know what the public_identifier will be, so you can query on that.

ttung (changes requested):

src/backend/aspen/workflows/pangolin/tests/test_pangolin_workflow.py
phoenixAja: i actually don't know what the public_identifier will be because the taxon is set to be entity_id (from when we renamed the sequences in export.py)

phoenixAja: should i change export script to output public_identifier instead of entity_id?

ttung: You don't need to, but there is no guarantee that entity_id will be 1-3.  Maybe you should just pull all the entries from the DB, and then compare it against what you expect to find.

src/backend/aspen/workflows/pangolin/save.py
phoenixAja: i think i fixed this

src/backend/aspen/workflows/pangolin/save.py
phoenixAja: added

src/backend/aspen/workflows/pangolin/tests/test_pangolin_workflow.py
phoenixAja: oh i thought you were talking about the overall logic and not the test, ok will fix

phoenixAja: ok i changed it

PR #331, opened 2021-05-04
Make sure frontend container gets rebuilt when package*.json changes.

jgadling:
### Description

This refactors our frontend Dockerfile so that it reliably rebuilds the frontend image whenever src/frontend/package*.json files change. This means we shouldn't need to explicitly run npm install inside the frontend container anymore.

#### Issue

### Test plan


phoenixAja (approved):
this works! thanks jess!

PR #334, opened 2021-05-04
Import non-dphczbid gisaid submissions

ttung:
### Description
We should import gisaid submissions for all czbids, not just dphczbids.

### Test plan
visual inspection.


shanaxel42 (approved):
nice!

PR #335, opened 2021-05-04
Update db connection Make targets

jgadling:
### Description

This updates the remote-db connection targets to support dev/stage/prod. 

Docs for how to use these are here: https://wiki.czi.team/display/SCI/Aspen+-+Engineering+How+To%27s#AspenEngineeringHowTo's-Connectingtoremote-dev,stagingandprodDB's:

#### Issue

### Test plan

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ export DOCKER_BUILDKIT:=1
 export COMPOSE_DOCKER_CLI_BUILD:=1
 export COMPOSE_OPTS:=--env .env.ecr
 export AWS_DEV_PROFILE=genepi-dev
+export AWS_PROD_PROFILE=genepi-prod
 export BACKEND_APP_ROOT=/usr/src/app
 
 ### DATABASE VARIABLES #################################################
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -30,18 +31,22 @@ rm-pycache: ## remove all __pycache__ files (run if encountering issues with pyc
...
-	export config=$$(aws --profile $(AWS_DEV_PROFILE) secretsmanager get-secret-value --secret-id $${ENV}/aspen-config | jq -r .SecretString ); \
+	export AWS_PROFILE=$(shell [ $(ENV) = prod ] && echo $(AWS_PROD_PROFILE) || echo $(AWS_DEV_PROFILE)); \
+	export config=$$(aws secretsmanager get-secret-value --secret-id $${ENV}/aspen-config | jq -r .SecretString ); \
 	export OSX_IP=$$(ipconfig getifaddr en0 || ipconfig getifaddr en1); \
-	export DB_URI=$$(jq -r '"postgresql://\(.DB.admin_username):\(.DB.admin_password)@'$${OSX_IP}':5555/$(STACK)"' <<< $$config); \
-	ssh -f -o ExitOnForwardFailure=yes -L $${OSX_IP}:5555:$$(jq -r .DB.address <<< $$config):5432 $$(jq -r .bastion_host <<< $$config) sleep 10; \
-	docker-compose $(COMPOSE_OPTS) run -e DB_URI backend aspen-cli db --remote interact --connect
+	export DB_URI=$$(jq -r '"postgresql://\(.DB.admin_username):\(.DB.admin_password)@'$${OSX_IP}':5555/$(DB)"' <<< $$config); \
+	echo Connecting to $$(jq -r .DB.address <<< $$config)/$(DB) via $$(jq -r .bastion_host <<< $$config); \
+	ssh -f -o ExitOnForwardFailure=yes -L $${OSX_IP}:5555:$$(jq -r .DB.address <<< $$config):5432 $$(jq -r .bastion_host <<< $$config) sleep 20; \
```
ttung, line 48: why is this binding on the public address and not just the loopback interface?  does docker need it on the public interface?

also, i noticed that the sleep time is different for the different commands. curious why.

ttung (approved):

Makefile
jgadling: Yes, docker (running in a VM) needs it on the public interface since it doesn't have access to OSX's loopback interface.
The sleep time is the amount of time we have to establish a connection with the db before the tunnel closes. psql creates a connection ~instantly, but the REPL connection might need to do a pip install (I'd like to fix this when we get a chance) before we make a db connection, so I've increased its timeout accordingly.

PR #336, opened 2021-05-05
GISAID WDL workflow

kislyuk:


ttung:
Is this still draft?

kislyuk:
Still being integration tested but comments welcome

workflows/pangolin/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+    gcc \
+    && locale-gen en_US.UTF-8
+
+RUN git clone --depth 1 git://github.com/chanzuckerberg/aspen
+RUN python3 -m venv /aspen/.venv
+RUN bash -c "source /aspen/.venv/bin/activate; pip3 install /aspen/src/backend /aspen/src/backend/third-party/sqlalchemy-enum-tables"
+
+RUN git clone --depth 1 git://github.com/nextstrain/ncov-ingest
+RUN python3 -m venv /ncov-ingest/.venv
+RUN bash -c "source /ncov-ingest/.venv/bin/activate; pip install -U pip pipenv; pipenv install"
```
phoenixAja, line 41: we don't need ncov ingest for pangolin, so i think this can be removed

phoenixAja: also need pangolin to be installed, [see this file for how i installed it in the run pangolin shell script](https://github.com/chanzuckerberg/aspen/blob/phoenix/pagolin-run/src/backend/aspen/workflows/pangolin/run_pangolin.sh)

should probably add installing pangolin to an entrypoint, as we want to make sure we have the most recent version when running

workflows/Dockerfile
```diff
@@ -0,0 +1,47 @@
+FROM ubuntu:20.04
```
ttung, line 1: so this model implies that the docker image needs to be rebuilt periodically?

ttung: And why not build off the nextstrain image?  That ensures that the libraries required for nextstrain are always installed correctly.

workflows/gisaid.wdl
```diff
@@ -0,0 +1,275 @@
+version 1.1
+
+workflow LoadGISAID {
+    input {
+        String docker_image_id = "aspen-gisaid"
+        String aws_region = "us-west-2"
+        String db_data_bucket = "aspen-db-data-dev"
```
ttung, line 7: how does this work if the bucket changes depending on prod/staging?

workflows/pangolin/Dockerfile
jgadling: Yeah, let's work together on getting a functional pangolin dockerfile. I think this is just a starting point.

workflows/gisaid.wdl
```diff
@@ -0,0 +1,275 @@
...
+        String gisaid_ndjson_export_url = "https://www.epicov.org/epi3/3p/exp3/export/export.json.bz2"
+        String gisaid_ndjson_staging_bucket = "aspen-data-dev"
+        String gisaid_ndjson_staging_key = "raw_gisaid_dump/test.zst"
```
ttung, line 10: why is this test.xst?  also, this implies that every run will write to the same place.

workflows/Dockerfile
kislyuk: I usually prefer to build from scratch and use plain ubuntu and versioned PyPI dependencies where possible, but I'm not opposed to using an upstream Docker image instead as long as it's built reproducibly. Where can I get the upstream nextstrain dockerfile? I can give it a try to see if it works as expected.

workflows/gisaid.wdl
kislyuk: This is a test placeholder; it is expected to be overridden at runtime.

workflows/gisaid.wdl
kislyuk: Also a test placeholder, expected to be overridden at runtime.

kislyuk: In general I agree these placeholder defaults should be removed, and test values should be set in the test suite instead.

However in the particular case of gisaid dump caching (see the RefreshGISAID task - the idea is not to download and decompress the large upstream file every time if it has not changed), if this input changes, that task will need a separate pointer to the most recently downloaded copy of the file as well, in order for the caching logic to work.

PR #338, opened 2021-05-05
Clean up terraform

ttung:
### Description
These are no longer necessary.

### Test plan
`make deploy-tf`


phoenixAja (approved):
thanks for cleaning

PR #340, opened 2021-05-05
thuang-hsts-csp-135163

tihuan:
### Description

1. Add `build` folder in `src/frontend/.stylelintignore`
2. Replace the usage of npm package - `serve` as our frontend server with Express, so we can add the security headers
3. Add HSTS header with 2 years TTL
4. Add CSP that allow all the resources we currently use to load properly. However, CSP is a living policy, which means whenever we add new resources (Google Analytics, etc.) we'll need to update our allowlist accordingly. Likely [Meta's CSP setting](https://github.com/chanzuckerberg/khepri/blob/develop/src/server/common/cspSettings.ts) will have examples we need down the road

BONUS:
1. Update tree modal outlink to Nextstrain.org as opening in a new tab instead of the current tab
2. Update agree ToS "Decline" button to redirect instead of making an API call to logout, which didn't work 

#### Issue
https://app.clubhouse.io/genepi/story/135163/frontend-security-add-hsts-and-csp

### Test plan

1. Visit https://thuang-hsts-1-frontend.dev.genepi.czi.technology/data/samples
2. Check the html request's response header and you should see content security policy and strict transport security headers in place!
3. 
Screen Shot 2021-05-05 at 11 59 58 AM

PTAL thank you!


src/frontend/Dockerfile
```diff
@@ -7,7 +7,6 @@ ENV NODE_ENV=development
 ENV BUILD_PATH=build
 RUN npm i --verbose
 RUN apt-get update && apt-get install make
-RUN npm install -g serve
```
phoenixAja, line 10: was this causing errors?

tihuan: Ah no, we had to replace the usage of `serve` with `Express`, so we can add custom security headers 😄 !

src/frontend/config/webpack.config.server.js
```diff
@@ -0,0 +1,36 @@
+const paths = require("./paths");
```
tihuan, line 1: Frontend server webpack config file. This is needed to transpile server side Typescript code to JavaScript

src/frontend/package.json
```diff
@@ -4,9 +4,12 @@
   "private": true,
   "proxy": "http://localhost:5000",
   "scripts": {
-    "serve": "npm run build && serve -s build -l tcp://0.0.0.0:3000",
+    "serve": "npm run build && npm run start-server",
```
tihuan, line 7: update commands here to build both client JS and server JS

src/frontend/public/index.html
```diff
@@ -5,39 +5,13 @@
     <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <meta name="theme-color" content="#000000" />
-    <meta
```
tihuan, line 8: remove Create React App tutorial text

src/frontend/src/common/api/index.tsx
```diff
@@ -119,7 +119,3 @@ const TREE_MAP = new Map<string, keyof Tree>([
 ]);
 export const fetchTrees = (): Promise<TreeResponse> =>
   apiResponse<TreeResponse>(["phylo_trees"], [TREE_MAP], API.PHYLO_TREES);
-
-export const logout = (): Promise<Response> => {
```
tihuan, line 123: This is dead code, since logout endpoint needs to be hit with redirect directly instead of an API call

src/frontend/src/common/components/library/button/index.tsx
```diff
@@ -27,7 +27,11 @@ const Button: FunctionComponent<Props> = ({
   if (link === undefined) {
     return buttonElements;
   }
-  return <a href={link}>{buttonElements}</a>;
+  return (
+    <a href={link} target="_blank" rel="noopener">
```
tihuan, line 31: Open Nextstrain.org in a new tab instead of the current tab

src/frontend/src/server/config/cspSettings.ts
```diff
@@ -0,0 +1,26 @@
+import { ContentSecurityPolicyOptions } from "helmet-csp";
```
tihuan, line 1: This is where we set Content Security Policy

src/frontend/src/server/index.ts
```diff
@@ -0,0 +1,29 @@
...
+import { cspSettings } from "./config/cspSettings";
+
+const buildDir = path.join(process.cwd() + "/build");
+const app = express();
+
+const ONE_YEAR_S = 365 * 24 * 60 * 60;
+
+// Middlewares
+// NOTE(thuang): `app.use()` injection order matters
+app.use(hsts({ maxAge: 2 * ONE_YEAR_S }));
```
tihuan, line 14: hsts and csp needs to be added first to Express app middleware chain BEFORE any other controllers, otherwise they won't take effect. 

E.g.,

```ts
// If we add `express.static()` controller first before HSTS and CSP, static file controller will return early before running those two middlewares, thus the response header lacking the security stuff
app.use(express.static(buildDir));

app.use(hsts({ maxAge: 2 * ONE_YEAR_S }));
app.use(csp(cspSettings));

```

src/frontend/src/server/tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
```
tihuan, line 1: server folder's own TypeScript config file to allow server code to use node API not available in the client code

src/frontend/src/views/AgreeTerms/index.tsx
```diff
@@ -120,15 +110,16 @@ export default function AgreeTerms(): JSX.Element {
...
-          <Button
-            disabled={isLoading}
-            color="primary"
-            variant="outlined"
-            isRounded
-            onClick={handleDeclineClick}
-          >
-            Decline
-          </Button>
+          <a href={process.env.API_URL + API.LOG_OUT}>
```
tihuan, line 113: On "Decline", we redirect the user to our logout endpoint

src/frontend/tsconfig.json
```diff
@@ -35,11 +35,12 @@
   "exclude": [
     "node_modules",
     "public",
-    ".cache"
+    ".cache",
+    "src/server"
```
tihuan, line 39: Make sure we don't include server code in client's TS transpilation

src/frontend/src/views/AgreeTerms/index.tsx
ttung, line 113: does js have a better way of concatenating url paths?  (akin to python's [`posixpath.join`](https://github.com/python/cpython/blob/3.9/Lib/posixpath.py#L71)

ttung (approved):
rest lgtm.

src/frontend/src/views/AgreeTerms/index.tsx
tihuan: Ooh interesting! I think FE typically only needs to join two parts, since the latter part is pretty much hard coded all the time, like so 😆 :

```ts
export enum API {
  USER_DATA = "/api/usergroup",
  SAMPLES = "/api/samples",
  LOG_IN = "/login",
  LOG_OUT = "/logout",
  PHYLO_TREES = "/api/phylo_trees",
}
```

tihuan:
Merging now! Thank you!

PR #342, opened 2021-05-05
Update the patch file

ttung:
### Description
https://github.com/nextstrain/ncov/pull/605/ added some of the functionality that we care about in our patch file (specifically the ability to warn on duplicate but non-matching sequence data).  However, we still want a different ordering of the files.

### Test plan
started a build, got past the patch phase.


phoenixAja (approved):
#thankful this suspenseful rollercoaster ride is about to be over lol

PR #346, opened 2021-05-06
Flatten *aspen-config

ttung:
### Description
SecretsManager doesn't have a nice facility for editing nested json.

### Test plan
grepped for `DB.`


jgadling (changes requested):
Keeping a "." in key names makes it really easy to misunderstand whether we're navigating a nested data structure or referring to a literal key.

ttung:
Replaced `.` with `/`.  Rationale: same separator as AWS parameter store.

jgadling:
Slashes in json keys is also pretty nonstandard. can we use - or _ ?

ttung:
Only `_` is allowed by jq.  Even `-` is not allowed.

ttung:
I guess if the criteria is "allowed by jq" then we will just have to use `_`.

ttung:
Switched to using `_`

shanaxel42 (approved):
looks good

PR #350, opened 2021-05-06
thuang-adjust-gisaid-status-133230

tihuan:
### Description

Update GISAID status message in FE

#### Issue
https://app.clubhouse.io/genepi/story/133230/adjust-gisaid-statuses

### Test plan

1. Visit https://thuang-gisaid-status-frontend.dev.genepi.czi.technology/data/samples
2. The status should now be one of: 

```
{
  accepted: "Accepted",
  no_info: "Not Yet Submitted",
  not_eligible: "Not Eligible",
  rejected: "Rejected",
  submitted: "Submitted",
}
```

Screen Shot 2021-05-06 at 2 56 04 PM


src/frontend/src/common/api/index.tsx
```diff
@@ -82,7 +82,7 @@ const USER_MAP = new Map<string, keyof User>([
   ["auth0_user_id", "auth0UserId"],
   ["group_admin", "groupAdmin"],
   ["system_admin", "systemAdmin"],
-  ["group_id", "groupdId"],
+  ["group_id", "groupId"],
```
tihuan, line 85: typo! My bad 🤦

src/frontend/src/common/types/bioinformatics.d.ts
```diff
@@ -3,14 +3,19 @@ interface BioinformaticsType {
   type: "BioinformaticsType";
 }
 
+interface GISAID {
```
tihuan, line 6: GISAID response is now an object instaed of a string

src/frontend/src/views/Data/cellRenderers.tsx
```diff
@@ -11,18 +11,32 @@ import {
...
 };
 
 const SAMPLE_CUSTOM_RENDERERS: Record<string | number, CellRenderer> = {
-  privateId: (value: JSONPrimitive): JSX.Element => (
-    <div className={dataTableStyle.cell}>
-      {<SampleIcon className={dataTableStyle.icon} />}
-      {value}
-    </div>
-  ),
+  gisaid(value): JSX.Element {
```
tihuan, line 22: Add custom renderer for `gisaid` cell

tihuan (commented):
PTAL thanks all!

src/frontend/src/common/types/network.d.ts
```diff
@@ -1,4 +1,4 @@
-type JSONPrimitive = string | number | boolean;
+type JSONPrimitive = string | number | boolean | Record<string, unknown>;
```
lvreynoso: hmm, it's not really a primitive any more is it? why modify this type instead of changing the type of the variable to `JSONPrimitive | Record<string, unknown>`?

src/frontend/src/common/types/bioinformatics.d.ts
lvreynoso: I think it would be better to just add two keys, `gisaid_status` and `gisaid_id`, instead of embedding an object just for two keys.

src/frontend/src/views/Data/cellRenderers.tsx
lvreynoso: In a previous iteration of this I added a reference to the parent object the value came from in case we needed to reference more than one value in a column, I think that would be a better approach.

src/frontend/src/views/Data/style.ts
```diff
@@ -8,3 +9,19 @@ export const Container = styled.div`
...
+    return `
+      color: ${colors?.gray[400]};
+    `;
+  }}
+`;
+
+export const GISAIDCell = styled.div`
+  flex-direction: column;
+  align-items: unset;
+`;
```
lvreynoso, line 27: If these are just styled divs, wouldn't they be better defined as CSS classes?

lvreynoso (commented):
Some comments

src/frontend/src/views/Data/cellRenderers.tsx
tihuan: Ah yes! In my Pangolin PR, renderer will accept an object as an argument `gisaid({ value, item, index })`, so that will be fixed there. Currently the `item` is the second argument of this render function. E.g, `gisaid(value, item, index)`

src/frontend/src/common/types/bioinformatics.d.ts
tihuan: Ah I was thinking an object here works out pretty well , since it naturally namespaces `gisaid_`. What do you like about `gisaid_status` and `gisaid_id`?

src/frontend/src/common/types/network.d.ts
tihuan: That's true! I'm thinking about changing `JSONPrimitive` to just `JSONValue` and expand it to:

```ts
type JSONValue = 
 | string
 | number
 | boolean
 | null
 | JSONValue[]
 | {[key: string]: JSONValue}
```

Since that's all the valid JSON types we can get

source: https://github.com/microsoft/TypeScript/issues/1897#issuecomment-822032151

What do you think?

src/frontend/src/views/Data/style.ts
tihuan: Ah yes these are styled using CSS in JS solution, which provides better developer experience, app performance, and styling flexibility, since it naturally works well with JS. There shouldn't be a need of us using CSS Modules anymore!

src/frontend/src/common/types/network.d.ts
lvreynoso: I think that's a good idea but it should be a separate type, since I am pretty sure I relied on constraining variable to just primitives where specified in the code.

src/frontend/src/views/Data/style.ts
lvreynoso: I think better developer experience is a matter of preference here, but would we really eliminate CSS modules? How would we handle pseudo classes or states?

src/frontend/src/common/types/bioinformatics.d.ts
lvreynoso: It keeps the response a flat object, which I find easier to reason about and perform transformations on, or write generic functions involving them.

src/frontend/src/views/Data/style.ts
tihuan: Yup! CSS in JS solution works for all pseudo classes and states: https://emotion.sh/docs/@emotion/jest#custom-matchers, and it's the same syntax as Sass

Meta and Single Cell both use [Styled Components](https://styled-components.com/) for styling as well, which is another CSS in JS solution. Extremely similar to Emotion, but Emotion is better React future proof 👌

src/frontend/src/common/types/bioinformatics.d.ts
tihuan: Ah key transformation and flattening 💡  

Meta went down the API response key transformation path 4 years ago that recursively transform key from snake_case to camelCase AND flatting the response, the engineers found it pretty confusing and bug prone, especially around typings, debugging, and namespace collision 

So we switched over to just having the single source of truth being the API response, and accepting snake_case for JSON responses, and keeping the nested structure. Then we type each response interface individually composed of entity interfaces. This made things a lot easier to reason and more predictable. Single Cell Team is doing the same thing too!

This also means that we no longer need the generic function for key transformation, which is one less layer of code to deal with, and we get to keep things simple!

tihuan (commented):
@lvreynoso I reverted `type JSONPrimitive = string | number | boolean;` and we can look out for adding `type JSONvalue` in the future if the needs arise!

PTAL again thank you!

shanaxel42:
WOOOOOOO! But @tihuan when I visit your remote branch I see all the samples with the `Not Eligable` tag even though it looks like there's a bunch of public ids?

tihuan:
@shanaxel42 Thanks for spotting this! It seems like the API response reflects the same thing:

Screen Shot 2021-05-07 at 2 57 31 PM

Wild guess: Is that a database snapshot issue?

CC: @ttung 

Thank you!

ttung:
https://github.com/chanzuckerberg/aspen/pull/358

tihuan:
Updated rdev with Tony's commits! Looks super good now 🎉 Thank you!  https://thuang-gisaid-status-frontend.dev.genepi.czi.technology/data/samples

Screen Shot 2021-05-10 at 10 21 32 AM

src/frontend/src/common/types/bioinformatics.d.ts
lvreynoso: I meant more data transformations, but :shrug:

lvreynoso (approved):

tihuan (commented):
Thanks Lucia!! Merging 🚀 🚀 🚀

PR #351, opened 2021-05-06
Remove a bunch of old elastic beanstalk setup code

ttung:
### Description
These configuration profiles and documentation pertain to the elastic beanstalk steup.

Depends on #346 

### Test plan
GHA


phoenixAja:
lol i was just thinking of doing this

src/backend/scripts/setup_dev_data.sh
```diff
@@ -29,5 +29,9 @@ ${local_aws} secretsmanager update-secret --secret-id aspen-config --secret-stri
   "AUTH0_USERINFO_URL": "connect/userinfo",
   "AUTH0_ACCESS_TOKEN_URL": "'"${OIDC_INTERNAL_URL}"'/connect/token",
   "AUTH0_AUTHORIZE_URL": "'"${OIDC_BROWSER_URL}"'/connect/authorize",
-  "AUTH0_CLIENT_KWARGS": {"scope": "openid profile email offline_access"}
+  "AUTH0_CLIENT_KWARGS": {"scope": "openid profile email offline_access"},
+  "FLASK_SECRET": "DevelopmentKey",
+  "DB.rw_username": "user_rw",
```
jgadling: Can we not put "." chars in keys? It's pretty confusing

ttung: Updated to use `/` as well.

src/backend/aspen/cli/db.py
```diff
@@ -38,9 +36,8 @@
 
 
 @cli.group()
-@click.option("--local", "config_cls", flag_value=DevelopmentConfig, default=True)
-@click.option("--remote", "config_cls", flag_value=RemoteDatabaseConfig)
-@click.option("--docker", "config_cls", flag_value=LocalConfig)
+@click.option("--remote", "config_cls", flag_value=Config)
+@click.option("--docker", "config_cls", flag_value=DockerComposeConfig, default=True)
```
phoenixAja: i like `local` as a flag here instead of `docker`, but that's a small nit

phoenixAja (approved):
this looks good to me, do we have to change the docs anywhere to reflect these changes?

src/backend/aspen/cli/db.py
ttung: I think of `--local` as without docker, but I can change it.

PR #352, opened 2021-05-07
presigned url link for auspice with tree rewrite

ttung:
### Description
Rewrite the tree data and save to S3.  Then return a presigned url to that key.

Modified existing docker-compose infrastructure to set up the bucket.

Depends on #351 

#### Issue
[ch133611](https://app.clubhouse.io/genepi/story/133611/rewrite-tree-data-with-private-identifiers)

### Test plan
Modified test to retrieve the data from the presigned url


ttung:
@jgadling if you could review the changes in ` src/backend/scripts/setup_dev_data.sh`, that would be great.

shanaxel42 (approved):

src/backend/aspen/app/views/tests/test_auspice_redirect.py
```diff
@@ -1,33 +1,63 @@
...
+    wrong_can_see_sample = sample_factory(
+        wrong_can_see_group,
+        private_identifier="private_identifer_3",
+        public_identifier="public_identifier_3",
+    )
+    no_can_see_sample = sample_factory(
+        no_can_see_group,
+        private_identifier="private_identifer_4",
+        public_identifier="public_identifier_4",
+    )
```
phoenixAja, line 55: i'm a little confused on the difference between `no_can_see` and `wrong_can_see`, like when would a `wrong_can_see` happen in real life?

ttung: `wrong_can_see` means like you can see trees of another group, but not the private identifiers.
`no_can_see` means there are no can-see relationships of any type.

PR #356, opened 2021-05-08
Fix the import of external samples

ttung:
### Description
#298 broke the import for external samples.  I forgot to add the list of external czbids to the list of external accessions to examine.

### Test plan
ran migrate-all.sh and it seemed to import a lot more samples.


shanaxel42 (approved):
I guess this is a repeat of #360 now

PR #358, opened 2021-05-08
Fix default for czb_failed_genome_recovery

ttung:
### Description
It should default to false and not true.

Added a migration that fixes existing databases.  Reversing the migration does not restore the previous data state!

### Test plan
Ran migration


shanaxel42 (approved):
Ah nice!! this fixes what timmy was seeing!

tihuan:
Amazing!! Thanks so much for the quick fix, Tony!!

tihuan (approved):
LGTM!!

PR #361, opened 2021-05-10
Replace Accessions/AccessionWorkflows with repository-specific types

ttung:
### Description
This is necessary to disambiguate between failed accession workflows to different repositories.

Note that the db migration isn't a "good" one in the sense that it preserves data.  It simply deletes existing accessions.

### Test plan
migrated from covidhub.


shanaxel42 (approved):
lgtm

PR #363, opened 2021-05-10
fix samples endpoint

ttung:
### Description
This allows us to not execute N queries during the samples endpoint for the gisaid data.

Depends on #361 

### Test plan
turned on profiling and verified that the samples endpoint executes 3 queries now.


shanaxel42 (approved):
ooof I knew I slowed this down!!

PR #365, opened 2021-05-10
add lineage to sample endpoint

phoenixAja:
### Description

add lineage info to sample endpoint

#### Issue
[ch136408](https://app.clubhouse.io/genepi/story/136408)

### Test plan
changed tests to reflect new endpoint changes


shanaxel42 (approved):
niccccce!!

src/backend/aspen/app/views/sample.py
```diff
@@ -84,6 +84,28 @@ def _format_gisaid_accession(
...
+                pathogen_genome.pangolin_last_updated
+            ),
+        }
+    else:
+        lineage = {
+            "lineage": None,
+            "probability": None,
+            "version": None,
+            "last_updated": None,
+        }
```
ttung, line 104: Can you add an issue that we need to handle this for `CalledConsensusGenomes`?

src/backend/aspen/app/views/sample.py
```diff
@@ -84,6 +84,28 @@ def _format_gisaid_accession(
     return {"status": "no_info", "gisaid_id": None}
 
 
+def _format_lineage(sample: Sample):
```
ttung:
```suggestion
def _format_lineage(sample: Sample) -> dict[str, Any]:
```

src/backend/aspen/app/views/tests/test_samples_view.py
```diff
@@ -25,7 +25,13 @@ def test_samples_view(
     group = group_factory()
     user = user_factory(group)
     sample = sample_factory(group)
-    uploaded_pathogen_genome = uploaded_pathogen_genome_factory(sample)
+    pathogen_genome = uploaded_pathogen_genome_factory(
```
ttung: i don't agree with the rename of `uploaded_pathogen_genome` -> `pathogen_genome`.

src/backend/aspen/app/views/tests/test_samples_view.py
```diff
@@ -25,7 +25,13 @@ def test_samples_view(
...
+        sample,
+        pangolin_lineage="B.1.590",
+        pangolin_probability=1.0,
+        pangolin_version="2021-04-23",
+        pangolin_last_updated=datetime.datetime.now(),
```
ttung: Consider adding this to the factory so it doesn't have to be repeated so many times.

phoenixAja: added

src/backend/aspen/app/views/tests/test_samples_view.py
phoenixAja: changed it back

src/backend/aspen/app/views/tests/test_samples_view.py
phoenixAja: done

src/backend/aspen/app/views/sample.py
phoenixAja: Sure will do, in that case will we be getting the lineage info from a `CalledPathogenGenome`?

phoenixAja: does [this](https://app.clubhouse.io/genepi/story/136767/lineage-info-for-when-we-have-a-called-consensus-genome-instead-of-an-uploaded-pathogen-genome) work?

ttung: > Sure will do, in that case will we be getting the lineage info from a `CalledPathogenGenome`?

Yes.

> does [this](https://app.clubhouse.io/genepi/story/136767/lineage-info-for-when-we-have-a-called-consensus-genome-instead-of-an-uploaded-pathogen-genome) work? 

Perfect!

PR #370, opened 2021-05-11
Fix for tuolumne county

ttung:
### Description
Build a list of all projects and try to import from the first one.

### Test plan
Commented out all except for tuolumne and ran migrate-all.sh

### DIFF
--- a/scripts/migrate-all.sh
+++ b/scripts/migrate-all.sh
@@ -17,7 +17,7 @@ COUNTY_INFO='{
         "ventura": {"external_project_id": "RR078e"},
         "humboldt": {"external_project_id": "RR075e"},
         "vrdl": {"external_project_id": "RR096e"},
-        "tuolumne": {"internal_project_ids": "RR095i"},
+        "tuolumne": {"internal_project_ids": ["RR095i"]},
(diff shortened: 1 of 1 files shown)


src/backend/aspen/covidhub_import/import_trees.py
```diff
@@ -128,7 +128,7 @@ def import_trees(
             workflow.inputs = list(all_uploaded_pathogen_genomes)
             workflow.outputs = [phylo_tree]
 
-            s3_dst.Bucket(phylo_tree.s3_bucket).Object(phylo_tree.s3_key).put(Body=data)
+            # s3_dst.Bucket(phylo_tree.s3_bucket).Object(phylo_tree.s3_key).put(Body=data)
```
shanaxel42: probs want this?

shanaxel42 (changes requested):

src/backend/aspen/covidhub_import/import_trees.py
ttung: whoops, this shouldn't be in there.

shanaxel42 (approved):

PR #374, opened 2021-05-11
Upgrade batch base image.

jgadling:
### Description

The base batch/miniwdl image is built against a [branch of miniwdl plugins](https://github.com/chanzuckerberg/miniwdl-plugins/compare/akislyuk-swipe-aspen-vars) that we've modified to allow some core aspen env vars to be passed through to worker steps. The remote-dev environment uses an env var called REMOTE_DEV_PREFIX, but the staging and prod envs do not. Our patched version of miniwdl-plugins was *requiring* that env var to be set. I've rebuilt the image with a patch that allows the REMOTE_DEV_PREFIX env var to be missing

#### Issue

### Test plan


ttung (approved):

shanaxel42 (approved):
nice!

PR #376, opened 2021-05-11
Rename the override var for secret name

ttung:
### Description
It's not just for auth0.

### Test plan
git grep AUTH0_CONFIG_SECRET_NAME


jgadling (approved):
Headsup that this will break batch jobs for a little bit until I get a new base image built. Should only be a few minutes.

jgadling:
Added a patch to this branch that means this shouldn't break batch anymore

PR #377, opened 2021-05-11
Don't import VRDL trees

ttung:
### Description
They don't exist!

Depends on #370 

#### Issue
[ch136615](https://app.clubhouse.io/genepi/story/136615/vrdl-trees-do-not-port)

### Test plan
removed all but vrdl and tulare and ran import.

PR #378, opened 2021-05-11
Mark samples with valid recovered genomes but no genome submission info as gisaid-rejected

ttung:
### Description
Add a failed workflow to signify a failed submission.

Depends on #377 

### Test plan
will import and test


shanaxel42 (approved):
Ahhhh the rejected case

PR #383, opened 2021-05-11
Gisaid workflow

jgadling:
### Description

1. Build a docker image capable of running the gisaid workflow steps (based on the nexstrain/base image)
2. Pin the version of ncov we use during the docker image build step, since the latest ncov commits have some breaking changes in them
3. Update our build pipeline to pass the current git SHA into the docker image at build time

#### Issue

### Test plan


tihuan (approved):
rubberstamping in case it's time sensitive 😆

PR #391, opened 2021-05-12
Small fixes, copy fixes

jshoe:
### Description

@oliviabholmes this changes `visualization on auspice.us, which is not controlled by Aspen` -> `visualization on Nextstrain, which is not controlled by Aspen` since auspice.us is a different website from Nextstrain.

### Test plan

- Tested locally.


tihuan (approved):
Amazing!! Thanks for the fixes, @jshoe 🙏 🙏 🙏

PR #392, opened 2021-05-12
Sort data by upload date and creation date

lvreynoso:
### Description

Sorts trees and samples by upload / creation date.

#### Issue
[ch137016](https://app.clubhouse.io/genepi/story/137016)

### Test plan

### DIFF
--- a/src/frontend/src/views/Data/index.tsx
+++ b/src/frontend/src/views/Data/index.tsx
@@ -12,9 +12,9 @@ import { Container } from "./style";
 import { TREE_TRANSFORMS } from "./transforms";
 
 const sortByKeys: Record<string, string> = {
-  "Samples": "uploadDate",
   "Phylogenetic Trees": "creationDate",
-}
+  Samples: "uploadDate",
+};
 
 const Data: FunctionComponent = () => {
   const [samples, setSamples] = useState<Sample[] | undefined>();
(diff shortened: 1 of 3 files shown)


src/frontend/src/views/Data/index.tsx
```diff
@@ -78,6 +83,20 @@ const Data: FunctionComponent = () => {
     category.data = transformedData;
   });
 
+  // sort data by creation date
+  dataCategories.forEach((category) => {
+    if (category.data === undefined) {
+      return;
+    }
+    const sortKey = sortByKeys[category.text]
+    console.log(sortKey)
```
tihuan: nit: unneeded console.log 😄

src/frontend/src/views/Data/index.tsx
```diff
@@ -78,6 +83,20 @@ const Data: FunctionComponent = () => {
     category.data = transformedData;
   });
 
+  // sort data by creation date
+  dataCategories.forEach((category) => {
+    if (category.data === undefined) {
```
tihuan, line 88: nit: `!category.data` would do instead of `category.data === undefined`

tihuan (approved):
LGTM after two small nits! Thanks @lvreynoso 🎉 !

src/frontend/src/views/Data/index.tsx
lvreynoso: Typescript needs you to explicitly rule out undefined in order for type checking to work though.

tihuan: ah right, `if (!category.data)` should have the same effect as `if (category.data === undefined)`, just less code? Or does TS do something different in those two cases? 

Thanks!

PR #398, opened 2021-05-12
default lineage display value

morsecodist:
### Description

Add a default display value if we don't have a lineage.

Selection_084

#### Issue
[ch-137263](https://app.clubhouse.io/genepi/story/137263)

### Test plan

Tested locally


tihuan (approved):
LGTM!! Love the fix 👍

PR #405, opened 2021-05-12
Update Auspice link hit area

j-x-han:
### Description

Updates the Auspice link hit area to a single clickable area that covers the icons and phylo tree name.

Everything in the orange outline is a single clickable area.
image

#### Issue
[ch137514](https://app.clubhouse.io/genepi/story/137514)

### Test plan

Verify that there is 1 hit area that covers the phylo icon, tree name, and link icon, and the cursor doesn't flicker when passing between them.


tihuan (approved):
Great solution! LGTM 🎉 !!

PR #406, opened 2021-05-12
Add right margins to table columns

j-x-han:
### Description

Adds `$space-m` to the right side of all table columns.

**Before:**
image

**After:**
(Might be worth getting a look from Janeece to see if we want to resisze any of these columns to give the Public ID more space? Although it might be resolved by ch137076)
image

#### Issue
[ch137092](https://app.clubhouse.io/genepi/story/137092)

### Test plan

Check that all table columns have a right margin of 10px and content doesn't run into each other anymore.


tihuan (approved):
LGTM! 🎉

PR #409, opened 2021-05-12
Fix External Accessions / Sample Name TSV for trees

lvreynoso:
### Description

Extracts external accessions both from the old key (`node["node_attrs"]["external_accession"]) and the new key (node["name"]). Leaves out names that begin with `NODE_`.

#### Issue
[ch137187](https://app.clubhouse.io/genepi/story/137187)

### Test plan

Tested in Python REPL with a staging tree json.

### DIFF
--- a/src/backend/aspen/app/views/phylo_trees.py
+++ b/src/backend/aspen/app/views/phylo_trees.py
@@ -1,5 +1,6 @@
 import json
 import os
+import re
 import uuid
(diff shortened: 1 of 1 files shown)


shanaxel42 (approved):
YASSSSSS

ttung:
Why are we dropping the nodes that start with `NODE_`?  The next person who reads this code will be very confused.

ttung:
Code seems ok.  Definitely this deserves a comment though.

PR #411, opened 2021-05-12
restrict usergroup data

phoenixAja:
### Description

restrict the information returned from usergroup api endpoint

#### Issue
[ch137500](https://app.clubhouse.io/genepi/story/137500)

### Test plan

changed test to expect new data


src/backend/aspen/app/views/tests/test_usergroup_views.py
```diff
@@ -12,7 +12,12 @@ def test_usergroup_view_get(session, app, client):
     with client.session_transaction() as sess:
         sess["profile"] = {"name": user.name, "user_id": user.auth0_user_id}
     res = client.get("/api/usergroup")
-    expected = {"user": user.to_dict(), "group": group.to_dict()}
+    expected = {
+        "user": {
+            k: v for k, v in user.to_dict().items() if k in ["name", "agreed_to_tos"]
+        },
+        "group": {k: v for k, v in user.group.to_dict().items() if k in ["name"]},
+    }
```
ttung: I would say that at this point, you might as well be explicit about what you're returning:

```
{"user": { "name": "xxx", "agreed_to_tos": "yyy"}, "group": {"name": "zzz"}}
```

ttung (changes requested):

src/backend/aspen/app/views/tests/test_usergroup_views.py
phoenixAja: i made some changes where i wrap this in a function, maybe it's a better approach

phoenixAja: also i tried doing:
`session.query(User.name, User.agreed_to_tos, Group.name).select_from(User).join(User.Group)` and 
`session.query(User).options(load_only(“agreed_to_tos”, “name”), joinedload(User.group).load_only(“name”))` to do the filtering in the query itself, but both methods had faults

src/backend/aspen/app/views/api_utils.py
```diff
@@ -8,6 +8,13 @@
 from aspen.database.models.usergroup import User
 
 
+def filter_usergroup_dict(
+    unfiltered_dict: Mapping[str, Union[str, bool]],
+    fields_to_keep: Iterable[Union[str, bool]],
```
ttung:
```suggestion
    fields_to_keep: Collection[str],
```

src/backend/aspen/app/views/tests/test_usergroup_views.py
```diff
@@ -12,7 +13,10 @@ def test_usergroup_view_get(session, app, client):
     with client.session_transaction() as sess:
         sess["profile"] = {"name": user.name, "user_id": user.auth0_user_id}
     res = client.get("/api/usergroup")
-    expected = {"user": user.to_dict(), "group": group.to_dict()}
+    expected = {
+        "user": filter_usergroup_dict(user.to_dict(), ["name", "agreed_to_tos"]),
+        "group": filter_usergroup_dict(user.group.to_dict(), ["name"]),
```
ttung: These allowed fields should be constants (tuples?  frozensets?) in `src/backend/aspen/app/views/usergroup.py`.

ttung (changes requested):
very close!

src/backend/aspen/app/views/tests/test_usergroup_views.py
phoenixAja: added some tuples!

src/backend/aspen/app/views/api_utils.py
phoenixAja: what's the main difference between iterable and collection? They seem the same to me

ttung: Collection supports the `in` operation.

ttung (approved):

src/backend/aspen/app/views/api_utils.py
phoenixAja: oh interesting that makes sense

PR #421, opened 2021-05-13
Switch to using nextalign for gisaid alignment

ttung:
### Description
It burns less memory.

### Test plan
successfully ran a gisaid alignment job

### DIFF
--- a/.happy/terraform/modules/sfn_config/gisaid.wdl
+++ b/.happy/terraform/modules/sfn_config/gisaid.wdl
@@ -234,7 +234,9 @@ task AlignGISAID {
     # fetch the gisaid dataset
     aws s3 cp --no-progress s3://"${processed_gisaid_s3_bucket}"/"${processed_gisaid_sequences_s3_key}" - | zstdmt -d > /ncov/data/sequences.fasta
     aws s3 cp --no-progress s3://"${processed_gisaid_s3_bucket}"/"${processed_gisaid_metadata_s3_key}" /ncov/data/metadata.tsv
-    (cd /ncov; snakemake --printshellcmds results/aligned.fasta --profile /usr/src/app/aspen/workflows/align_gisaid 1>&2)
+    mkdir /ncov/my_profiles/align_gisaid/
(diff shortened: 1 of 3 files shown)


ttung:
So this wasn't enough alone to get the gisaid workflow running.  I also had to hack the wdl to inject the environment variables `ASPEN_CONFIG_SECRET_NAME` and `REMOTE_DEV_PREFIX`.  I did not include that in my commit because I'm pretty sure what I'm doing is terrible and non-portable, and @jgadling probably has a better solution for that.

ttung:
For reference: https://gist.github.com/ttung/a4ff4831117e7a6860455985259209f4

shanaxel42 (approved):

.happy/terraform/modules/sfn_config/gisaid.wdl
```diff
@@ -234,7 +234,9 @@ task AlignGISAID {
     # fetch the gisaid dataset
     aws s3 cp --no-progress s3://"${processed_gisaid_s3_bucket}"/"${processed_gisaid_sequences_s3_key}" - | zstdmt -d > /ncov/data/sequences.fasta
     aws s3 cp --no-progress s3://"${processed_gisaid_s3_bucket}"/"${processed_gisaid_metadata_s3_key}" /ncov/data/metadata.tsv
-    (cd /ncov; snakemake --printshellcmds results/aligned.fasta --profile /usr/src/app/aspen/workflows/align_gisaid 1>&2)
+    mkdir /ncov/my_profiles/align_gisaid/
+    cp /usr/src/app/aspen/workflows/align_gisaid/{builds.yaml,config.yaml} /ncov/my_profiles/align_gisaid/
```
jgadling, line 238: any reason to copy the profile files into a different directory?

ttung: the config files referenced in `config.yaml` is relative to ncov's root.  I didn't want to hard code an absolute path into the config file.  As such, I would need to copy `builds.yaml` to the ncov tree.  And if I'm doing that, I might as well copy the `config.yaml` to the prescribed place as well.  This is the way nextstrain organizes its profiles, and I don't see why we need to deviate from that.

PR #423, opened 2021-05-13
Generate SFN's with and without trying on SPOT instances first.

jgadling:
### Description

Create two SFN's: 
- one that tries running a job on SPOT instances first, and then retries on on-demand EC2 instances if that fails. This is useful for cost savings for any jobs that are re-entrant.
- another that runs jobs on ECS hosts without trying SPOT first. These hosts are more expensive, but also more reliable. They're the best choice for any jobs that are difficult to re-run when they fail partway through

#### Issue

### Test plan

I applied these on my remote-dev env and invoked them via the AWS console.


shanaxel42 (approved):
nice!!

PR #426, opened 2021-05-13
[super easy] remove eb ignore

phoenixAja:
### Description

removing stray .ebignore file because i don't think we need it anymore? 

#### Issue
no issue

### Test plan
nothing


ttung (approved):
so easy.

PR #431, opened 2021-05-14
fix for internal sample dupes

shanaxel42:
### Description

There's an edge case of duplicate qpcr side accessions matching to a czb_id, one with a submitter_id and one without. This picks the one with the id

#### Issue

### Test plan

### DIFF
--- a/src/backend/aspen/covidhub_import/import_projects.py
+++ b/src/backend/aspen/covidhub_import/import_projects.py
@@ -215,23 +215,27 @@ def import_project(
 
         # load the existing samples, store as a mapping between external accession to
         # sample.
-        external_accessions_to_samples: Mapping[str, Sample] = {
-            sample.private_identifier: sample
-            for sample in (
-                session.query(Sample)
-                .filter(Sample.submitting_group == group)
-                .options(
-                    joinedload(Sample.uploaded_pathogen_genome)
-                    .joinedload(
-                        UploadedPathogenGenome.consuming_workflows.of_type(  # type: ignore
-                            GisaidAccessionWorkflow
-                        )
-                    )
-                    .joinedload(
-                        GisaidAccessionWorkflow.outputs.of_type(GisaidAccession)  # type: ignore
+        samples = (
+            session.query(Sample)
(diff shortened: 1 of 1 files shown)


src/backend/aspen/covidhub_import/import_projects.py
```diff
@@ -234,6 +234,27 @@ def import_project(
...
+                            GisaidAccessionWorkflow
+                        )
+                    )
+                    .joinedload(
+                        GisaidAccessionWorkflow.outputs.of_type(GisaidAccession)  # type: ignore
+                    )
+                )
+            )
+        }
+
```
ttung: This should be combined with the immediately preceding block as one query:

```
samples = (session.query(Sample)
               .filter(Sample.submitting_group == group)
               .options(
                   joinedload(Sample.uploaded_pathogen_genome)
                   .joinedload(
                       UploadedPathogenGenome.consuming_workflows.of_type(  # type: ignore
                           GisaidAccessionWorkflow
                       )
                   )
                   .joinedload(
                       GisaidAccessionWorkflow.outputs.of_type(GisaidAccession)  # type: ignore
                   )
               )
               .all()
)
external_accessions_to_samples: Mapping[str, Sample] = {
    sample.private_identifier: sample
    for sample in samples
}
public_id_to_samples: Mapping[str, Sample] = {
    sample.public_identifier: sample
    for sample in samples
}
```

ttung (changes requested):
rest lgtm.

src/backend/aspen/covidhub_import/import_projects.py
shanaxel42: done!

ttung:
Wait, I don't get this. Wouldn't a better solution just be to exclude anything without private identifiers?

shanaxel42:
> Wait, I don't get this. Wouldn't a better solution just be to exclude anything without private identifiers?

we could....there are a few internal samples that were cherry picked without location_submitter_ids so it would then exclude those. But going forward it would probably be fine to just do that

PR #428, opened 2021-05-13
Initial support for scheduled tasks.

jgadling:
### Description
This is an outline for implementing scheduled jobs in prod. I'm not sure what the specific schedule needs to look like though.

#### Issue
[ch<fill_in_issue_number>](https://app.clubhouse.io/genepi/story/<fill_in_issue_number>)

### Test plan

I did some spot-checks in my remote-dev environment, and the cloudwatch rules/targets are created, and SFN's invoked at the scheduled times

PR #432, opened 2021-05-14
Refactor the gisaid workflow

ttung:
### Description
1. Source of truth is aspen config / remote dev stack name.  Everything eles is junked.
2. Don't name the key the same every time.  That angers the db as we imposed a unique constraint.
3. Pass only the entity ids between subroutines.

### Test plan
<strike>running now.</strike> Successfully ran gisaid workflow in rdev.

### DIFF
--- a/.happy/terraform/modules/sfn_config/gisaid.wdl
+++ b/.happy/terraform/modules/sfn_config/gisaid.wdl
@@ -4,44 +4,48 @@ workflow LoadGISAID {
     input {
         String docker_image_id = "aspen-gisaid"
         String aws_region = "us-west-2"
-        String db_data_bucket = "aspen-data-rdev"
-        String gisaid_ndjson_export_url = "https://www.epicov.org/epi3/3p/exp3/export/export.json.bz2"
-        String gisaid_ndjson_staging_bucket = "aspen-data-rdev"
-        String gisaid_ndjson_staging_key = "raw_gisaid_dump/test.zst"
+        String aspen_config_secret_name
+        String remote_dev_stack_name = ""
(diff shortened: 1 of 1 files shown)


.happy/terraform/modules/sfn_config/gisaid.wdl
```diff
@@ -4,44 +4,48 @@ workflow LoadGISAID {
     input {
         String docker_image_id = "aspen-gisaid"
         String aws_region = "us-west-2"
-        String db_data_bucket = "aspen-data-rdev"
-        String gisaid_ndjson_export_url = "https://www.epicov.org/epi3/3p/exp3/export/export.json.bz2"
-        String gisaid_ndjson_staging_bucket = "aspen-data-rdev"
-        String gisaid_ndjson_staging_key = "raw_gisaid_dump/test.zst"
+        String aspen_config_secret_name
+        String remote_dev_stack_name = ""
```
jgadling, line 8: looks great! I think we need to update our terraform to pass these values in to the SSM parameters as well though!

ttung: I _think_ we only need to pass in aspen_config_secret_name.  Everything else is kind of optional.

jgadling: yeah, but we may as well set the stack name as well -- right now we're writing the inputs for a given env to an SSM parameter (#428 promotes these to cloudwatch event inputs) to make it easier to know how to invoke these SFN's: https://github.com/chanzuckerberg/aspen/blob/trunk/.happy/terraform/modules/sfn_config/main.tf#L8

They should be updated to add the additional params.

PR #437, opened 2021-05-14
Send human-readable GISAID status from the api response

lvreynoso:
### Description

Returns human-readable GISAID statuses from the samples API instead of a snake_case value. Since sample search keys on the values of the sample object and not any transformations in rendering, this ensures that users are able to search on what they actually see.

(The statuses displayed are the same; they are just now searchable, too.)

#### Issue
[ch136817](https://app.clubhouse.io/genepi/story/136817)

### Test plan


shanaxel42 (approved):
awesome!

PR #438, opened 2021-05-14
Update ssm parameters to reflect wdl inputs.

jgadling:
### Description

Add ssm parameter config for nextstrain, and update config for gisaid

#### Issue

### Test plan

I tested launching SFN's in rdev with these parameters and they seemed to be parsing input and starting up as expected, though I didn't let the jobs run to completion.


shanaxel42 (approved):
woooo!

PR #442, opened 2021-05-15
Fix rdev name for align step

ttung:
### Description
It should be **not** prefixed with a `/`.

### Test plan
Run locally again.


jgadling (changes requested):
This value already has a leading / in it

ttung:
Oh.  Well then it needs to be removed in the other workflows then.

jgadling (approved):

PR #445, opened 2021-05-17
feat[sastisfaction]: adding non-blocking security linting to aspen

jakeyheath:
#### Description
This PR introduces a security tool that is being worked on by @edulop91, @jacobrheath, and the [prodsec team](https://wiki.czi.team/display/TECH/Product+Security).
The tool combines reviewdog and semgrep to make educational security comments about dangerous functions and potentially bad code patterns relating to security.
The main priority of the tool is not to be annoying or blocking, but to bring security learning closer to the development lifecycle. We will do our best to make this tool non-blocking.

#### Notes
- This is a non-blocking tool, and purely for educational purposes. We encourage you to read any comments that are produced and reach out to the prodsec team if you'd like to learn more information.
- If you see a mistake (a comment was created that seems to not be relevant to the code or something similar), let us know! We want to make this as high-signal and non-noisy as possible.
- The tool only runs semgrep rules on the diff from the PR.
- We only have a short number of rules that run right now, but we plan to bring in more over time.
- We have no Python rules at this time, but plan to create some in the future, so expect no comments until those rules are made.

Comments will look like this:
image

- Each comment will have a short description, as well as links to more educational material.
- The addition of new rules or changes to the tool do not require modifications to this codebase.
- If you have rule suggestions or patterns you'd like to prevent in your app please let us know!
- We collect some data around which rules are getting triggered and what action they prompt. This will help us tune our rules, avoid false positives, and hopefully make this a high-quality integration. 

#### FAQs
- *How does this compare with other static analysis tools?*
We opted to use `semgrep` for now (with room to grow into possible different tools) since it gives us the chance to write rules that are not necessarily framework/language specific. We can share rules with other projects at CZI and write rules for various languages with the same format.

- *Will this block my work?*
Our primary goal with this tool is to bring "Security Awareness" earlier into to the development lifecycle. Our hope is this project will help us highlight certain patterns we hope to avoid by tying in external documentation, example code snippets, etc directly into the development cycle. We've spent some time making sure tests run quickly, they are non-blocking, and we collect metrics to make sure we drive down false positives. Data we collect includes the rule that was triggered, the line that triggered it, the repo/SHA/branch, and any corresponding comments/responses in the GitHub PR review.

### DIFF
--- a/.github/workflows/sastisfaction.yml
+++ b/.github/workflows/sastisfaction.yml
@@ -1,0 +1,29 @@
+name: Run SASTisfaction
+on:
+  - pull_request
+
+jobs:
+  sastisfaction:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/checkout@v2
+        with:
+          repository: chanzuckerberg/sastisfaction
+          ref: main
+          path: .github/actions/sastisfaction
+          ssh-key: ${{ secrets.SASTISFACTION_READ_KEY }}
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          # NOTE(el): There is a bug in GitHub <-> ghcr.io that prevents reading "internal" images using secrets.GITHUB_TOKEN
+          #           for now, use a scoped PAT
+          username: cziprodsec
+          password: ${{ secrets.SASTISFACTION_DOCKER_READ_KEY }}
+      - name: Docker pull
+        run: docker pull ghcr.io/chanzuckerberg/sastisfaction:main
+      - name: Run SASTisfaction
+        uses: ./.github/actions/sastisfaction
+        with:
+          snowflake_private_key: ${{ secrets.SASTISFACTION_READ_KEY }}


tihuan (approved):
Looks super cool! Thanks for adding this, @jacobrheath 🙏 !!

CC: @shanaxel42

tihuan:
@jacobrheath clarifying question!

I see that https://semgrep.dev/ supports Python and JavaScript, but you're saying there's no default rules we enabled for Python, so there won't be any comments for our Python code 

How about JavaScript, JSX, TypeScript, and TSX? Do we have rules for them or are there default rules we could use already?

Thank you!

jakeyheath:
@tihuan we do have plans for all these languages and it was one of the reasons for using semgrep (CodeQL was another option, but many of the CZI projects use Ruby which is not supported by CodeQL). They have a lot of language support. Although, we don't want to just throw every rule in the book at all our repos since we were afraid it would create a lot of noise and bug people.

Our plan is to collect metrics from the Github Action on every rule that is run and only turn on rules that we know are high signal and valuable to our developers. We started with 3 types of bugs (command injection, CSRF, and SSRF), but will plan to build these rulesets over time. So even though there aren't any Python or JS/TS rules yet, they will get turned on in time. And once this gets merged, you won't have to do anything, they new rules will automatically get applied. 

Once we start pushing new Python and JS/TS rules, I will be sure to give you a heads up, too! And if you have ideas for security things you'd want to lint for, please reach out and let us know! We can either find an existing rule someone made online, or we can make a custom one.

tihuan:
@jacobrheath ahh got it! Love the strategy 🔥 That sounds awesome! Thanks so much for the explanation 🤩

PR #446, opened 2021-05-17
adding POST to usergroup view

phoenixAja:
### Description

adding POST method to usergroup api endpoint

#### Issue
[ch138281](https://app.clubhouse.io/genepi/story/138281)

### Test plan
Added some tests

### DIFF
--- a/src/backend/aspen/app/views/usergroup.py
+++ b/src/backend/aspen/app/views/usergroup.py
@@ -10,7 +10,14 @@ from aspen.database.models.usergroup import User
 GET_USER_FIELDS: Collection[str] = ("name", "agreed_to_tos")
 GET_GROUP_FIELDS: Collection[str] = ("name",)
 PUT_USER_FIELDS: Collection[str] = ("name", "email", "agreed_to_tos")
-POST_USER_FIELDS: Collection[str] = ("name", "email", "auth0_user_id", "group_admin", "system_admin", "agreed_to_tos", "group_id")
+POST_USER_FIELDS: Collection[str] = (
+    "name",
+    "email",
+    "auth0_user_id",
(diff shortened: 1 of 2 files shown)


src/backend/aspen/app/views/usergroup.py
```diff
@@ -43,3 +51,28 @@ def usergroup():
...
+                if missing_required_fields:
+                    return Response(
+                        f"Insufficient information required to create new user, {missing_required_fields} are required",
+                        400,
+                    )
+                else:
+                    user = User(**new_user_data)
+                    db_session.add(user)
+                    db_session.flush()
+                    return jsonify(success=True)
```
ttung, line 73: Do we want to interface with Auth0 to create the user there?

phoenixAja: that would be fancy! I wasn't sure if that was what we wanted here or not @shanaxel42 what do you think? Was that part of the AC for this?

ttung: FWIW, the auth0 management credentials are in `aspen-config` so you could theoretically do this.

phoenixAja: yes i can definitely do this, do you think i should do that in this PR? or should i merge this and then add that functionality in another PR? I think it might be cleaner to add it in a follow up PR

ttung: Subsequent PR makes sense.

ttung (approved):

phoenixAja:
added an issue here: https://app.clubhouse.io/genepi/story/138635/add-in-auth0-integration-to-create-new-user-in-api-endpoint

PR #449, opened 2021-05-18
Provide a file stem for the trees

ttung:
### Description
This is necessary to customize the tree filename.

### Test plan
will try in rdev


shanaxel42 (approved):
Chill! So what does this name the files then?

ttung:
It comes in as an input.

PR #451, opened 2021-05-18
Set up nextstrain runs for scc

ttung:
### Description
I plan to do this for all the counties we want to generate nightly trees, but want to try this first.

### Test plan
deploy and test?


shanaxel42 (approved):
exciting!

PR #453, opened 2021-05-18
[fix] TFE's type checker can't handle map values with different types.

jgadling:
### Description

Since TFE's type checker can't handle map values with different types (in our case, some values are strings, and others are maps) we can just disable its typechecking for this module

#### Issue

### Test plan


ttung:
Did you mean to check in the nextstrain wdl again?

ttung (approved):

PR #456, opened 2021-05-19
Add all counties

ttung:
### Description
This PR adds all counties and moves the run time to 3AM UTC (so 8PM PDT).  This is 3 hours after the gisaid run to ensure that the gisaid alignment is complete.

Depends on #455 

### Test plan
will test in staging!


shanaxel42 (approved):
ALLL THE TREES!!

ttung:
I would be surprised if there isn't a typo in here, but we can fix that after deploying.

PR #458, opened 2021-05-19
Remove error-prone caching step.

ttung:
### Description
It is error prone because if the curl request fails, the cache might have corrupt data.  Furthermore, while a `aws s3 cp` operation is running, the file is not in a consistent state.

### Test plan
test in sandbox


ttung:
I tested this:

`curl <gisaid_url> | bunzip2 | zstd | aws s3 cp - <blah>`

Killed curl, and even though aws s3 cp got the message that the pipe failed, it left a partial dump on s3.  This is really bad.

ttung:
Literally, the only reason for this to exist is during development, and there's wdl caching to handle that.  No need to complicate production workflows!

shanaxel42 (approved):

PR #459, opened 2021-05-19
[easy] Print disk availability and meminfo at start of job

ttung:
### Description
This might be helpful for debugging.

Removed a bit of cruft from past debugging runs.

### Test plan
YOLO

### DIFF
--- a/.happy/terraform/modules/sfn_config/nextstrain.wdl
+++ b/.happy/terraform/modules/sfn_config/nextstrain.wdl
@@ -43,6 +43,9 @@ task nextstrain_workflow {
     set -Eeuxo pipefail
     shopt -s inherit_errexit
 
+    df 1>&2
+    cat /proc/meminfo 1>&2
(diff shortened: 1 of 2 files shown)


.happy/terraform/modules/sfn_config/nextstrain.wdl
```diff
@@ -43,6 +43,9 @@ task nextstrain_workflow {
     set -Eeuxo pipefail
     shopt -s inherit_errexit
 
+    df
```
jgadling: gotta add `1>&2` so miniwdl doesn't eat stdout!

ttung: damn it you're right.

ttung: ok fixed.

jgadling (approved):

PR #460, opened 2021-05-19
Update covid import image tag

shanaxel42:
### Description

Forgot to update tis tag to the one with fixes in it! 

#### Issue

### Test plan


ttung:
The other images' tags are from the deploy, I believe.  Do we want to do that for this image as well?

PR #461, opened 2021-05-19
Increase memory requirements to spread out runs over more machines

ttung:
### Description
This should ideally solve the disk usage problem.

Also changed the schedule to launch at 12:15pm PDT.

### Test plan
TEST IN PROD.


shanaxel42 (approved):
MAX MEM

PR #463, opened 2021-05-19
Fix nextstrain build

ttung:
### Description
Nextstrain has been modified to expect the intermediate aligned sequenced to be [compressed with xz](https://github.com/nextstrain/ncov/pull/636).

### Test plan
Test in my sandbox


ttung:
I also updated the cron job so it'll run at 4:15pm PDT. #chasingtheclock

shanaxel42 (approved):

PR #464, opened 2021-05-20
Allow jobs to parameterize how many VCPUs are allocated

ttung:
### Description
This may make the jobs a taaaaaaaaaaad faster.

### Test plan
[Test in my sandbox](https://us-west-2.console.aws.amazon.com/states/home?region=us-west-2#/executions/details/arn:aws:states:us-west-2:473004499091:execution:aspen-rdev-ttung-nextstrain-swipe-spot-sfn:gisaid-more-vcpu)

### DIFF
--- a/.happy/terraform/modules/ecs-stack/county_runs.tf
+++ b/.happy/terraform/modules/ecs-stack/county_runs.tf
@@ -4,6 +4,7 @@ module nextstrain_scc_local_sfn_config {
   source   = "../sfn_config"
   app_name = "nextstrain-scc-local-sfn"
   image    = "${local.nextstrain_image_repo}:${local.image_tag}"
+  vcpus    = 4
   memory   = 64000
   wdl_path = "workflows/nextstrain.wdl"
   custom_stack_name     = local.custom_stack_name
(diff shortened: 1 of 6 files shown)


shanaxel42 (approved):
approved!

jgadling:
I'm a little curious at how this impacts provisioning. Based on the instance types available for R5d, it looks like AWS is allocating 8 vcpus per 64GB of memory, so will this just be a no-op?

ttung:
I believe that it allocates you the cores you request.  So as it is, you would get a 1 core 64 GB container.

jgadling:
Alright, I looked this up: https://docs.aws.amazon.com/batch/latest/userguide/job_definition_parameters.html

```
vcpus
    The number of vCPUs reserved for the container. This parameter maps to CpuShares in the Create a container
section of the Docker Remote API and the --cpu-shares option to docker run. Each vCPU is equivalent to 1,024 CPU shares
```

TL;DR: I don't think this will hurt anything, but I don't think it'll help either.

[It's a longer story](https://en.wikipedia.org/wiki/Completely_Fair_Scheduler), but this essentially sets a weight for each container running on the system. Since we've set the weight to the same value for all jobs, this means that if all containers on the system are requesting CPU time at the same time, they'll do an even split of CPU resources.

ttung:
Our tree builds are vcpu=4, the gisaid ingest is vcpu=32, and everything else is vcpu=1.

ttung:
But you're right, this is a different issue!  Turns out upstream ncov is playing with compression, which has huge perf implications.

PR #468, opened 2021-05-20
Data table sorting without breaking search

lvreynoso:
### Description

Turns out I overcomplicated state management in my previous PR and was not sorting on updated data fed into the `DataSubview`. Sorting logic is now clearer and simpler, and search still works!

Screenshot from 2021-05-20 10-04-58

#### Issue
[ch136816](https://app.clubhouse.io/genepi/story/136816)

### Test plan

Navigate to data tables, interact with them in _all_ the ways you interact with them normally.


tihuan (approved):
LGTM!! Thanks for the quick fix 🎉

PR #471, opened 2021-05-20
miniwdl 1.1.5

morsecodist:
### Description

Applying https://github.com/chanzuckerberg/swipe/pull/4/files which applies https://github.com/chanzuckerberg/idseq/pull/313 (not great I know but better than not getting the fixes).

This updates miniwdl and the treatment of the file system cache to fix an issue where a full cache can lead to a failure even if there is still space on the device.

#### Issue
None

### Test plan

This change has been in production in IDSeq for some time.

### DIFF
--- a/.happy/terraform/modules/batch/container_properties.yml
+++ b/.happy/terraform/modules/batch/container_properties.yml
@@ -33,14 +33,14 @@ command:
     put_metric MemoryInUse $(python3 -c 'import psutil; m=psutil.virtual_memory(); print(100*(1-m.available/m.total))');
     sleep 60;
     done &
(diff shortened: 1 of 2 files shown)


jgadling (approved):

morsecodist:
What does the failed patch test mean here? @jgadling

jgadling:
> What does the failed patch test mean here? @jgadling

Something to do with patching nexstrain. Almost certainly not failing due to this PR. @ttung

ttung:
> Something to do with patching nexstrain. Almost certainly not failing due to this PR. @ttung

yup, ignore it.  nextstrain accepted one of our patches upstream, so we no longer need to patch their code when we run it.  https://github.com/chanzuckerberg/aspen/pull/472

PR #472, opened 2021-05-20
Remove hack to order sequence data correctly

ttung:
### Description
Once https://github.com/nextstrain/ncov/pull/639 lands, we can remove this hack.

### Test plan
Will test once upstream has merged.


danrlu (approved):

danrlu:
Note to self: we'll need to keep aspen to be [the 2nd input](https://github.com/chanzuckerberg/aspen/blob/2e87d9b4ca2e50384171c07fa36c3efbdec20d72/src/backend/aspen/workflows/nextstrain_run/builds_templates/group.yaml#L1) so the aspen sequences get kept during de-duplication.

PR #473, opened 2021-05-21
Rename samples for admin users

ttung:
### Description
Rename all samples for admin users.

### Test plan
added a test to cover this.


shanaxel42 (approved):
Nice!

PR #475, opened 2021-05-21
thuang-next-js-138313

tihuan:
### Description

This PR seems like a huge change, but it can be boiled down to the following main points:

1. Replace Create React App related dependencies and scripts, and use Next.js to manage all that for us
2. Create `pages/` directory required by Next.js to host all pages, since they use file system based routing for both static and dynamic routes: https://nextjs.org/docs/routing/introduction
3. Update `docker-compose` to pass `HAPPY_COMMIT` to use as consistent build id for all Next.js servers in a cluster
4. Next.js has its own `next/router`, so we no longer need `react-router`
5. Add `pages/_document.tsx` and `pages/_app.tsx` for hosting global Next.js page configuration
6. Add `src/frontend/next.config.js` to configure Next.js to process `SVG` files as React components, add CSP/HSTS, specify prod build directory (default is `.next/`, but Infra expects `build/`), add `HAPPY_COMMIT` as consistent build id, and make `envVariable.js` available in the browser (see next point)
7. In Next.js, by default `process.env` variables are only available in Node.js and not in the browser, and they allow exposing env variables to browser through prefixing a variable with `NEXT_PUBLIC_`, which I find noisy and affecting across too many stacks and files (Docker files, entrypoint.js, and FE files). So I created `envVariable.js` as the single source of truth for all env variables used in the FE app and export them as process.env variables to the browser, so we don't need `NEXT_PUBLIC_` prefix!
8. Update `package.json` scripts to use Next.js' commands
9. Add [`react-query`](https://react-query.tanstack.com/) for handling API calls, which handles caching, garbage collection, refetch for us automatically
10. Use `import Head from 'next/head'` to add page titles

That's it! PTAL thank you 😄 

#### Issue
https://app.clubhouse.io/genepi/story/138313/migrate-fe-to-next-js

### Test plan

1. Visit: https://thuang-next-1-frontend.dev.genepi.czi.technology
2. The site should be fully functional
3. Some things I explicitly checked are:

1. Aspen logo works in Auth0 - inferred through https://thuang-next-1-frontend.dev.genepi.czi.technology/AspenLogoBlack.svg works
2. All routes work - /samples, /trees, /agreeTerms, /terms, /privacy/, /faq, / (homepage)
3. Styling looks fine with both scss and emotion.js
4. Pages now have page title (e.g., `Aspen | Samples`, `Aspen | Phylogenetic Trees`
5. Tests pass
6. HSTS and CSP work
7. Remove dependencies no longer needed
8. Use `HAPPY_COMMIT` for consistent build id in `next.config.js`. This is needed because FE servers run in a cluster, while each Next.js build by default has a random id, so if Server A sends the browser some static files fingerprinted with its unique build id, and then the browser makes a request for another static file, and such request gets routed to Server B, Server B won't  recognize Server A's build id and thus not being able to find the static asset the browser needs. [Source](https://nextjs.org/docs/api-reference/next.config.js/configuring-the-build-id)
9. Replace `<ProtectedRoute />` component with `useProtectedRoute` hook for AgreeTerms page and Sample page
10. Move font import to root html page with preconnect, so the browser is able to download the font faster!


src/frontend/babel.config.js
```diff
@@ -0,0 +1,19 @@
...
+  plugins: ["@emotion", transformImports()],
+  presets: ["next/babel"],
+};
+
+function transformImports() {
+  const materialUIPackages = ["core", "icons", "lab", "styles"];
+
+  const transform = {};
+
+  for (const package of materialUIPackages) {
```
tihuan, line 11: Material UI (MUI) has treeshaking for prod build, so we won't ship unneeded MUI code in prod mode. But we still need to transform imports to speed up dev mode build time! https://material-ui.com/guides/minimizing-bundle-size/#option-2

src/frontend/jest/config.js
```diff
@@ -0,0 +1,35 @@
+module.exports = {
```
tihuan, line 1: replace `jest` config in `package.json` to its own file and update the content to replace Create React App specific settings

src/frontend/pages/_document.tsx
```diff
@@ -0,0 +1,57 @@
...
+  NextScript,
+} from "next/document";
+import React from "react";
+import { theme } from "../src/common/styles/theme";
+
+export default class MyDocument extends Document {
+  static async getInitialProps(
+    context: DocumentContext
+  ): Promise<DocumentInitialProps> {
+    const sheets = new ServerStyleSheets();
```
tihuan, line 17: This is needed to set up MUI's server side rendering capability: https://material-ui.com/guides/server-rendering/#server-rendering

src/frontend/src/common/api/index.tsx
```diff
@@ -69,33 +75,12 @@ async function apiResponse<T extends APIResponse>(
     }
     return [key, resultData];
   });
+
   return Object.fromEntries(convertedData);
 }
 
 /** Calls to specific API endpoints **/
 
-interface UserResponse extends APIResponse {
```
tihuan, line 77: Extracted auth related API calls to its own file `src/frontend/src/common/queries/auth.ts`

src/frontend/src/common/queries/auth.ts
```diff
@@ -0,0 +1,67 @@
...
+  });
+};
+
+export function useUserInfo(): UseQueryResult<UserResponse, unknown> {
+  return useQuery([USE_USER_INFO], fetchUserInfo, {
+    retry: false,
+  });
+}
+
+export function useProtectedRoute(): void {
```
tihuan, line 58: For pages that are protected behind auth, use this hook in the component! Examples in `src/frontend/src/views/AgreeTerms/index.tsx` and `src/frontend/src/views/Data/index.tsx`

src/frontend/src/common/utils/TreeModal/ConfirmButton/index.tsx
```diff
@@ -74,7 +75,7 @@ export const createConfirmButton = (treeId: number): FC => {
 
 async function getTreeUrl(treeId: number) {
   const result = await fetch(
-    `${process.env.API_URL}/api/auspice/view/${treeId}`,
+    `${envVariables.API_URL}/api/auspice/view/${treeId}`,
```
tihuan: use `envVariables.js` to access `process.env` variables, so we have single source of truth and know what variables are available to use!

src/frontend/src/components/NavBar/components/UserMenu/index.tsx
```diff
@@ -49,16 +49,16 @@ const UserMenu = ({ user }: UserMenuProps): JSX.Element => {
         <a href={ROUTES.CONTACT_US_EMAIL} target="_blank" rel="noopener">
           <MenuItem onClick={handleClose}>Contact us</MenuItem>
         </a>
-        <Link to={ROUTES.TERMS} target="_blank" rel="noopener">
+        <a href={ROUTES.TERMS} target="_blank" rel="noopener">
           <MenuItem onClick={handleClose}>Terms of Use</MenuItem>
-        </Link>
-        <Link to={ROUTES.PRIVACY} target="_blank" rel="noopener">
+        </a>
+        <a href={ROUTES.PRIVACY} target="_blank" rel="noopener">
```
tihuan, line 55: since we're opening these links in new tabs, we don't need client routing Link component

src/frontend/src/components/NavBar/index.tsx
```diff
@@ -54,9 +56,11 @@ const NavBar = ({ org, user }: Props): JSX.Element => {
   return (
     <div className={style.bar}>
       <div className={style.left}>
-        <LogoAnchor to={route}>
-          <Logo alt="logo" src={String(LogoImage)} />
-        </LogoAnchor>
+        <Link href={route} passHref>
```
tihuan, line 59: Use `next/link` for client side routing

src/frontend/src/views/AgreeTerms/index.tsx
```diff
@@ -1,23 +1,29 @@
...
+  useProtectedRoute();
   const [isLoading, setIsLoading] = useState(false);
   const [shouldRedirect, setShouldRedirect] = useState(false);
   const [hasAcceptedTerms, setHasAcceptedTerms] = useState(false);
+  const router = useRouter();
 
   useEffect(() => {
     if (shouldRedirect) {
-      window.location.href = process.env.PUBLIC_URL + ROUTES.DATA;
+      router.push(ROUTES.DATA);
```
tihuan, line 24: use `next/router` to redirect

src/frontend/src/views/AgreeTerms/index.tsx
```diff
@@ -36,91 +42,96 @@ export default function AgreeTerms(): JSX.Element {
   }
 
   return (
-    <Container>
```
tihuan, line 39: This looks like a huge change, but the only thing I added is `<Head><title>...</title></Head>`

src/frontend/src/views/Data/index.tsx
```diff
@@ -87,58 +107,59 @@ const Data: FunctionComponent = () => {
...
+
     dataJSX.menuItems.push(
-      <Link to={category.to} key={category.text}>
-        <Menu.Item className={style.menuItem}>
-          <div className={style.category}>
-            <div className={cx(style.title, focusStyle)}>{category.text}</div>
-            <div className={style.count}>{category.data?.length}</div>
-          </div>
-        </Menu.Item>
+      <Link href={category.to} key={category.text} passHref>
```
tihuan, line 121: `next/link` needs `passHref` prop to pass its `href` attribute to its direct child anchor tag `<a />`, so we add `<a href="passHref"` as a convention that it will get the `href` from its Link parent

src/frontend/src/views/Data/index.tsx
```diff
@@ -87,58 +107,59 @@ const Data: FunctionComponent = () => {
...
-          />
-        )}
-      />
-    );
   });
 
+  const title = TITLE[router.asPath];
+
+  const category =
+    dataCategories.find((category) => category.to === router.asPath) ||
```
tihuan, line 137: use `router.asPath` to find the corresponding `category`

src/frontend/src/views/Homepage/index.test.tsx
```diff
@@ -0,0 +1,11 @@
+import { render, screen } from "@testing-library/react";
+import React from "react";
+import Homepage from "./index";
+
+test("renders the Homepage", () => {
+  render(<Homepage />);
```
tihuan, line 6: update test to test `Homepage` component directly

src/frontend/src/views/Homepage/index.tsx
```diff
@@ -22,49 +22,54 @@ import {
...
-                Aspen is a new tool that helps you manage, analyze, and share
-                your pathogen data for genomic epidemiology.
-              </Main>
-              <Details>
-                This is also where you’ll receive all of your sequencing results
-                from the COVIDTracker project at the CZ Biohub.
-              </Details>
-            </Left>
+    <>
+      <Head>
```
tihuan, line 26: Also only added the `Head` component here and moved `ConsensusGenomes` SVG into `Virus` styled component

src/frontend/src/common/constants/envVariables.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  API_URL: process.env.API_URL,
+  BUILD_PATH: process.env.BUILD_PATH,
+  COMMIT_SHA: process.env.COMMIT_SHA,
+  NODE_ENV: process.env.NODE_ENV,
+};
```
lvreynoso: I think this would be more attractive to use if the object was simply ENV, e.g. `ENV.API_URL`

lvreynoso (approved):
This is a big change, but I only could really think of one suggestion. I checked out the site and everything seems to work. I trust you for all the architecture changes

src/frontend/src/common/constants/envVariables.js
tihuan: Ooh this is a great idea! Thanks for the suggestion 🔥 Updated everything to use `ENV` 🚀 👏

tihuan:
Thanks so much for the quick review, @lvreynoso 🙏 !! This is the second time I do Next.js migration now, so all the learnings from Single Cell have been applied here 😆 I don't expect any surprises, but will be ready to resolve them if they happen 💪 Merging!!

PR #478, opened 2021-05-24
Redirect stdout to stderr

ttung:
### Description
wdl only provides stderr, so redirect our stdout to stderr.

### Test plan
TEST IN PROD YO.


ttung:
nm duped by #479

PR #487, opened 2021-05-25
write snakemake log to S3

danrlu:
### Description

Nextstrain snakemake log is in `ncov/.snakemake/log/` folder.
The edits will write the log to S3 where the tree json locates so we can see the content.

### Issue

Nextstrain is giving an error that we don't know what it is.

### Other notes

- I'm not confident in the `aws s3 cp` syntax. Please check!

- Also the `${key}` variable replacement doesn't look pretty. Will fix if we decide to keep the change.


shanaxel42 (approved):
yay!

PR #488, opened 2021-05-26
Add Github Action for scheduled prod deploy

lvreynoso:
### Description

Adds github action to pull HEAD^ and deploy to prod using `./scripts/happy`. Also uses `act10ns/slack@v1` to provide slack updates during the deployment process.

The cron schedule `'0 15 * * 1,4'` translates to "Run at 15:00 UTC on days 1 and 4 of the week." This is 8am PST Mondays and Thursdays.

#### Issue
[ch132910](https://app.clubhouse.io/genepi/story/132910)

### Test plan


.github/workflows/scheduled-prod-deploy.yml
```diff
@@ -0,0 +1,52 @@
...
+          fetch-depth: 2
+        run: git checkout HEAD^
+      - name: Install happy dependencies
+        run: |
+          pip install -r .happy/requirements.txt
+      - name: Update deployment
+        env:
+          TFE_TOKEN: ${{ secrets.TFE_TOKEN }}
+        run: |
+          ./scripts/happy --profile="" --env prod update prodstack --tag HEAD
```
jgadling: the --tag argument has to refer to a valid docker image tag. We push images to our ECR repos on every successful commit to trunk with the tags "branch-trunk" and "sha-${GITHUB_SHA:0:8}" ... the scripts/happy_deploy.py script basically looks for the latest successful build & push. We might want invoke that instead of the happy CLI directly here.

lvreynoso: The happy deploy script would launch another github action, and then this action would only know if the script successfully ran and not if the deployment was successful, right? Could I just pass in the tag `branch-trunk` or `sha-$(git rev-parse --short HEAD)`?

I tried passing HEAD on the cli and it worked so I thought it would work in the script.

jgadling: ahhh ... good point about the deploy script -- maybe we can update it to output a valid tag instead of invoking the action, if we pass in a flag?

lvreynoso: that's a good idea - i'll expose the functionality of `get_latest_successful_deployment()`.

lvreynoso:
Updated to get the short sha of the latest successful staging deployment, and use that as the `--tag` argument.

.github/workflows/scheduled-prod-deploy.yml
```diff
@@ -0,0 +1,53 @@
...
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-west-2
+          role-to-assume: ${{ secrets.AWS_PROD_ROLE_TO_ASSUME }}
+          role-duration-seconds: 900
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+      - run: git checkout HEAD^
```
jgadling: I think the actions/checkout step above clones the repo at the head of the default branch - do we need this step?

jgadling (approved):

.github/workflows/scheduled-prod-deploy.yml
lvreynoso: nope, I saw that too. got rid of it

PR #489, opened 2021-05-26
add uploaded_by field to sample table

phoenixAja:
### Description

adding an uploaded_by field to sample table

#### Issue
[ch139023](https://app.clubhouse.io/genepi/story/139023)

### Test plan

run migration, check that field was properly added 

### Note
~~i set this fields to be nullable for now, we need to update the local and remote env DBs to be populated with the aspin admin account before we set this field to be non nullable, figured we could do this in a subsequent PR~~
added the field as non nullable and backfilled with the ASPEN ADMIN account


src/backend/aspen/covidhub_import/import_projects.py
```diff
@@ -267,13 +268,19 @@ def import_project(
...
             sample = external_accessions_to_samples.get(external_accession, None)
+            uploaded_by_user = (
+                session.query(User).filter(User.name == "ASPEN ADMIN").one()
+            )
             if sample is None:
                 sample = Sample(
-                    submitting_group=group, private_identifier=external_accession
+                    submitting_group=group,
+                    private_identifier=external_accession,
+                    uploaded_by=uploaded_by_user,
```
shanaxel42, line 278: I think you only need the line below since it always runs!

src/backend/database_migrations/versions/20210526_162747_add_uploaded_by_to_samples.py
```diff
@@ -0,0 +1,69 @@
...
+        ["id"],
+        source_schema="aspen",
+        referent_schema="aspen",
+    )
+
+    conn = op.get_bind()
+
+    # create aspen admin user group
+    create_group_sql = sa.sql.text(
+        "INSERT INTO aspen.groups (name, address) VALUES ('ASPEN ADMIN GROUP', 'stub')"
```
shanaxel42, line 36: what does the 'stub' do?

shanaxel42 (approved):
Looks great!

src/backend/database_migrations/versions/20210526_162747_add_uploaded_by_to_samples.py
phoenixAja: oh i just didn't want to enter an actual address so i just said `'stub'`. Is that ok?

src/backend/aspen/covidhub_import/import_projects.py
phoenixAja: I'm confused by what you mean

PR #493, opened 2021-05-26
thuang-fe-dockerfile-update

tihuan:
### Description

Move `npm ci` command after `USER node`, so the files get stored in `/home/node/.cache/` instead of `/root/.cache/`, since we run `npm` commands with `USER node`, which looks for executables in `/home/node/.cache/`

#### Issue

### Test plan


jgadling (approved):

tihuan:
Thank you!!

PR #498, opened 2021-05-27
samples post 

phoenixAja:
### Description

sample post upload endpoint

#### Issue
[ch119809](https://app.clubhouse.io/genepi/story/119809)

### Test plan
wrote some new tests

### DIFF
--- a/src/backend/aspen/app/views/api_utils.py
+++ b/src/backend/aspen/app/views/api_utils.py
@@ -35,3 +35,28 @@ def format_datetime(dt: Optional[datetime.datetime], format="%Y-%m-%d %I:%M%p") 
         return dt.strftime(format)
     else:
         return "N/A"
+
+
+def check_data(
+        data: list[str],
+        required_fields:tuple,
+        optional_fields: tuple
+) -> tuple[bool, Optional[list[str]], Optional[list[str]]]:
+
+    # check all required fields are present:
+    missing_required = [i for i in required_fields if i not in data]
+
+    # check no fields were added that are unexpected
+    unexpected = [i for i in data if i not in required_fields+optional_fields]
+
+    if missing_required or unexpected:
+        return False, missing_required, unexpected
+
+    else:
+        # data is clean
+        return True
+
+
+
+
+
--- a/src/backend/aspen/app/views/api_utils.py
+++ b/src/backend/aspen/app/views/api_utils.py
@@ -59,3 +59,8 @@ def check_data(
     else:
         # data is clean
         return True, [], []
+
+
+# def create_public_id(user: User, sample_id: int) -> str:
+#     current_year = datetime.datetime.today().strftime("%Y")
+#     return f"USA/{user.group.prefix}-{sample_id}/{current_year}"
--- a/src/backend/aspen/app/views/api_utils.py
+++ b/src/backend/aspen/app/views/api_utils.py
@@ -1,5 +1,5 @@
 import datetime
-from typing import Collection, Mapping, Optional, Union
+from typing import Collection, List, Mapping, Optional, Tuple, Union
 
 from sqlalchemy.orm import joinedload
 from sqlalchemy.orm.query import Query
@@ -42,14 +42,16 @@ def check_data(
     pathogen_genome_fields: list[str],
     required_fields: tuple,
     optional_fields: tuple,
-) -> tuple[bool, Optional[list[str]], Optional[list[str]]]:
+) -> Tuple[bool, List[str], List[str]]:
 
     combined_fields: list[str] = sample_fields + pathogen_genome_fields
     # check all required fields are present:
-    missing_required = [i for i in required_fields if i not in combined_fields]
+    missing_required: list[str] = [
+        i for i in required_fields if i not in combined_fields
+    ]
 
     # check no fields were added that are unexpected
-    unexpected = [
+    unexpected: list[str] = [
         i for i in combined_fields if i not in required_fields + optional_fields
     ]
 
@@ -59,4 +61,3 @@ def check_data(
     else:
         # data is clean
         return True, [], []
-
--- a/src/backend/aspen/app/views/api_utils.py
+++ b/src/backend/aspen/app/views/api_utils.py
@@ -60,7 +60,3 @@ def check_data(
         # data is clean
         return True, [], []
 
-
-# def create_public_id(user: User, sample_id: int) -> str:
-#     current_year = datetime.datetime.today().strftime("%Y")
-#     return f"USA/{user.group.prefix}-{sample_id}/{current_year}"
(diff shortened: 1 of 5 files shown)


src/backend/aspen/app/views/api_utils.py
```diff
@@ -35,3 +35,29 @@ def format_datetime(dt: Optional[datetime.datetime], format="%Y-%m-%d %I:%M%p")
...
+
+    # check no fields were added that are unexpected
+    unexpected: list[str] = [
+        i for i in combined_fields if i not in required_fields + optional_fields
+    ]
+
+    if missing_required or unexpected:
+        return False, missing_required, unexpected
+
+    else:
```
shanaxel42: I don't think you need the else here!

src/backend/aspen/app/views/sample.py
```diff
@@ -230,3 +263,61 @@ def samples():
...
+            get_usergroup_query(db_session, profile["user_id"])
+            .options(joinedload(User.group))
+            .one()
+        )
+        request_data = request.get_json()
+
+        for data in request_data:
+            data_ok: bool
+            missing_fields: Optional[list[str]]
+            unexpected_fields: Optional[list[str]]
```
shanaxel42, line 278: I don't think you need to instantiate these values before the check_data call

src/backend/aspen/app/views/sample.py
```diff
@@ -230,3 +263,61 @@ def samples():
...
+            data_ok, missing_fields, unexpected_fields = api_utils.check_data(
+                list(data["sample"].keys()),
+                list(data["pathogen_genome"].keys()),
+                SAMPLES_POST_REQUIRED_FIELDS,
+                SAMPLES_POST_OPTIONAL_FIELDS,
+            )
+            if data_ok:
+                sample_args: Mapping[str, Any] = {
```
shanaxel42, line 286: not as important for the first release of this endpoint but would we check here if any of the request data had any of these optional fields in it?

src/backend/aspen/database/models/sample.py
```diff
@@ -46,6 +48,29 @@ class RegionType(enum.Enum):
 )
 
 
+def create_public_id(context) -> str:
+    current_parameters: Mapping[
```
shanaxel42, line 53: Could we just make this a helper method that just takes in a `group_id`?

src/backend/aspen/database/models/sample.py
```diff
@@ -46,6 +48,29 @@ class RegionType(enum.Enum):
...
+        str, Union[str, bool, datetime]
+    ] = context.get_current_parameters()
+    interface = SqlAlchemyInterface(context.engine)
+    with session_scope(interface) as session:
+        group: Group = (
+            session.query(Group)
+            .filter(Group.id == current_parameters["submitting_group_id"])
+            .one()
+        )
+        all_samples: Collection[Sample] = session.query(Sample).all()
```
shanaxel42: This is going to be pretty computationally expensive if we're calling this on every new sample we make. Could we do some pre calculating of how many ids we're going to need to generate outside of the for loop in the post submission and then generate them in bulk?

shanaxel42: ACTUALLY scratch that, looks like sqlalchmey has a max id functionality https://stackoverflow.com/questions/3750118/sqlalchemy-0-5-8-max-function 
let's use this instead of running this query.

shanaxel42 (changes requested):
Only required change is around the query in the create_public_id method to avoid an O(n^2) post endpoint

src/backend/aspen/app/views/api_utils.py
phoenixAja: so my thinking was that i kind of liked the true being returned because then we can say` if data_ok` in the view otherwise we would have to say if `data_ok ==None` in the view? I can take it out though if it's confusing!

src/backend/aspen/app/views/sample.py
phoenixAja: oh yeah we have that! so we so check in the `check_data` function if there were any optional fields in the request data.

src/backend/aspen/database/models/sample.py
phoenixAja: so i looked into that, the issue is that if you don't have the context passed in you can't create a session from the context engine and then we couldn't query to find the group prefix :(

src/backend/aspen/database/models/sample.py
phoenixAja: oh nice!!

src/backend/aspen/app/views/sample.py
phoenixAja: apparently this is how you add type hints for when you have a tuple return value, we can remove the type hints though, i personally find them a little too much sometimes lol

src/backend/aspen/database/models/sample.py
phoenixAja: added!

src/backend/aspen/app/views/sample.py
shanaxel42: oh interesting, let's leave them then! extra info : D

src/backend/aspen/app/views/api_utils.py
shanaxel42: oh yea keep the return! I just meant you didn't need the else. Could just be 
```
if missing_required or unexpected:
        return False, missing_required, unexpected
return True, [], []
```

src/backend/aspen/database/models/sample.py
```diff
@@ -46,6 +49,30 @@ class RegionType(enum.Enum):
...
+        str, Union[str, bool, datetime]
+    ] = context.get_current_parameters()
+    interface = SqlAlchemyInterface(context.engine)
+    with session_scope(interface) as session:
+        group: Group = (
+            session.query(Group)
+            .filter(Group.id == current_parameters["submitting_group_id"])
+            .one()
+        )
+        next_id: Union[int, None] = session.query(func.max(Sample.id)).scalar()
```
shanaxel42, line 63: nice!

src/backend/aspen/app/views/sample.py
```diff
@@ -21,16 +21,46 @@
...
+    "country",
+    "region",
+    "organism",
+    "host",
+    "purpose_of_sampling",
+    "specimen_processing",
+    "czb_failed_genome_recovery",
+    # following fields from PathogenGenome
+    "sequencing_depth",
+    "num_unambiguous_sites",
```
shanaxel42: not huge but not sure we'd ever accept these fields from the user since we calculate them from the sequence now

shanaxel42 (approved):
exciting!!

src/backend/aspen/app/views/sample.py
phoenixAja: right i'll take those out

PR #499, opened 2021-05-27
breaking: unify local dev URL's as [service].genepinet.local

jgadling:
#### TL;DR:
Instead of connecting to the frontend/backend local dev at http://localhost:[port], this PR moves them to http://frontend.genepinet.local:8000/ and http://backend.genepinet.local:3000 ... This is a breaking change, so please run `make local-init` when updating to a commit that includes these changes to update your local-dev environment.

#### More:
The DNS context inside the docker containers in our local dev env, and on our laptops is different. From our *laptops* we can use the `localhost` alias to connect to each container on a specific port since we've bound ports in the containers to our laptop's loopback interface.

However, these containers cannot connect to *each other* using the `localhost` name. They have to refer to each other by the service names they're given in `docker-compose.yml` or via network name aliases (like the ones added by this PR). If a web browser (playwright) running inside the frontend container gets an OAuth redirect to `localhost`, that breaks its login flow.

This PR unifies the names we use to connect to each container, whether we're connecting from a browser / curl on our laptops, or connecting from another container. It does this by assigning a DNS name to each container in docker-compose.yml for inter-container connections, and adding matching /etc/hosts entries on OSX for laptop -> container connections.

#### Issue

### Test plan


tihuan (approved):
Tested with Jessica offline and can confirm this works!!

PTAL https://github.com/chanzuckerberg/aspen/pull/500 that will merge into this PR 🙏 

Thanks again for all your help, Jessica 🤩 👍 ✨ !!

PR #505, opened 2021-06-02
thuang-e2e-part-1-123908

tihuan:
### Description

This PR is part 1 of 3 for complete E2E test setup!

Part 2 is getting E2E to run per PR
Part 3 is getting E2E to run after each staging and prod deployment (which also runs `describeIfDeployed` tests)

Changes:
1. Set up E2E tests using Jest + [Playwright](https://playwright.dev/), which is a rewrite of Puppeteer from the original Google team that's at Microsoft now 🎭 
1. Add tests for the following features:
    1. Assert Samples Page and Trees Page render some expected elements
    1. Login works (which is implicit, since many tests require login first)
    1. Search works (we will search for the first public id, and assert the table only contains one row)
    2. Sort works (we sort by public id header, and assert the order is different from the default sort order (upload date))
    1. Test Nextstrain link is created (only run this test in deployed environment. E.g., staging and prod)
    1. Test TSV file link is generated when Samples Page's table is rendered
1. Update `npm run test` to `npm run tests`, since autocorrect always bugs me about using plural 
2. Add `npm run e2e` command to run against local FE server: `http://frontend.genepinet.local:8000`
3. Add `npm run e2e-staging-from-local` to run E2E tests against staging: `https://staging.genepi.czi.technology`. **THIS IS HELPFUL FOR ONCALL ENGINEERS TO SMOKE TEST STAGING!** After part 3 is merged, we won't need to do this anymore
4. Update unit tests to accommodate [`react-query`](https://react-query.tanstack.com/), which we use for fetching and managing data from API endpoints. This involves setting up [Mock Server Worker](https://mswjs.io/) to return mocked data

#### Issue
https://app.clubhouse.io/genepi/story/123908/cross-browser-testing-functionality

### Test plan

1. Pull the branch
2. Run `make local-sync && make local-init && make frontend-e2e`

All tests should pass!

PTAL thanks all 😄 🎉 !


src/frontend/jest/playwright.setup.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * `frontend/jest-playwright.config.js` is for configuring Playwright's launch config options
+ * `jest/playwright.setup.ts` is for configuring `jest`, `browser`, and `page` objects
+ */
+
+// (thuang): This is the max time a test can take to run.
+// Since when debugging, we run and !headless, this means
+// a test can take more time to finish, so we don't want
+// jest to shut off the test too soon
```
valenzuelaomar, line 9: Can you please clarify this a bit? 🙏

valenzuelaomar:
Exciting stuff!! Thanks for tagging us IDseq folk :)

src/frontend/jest/playwright.setup.ts
tihuan: Absolutely! So when you run a test with headful mode (which means Playwright will open a browser window for you to see what it is clicking) often times you will add `await page.waitForTimeout(30 * 1000)` at a certain step, so you have 30 seconds to inspect the browser's state and see why a step is not doing what you expect to do, etc.. 

But your extra 30 seconds still counts towards the max timeout a test is allowed to run, which by default is 30 seconds. So if it takes the test 10 seconds to get to the state you want to inspect, then you'll only have 20 seconds left to do so, even though your code sets `await page.waitForTimeout(30 * 1000)`

Thus we just expand the default timeout to 2 mins for each test, so you likely won't run into that problem. And if so, you know where to increase the timeout limit to a higher number!

phoenixAja (approved):
this LG to me!! thanks for adding! 🎉

tihuan:
Fantastic!! Thanks for taking a look, everyone!! Please ask more questions if you have 🙏 I'll merge first!

PR #506, opened 2021-06-02
Add lint check for WDL files

jgadling:
### Description

This adds a simple lint check for WDL files. There are a few warnings we should probably fix, but warnings are allowed for now. We can look into making them more strict when we're ready.

### Test plan

There's an example of a successful / failed lint check below in the commits listed as part of this PR.


phoenixAja (approved):
awesome thanks!!

PR #508, opened 2021-06-02
Phoenix/group prefixes

phoenixAja:
### Description

adding prefix to group table for easier public id generation

#### Issue
[ch141263](https://app.clubhouse.io/genepi/story/141263)

### Test plan

run migration, run tests


src/backend/aspen/database/models/usergroup.py
```diff
@@ -20,6 +20,11 @@ class Group(idbase, DictMixin):  # type: ignore
 
     name = Column(String, unique=True, nullable=False)
     address = Column(String, nullable=True)
+    prefix = Column(
+        String,
+        nullable=True,
```
shanaxel42: Do we want this to be nullable?

shanaxel42: Probably also want to enforce it as unique

shanaxel42 (changes requested):
Just one change about enforcing uniqueness!

src/backend/aspen/database/models/usergroup.py
phoenixAja: yeah uniqueness sounds good! I left it as nullable in case we had a group that was just for viewing (like how we have the admin group, i imagine we could also have some view only groups?), do you think that assessment is right?

phoenixAja: i changed it to be not null

src/backend/database_migrations/versions/20210602_224122_adding_prefix_to_group.py
```diff
@@ -0,0 +1,80 @@
...
+        "Orange County Health Care Agency": "CA-OC-",
+        "San Bernardino County Public Health": "CA-SBC-",
+        "San Joaquin County Public Health Services": "CA-SJC-",
+        "San Luis Obispo County Health Agency, Public Health Laboratories": "CA-SLOC-",
+        "Santa Clara County Public Health": "CA-SC-",
+        "San Francisco Public Health Laboratory": "CA-SFC-",
+        "Tulare County Public Health Lab": "CA-TCPHL-",
+        "Tuolumne County Public Health": "CA-TLMNC-",
+        "Ventura County Public Health Laboratory": "CA-VC-",
+        "CDPH": "CA-CDPH",
```
shanaxel42, line 47: These ones don't have the trailing `-`?

sidneymbell:
No, they don't. CDPH is weird that way.

shanaxel42 (approved):

PR #510, opened 2021-06-08
fix county location filter

danrlu:
### Description

1. Tulare submitted their samples with location being "California/Southern San Joaquin Valley". However when `ncov-ingest` parses out the metadata from gisaid, they would convert that location to blank. see https://github.com/nextstrain/ncov-ingest/issues/179. Modified this in `32fb423`

2. `ncov-ingest` will also append "CA" to Orange County and Monterey County locations. So modified those so Gisaid data get sorted properly to county trees in `9032a8b`
image
Also see [how we did this](https://github.com/czbiohub/covidhub/blob/61001a4ea917ccd30e13110d59544827e128be4d/code/genome_tracking/scripts/prepare_ncov_input.py#L192) in covidhub.

### Issues

Note currently `location` is used to filter both Aspen and Gisaid data. So this change will remove Aspen data from these county trees, which should be fixed after this change: https://app.clubhouse.io/genepi/story/141787/split-subsampling-for-aspen-and-gisaid-data

### Test plan

I tested the `sed` line in `32fb423` on bash in EC2. We'll see whether the samples will show up in Thursday tree after the changes are deployed.

### DIFF
--- a/.happy/terraform/modules/ecs-stack/county_runs.tf
+++ b/.happy/terraform/modules/ecs-stack/county_runs.tf
@@ -372,7 +372,7 @@ module nextstrain_monterey_local_sfn_config {
     template_filename        = "group.yaml"
     template_args            = {
      division = "California"
-     location = "Monterey County"
+     location = "Monterey County CA"
     }
   }
 }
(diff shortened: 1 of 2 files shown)


danrlu:
I noticed that `location` is also used as the title of the tree, so this won't work. I'll try a different approach.

PR #517, opened 2021-06-11
Use ambiguity score, not `conflict` to assign pango confidence

sidneymbell:
### Description

Should be using the `ambiguity_score` to estimate pangolin confidence, rather than `conflicts`. Hopefully this also solves the  weird issue where some pangolin lineages weren't able to be parsed and assigned because of empty string values in `conflicts`. 

#### Issue

### Test plan

Does someone mind kicking off the pangolin job for me and checking that:
1.  `USA/CA-CZB-32366/2021` gets correctly assigned as `B.1.617.something`
2. The `probability` values in the app now range from 0 -> 100, and we see a range of these values (although heavily skewed towards 100)

### DIFF
--- a/src/backend/aspen/workflows/pangolin/save.py
+++ b/src/backend/aspen/workflows/pangolin/save.py
@@ -15,9 +15,15 @@ from aspen.database.connection import (
 from aspen.database.models import PathogenGenome, UploadedPathogenGenome
 
 
-def get_probability(conflict: float) -> float:
-    assert conflict <= 1
-    return 1.0 - conflict
+def get_probability(ambiguity_score: float) -> int:
+    """
+    Estimate confidence percentage based on the `ambiguity_score`.
+    Per pangolin docs, this is basically the number of lineage-defining sites
+    that had to be imputed from the reference sequence.
+    Round and multiply by 100 --> percentage for easier user comprehension.
+    """
+    assert 0 <= ambiguity_score <= 1
+    return round((1 - ambiguity_score) * 100)
 
 
 @click.command("save")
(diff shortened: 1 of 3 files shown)


sidneymbell:
NOTE: this test currently fails because the "correct" probability value needs to be updated, but I'm not sure how the demo CSV is being generated (if we're not already, should we update these tests to run on dummy genomes --> create the test CSV with each run? would catch similar bugs in the future)
```
        assert pathogen_genome.pangolin_probability == 1
```

src/backend/aspen/workflows/pangolin/save.py
```diff
@@ -33,7 +39,7 @@ def cli(pangolin_fh: io.TextIOBase, pangolin_last_updated: datetime):
         taxon_to_pango_info: Mapping[int, Mapping[str, Union[str, float]]] = {
             int(row["taxon"]): {
                 "lineage": row["lineage"],
-                "probability": get_probability(float(row["conflict"])),
+                "probability": get_probability(float(row["ambiguity_score"])),
```
shanaxel42: will this always be set?

shanaxel42: looks like in the docs it also might be blank https://cov-lineages.org/pangolin_docs/output.html

shanaxel42: in which case we might want to just add 
```
"probability": get_probability(float(row["ambiguity_score"])) if row["conflict"] != "" else 1,
```
so it doesn't error on the float cast

PR #524, opened 2021-06-15
Update confidence scores for 'designation hash' sequences

sidneymbell:
per https://github.com/cov-lineages/pangolin/issues/242#issuecomment-861282358

#### Issue
[ch](https://app.clubhouse.io/genepi/story/<fill_in_issue_number>)

### Test plan

### DIFF
--- a/src/backend/aspen/workflows/pangolin/save.py
+++ b/src/backend/aspen/workflows/pangolin/save.py
@@ -23,7 +23,7 @@ def get_probability(row) -> int:
     Round and multiply by 100 --> percentage for easier user comprehension.
     """
     if row["ambiguity_score"] != "":
-        return round(ambiguity_score * 100)
+        return round(row["ambiguity_score"] * 100)
     elif "Assigned using designation hash" in row["note"]:
         return 100
     else:
--- a/src/backend/aspen/workflows/pangolin/save.py
+++ b/src/backend/aspen/workflows/pangolin/save.py
@@ -23,7 +23,7 @@ def get_probability(row: dict) -> Union[int, None]:
     Round and multiply by 100 --> percentage for easier user comprehension.
     """
     if row["ambiguity_score"]:
-        return round(row["ambiguity_score"] * 100)
+        return round(float(row["ambiguity_score"]) * 100)
     elif "Assigned using designation hash" in row["note"]:
         return 100
     else:
--- a/src/backend/aspen/workflows/pangolin/save.py
+++ b/src/backend/aspen/workflows/pangolin/save.py
@@ -15,14 +15,19 @@ from aspen.database.connection import (
 from aspen.database.models import PathogenGenome, UploadedPathogenGenome
 
 
-def get_probability(ambiguity_score: float) -> int:
+def get_probability(row) -> int:
     """
     Estimate confidence percentage based on the `ambiguity_score`.
     Per pangolin docs, this is basically the number of lineage-defining sites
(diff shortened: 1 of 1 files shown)


src/backend/aspen/workflows/pangolin/save.py
```diff
@@ -15,14 +15,19 @@
...
-def get_probability(ambiguity_score: float) -> int:
+def get_probability(row: dict) -> Union[int, None]:
     """
     Estimate confidence percentage based on the `ambiguity_score`.
     Per pangolin docs, this is basically the number of lineage-defining sites
     that had to be imputed from the reference sequence.
     Round and multiply by 100 --> percentage for easier user comprehension.
     """
-    return round(ambiguity_score * 100)
+    if row["ambiguity_score"] != "":
```
shanaxel42: `
```suggestion
    if row["ambiguity_score"]:
```
This will also work for an empty string check!

src/backend/aspen/workflows/pangolin/save.py
```diff
@@ -15,14 +15,19 @@
...
+        return round(row["ambiguity_score"] * 100)
+    elif "Assigned using designation hash" in row["note"]:
+        return 100
+    else:
+        return None
```
shanaxel42, line 30: could just be:
```
elif "Assigned using designation hash" in row["note"]:
        return 100
 return None
```

src/backend/aspen/workflows/pangolin/save.py
```diff
@@ -15,14 +15,19 @@
...
+def get_probability(row: dict) -> Union[int, None]:
     """
     Estimate confidence percentage based on the `ambiguity_score`.
     Per pangolin docs, this is basically the number of lineage-defining sites
     that had to be imputed from the reference sequence.
     Round and multiply by 100 --> percentage for easier user comprehension.
     """
-    return round(ambiguity_score * 100)
+    if row["ambiguity_score"] != "":
+        return round(row["ambiguity_score"] * 100)
```
shanaxel42: I think you still need the float cast here? `round(float(row["ambiguity_score"]) * 100)`

src/backend/aspen/workflows/pangolin/save.py
sidneymbell: true! this is just a personal preference thing / I tend to make these things explicit. But if it's better practice, happy to make the change :)

src/backend/aspen/workflows/pangolin/save.py
sidneymbell: Maybe? Only ever encountered values that are floats or strings here, but agree it's safer to explicitly call it out. Will fix!

src/backend/aspen/workflows/pangolin/save.py
```diff
@@ -15,14 +15,19 @@
...
+def get_probability(row: dict) -> Union[int, None]:
     """
     Estimate confidence percentage based on the `ambiguity_score`.
     Per pangolin docs, this is basically the number of lineage-defining sites
     that had to be imputed from the reference sequence.
     Round and multiply by 100 --> percentage for easier user comprehension.
     """
-    return round(ambiguity_score * 100)
+    if row["ambiguity_score"]:
+        return round(row["ambiguity_score"] * 100)
```
sidneymbell:
```suggestion
        return round(float(row["ambiguity_score"]) * 100)
```

shanaxel42 (approved):
looks good!

PR #525, opened 2021-06-15
sample post fixes

phoenixAja:
### Description

format sequencing_date if empty string, also check if any public or private_identifiers are already in the DB before inserting data

#### Issue
no issue

### Test plan

wrote tests


shanaxel42 (approved):
awesome!

PR #532, opened 2021-06-15
thuang-metadata-private-id-match

tihuan:
### Description

Screen Shot 2021-06-15 at 3 42 15 PM

#### Issue

### Test plan


Janeece:
@tihuan I think the message is clear and that the format work now, but also I think we'll want to re-address the sample list design in the future to account for the edge case of 100s of samples.

tihuan:
Oooh really great point, @Janeece ! That'd be a realllly long list ahahaha sounds great!

PR #538, opened 2021-06-16
refactors needed for sample api

shanaxel42:
### Description

Write a description of your pull request here

#### Issue

### Test plan


phoenixAja:
ugh creating the public ids from the id is weirdly challenging.. i guess we have to go the way of batch generation for the ids, i saw that was how covidhub did the czb_id generation, but i then i thought it would be nicer to add the default option instead on the model but looks like it was having issues. Weird that the test was working though, maybe it was a threading issue when saving the model through the frontend.

PR #546, opened 2021-06-17
adding catch for all cases of yes/no

phoenixAja:
### Description

adding check for boolean fields for all cases of yes/no

#### Issue
[ch143963](https://app.clubhouse.io/genepi/story/143963)

### Test plan

tested locally, data was being parsed correctly
rdev here: https://upload-boolean-fix-frontend.dev.genepi.czi.technology/

### DIFF
--- a/src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
+++ b/src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
@@ -90,7 +90,7 @@ function buildMetadata({ headers, row, warningMessages }: RowInfo) {
     const key = HEADERS_TO_METADATA_KEYS[headers[i]];
 
     if (key) {
-      metadata[key] = convertValue(value);
+      metadata[key] = convertValue(key, value);
     }
   }
 
@@ -121,7 +121,11 @@ function buildMetadata({ headers, row, warningMessages }: RowInfo) {
   return metadata;
 }
 
-function convertValue(value: string) {
+function convertValue(key: string, value: string) {
+  if (key === "keepPrivate") {
+    console.log("KEY check cleared: ", key);
+    if (value === "") return true;
+  }
   if (value.toUpperCase() === "YES") return true;
   if (value.toUpperCase() === "NO") return false;
 
--- a/src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
+++ b/src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
@@ -121,15 +121,10 @@ function buildMetadata({ headers, row, warningMessages }: RowInfo) {
 }
 
 function convertValue(key: string, value: string) {
-  if (key === "keepPrivate") {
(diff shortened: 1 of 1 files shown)


src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
```diff
@@ -117,13 +117,13 @@ function buildMetadata({ headers, row, warningMessages }: RowInfo) {
...
   }
-
   return metadata;
 }
 
-function convertValue(value: string) {
-  if (value === "Yes") return true;
-  if (value === "No") return false;
+function convertValue(key: string, value: string) {
+  if (key === "keepPrivate" && value === "") return true;
```
shanaxel42: I think the default should be false

src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
```diff
@@ -117,13 +117,13 @@ function buildMetadata({ headers, row, warningMessages }: RowInfo) {
...
+  if (value.toUpperCase() === "YES") return true;
+  if (value.toUpperCase() === "NO") return false;
 
   return value;
```
shanaxel42, line 128: And then as the final return can we do `return False` for the default again. Since right now what happens is if a user input garbage like "asdasda" it will pass all these cases and then that get's passed to the backend

shanaxel42 (changes requested):

src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
phoenixAja: changed

src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
```diff
@@ -117,14 +117,14 @@ function buildMetadata({ headers, row, warningMessages }: RowInfo) {
...
-function convertValue(value: string) {
-  if (value === "Yes") return true;
-  if (value === "No") return false;
-
+function convertValue(key: string, value: string) {
+  if (key === "keepPrivate" || key === "submittedToGisaid") {
+    if (value.toUpperCase() === "YES") return true;
+    return false;
+  }
   return value;
```
shanaxel42, line 128:
```suggestion
  return False;
```

shanaxel42 (changes requested):

src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
tihuan: Ah yes we take that into account at line 126 now 👌 Thank you!

phoenixAja: this one we want to return value since this is parsing the other fields too, otherwise we would return false for public_id, private_id, etc

shanaxel42 (approved):

PR #549, opened 2021-06-17
thuang-autocorrect-private

tihuan:
### Description

If we detect Public ID or Accession Number exists, we will check if we need to autocorrect the boolean fields!

Screen Shot 2021-06-17 at 12 55 26 PM

#### Issue
https://app.clubhouse.io/genepi/story/143965/you-shouldn-t-be-able-to-mark-a-sample-as-private-and-give-it-a-public-id-through-the-metadata-upload

### Test plan


phoenixAja (approved):
cool looks good!

tihuan:
Thanks for the quick review, @phoenixAja ! Will merge after tests pass 🎉 !!

PR #553, opened 2021-06-23
[EZ-PZ] add automatic margin resizing

phoenixAja:
### Description

fixed margin issues on data views. 

#### Issue
[ch137096](https://app.clubhouse.io/genepi/story/137096)

### Test plan

tested on rdev here: https://data-margin-fix-frontend.dev.genepi.czi.technology/data/samples

### DIFF
--- a/src/frontend/src/views/Data/index.module.scss
+++ b/src/frontend/src/views/Data/index.module.scss
@@ -55,6 +55,6 @@
     // This is needed to ensure the div doesn't disappear
     // when the viewport height is too short
     height: calc(100% - 74px);
-    margin: 0 10em;
+    margin: 0 auto; width: calc(100% - 10em);
   }
 }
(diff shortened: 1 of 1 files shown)


Janeece:
Hi @phoenixAja! This is **such** an improvement. My feedback is nit-picky: 

1. Can we add a content max width? `max-width:1308px;` would align us with IDseq's, but we could round that to 1300px. 
Page Margins 1 - max width

2. Can we adjust the min-width on the margin to equal $space-l (14px)? That way we'd get more room when content is under its max-width. 
Page Margins 2 - margin min-width cropped

3. If possible, can we have the margin in the tab bar match the content margins, so they are aligned with each other?
Page Margins 3 - header and content alignment

shanaxel42 (approved):
woo!

PR #555, opened 2021-06-24
Adding backend endpoint for fasta file downloads

shanaxel42:
### Description

Backend endpoint for fasta file downloads. 

I didn't end up refactoring the cli code into a common api since this endpoint required more permission verification 
while writing the file 

#### Issue

### Test plan

Wrote some tests

### DIFF
--- a/src/backend/aspen/app/views/sample.py
+++ b/src/backend/aspen/app/views/sample.py
@@ -1,8 +1,18 @@
 import datetime
+import tempfile
 from collections import defaultdict
-from typing import Any, Mapping, MutableSequence, Optional, Sequence, Set, Union
+from typing import (
+    Any,
+    Iterable,
+    Mapping,
+    MutableSequence,
+    Optional,
+    Sequence,
+    Set,
+    Union,
+)
 
-from flask import jsonify, request, Response, session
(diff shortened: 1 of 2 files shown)


src/backend/aspen/app/views/sample.py
```diff
@@ -144,6 +154,95 @@ def _format_lineage(sample: Sample) -> dict[str, Any]:
...
+                        f.write(f">{sample.public_identifier}\n")
+                    f.write(stripped_sequence)
+                    f.write("\n")
+
+        fasta_filename = f"{user.group.name}_sample_sequences.fasta"
+        res = send_file(
+            fasta_file, attachment_filename=fasta_filename, as_attachment=True
+        )
+        # remove the file
+        f.close()
```
phoenixAja, line 242: interesting, we don't need to close the file before sending it?

phoenixAja (approved):
this looks really solid 👍 thanks for adding this!

src/backend/aspen/app/views/sample.py
shanaxel42: oh yea the send_file method needs to read the file so it failed if we had already closed the file

PR #557, opened 2021-06-24
simplfy the samples query

shanaxel42:
### Description

Since we don't currently (or will anytime in the near future have any non UploadedPathogenGenome samples) we can simplify this query for now

#### Issue

### Test plan


phoenixAja (approved):
looks good

PR #560, opened 2021-06-28
fix public ID not converting to private ID on tree

danrlu:
### Description

Trees' public ID gets converted to private ID before uploaded to Aspen. The public ID Humboldt uploaded is the full Gisiad ID that looks like "hCoV-19/USA/CA-HC-P2117185/2021", but the "hCoV-19/" part got stripped by Nextstrain in the tree building. So the step to convert public ID to private ID didn't work b/c the ID in tree json doesn't match the public ID in our database. 

I don't know whether it makes more sense to strip "hCoV-19/" in our database instead of the fix proposed here.

#### Issue

### Test plan

The added piece worked in Python console.

### DIFF
--- a/src/backend/aspen/app/views/phylo_trees.py
+++ b/src/backend/aspen/app/views/phylo_trees.py
@@ -172,7 +172,7 @@ def _process_phylo_tree(
             return sample.submitting_group_id in can_see_group_ids_pi
 
     identifier_map: Mapping[str, str] = {
-        sample.public_identifier: sample.private_identifier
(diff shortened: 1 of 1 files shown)


shanaxel42:
This seems like a good fix to me, let me know if you want me to take over this branch to fix the linting stuff!

danrlu:
Thanks for fixing the py-test! I'll leave the rest to you then - since I ran out of ideas what else to try. Thanks so much Shannon!

shanaxel42 (approved):

PR #562, opened 2021-06-29
thuang-fix-fasta-parser

tihuan:
### Description

Turned out that some fasta files have empty last line and some don't, so I need to take that into account!
Also refactored the code a little for readability

PTAL thank you!

#### Issue
https://app.clubhouse.io/genepi/story/146144/fasta-file-freezes-and-hangs-upload-step

### Test plan

### DIFF
--- a/src/frontend/src/views/Upload/components/Samples/utils.ts
+++ b/src/frontend/src/views/Upload/components/Samples/utils.ts
@@ -129,25 +129,49 @@ function handleFastaText(text: string, filename: string): ParseOutcome {
       continue;
     }
 
-    // Aggregate sequence line by line
-    for (let j = i + 1; j < lines.length; j++) {
-      const jLine = lines[j];
+    const { newIndex, sequence } = aggregateSequence(i, lines);
 
-      if (!jLine || jLine.includes(">")) {
-        i = j;
-        break;
-      }
+    i = newIndex;
 
+    if (sequence) {
       const id = iLine.replace(">", "");
 
       result = {
         ...result,
-        [id]: (result[id] || "") + jLine,
+        [id]: sequence,
       };
     }
   }
 
   return { errors, result };
+}
+
+function aggregateSequence(
+  i: number,
+  lines: string[]
+): { newIndex: number; sequence: string } {
+  let newIndex = i;
+  let sequence = "";
+
+  // (thuang): `j` starts at `i + 1` to be on the first line of sequence,
+  // since `i` should be the `id` line
+  for (let j = i + 1; j < lines.length; j++) {
+    const jLine = lines[j];
+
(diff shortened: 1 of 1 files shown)


src/frontend/src/views/Upload/components/Samples/utils.ts
```diff
@@ -129,27 +129,49 @@ function handleFastaText(text: string, filename: string): ParseOutcome {
...
+    const jLine = lines[j];
+
+    if (!jLine || jLine.includes(">")) {
+      newIndex = j;
+      break;
+    }
+
+    sequence += jLine;
+
+    // (thuang): j has reached the end of file
```
tihuan, line 168: this is the only additional check in the algo. Once `j` reaches the end of file, we need to move `i` to the same index as well, otherwise the loop won't end because it never detects an empty line as available in some fasta files (we check the empty line at line `159`)

src/frontend/src/views/Upload/components/Samples/utils.ts
```diff
@@ -129,27 +129,49 @@ function handleFastaText(text: string, filename: string): ParseOutcome {
...
       };
     }
   }
 
   return { errors, result };
 }
 
+function aggregateSequence(
+  i: number,
+  lines: string[]
```
phoenixAja, line 151: why `string[]` and not `string` here?

phoenixAja: oh because it's an array of strings?

src/frontend/src/views/Upload/components/Samples/utils.ts
```diff
@@ -129,27 +129,49 @@ function handleFastaText(text: string, filename: string): ParseOutcome {
...
+): { newIndex: number; sequence: string } {
+  let newIndex = i;
+  let sequence = "";
+
+  for (let j = i + 1; j < lines.length; j++) {
```
phoenixAja, line 158: can you add a quick comment on why we need` j=i+1` here? Is that to move past the sample id?

tihuan: That's right! Just added a comment to capture this 😄

src/frontend/src/views/Upload/components/Samples/utils.ts
tihuan: Yup! We break fasta file into a collection of strings via line break `const lines = text.split(/\r?\n/)`, so each string in the array represents a line in the fasta file. And normally line break is just `\n` in macOS, but Windows could use `\r\n`, thus taking that into account in the `.split()` method!

phoenixAja (approved):
yay thanks timmy!

PR #564, opened 2021-06-29
fix svg

tihuan:
### Description

turned out that the gap was coming from svg directly! So I had to trim the SVG using https://editor.method.ac/ and set size `fit to content`, and then I downloaded the svg from that editor and ran through https://vecta.io/nano to minify the SVG further

also I renamed the SVG files to be without Primary part, since we’re setting colors directly in styled component!

#### Issue

### Test plan


src/frontend/src/views/Upload/components/Review/components/Upload/style.ts
```diff
@@ -43,3 +44,22 @@ export const StyledDialogActions = styled(DialogActions)`
...
+    `;
+  }}
+`;
+
+export const StyledUploadFailedImage = styled(UploadFailedImage)`
+  ${(props) => {
+    const colors = getColors(props);
+
+    return `
+      fill: ${colors?.error[400]};
```
phoenixAja, line 62: oh cool you got the colors working without modifying the svg directly! How did you fix that? Or was i just doing something wrong? lol

phoenixAja (approved):
looks great!! ty!!

src/frontend/src/views/Upload/components/Review/components/Upload/style.ts
tihuan: Ah yes, I removed `fill` from SVG file, so we can overwrite it through our styled components 🎉 !

PR #569, opened 2021-06-30
[DO NOT MERGE] smart retries

morsecodist:
### Description

**DO NOT MERGE**: this PR https://github.com/chanzuckerberg/genepi-infra/pull/54 must be deployed at the same time. It contains the lambda elements of this change.

This ports smart retries to aspen: https://github.com/chanzuckerberg/swipe/pull/8.

Right now if the job fails on spot for any reason (except one of our hard-coded idseq-specific error types) it will retry on-demand. This is not ideal because retrying on-demand is only helpful if the failure happened because of spot termination. This change only retries on demand if the job was terminated because the instance shut down (which should only be due to a spot instance termination). This prevents costly, wasteful retries.

I am trying to merge this now because I am about to do the error notifications and I want to bring those into swipe then aspen as well and I didn't want to apply stuff out of order and have things drift out of sync.

#### Issue
None

### Test plan

This change has worked pretty well on the idseq side. There was one bug but it has since been fixed so I am pretty confident. Any test that tests the workflows will also be testing this.

### DIFF
--- a/.happy/terraform/modules/swipe-sfn/sfn.yml
+++ b/.happy/terraform/modules/swipe-sfn/sfn.yml
@@ -43,33 +43,29 @@ States:
           - Name: SFN_CURRENT_STATE
             Value.$: $$.State.Name
     ResultPath: $.BatchJobDetails.Run
-    Next: RunSPOTReadOutput
+    Next: RunReadOutput
     Retry: &BatchRetryConfig
       - ErrorEquals: ["Batch.AWSBatchException"]
         IntervalSeconds: 15
         MaxAttempts: 3
+        BackoffRate: 2.0
     Catch:
       - ErrorEquals: ["States.ALL"]
         ResultPath: $.BatchJobError.RunSPOT
-        Next: RunSPOTReadOutput
-  RunSPOTReadOutput:
-    Type: Task
-    Resource: arn:aws:states:::lambda:invoke
-    Parameters: &ReadOutputParameters
-      FunctionName: "${process-stage-output}"
-      Payload: *PassthroughStatePayload
-    OutputPath: $.Payload
-    Next: HandleSuccess
-    Catch:
-      - ErrorEquals: &InputErrors
-          - InvalidInputFileError
-          - InsufficientReadsError
-          - BrokenReadPairError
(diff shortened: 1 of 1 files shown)


phoenixAja:
@morsecodist (or others that know the answer) what causes spot termination generally? I'm guessing this type of termination doesn't happen due to pipeline errors? Along those same lines why does retrying on-demand only make sense when we have spot termination?

morsecodist:
@phoenixAja good question. The idea behind spot is you get cheaper compute in exchange for letting AWS terminate your instance to serve someone else's on demand needs. For background compute like this it can save you a lot of money without much impact. It is on you as the developer to handle that termination. Some people retry again on spot after a wait but we try again on-demand. This guarantees the workflow will eventually finish in 2n where n is the time of a single workflow. Spot terminations are pretty rare so even when you retry on demand like this you still come out pretty far ahead.

.happy/terraform/modules/swipe-sfn/sfn.yml
```diff
@@ -43,33 +43,29 @@ States:
           - Name: SFN_CURRENT_STATE
             Value.$: $$.State.Name
     ResultPath: $.BatchJobDetails.Run
-    Next: RunSPOTReadOutput
+    Next: RunReadOutput
     Retry: &BatchRetryConfig
       - ErrorEquals: ["Batch.AWSBatchException"]
         IntervalSeconds: 15
         MaxAttempts: 3
+        BackoffRate: 2.0
```
phoenixAja, line 51: what is `BackoffRate` used for?

phoenixAja (approved):

phoenixAja:
lgtm, thank you for explaining!

.happy/terraform/modules/swipe-sfn/sfn.yml
morsecodist: This means to double the wait interval between each backup. This is helpful if failures happen due to some sort of rate limiting since it will dynamically slow the total rate to fit the limit. This value is actually the default I just realized but it was set on our end and I am syncing them so I ported this too even though I guess it is redundant. At least it makes it more explicit what the rate is to the reader.

PR #575, opened 2021-07-09
Basic Sentry integration

lvreynoso:
### Description

This adds Sentry reporting to the Aspen python backend. Sentry will report errors / exceptions.

Follow-up PRs should add code to distinguish reporting environments.

#### Issue
[ch132909](https://app.clubhouse.io/genepi/story/132909)

### Test plan

Tested remotely in an rdev happy stack.

Screen Shot 2021-07-08 at 17 44 41


lvreynoso:
Not sure why the CI server is having trouble, will investigate in the morn

Screen Shot 2021-07-08 at 17 52 14

phoenixAja:
is pytest/ rebuilding the containers working locally for you? (`make backend-test`, `make local-rebuild`) I wonder if adding the new dependency downgraded or upgraded six to a version that has issues.

src/backend/aspen/app/app.py
```diff
@@ -20,6 +21,17 @@
...
+
+# We should be able to allow this in all environments and only alert on prod.
+# Init as early as possible to catch more
+sentry_sdk.init(
+    aspen_config.SENTRY_URL,
+    # Set traces_sample_rate to 1.0 to capture 100%
+    # of transactions for performance monitoring.
+    # We recommend adjusting this value in production.
+    traces_sample_rate=1.0,
+)
```
phoenixAja, line 33: based on the comment should we reduce the sample rate? You could also probably remove the `# We recommend adjusting this value in production.` comment.

lvreynoso: Right now we are only catching exceptions and AFAIK we shouldn't experience too many of them. And events seem to be grouped. But it could get overwhelming. Any suggestions for a good value? We might actually need to let Sentry run in prod for a week or so to get an idea of what to adjust it to.

phoenixAja: maybe 0.7?

morsecodist (approved):
Looks good to me though not sure why the tests are failing

lvreynoso:
Hmm, trying a few quick things didn't fix the CI issues so I'll just merge this code, and then work on the CI issues

PR #577, opened 2021-07-12
[Small] Explicitly exclude .terraform folder in gitignore

lvreynoso:
### Description

The previous patterns to exclude Terraform folders did not filter out the `.terraform/` folder on my machine, so this explicitly excludes it.

#### Issue

### Test plan

Run `git add .` then `git status` and see that the folder is not included.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -312,6 +312,7 @@ static
 /.terraform.*/
 /.terraform.lock.hcl
 /terraform/variables.tf
+.terraform
 
 # VS Code
 .vscode


.gitignore
```diff
@@ -312,6 +312,7 @@ static
 /.terraform.*/
 /.terraform.lock.hcl
 /terraform/variables.tf
+.terraform
```
jgadling, line 315: can/should we also remove the terraform directory from this repo?

lvreynoso: I'm not sure, are the batch jobs moved over? I'm not sure if anything still relies on that folder

jgadling: Hm, I think the only terraform code that's still (somewhat) in use from that directory is the s3 buckets. We should import those into our infra repo when we get a chance.

jgadling (approved):

phoenixAja:
should we get the CI and lint fixed first?

PR #580, opened 2021-07-14
Add undefined guard to table row rendering

lvreynoso:
### Description

Prevents the `sampleRow()` functioned from being called with undefined data, which might create errors (and @phoenixAja has run into problems with this and expanding table functionality)

#### Issue
[ch146170](https://app.clubhouse.io/genepi/story/146170)

### Test plan

Tested with console logging.


phoenixAja (approved):

phoenixAja:
ty!

PR #581, opened 2021-07-14
Add Sentry-Flask extra & send deployment_stage identifier

lvreynoso:
### Description

Adds the Sentry Flask integration, sends the `deployment_stage` identifier to Sentry with alerts. This way we should be able to filter out alerts based on rdev/staging/prod.

#### Issue
[ch132909](https://app.clubhouse.io/genepi/story/132909)

### Test plan

Tested in rdev:
Screen Shot 2021-07-14 at 16 17 13

### DIFF
--- a/src/backend/aspen/app/app.py
+++ b/src/backend/aspen/app/app.py
@@ -4,6 +4,7 @@ from pathlib import Path
 from typing import Optional
 
 import sentry_sdk
+from sentry_sdk.integrations.flask import FlaskIntegration
 from authlib.integrations.flask_client import OAuth
 from flask import redirect, session
 from flask_cors import CORS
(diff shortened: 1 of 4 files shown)


src/backend/aspen/app/app.py
```diff
@@ -22,24 +23,28 @@
...
     aspen_config = DockerComposeConfig()
 
+deployment = os.getenv("DEPLOYMENT_STAGE")
+
 # We should be able to allow this in all environments and only alert on prod.
 # Init as early as possible to catch more
 sentry_sdk.init(
-    aspen_config.SENTRY_URL,
+    dsn=aspen_config.SENTRY_URL,
+    integrations=[FlaskIntegration()],
```
jgadling, line 32: I think we want to set the environment here? (see https://docs.sentry.io/platforms/python/configuration/environments/)

src/backend/aspen/app/app.py
```diff
@@ -22,24 +23,28 @@
...
     # Set traces_sample_rate to 1.0 to capture 100%
     # of transactions for performance monitoring.
     # We recommend adjusting this value in production.
     traces_sample_rate=1.0,
 )
+sentry_sdk.set_user(
+    {"id": os.getenv("EC2_INSTANCE_ID"), "deployment_stage": deployment}
+)
```
jgadling: I think the user is supposed to be a website user. Maybe this would be better to include near here:
https://github.com/chanzuckerberg/aspen/blob/main/src/backend/aspen/app/app.py#L81

lvreynoso:
Added the environment to the `init()` call, and moved `set_user()` call to the `decorated()` function so it is set based on user sessions

jgadling (approved):

PR #583, opened 2021-07-15
Initial support for device authorization flow

jgadling:
### Description

This PR includes initial support for a second backend API authentication scheme: [Device Authorization Flow](https://auth0.com/docs/flows/device-authorization-flow). We'll need to update our Auth0 applications to enable this flow, but it's a small config change.

Now that have two different ways to authenticate to our API: via session cookie, and via a bearer token, our old method of looking for a user ID in session data is unreliable. I've updated our authentication code to put a User record in the [flask request context](https://flask.palletsprojects.com/en/1.1.x/api/#flask.g) regardless of which authentication type is used for a given request.

I've also updated all authenticated endpoints so that they automatically have access to a DB session and don't need to call the session_scope() context manager.

There's a stub of a new Aspen CLI that just acts as an API client using device authorization flow. It will be be fleshed out further as we need to make more API calls from the command line.

I tested these changes against a temporary (just-for-testing) auth0 app. The OIDC provider we use for local development doesn't currently support device authorization flow -- we'll need to look for an alternative that does, so we can have access to this functionality in dev.

#### Issue

### Test plan

### DIFF
--- a/src/backend/aspen/app/app.py
+++ b/src/backend/aspen/app/app.py
@@ -4,21 +4,21 @@ from pathlib import Path
 from typing import Optional
 
 import sentry_sdk
+from auth0.v3.authentication.token_verifier import (
+    AsymmetricSignatureVerifier,
+    TokenVerifier,
+)
+from auth0.v3.exceptions import TokenValidationError
 from authlib.integrations.flask_client import OAuth
-from flask import request, redirect, session, g
+from flask import g, redirect, request, session
 from flask_cors import CORS
 from sentry_sdk.integrations.flask import FlaskIntegration
 
 from aspen.app.aspen_app import AspenApp
+from aspen.app.views.api_utils import get_usergroup_query
 from aspen.config.config import Config
 from aspen.config.docker_compose import DockerComposeConfig
 from aspen.config.production import ProductionConfig
-from auth0.v3.exceptions import TokenValidationError
-from auth0.v3.authentication.token_verifier import (
-    TokenVerifier,
-    AsymmetricSignatureVerifier,
-)
-from aspen.app.views.api_utils import get_usergroup_query
 from aspen.database.connection import session_scope
 
 static_folder = Path(__file__).parent.parent / "static"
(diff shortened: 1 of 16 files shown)


src/backend/aspen/app/app.py
```diff
@@ -70,13 +77,59 @@
...
+    return payload
+
+
+def setup_userinfo(user_id):
+    user = get_usergroup_query(g.db_session, user_id).one()
+    g.auth_user = user
+
+
 # use this to wrap protected views
 def requires_auth(f):
```
jgadling, line 121: This method is where the bulk of the changes in this PR start!

src/backend/aspen/app/app.py
```diff
@@ -70,13 +77,59 @@
...
     @wraps(f)
     def decorated(*args, **kwargs):
-        if "profile" not in session:
-            # Redirect to Login page
+        auth_header = request.headers.get("authorization")
+        auth0_user_id = None
```
jgadling, line 125: This if/elif switches between our two authorization flows (bearer token or session profile)

src/backend/aspen/app/app.py
```diff
@@ -70,13 +77,59 @@
...
+            except TokenValidationError as err:
+                application.logger.debug(f"Token validation error: {err}")
+        elif "profile" in session:
+            auth0_user_id = session["profile"]["user_id"]
+        # Redirect to Login page
+        if not auth0_user_id:
             return redirect("/login")
-        return f(*args, **kwargs)
+        with session_scope(application.DATABASE_INTERFACE) as db_session:
+            g.db_session = db_session
```
jgadling, line 138: Here's where we automatically wrap all endpoints with a @requires_auth in a session_scope. We make the session object available via flask's request context manager.

src/backend/aspen/app/app.py
```diff
@@ -70,13 +77,59 @@
...
+            setup_userinfo(auth0_user_id)
```
jgadling, line 139: And here we make a user object (with groups & cansee rows attached) to all authenticated endpoints.

PR #584, opened 2021-07-15
[small] Set search_path=aspen when connecting to psql

jgadling:
### Description

I got tired of typing 'set search_path=aspen' every time I connected to psql. This makes the search path configuration automatic.

### Test plan

I ran `make local-pgconsole` and `make remote-pgconsole` and both commands still work, and I can query aspen-related tables immediately.


shanaxel42 (approved):
Ah thank you!

PR #589, opened 2021-07-19
Stream fasta files

jgadling:
### Description

This PR all endpoints to accept credentials via OAuth [device authorization flow](https://auth0.com/docs/flows/device-authorization-flow), and updates the sample download endpoint to stream fasta files from the DB in chunks. This uses SQLAlchemy's [yield_per functionality](https://docs.sqlalchemy.org/en/14/core/connections.html?highlight=yield_per#sqlalchemy.engine.Result.yield_per) to only load a few samples at a time into memory, and Flask's [response generator](https://flask.palletsprojects.com/en/2.0.x/patterns/streaming/) functionality to stream results without writing local files.

#### Issue

### Test plan

This has been tested in local dev with a test oauth application, but let's double check that the auth changes haven't affected existing endpoints in staging.

### DIFF
--- a/src/backend/aspen/app/app.py
+++ b/src/backend/aspen/app/app.py
@@ -4,21 +4,21 @@ from pathlib import Path
 from typing import Optional
 
 import sentry_sdk
+from auth0.v3.authentication.token_verifier import (
+    AsymmetricSignatureVerifier,
+    TokenVerifier,
+)
+from auth0.v3.exceptions import TokenValidationError
 from authlib.integrations.flask_client import OAuth
-from flask import request, redirect, session, g
+from flask import g, redirect, request, session
 from flask_cors import CORS
 from sentry_sdk.integrations.flask import FlaskIntegration
 
 from aspen.app.aspen_app import AspenApp
+from aspen.app.views.api_utils import get_usergroup_query
 from aspen.config.config import Config
 from aspen.config.docker_compose import DockerComposeConfig
 from aspen.config.production import ProductionConfig
-from auth0.v3.exceptions import TokenValidationError
-from auth0.v3.authentication.token_verifier import (
-    TokenVerifier,
-    AsymmetricSignatureVerifier,
-)
-from aspen.app.views.api_utils import get_usergroup_query
 from aspen.database.connection import session_scope
 
 static_folder = Path(__file__).parent.parent / "static"
--- a/src/backend/aspen/app/app.py
+++ b/src/backend/aspen/app/app.py
@@ -5,7 +5,7 @@ from typing import Optional
 
 import sentry_sdk
 from authlib.integrations.flask_client import OAuth
-from flask import redirect, session
+from flask import request, redirect, session, g
 from flask_cors import CORS
 from sentry_sdk.integrations.flask import FlaskIntegration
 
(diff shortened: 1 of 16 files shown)


src/backend/aspen/app/app.py
```diff
@@ -73,20 +80,65 @@
...
-
-        sentry_sdk.set_user(
-            {
-                "id": session["profile"]["user_id"],
-                "name": session["profile"]["name"],
-            }
-        )
-        return f(*args, **kwargs)
+        with session_scope(application.DATABASE_INTERFACE) as db_session:
+            g.db_session = db_session
```
shanaxel42, line 138: Wait so are we refactoring to have one db_session open when a user logs in? Instead of for each endpoint?

shanaxel42 (commented):
Wait I don't actually see the streaming changes in this pr?

src/backend/aspen/app/views/sample.py
```diff
@@ -150,57 +149,39 @@ def _format_lineage(sample: Sample) -> dict[str, Any]:
 @application.route("/api/sequences", methods=["GET"])
 @requires_auth
 def prepare_sequences_download():
```
jgadling, line 151: @shanaxel42 I think github wasn't expanding this by default.

src/backend/aspen/app/app.py
jgadling: Yes - we'll have one db_session for each request from an authenticated user.

src/backend/aspen/app/views/sample.py
shanaxel42: AH MY BAD, sorry about that

src/backend/aspen/app/views/sample.py
```diff
@@ -221,251 +202,269 @@ def prepare_sequences_download():
...
-            db_session.query(GisaidAccessionWorkflow)
-            .join(GisaidAccessionWorkflow.inputs)
-            .filter(Entity.id.in_(sample_entity_ids))
-            .options(
-                joinedload(GisaidAccessionWorkflow.inputs),
-                joinedload(GisaidAccessionWorkflow.outputs.of_type(GisaidAccession)),
+        # check that user has access to the requested samples
+        for denied_sample in denied_samples:
+            return Response(
+                "User does not have access the requested sequences",
```
shanaxel42:
```suggestion
                "User does not have access to the requested sequences",
```

src/backend/aspen/app/views/sample.py
```diff
@@ -150,57 +149,39 @@ def _format_lineage(sample: Sample) -> dict[str, Any]:
...
-                        Sample.public_identifier.in_(sample_ids),
+    # stream output file
+    user = g.auth_user
+    fasta_filename = f"{user.group.name}_sample_sequences.fasta"
+    request_data = request.get_json()
+    sample_ids = request_data["requested_sequences"]["sample_ids"]
+
+    @stream_with_context
+    def stream_samples(cansee_groups_private_identifiers):
+        with session_scope(application.DATABASE_INTERFACE) as db_session:
```
shanaxel42, line 160: What's the reason for having to create a new db_session here instead of passing in the existing one?

shanaxel42 (approved):
Big change! Looks fair to me, let's make sure to do some extra staging manual testing after this get's merged

src/backend/aspen/app/views/sample.py
jgadling: As far as I know, ~all API endpoints will require DB access so creating the session here declutters the endpoint code a little bit. I think the main drawback to managing this centrally is that it can make advanced db session management (explicitly closing/reopening sessions) more confusing if any endpoints need to do that.

jgadling: prepare_sequences_download() returns this generator, and then the requires_auth() decorator returns it, closing the db session scope we opened in requires_auth. Even our request context would normally be out of scope, except that we're using flask's [stream_with_context](https://flask.palletsprojects.com/en/2.0.x/patterns/streaming/#streaming-with-context) decorator to keep it around.

PR #601, opened 2021-07-26
Convert the sample download endpoint to a POST request

jgadling:
### Description

- [CloudFront doesn't allow GET requests with a body](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/RequestAndResponseBehaviorCustomOrigin.html#RequestCustom-get-body) (this is pretty annoying since it's a pretty common pattern for REST api's)
- Most web browsers / server software has a limit on the length of a valid URL - 2000 characters is common.
- If our sample ID's are typically about 20 characters long, this limits us to being able to download about 100 samples at a time, and I think our goal is to be able to support larger batch downloads.

#### Issue

### Test plan

I've tested the CLI and API changes locally, I'm still able to download sequences

### DIFF
--- a/src/backend/aspen/app/views/sample.py
+++ b/src/backend/aspen/app/views/sample.py
@@ -146,7 +146,7 @@ def _format_lineage(sample: Sample) -> dict[str, Any]:
     return lineage
 
 
-@application.route("/api/sequences", methods=["GET"])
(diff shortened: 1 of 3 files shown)


phoenixAja (approved):
LG!

PR #603, opened 2021-07-26
Quickfix to reduce query time for phylo tree view.

jgadling:
### Description

This is a quick patch to improve the performance of the phylo tree endpoint.

#### Issue
[ch150554](https://app.clubhouse.io/genepi/story/150554)

### Test plan

Tests pass

### DIFF
--- a/src/backend/aspen/app/views/phylo_trees.py
+++ b/src/backend/aspen/app/views/phylo_trees.py
@@ -2,12 +2,12 @@ import json
 import os
 import re
 import uuid
-from typing import Any, Callable, Iterable, Mapping, MutableSequence, Set, Tuple, Union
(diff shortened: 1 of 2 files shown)


src/backend/aspen/app/views/phylo_trees.py
```diff
@@ -53,16 +53,6 @@ def phylo_trees():
     phylo_runs: Iterable[Tuple[PhyloRun, int]] = (
```
shanaxel42:
```suggestion
    phylo_runs: Iterable[PhyloRun] = (
```

This is probably what py-lint is mad about

shanaxel42 (approved):
Looks good!

PR #605, opened 2021-07-26
Get associated phylo trees in the main query.

jgadling:
### Description

This should resolve the n+1 query issue in our phylo tree endpoint. [We use a similar trick here](https://github.com/chanzuckerberg/aspen/blob/trunk/src/backend/aspen/app/views/sample.py#L300).

#### Issue
[ch150554](https://app.clubhouse.io/genepi/story/150554)

### Test plan

Tests pass, and I verified via logs in local dev that we're only running one query for this endpoint.


shanaxel42 (approved):
Awesome!!

PR #611, opened 2021-07-28
Log and handle db user queries where None is found

lvreynoso:
### Description

We had a sudden burst of events in the past two days where somehow some auth0 headers were transmitted when trying to access a protected route but no corresponding user was found in the db, raising an exception. This PR adds exception handling to that case and sends a sentry message to capture the requesting user_id.

#### Issue

### Test plan

I'm not sure how to reproduce the circumstances.


src/backend/aspen/app/app.py
```diff
@@ -107,7 +107,17 @@ def validate_auth_header(auth_header):
 
 
 def setup_userinfo(user_id):
-    user = get_usergroup_query(g.db_session, user_id).one()
+    sentry_sdk.set_user(
+        {
+            "requested_user_id": user_id,
+        }
+    )
+    user = get_usergroup_query(g.db_session, user_id).one_or_none()
```
jgadling: Can we leave this as a .one() call inside a try / except NoResultFound block instead? I think it's better to be explicit that getting no rows back is an exception.

lvreynoso: Done!

jgadling (approved):

PR #612, opened 2021-07-28
Migration script to backfill missing epi ids

shanaxel42:
### Description

Had to remake this PR because alembic weirdness from the last migration

This migration script populates the remaining missing epi_isl ids from the covidhub db. Once this runs will follow up with a pr to remove all the covidhub import code 

#### Issue
[ch150722](https://app.clubhouse.io/genepi/story/150722)

### Test plan

Tested locally on some ids within my local db. Also manually queried to make sure non of these ids existed yet.


jgadling (approved):

src/backend/database_migrations/versions/20210728_213140_backfill_gisaid_epi_ids.py
```diff
@@ -0,0 +1,1510 @@
...
+            repository_type=PublicRepositoryType.GISAID,
+            public_identifier=epi_isl,
+            workflow_start_datetime=datetime.datetime.now(),
+            workflow_end_datetime=datetime.datetime.now(),
+        )
+    session.commit()
+
+
+def downgrade():
+    ...
```
jgadling: nit - I think `pass` is a little more commonly used than the ellipsis (`...`) to declare an empty block.

shanaxel42: changed!

src/backend/database_migrations/versions/20210728_213140_backfill_gisaid_epi_ids.py
```diff
@@ -0,0 +1,1510 @@
...
+    )
+    for sample in samples_to_update:
+        epi_isl = MISSING_EPI_ISL_MAPPING[sample.public_identifier]
+        sample.uploaded_pathogen_genome.add_accession(
+            repository_type=PublicRepositoryType.GISAID,
+            public_identifier=epi_isl,
+            workflow_start_datetime=datetime.datetime.now(),
+            workflow_end_datetime=datetime.datetime.now(),
+        )
+    session.commit()
```
phoenixAja, line 1506: so i think when writing migrations we should only be using raw sql because otherwise if our model api changes then this migration will fail (or at least that's what tony told me awhile ago)

src/backend/database_migrations/versions/20210728_213140_backfill_gisaid_epi_ids.py
```diff
@@ -0,0 +1,1510 @@
...
+
+
+def downgrade():
+    pass
```
phoenixAja, line 1510: also i think next time we should add the NotImplemented error to keep things consistant?

PR #617, opened 2021-08-03
Add API pathing and batch support for on-demand tree building

lvreynoso:
### Description

Adds the SFN execution call to `start_phylo_run()` including building needed input JSON. Adds support for getting information from the SSM Parameter Store in the Config object.

#### Issue
[ch150139](https://app.clubhouse.io/genepi/story/150139)

### Test plan

* Local tests cover the API route with valid, invalid, and unauthorized requests
* End-to-end integration testing done in RDEV


src/backend/aspen/app/views/phylo_runs.py
```diff
@@ -143,5 +147,41 @@ def start_phylo_run():
...
+
+    session = aws.session()
+    client = session.client(
+        service_name="stepfunctions",
+        endpoint_url=os.getenv("BOTO_ENDPOINT_URL") or None
+    )
+
+    try:
+        response = client.start_execution(
+            stateMachineArn=os.environ.get("NEXTSTRAIN_SFN_ARM"),
```
jgadling: This should be ARN

lvreynoso: oops got excited about ARM computing for a second there

src/cli/aspencli.py
```diff
@@ -211,6 +216,22 @@ def download_samples(ctx, sample_ids):
...
+def phylo_run():
+    pass
+
+@phylo_run.command(name="start")
+@click.option("-n","--name", required=True, type=str)
+@click.option("-t","--type", "tree_type", required=True, type=click.Choice(["local", "contextual"], case_sensitive=False))
+@click.argument("sample_ids", nargs=-1)
+@click.pass_context
+def start_phylo_run(ctx, name, tree_type, sample_ids):
+    # print("Name:", name, "Tree type:", tree_type, "Sample ids:", sample_ids)
```
phoenixAja: can you remove this comment?

src/cli/aspencli.py
```diff
@@ -211,6 +216,22 @@ def download_samples(ctx, sample_ids):
...
+    api_client = ctx.obj["api_client"]
+    payload = { "name": name, "tree_type": tree_type, "samples": sample_ids }
+    resp = api_client.post("/api/phylo_runs", json=payload)
+    print(resp.headers)
```
phoenixAja: do you still need these print statements?

src/backend/scripts/setup_dev_data.sh
```diff
@@ -37,8 +37,43 @@ ${local_aws} secretsmanager update-secret --secret-id aspen-config --secret-stri
   "S3_external_auspice_bucket": "aspen-external-auspice-data"
 }' || true
 
+echo "Creating SSM Parameter"
```
phoenixAja, line 40: dumb question, why are we creating a new SSM parameter/ IAM role here and not in the terraform code?

src/backend/aspen/workflows/nextstrain_run/export.py
```diff
@@ -127,146 +131,135 @@ def cli(
...
+
+        upload_date = None
+        if sample.sequencing_reads_collection is not None:
+            upload_date = sample.sequencing_reads_collection.upload_date.strftime(
+                "%Y-%m-%d"
+            )
+        elif sample.uploaded_pathogen_genome is not None:
+            upload_date = sample.uploaded_pathogen_genome.upload_date.strftime(
+                "%Y-%m-%d"
+            )
```
phoenixAja, line 229: why not put upload_date = None in an else block? I think that's a little easier to read

src/backend/aspen/app/views/phylo_runs.py
```diff
@@ -0,0 +1,188 @@
...
+
+    template_path_prefix = (
+        "/usr/src/app/aspen/workflows/nextstrain_run/builds_templates"
+    )
+    builds_template_file = (
+        f"{template_path_prefix}/{PHYLO_TREE_TYPES[request_data['tree_type']]}"
+    )
+    builds_template_args = {
+        "division": group.division,
+        "location": group.location,
```
phoenixAja, line 86: i think this should be the sample division/ location and not the group division /location?

src/backend/aspen/app/views/phylo_runs.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+class PhyloRunResponseSchema(Schema):
+    id = fields.Int()
+    start_datetime = fields.DateTime()
+    end_datetime = fields.DateTime()
+    workflow_status = fields.Pluck(WorkflowStatusSchema, "name")
+    group = fields.Nested(GroupResponseSchema, only=("id", "name"))
+    template_file_path = fields.String()
+    template_args = fields.Nested(GroupResponseSchema, only=("division", "location"))
```
phoenixAja: this is a huge nit, but if we're starting to use serializers like this it might be nice to separate them out somewhere away from our views (like on the same level as models/views we could also have serializers)

src/backend/aspen/app/views/api_utils.py
```diff
@@ -133,3 +134,49 @@ def check_valid_sequence(sequence: str) -> bool:
...
+    cansee_groups: Set[int] = {
+        cansee.owner_group_id
+        for cansee in user.group.can_see
+        if cansee.data_type == DataType.SEQUENCES
+    }
+    # add the user's own group
+    cansee_groups.add(user.group_id)
+
+    # Which groups can this user query private identifiers for?
+    # NOTE - this asssumes PRIVATE_IDENTIFIERS permission is a superset of SEQUENCES
```
phoenixAja, line 160: all these comments are great btw!

Makefile
```diff
@@ -165,7 +165,6 @@ local-clean: local-nohostconfig ## Remove everything related to the local dev en
 	-rm -rf ./oauth/pkcs12/server*
 	-rm -rf ./oauth/pkcs12/certificate*
 	docker-compose rm -sf
-	-docker volume rm aspen_database
```
phoenixAja, line 168: was this intentional?

phoenixAja (commented):
THIS IS AMAZING!!!!! EXCELLENT WORK!!! 

The only small question that might a blocker was if we should be using group division/location instead of sample division/location, but besides that this is really looking good 👍 

I'll leave this as a comment for now and let @shanaxel42 do her review to officially approve

Makefile
jgadling: Yep, this volume doesn't actually exist

src/backend/aspen/app/views/phylo_runs.py
```diff
@@ -0,0 +1,188 @@
...
+                ),
+                "aws_region": aws.region(),
+                "docker_image_id": aspen_config.NEXTSTRAIN_DOCKER_IMAGE_ID,
+                "group_name": group.name,
+                "remote_dev_prefix": os.getenv("REMOTE_DEV_PREFIX"),
+                "s3_filestem": f"{group.location}/{request_data['tree_type'].capitalize()}",
+                "workflow_id": workflow.id,
+            },
+        },
+        "OutputPrefix": aspen_config.NEXTSTRAIN_OUTPUT_PREFIX,
```
jgadling: TODO -- we need to specify a different output prefix for every SFN call. We should update this to include the workflow ID.

src/backend/aspen/app/views/phylo_runs.py
jgadling: Yeah - this is a little sticky actually. Theoretically each sample can have a different division/location, but the value in this part of the code is what get's populated in the nextstrain config for filtering, and we can only specify a single division/location there, which is why I was leaning towards using the group's division/location. Thoughts? @danrlu ?

src/cli/aspencli.py
lvreynoso: nope!

lvreynoso: yes, this is part of the CLI so we need to let the end-user (us) know what the response was.

src/backend/scripts/setup_dev_data.sh
lvreynoso: The ones we use in rdev/staging/prod are in terraform code, this is for mocking the params in the localstack.

src/backend/aspen/app/views/phylo_runs.py
danrlu: Very good question. I agree with what Jessica said!

src/backend/aspen/workflows/nextstrain_run/export.py
lvreynoso: @jgadling wrote this part, but I also like it this way because it's a lot more clear that `upload_date` is in the scope of the larger block; and it shows the default value first.

jgadling: I much prefer a default value above a check for a special case!

That said though, I think we're skipping support for sequencing_reads_collections at the moment, so we can assume we'll always have an uploaded pathogen genome?

src/cli/aspencli.py
jgadling: we should probably add some sort of --debug flag to en/disable displaying headers. Since the CLI is currently mainly used for debugging it's probably not an urgent fix.

src/backend/aspen/app/views/phylo_runs.py
lvreynoso: updated!

src/backend/aspen/app/views/phylo_runs.py
```diff
@@ -0,0 +1,148 @@
...
+    except ValidationError as verr:
+        return Response(str(verr), 400)
+    sample_ids = request_data["samples"]
+
+    # Step 2 - prepare big sample query per the old db cli
+    all_samples: Iterable[Sample] = g.db_session.query(Sample).options(
+        joinedload(Sample.uploaded_pathogen_genome, innerjoin=True),
+    )
+    # Step 3 - Enforce AuthZ
+    all_samples = authz_sample_filters(all_samples, sample_ids, user)
```
shanaxel42, line 51: Nice! I like this auth flow

src/backend/aspen/workflows/nextstrain_run/export.py
shanaxel42: There's the rare occurrence of a legacy failed sample that wouldn't have the uploaded pathogen genome so this is a good check

shanaxel42 (approved):
This looks awesome guys!! Great work excited to get this functionality in 😄

.happy/terraform/modules/sfn_config/nextstrain-ondemand.wdl
```diff
@@ -0,0 +1,114 @@
...
+    aligned_gisaid_metadata_s3_key=$(echo "${aligned_gisaid_location}" | jq -r .metadata_key)
+
+    # fetch the gisaid dataset
+    aws s3 cp --no-progress "s3://${aligned_gisaid_s3_bucket}/${aligned_gisaid_sequences_s3_key}" - | zstdmt -d | xz -2 > ncov/results/aligned_gisaid.fasta.xz
+    aws s3 cp --no-progress "s3://${aligned_gisaid_s3_bucket}/${aligned_gisaid_metadata_s3_key}" ncov/data/metadata_gisaid.tsv
+
+    (cd ncov && snakemake --printshellcmds auspice/ncov_aspen.json --profile my_profiles/aspen/ --resources=mem_mb=312320) || aws s3 cp ncov/.snakemake/log/ "s3://${aspen_s3_db_bucket}/phylo_run/${build_id}/" --recursive
+
+    # upload the tree to S3
+    key="phylo_run/${build_id}/~{s3_filestem}.json"
```
jgadling: Can/Should we add a workflow ID to this path? Multiple jobs that start at the ~same time could clobber each other's result files.

jgadling: I updated the s3 paths.

PR #618, opened 2021-08-04
Remove unused covidhub import module

lvreynoso:
### Description

We deleted the covidhub import wdl file but left the terraform module that requests it in place, causing an error when applying it. This PR removes the module. 

#### Issue

### Test plan


jgadling (approved):

shanaxel42 (approved):
Ah good catch!!

PR #619, opened 2021-08-04
Fix backfill migration to check for uploaded pathogen genomes

lvreynoso:
### Description

The 20210728 backfill migration is breaking in rdev because it's trying to call `add_accession()` on a sample's `uploaded_pathogen_genome` but it seems that at least one of them doesn't exist. This just adds a conditional statement to safely run the migration.

#### Issue

### Test plan

Tested by creating an rdev stack and successfully applying the migrations.


jgadling (approved):

phoenixAja:
when writing migrations you should try to only use raw sql otherwise if our model api definitions change then these migrations will fail.

lvreynoso:
Yeah apologies for taking the easy route, I'm just trying to get rdev up

phoenixAja:
@lvreynoso np! Sorry i didn't mean to sound harsh, was just letting you know just in case 👍

PR #621, opened 2021-08-06
[EZPZ] rename metadata tsv download

phoenixAja:
### Description

Match tsv download to sequences download file name: 
ex: 
`czi_sample_sequences_2021-08-06_metadata.tsv`
`czi_sample_sequences_2021-08-06.fasta`

#### Issue
no issue

### Test plan
tested locally (can make an rdev, but thought that might be overkill, lmk if you want one!)

### DIFF
--- a/src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
@@ -220,12 +220,16 @@ const DownloadModal = ({
 
   function getDownloadButton(): JSX.Element | undefined {
     // button will have different functionality depending on download type selected
+    const metadataFilename = `${groupName}_sample_sequences_${downloadDate
+      .toISOString()
+      .slice(0, 10)}_metadata.tsv`;
+
     if (isMetadataSelected && !isFastaSelected) {
       return (
         <CSVLink
           data={tsvRows}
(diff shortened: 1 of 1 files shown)


mmmmmmaya (approved):

src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
```diff
@@ -220,12 +221,14 @@ const DownloadModal = ({
 
   function getDownloadButton(): JSX.Element | undefined {
     // button will have different functionality depending on download type selected
+    const metadataFilename = `${downloadPrefix}_metadata.tsv`;
+
     if (isMetadataSelected && !isFastaSelected) {
```
mmmmmmaya, line 226: It seems like there are 4 separate `CSVLink` definitions here depending on various permutations of `isMetadataSelected` and `isFastaSelected`, but the only the that (seems to?) varies is the click handler. I wonder if we can determine the click handler using conditional logic before render, and then eliminate 3/4 of the `CSVLink` components here.

phoenixAja: yeah i think that could definitely be improved.. were you thinking make the csvlink it's own component and pass the click handler as props?

PR #622, opened 2021-08-06
[ch143538] hide columns when no samples have been submitted to gisaid

mmmmmmaya:
### Summary
- **What:** Hide Sample ID and ISL # fields on metadata upload page if no samples have been submitted to GISAID yet
- **Why:** design request
- **Ticket**: [ch143538](https://app.clubhouse.io/genepi/story/143538/hide-public-id-and-isl-columns-if-no-samples-have-been-submitted-to-gisaid)

#### Before: 
Screen Shot 2021-08-06 at 1 46 05 PM

#### After: 
after

### Notes
Since the `isShown` property was only used in this particular case, I removed it to keep things clean :)

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/views/Upload/components/Metadata/components/Table/index.tsx
+++ b/src/frontend/src/views/Upload/components/Metadata/components/Table/index.tsx
@@ -53,7 +53,7 @@ export default function Table({
     );
 
     return () => clearTimeout(timeout);
-  }, []);
+  }, [metadata]);
 
   useEffect(() => {
     if (hasImportedFile) {
@@ -65,7 +65,7 @@ export default function Table({
     const isValid = Object.values(rowValidation).every((isValid) => isValid);
 
     setIsValid(isValid);
-  }, [rowValidation]);
+  }, [rowValidation, setIsValid]);
 
   const handleRowValidation_ = (id: string, isValid: boolean) => {
     if (rowValidation[id] === isValid) return;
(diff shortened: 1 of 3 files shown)


phoenixAja (approved):
LG

src/frontend/src/views/Upload/components/Metadata/components/Table/index.tsx
```diff
@@ -1,4 +1,5 @@
 import { Table as MuiTable, TableBody, TableHead } from "@material-ui/core";
+import { reduce } from "lodash";
```
tihuan, line 2: I think we might need `"lodash/reduce"` to prevent importing the whole `lodash` library here!

[source](https://dev.to/icokie/lodash-import-done-right-290n)

src/frontend/src/views/Upload/components/Metadata/components/Table/index.tsx
```diff
@@ -110,6 +113,16 @@ export default function Table({
     );
   }
 
+  const shouldShowGISAIDFields = reduce(
```
tihuan, line 116: I'm thinking maybe the following would do? 

```ts
shouldShowGISAIDFields = metadata.some((data) => Boolean(data?.submittedToGisaid))
```

This way we can early return `true` when we find the first element that has `submittedToGisaid === true` without going through the whole array like `reduce()` does. Secondly we avoid importing `reduce` and uses the native `Array.some()` method 😄 

[source](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)

tihuan (commented):
Sorry for the late review, @mmmmmmaya ! Was out on Friday 😆 

Nothing urgent, just two small nits! Thank you!

PR #631, opened 2021-08-11
Cleanly handle the case where an auth0 user is not in the db

lvreynoso:
### Description

We have a few login attempts from auth0 users who are not in the db. Currently the `setup_userinfo()` function returns a redirect but does not stop execution in the parent `requires_auth()` function, raising an `AttributeError` when we check for `if not g.auth_user` since that attribute does not exist on `g`. This PR properly sets `g.auth_user` to `None`.

#### Issue

### Test plan

Tested locally, logged in and logged out cases.

### DIFF
--- a/src/backend/aspen/app/app.py
+++ b/src/backend/aspen/app/app.py
@@ -180,7 +180,14 @@ def requires_auth(f):
             return redirect("/login")
         with session_scope(application.DATABASE_INTERFACE) as db_session:
             g.db_session = db_session
-            g.auth_user = setup_userinfo(auth0_user_id)
+            found_auth_user = setup_userinfo(auth0_user_id)
+            if not found_auth_user:
+                # login attempt from user not in DB
+                # TODO: return response to user that they do not have an account with aspen.
+                return redirect("/login")
+            else:
+                g.auth_user = found_auth_user
+            
             if not g.auth_user:
                 return redirect("/login")
             return f(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


src/backend/aspen/app/app.py
```diff
@@ -151,14 +151,14 @@ def setup_userinfo(user_id):
         sentry_sdk.capture_message(
             f"Requested auth0_user_id {user_id} not found in usergroup query."
```
phoenixAja, line 152: i think it would be nice to  also log their login email? That way we can more easily tell if we need to add a new user to our DB  or if we're getting unexpected login attempts (without involving detective @shanaxel42 lol)

phoenixAja: if we have that information in the request

src/backend/aspen/app/app.py
```diff
@@ -180,7 +180,7 @@ def decorated(*args, **kwargs):
             return redirect("/login")
         with session_scope(application.DATABASE_INTERFACE) as db_session:
             g.db_session = db_session
-            setup_userinfo(auth0_user_id)
+            g.auth_user = setup_userinfo(auth0_user_id)
```
phoenixAja: I think this is a little clearer what's going on (imo, totally cool if you think this is not a better idea lol)
```suggestion
            found_auth_user = setup_userinfo(auth0_user_id)
            if not found_auth_user:
                # login attempt from user not in DB
                # TODO: return response to user that they do not have an account with aspen.
                return redirect("/login")
            else:
                g.auth_user = found_auth_user
            
```

phoenixAja (approved):

phoenixAja:
test this to see if it works properly before merging? (Not sure if you already did that or not based on PR description)

src/backend/aspen/app/app.py
lvreynoso: merged!

PR #632, opened 2021-08-11
add interactive filtering to bulk download

phoenixAja:
### Description

allow users to filter and add/ remove data from returned filtering results

#### Issue
[ch154101](https://app.clubhouse.io/genepi/story/154101)

### Test plan

tested locally and in rdev: https://bulk-download-w-filtering-frontend.dev.genepi.czi.technology/data/samples

### DIFF
--- a/src/frontend/src/common/components/library/data_subview/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/index.tsx
@@ -131,10 +131,7 @@ const DataSubview: FunctionComponent<Props> = ({
   });
 
   const [checkedSamples, setCheckedSamples] = useState<any[]>([]);
-  const [isHeaderChecked, setIsHeaderChecked] = useState<boolean>(false);
   const [showCheckboxes, setShowCheckboxes] = useState<boolean>(false);
-  const [isHeaderIndeterminant, setHeaderIndeterminant] =
-    useState<boolean>(false);
   const [open, setOpen] = useState(false);
   const [isDownloadDisabled, setDownloadDisabled] = useState<boolean>(true);
   const [failedSamples, setFailedSamples] = useState<any[]>([]);
@@ -175,48 +172,6 @@ const DataSubview: FunctionComponent<Props> = ({
       setOpen(false);
     }
   }, [downloadFailed]);
-
-  function handleHeaderCheckboxClick(
-    newPublicIds: string[],
-    newFailedIds: string[]
-  ) {
-    if (isHeaderIndeterminant || isHeaderChecked) {
-      // remove samples in current data selection when selecting checkbox when indeterminate
-      const newCheckedSamples = checkedSamples.filter(
-        (el) => !newPublicIds.includes(el)
-      );
-      const newFailedSamples = failedSamples.filter(
-        (el) => !newFailedIds.includes(el)
-      );
-      setCheckedSamples(newCheckedSamples);
-      setFailedSamples(newFailedSamples);
-      setIsHeaderChecked(false);
-      setHeaderIndeterminant(false);
-    }
-    if (!isHeaderChecked && !isHeaderIndeterminant) {
-      // set isHeaderChecked to true, add all samples in current view
-      setCheckedSamples(checkedSamples.concat(newPublicIds));
-      setFailedSamples(failedSamples.concat(newFailedIds));
-      setIsHeaderChecked(true);
-    }
-  }
-
-  function handleRowCheckboxClick(
-    sampleId: string,
-    failedGenomeRecovery: boolean
-  ) {
-    if (checkedSamples.includes(sampleId)) {
-      setCheckedSamples(checkedSamples.filter((id) => id !== sampleId));
-      if (failedGenomeRecovery) {
-        setFailedSamples(failedSamples.filter((id) => id !== sampleId));
-      }
-    } else {
-      setCheckedSamples([...checkedSamples, sampleId]);
-      if (failedGenomeRecovery) {
-        setFailedSamples([...failedSamples, sampleId]);
-      }
-    }
-  }
 
   function handleDismissErrorClick() {
     setDownloadFailed(false);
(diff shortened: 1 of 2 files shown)


src/frontend/src/common/components/library/data_table/index.tsx
```diff
@@ -112,13 +151,45 @@ export const DataTable: FunctionComponent<Props> = ({
   handleHeaderCheckboxClick,
   handleRowCheckboxClick,
   isHeaderChecked,
+  setIsHeaderChecked,
```
mmmmmmaya: I think it is a react anti pattern to allow a component to update another component's state in this way. Something feels weird about passing in the state modification functions.

src/frontend/src/common/components/library/data_table/index.tsx
```diff
@@ -112,13 +151,45 @@ export const DataTable: FunctionComponent<Props> = ({
...
   isHeaderIndeterminant,
+  setHeaderIndeterminant,
 }: Props) => {
   const [state, dispatch] = useReducer(reducer, {
     ascending: false,
     sortKey: defaultSortKey,
   });
 
+  useEffect(() => {
```
mmmmmmaya, line 158: I'm a little confused why this all needs to happen in DataTable instead of in the parent component. It seems like `data`, and the state modifiers are all passed into this function. Are you able to do all this calculation in the parent component and then pass it down as a simplified/reduced prop instead of passing down all the info and functions needed to calculate it?

src/frontend/src/common/components/library/data_subview/index.tsx
```diff
@@ -207,14 +176,28 @@ const DataSubview: FunctionComponent<Props> = ({
...
-      setCheckedSamples([]);
-      setFailedSamples([]);
+  function handleHeaderCheckboxClick(
+    newPublicIds: string[],
+    newFailedIds: string[]
+  ) {
+    if (isHeaderIndeterminant || isHeaderChecked) {
+      // remove samples in current data selection when selecting checkbox when indeterminate
+      const newCheckedSamples = checkedSamples.filter(
+        (el) => !newPublicIds.includes(el)
```
mmmmmmaya: is this comparing html elements? Typically I think `el` as a variable name is reserved for use with html elements. Is there another variable name that is applicable to our use case that would work here?

mmmmmmaya:
I took a brief read through and I both understand what you are trying to do (from a product standpoint), and what you're trying to do from a code standpoint. I think I may be missing some piece of your mental model that is preventing me from understanding this solution. I'm going to pull the branch now and play with it a little bit, and see if that generates any insight for me.

I'll follow up with another comment soon!

src/frontend/src/common/components/library/data_table/index.tsx
phoenixAja: so this needs to happen in data table because that's where the table rows are being rendered based on filtering (and we need  access to this subset of data to tell which state the header should be (indeterminate or checked)), data in the subview is a list of all available rows (prior to any filtering that's done).

src/frontend/src/common/components/library/data_subview/index.tsx
phoenixAja: no this is comparing publicId values in each TableItems (data) from the current filtered view, and removing or adding them to the checkedSamples.

src/frontend/src/common/components/library/data_table/index.tsx
```diff
@@ -112,13 +151,45 @@ export const DataTable: FunctionComponent<Props> = ({
...
+      if (publicIds.length === sizeData || publicIds.length === 0) {
+        setHeaderIndeterminant(false);
+      } else {
+        setHeaderIndeterminant(true);
+      }
+    }
+  }, [data, checkedSamples, setHeaderIndeterminant]);
+
+  useEffect(() => {
+    // used to determine if header is checked
```
mmmmmmaya: I feel mixed about suggesting this, but I'm wondering if you considered combining these two useEffect. They both begin with the same couple of lines of code that iterate through a potentially pretty large data object. They both also deal with checkbox state, and although checked vs indeterminant is not exactly the same thing, I think they are similar enough you _could_ combine them if you wanted.

No pressure. I could go either way on it. Wonder what you think.

phoenixAja:
@mmmmmmaya i reorganized things a little so that this makes more sense. I moved checkbox callback functions and header state props to the datatable view, and am now passing in checkedSamples, setCheckedSamples, failedSamples, and setFailedSamples as props into data_table, Unfortunately cannot move those into data_table since the download modal also needs access to those props. Hopefully this makes more sense now, otherwise i can change it back to how it was.

src/frontend/src/common/components/library/data_table/index.tsx
phoenixAja: sure i combined them

src/frontend/src/common/components/library/data_table/index.tsx
phoenixAja: i reorganized this part to be more straightforwards

src/frontend/src/common/components/library/data_table/index.tsx
```diff
@@ -91,6 +96,35 @@ function sortData(
...
+): string[] {
+  const publicIds: string[] = [];
+  for (const key in data) {
+    const id = String(data[key as any].publicId);
+    if (onlyCheckedSamples) {
+      if (checkedSamples.includes(id)) {
+        publicIds.push(id);
+      }
+    } else {
+      publicIds.push(String(data[key as any].publicId));
```
mmmmmmaya:
```suggestion
      publicIds.push(id);
```

mmmmmmaya (approved):

src/frontend/src/common/components/library/data_table/index.tsx
phoenixAja: ty! i missed this

PR #633, opened 2021-08-12
[ch150281] filter panel toggle

mmmmmmaya:
### Summary
- **What:** 
  - Add a toggle button for the filter panel
  - Create an `enum` type def for data table view
  - Genericize styling for the IconButton grey background circle
- **Why:** 
  - So users can show/hide the panel
  - In case we ever need to change the view name, we only need to update one location
  - Reusability = DRY
- **Ticket:** [[ch150281]](https://app.clubhouse.io/genepi/story/150281)

### Testing
1. Click the button
1. Ensure filter panel disappears
2. Click again
3. Ensure panel is shown

### Demos
after

### Checklist
- [ ] I merged latest `trunk`
- [x] I added labels to my PR


src/frontend/src/common/icons/IconFilters.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="32" height="32" viewBox="0 0 32 32" fill="none" xmlns="http://www.w3.org/2000/svg">
```
phoenixAja, line 1: should we take out these attrs and add them to the style.ts instead? (that's how i've been styling the other svgs)

phoenixAja (approved):

src/frontend/src/common/icons/IconFilters.svg
mmmmmmaya: My instinct says to leave them in the svg. What is the benefit of pulling them out into a style sheet?

PR #638, opened 2021-08-17
tree build button/ modal / api request

phoenixAja:
### Description

adding in tree build button, basic place holder modal, and "create tree" button to send tree build request to BE phyloruns post endpoint

#### Issue

- [x] [ch156147](https://app.clubhouse.io/genepi/story/156147) (add api call to phyloruns endpoint to start tree build)

- [x] [ch150269](https://app.clubhouse.io/genepi/story/150269) (Create tree build button with tooltip hover)

### Test plan

tested locally

treebuildbutton


src/frontend/src/common/components/library/data_subview/components/IconButton/style.ts
```diff
@@ -0,0 +1,12 @@
+import styled from "@emotion/styled";
+import { getSpacings } from "czifui";
+
+export const StyledSpan = styled.span`
+  display: flex;
+  ${(props) => {
+    const spacings = getSpacings(props);
+    return `
+    margin-left: ${spacings?.m}px;
+  `;
```
mmmmmmaya:
```suggestion
      margin-left: ${spacings?.m}px;
    `;
```

src/frontend/src/common/components/library/data_subview/components/createTreeModal/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+        <StyledIconButton onClick={onClose}>
+          <CloseIcon />
+        </StyledIconButton>
+        <Header>Create New Phylogenetic Tree</Header>
+        <Title>
+          {sampleIds.length} Sample{sampleIds.length > 1 && "s"} Selected
+        </Title>
+      </DialogTitle>
+      <DialogContent>
+        There are {failedSamples.length} failed samples selected
```
mmmmmmaya, line 62: may want to do the `{conditional && "s"}` thing here for both `is`/`are` and `sample`/`samples`. Not sure how far into it you want to get.

src/frontend/src/common/components/library/data_subview/components/createTreeModal/style.ts
```diff
@@ -0,0 +1 @@
+export {};
```
mmmmmmaya, line 1: I assume you're going to add some stuff in here immediately after merging this, so I don't mind it here. I'm wondering if, generally, we can avoid creating files that don't have anything in them? (Not sure whether linter requires a style sheet for every component or something).

src/frontend/src/common/components/library/data_subview/index.tsx
```diff
@@ -141,11 +142,22 @@ const DataSubview: FunctionComponent<Props> = ({
   const [isMetadataSelected, setMetadataSelected] = useState<boolean>(false);
   const [isFastaSelected, setFastaSelected] = useState<boolean>(false);
   const [isFastaDisabled, setFastaDisabled] = useState<boolean>(false);
+  const [createTreeModalOpen, setCreateTreeModalOpen] =
```
mmmmmmaya: idk if this is much of a pattern at czi, but one nice thing you can do for booleans is to prefix the variable name with something that indicates the value would be true/false. For example `didCreateTreeFailed` vs `createTreeFailed`. You can also see an example of this a few lines above with `isFastaDisabled`. Good prefixes for booleans (not all inclusive!) are `is`, `did`, `has`, `can`. etc. Basically if it's a word you can use to start a yes/no question in English 😂 .

Would love to see that pattern used here as well.

src/frontend/src/common/components/library/data_subview/index.tsx
```diff
@@ -224,50 +257,57 @@ const DataSubview: FunctionComponent<Props> = ({
...
+              setFastaSelected={setFastaSelected}
+              isFastaDisabled={isFastaDisabled}
+              setFastaDisabled={setFastaDisabled}
+              tsvData={tsvDataMap(
+                checkedSamples,
+                tableData,
+                headers,
+                subheaders
+              )}
+              open={open}
```
mmmmmmaya: Not necessary to fix this one, but since there are now two modals, I might consider renaming this to something more descriptive. It _is_ distinct enough from the other one to tell them apart, but no longer gives necessary context on what is open to be a good stand alone var name.

src/frontend/src/common/components/library/data_subview/index.tsx
phoenixAja: yeah i can change them, i think that makes sense. I named the other boolean values in that way, interesting that i didn't do the same here..  lol

src/frontend/src/common/components/library/data_subview/components/createTreeModal/style.ts
phoenixAja: I was a little lax because we were adding this to the feature branch and I was going to start using this file right away so i left it in. I won't be this lazy in the next PR 😆

src/frontend/src/common/components/library/data_subview/components/createTreeModal/index.tsx
phoenixAja: this is actually just a placeholder for now so that the linter wouldn't complain about failedSamples not being used.

src/frontend/src/common/components/library/data_subview/components/IconButton/style.ts
phoenixAja: interesting the linter didn't catch this

src/frontend/src/common/components/library/data_subview/index.tsx
phoenixAja: i for sure agree that this should be something more descriptive, I can add that now.

mmmmmmaya (approved):

PR #639, opened 2021-08-18
Fasta URL Generator (primarily for UShER integration)

jgadling:
### Description

This adds an API endpoint that's accepts a list of sample ID's `{"samples": ["USA/12345", "USA/6789"]}`, writes a fasta file containing the sequence data for those samples to S3, and returns a signed URL for that file. UShER is capable of accepting this signed URL as input.

There are a few things worth calling out here:
- We're using a library called [smart_open](https://pypi.org/project/smart-open/) to stream the fasta file to S3 efficiently.
- I refactored most of the code from the fasta download endpoint into a library that's used by both the fasta download & URL generator endpoint

I also made some small modifications to  .gitignore / .dockerignore files that hopefully improve image build times and developer experience a little bit.

#### Issue
[ch153595](https://app.clubhouse.io/genepi/story/153595)

### Test plan

Tests are passing, and the aspen CLI utility is able to generate a URL that successfully loads samples into UShER in rdev

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -308,6 +308,9 @@ static
 /terraform/variables.tf
 .terraform
 
+# Jetbrains
+.idea
+
 # VS Code
 .vscode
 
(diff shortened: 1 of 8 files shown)


lvreynoso (approved):
lgtm

PR #640, opened 2021-08-19
Ondemand Tree builds v2

jgadling:
### Description

Support for ondemand tree builds v2.
- Import gisaid metadata into our db nightly
- PhyloRun start API accepts and validates gisaid ID's
- Phylo run script writes aspen & gisaid ID's to include.txt 

#### Issue
[ch153615](https://app.clubhouse.io/genepi/story/153615)

### Test plan


src/backend/database_migrations/versions/20210813_172344_add_gisaid_metadata_table.py
```diff
@@ -0,0 +1,35 @@
...
+    op.create_table(
+        "gisaid_metadata",
+        sa.Column("strain", sa.String(), nullable=False),
+        sa.Column("pango_lineage", sa.String(), nullable=True),
+        sa.Column("gisaid_clade", sa.String(), nullable=True),
+        sa.Column("date", sa.DateTime(), nullable=True),
+        sa.Column("region", sa.String(), nullable=True),
+        sa.Column("division", sa.String(), nullable=True),
+        sa.Column("location", sa.String(), nullable=True),
+        sa.Column("import_id", sa.String(), nullable=False),
```
lvreynoso, line 27: This `import_id` column was removed from the model, do we still need it in the migration?

src/backend/database_migrations/versions/20210818_233848_remove_import_id.py
```diff
@@ -0,0 +1,25 @@
...
+    op.drop_column(
+        "gisaid_metadata",
+        "import_id",
+        schema="aspen",
+    )
+
+
+def downgrade():
+    # Don't downgrade
+    pass
```
lvreynoso, line 25: We can remove this whole file if we remove the `import_id` column from the other migration.

lvreynoso (changes requested):
Looks good, just a question about the `import_id` column. Since I also wrote code in this PR and paired on it, would love for @shanaxel42 to also take a look.

src/backend/database_migrations/versions/20210813_172344_add_gisaid_metadata_table.py
jgadling: It's generally best not to modify db migration scripts once they've been published. Anyone who runs the migration in an earlier form won't have future changes applied to their DB and thus wind up out of sync. There are tradeoffs to be made here - performance issues, the likelihood of other db's having already applied a migration, etc.

src/backend/aspen/app/views/phylo_runs.py
```diff
@@ -58,13 +62,40 @@ def start_phylo_run():
...
+        response.headers["Content-Type"] = "application/json"
+        return response
 
+    # These are the sample ID's that don't match the aspen db
     missing_sample_ids = set(sample_ids) - found_sample_ids
+
+    # Store the list of matching ID's so we can stuff it in our jsonb column later
+    gisaid_ids = set()
+    # See if these missing samples match Gisaid ID's
+    gisaid_matches: Iterable[GisaidMetadata] = g.db_session.query(
```
danrlu, line 80: How long does such a query usually take against 3 million Gisaid records? Just curious

jgadling: The strain names are indexed, so it's O(nlogn) ... but 3m rows isn't much, these queries are very fast.

danrlu: LOL "but 3m rows isn't much"

danrlu:
I don't understand everything but tree part looks good to me! Exciting!

lvreynoso (approved):

PR #641, opened 2021-08-19
disable tree build button when greater than 2000 samples are selected

phoenixAja:
### Description

disable tree build button when checkedSamples length is greater than 2000

#### Issue
no issue

### Test plan

tested locally

### DIFF
--- a/src/frontend/src/common/components/library/data_subview/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/index.tsx
@@ -137,6 +137,7 @@ const DataSubview: FunctionComponent<Props> = ({
   const [showCheckboxes, setShowCheckboxes] = useState<boolean>(false);
   const [isDownloadModalOpen, setDownloadModalOpen] = useState(false);
   const [isDownloadDisabled, setDownloadDisabled] = useState<boolean>(true);
+  const [isCreateTreeDisabled, setCreateTreeDisabled] = useState<boolean>(true);
   const [failedSamples, setFailedSamples] = useState<any[]>([]);
   const [downloadFailed, setDownloadFailed] = useState<boolean>(false);
   const [isMetadataSelected, setMetadataSelected] = useState<boolean>(false);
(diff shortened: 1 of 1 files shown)


src/frontend/src/common/components/library/data_subview/index.tsx
```diff
@@ -259,7 +266,7 @@ const DataSubview: FunctionComponent<Props> = ({
           <Divider />
           <IconButton
             onClick={handleCreateTreeClickOpen}
-            disabled={isDownloadDisabled}
+            disabled={isCreateTreeDisabled}
```
mmmmmmaya, line 269: it seems like making this change causes `isDownloadDisabled` to not be read/used anywhere. is that intentional? If so, you may want to remove that state variable.

mmmmmmaya, line 269: it seems like making this change causes `isDownloadDisabled` to not be read/used anywhere. is that intentional? If so, you may want to remove that state variable.

phoenixAja: isDownloadDisabled is used in the following IconButton component for the DownloadModal

mmmmmmaya: I think I might be confused, but I don't see it there. Maybe we can pair real quick and sync on it ...

phoenixAja: it's on line 277 in this file?

mmmmmmaya: LOL idk what happened there. I searched in my editor and it wasnt coming up

mmmmmmaya (approved):

src/frontend/src/common/components/library/data_subview/index.tsx
phoenixAja: haha no worries at all!!

PR #644, opened 2021-08-24
Phoenix/fix tos

phoenixAja:
### Description

fix trigger to redirect new users to agreed to tos page. 

#### Issue
[ch157533](https://app.clubhouse.io/genepi/story/157533)

### Test plan

tested locally and in rdev (https://fix-tos-frontend.dev.genepi.czi.technology)

### DIFF
--- a/src/frontend/src/common/queries/auth.ts
+++ b/src/frontend/src/common/queries/auth.ts
@@ -1,4 +1,4 @@
-import { useRouter } from "next/router";
+import { router } from "next/router";
 import { useQuery, UseQueryResult } from "react-query";
 import ENV from "src/common/constants/ENV";
 import {
@@ -55,15 +55,14 @@ export function useUserInfo(): UseQueryResult<UserResponse, unknown> {
   });
 }
 
-export function useProtectedRoute(): UseQueryResult<UserResponse, unknown> {
-  const router = useRouter();
-  const result = useUserInfo();
 
+export function protectedRoute(result: UserResponse, router: router): void {
   const { isLoading, data } = result;
 
+  if (!data?.user?.agreedToTos) {
+    router.push(ROUTES.AGREE_TERMS);
+  }
   if (!isLoading && !data) {
     router.push(ROUTES.HOMEPAGE);
   }
-
-  return result;
 }
--- a/src/frontend/src/common/queries/auth.ts
+++ b/src/frontend/src/common/queries/auth.ts
@@ -61,13 +61,12 @@ export function useProtectedRoute(): UseQueryResult<UserResponse, unknown> {
 
   const { isLoading, data } = result;
   const agreedToTOS = data?.user?.agreedToTos;
-  console.log("agreedToTOS: ", agreedToTOS);
 
   if (!isLoading && !data) {
     router.push(ROUTES.HOMEPAGE);
   }
 
-  if (!isLoading && !agreedToTOS) {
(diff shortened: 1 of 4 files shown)


src/frontend/src/common/queries/auth.ts
```diff
@@ -55,15 +55,16 @@ export function useUserInfo(): UseQueryResult<UserResponse, unknown> {
...
 }
 
-export function useProtectedRoute(): UseQueryResult<UserResponse, unknown> {
-  const router = useRouter();
-  const result = useUserInfo();
-
+export function protectedRoute(
+  result: UseQueryResult<UserResponse, unknown>,
+  router: NextRouter
+): void {
```
phoenixAja: have to pass result and router in like so because we're wrapping this function in a useEffect hook (only way we can use next JS router to push to new route if not wrapped in button link (client side redirecting)). Or at least that was my understanding of this issue. Would be nice to run by @mmmmmmaya and @tihuan to verify.

phoenixAja:
converting to draft for now, sorry

phoenixAja:
to test this sign in as ventura, tulare, san bernadino or sfdph

phoenixAja:
for some reason this is making you agree to tos twice before redirecting you back to previous page, @mmmmmmaya or @tihuan do you happen to know what's going on?

src/frontend/src/views/Data/index.tsx
```diff
@@ -5,9 +5,10 @@ import { useRouter } from "next/router";
 import React, { FunctionComponent, useEffect, useState } from "react";
 import { Menu } from "semantic-ui-react";
 import { fetchSamples, fetchTrees } from "src/common/api";
-import { useProtectedRoute } from "src/common/queries/auth";
+import { protectRoute } from "src/common/queries/auth";
 import { DataSubview } from "../../common/components";
 import { EMPTY_OBJECT } from "../../common/constants/empty";
+import { useUserInfo } from "../../common/queries/auth";
```
mmmmmmaya: can you combine this with the import on line 8?

src/frontend/src/common/queries/auth.ts
mmmmmmaya: I'm not an expert here, so I would love to hear @tihuan's opinion

mmmmmmaya:
I pulled this down and I only see the TOS one time before being redirected to the data page (after clicking the button, ofc). I even tried setting `agreed_to_tos` and `hasAcceptedTerms` to `false` after clearing my cookies and I still only see the modal one time.

mmmmmmaya:
The front end cache does not update before you try to load `Data`, so the protected page check fails, and you get sent to agree to tos page again. After you click the button the first time, the db is updated, but the front end still thinks the user has not agreed to terms. 

You can test this by clicking the button the first time, then trying to visit the samples page again by clicking the aspen logo. It should load the page without showing terms, even though it did ask a second time.

Unfortunately, I'm not familiar with how react-query expires the FE cache. But that's where I would start looking.

phoenixAja:
thanks @mmmmmmaya that's helpful info, i'll resume this when i'm done addressing design feedback for the create tree modal

phoenixAja:
sorry closed because i thought timmy merged his branch into trunk 😆 , will merge once tests pass, thank you again for your help @tihuan and @mmmmmmaya you all are stars ⭐

mmmmmmaya (approved):

PR #647, opened 2021-08-27
thuang-fix-tos

tihuan:
### Description

Hi all!

I think the bug should be fixed in this PR 🙆‍♂️ 

@mmmmmmaya is right that we need to invalidate the userInfo query once we set `agreeTos` to `true`, otherwise React Query will return the cached response by default. SORRY FOR MISSING THAT 🤦 !

I made the following changes:

1. Add `useUpdateUserInfo()`, which is a tiny wrapper to use `useMutation` from `react-query`, so we can invalidate `userInfo` cache on success
2. `useProtectedRoute()` push route pushes inside `useEffect()`, since pushes are side effects and this helps preventing unintended route pushes if `useProtectedRoute()` is called by the parent component on re-render
3. Use `useUpdateUserInfo()` in `AgreeTerms` page to redirect only on `isSuccess`

More on React Query invalidation is available here:

https://react-query.tanstack.com/guides/invalidations-from-mutations

Thanks all!

#### Issue

### Test plan

1. Go to http://frontend.genepinet.local:8000/agreeTerms
2. Go to dev console network tab, and turn on `Preserve log`
2. Click on `agree`
2. Find and copy the CURL for `PUT userinfo` and change payload to `--data-raw '{"agreed_to_tos":false}' \` 
    Screen Shot 2021-08-27 at 11 22 12 AM
1. Now you can run the CURL in your terminal any time you want to reset the ``agreed_to_tos flag to false
2. Once it's false, refresh the ToS page and click agree again, it should now redirect as expected

### DIFF
--- a/src/frontend/src/common/queries/auth.ts
+++ b/src/frontend/src/common/queries/auth.ts
@@ -1,5 +1,12 @@
 import { useRouter } from "next/router";
-import { useQuery, UseQueryResult } from "react-query";
+import { useEffect } from "react";
+import {
+  useMutation,
+  UseMutationResult,
+  useQuery,
+  useQueryClient,
+  UseQueryResult,
+} from "react-query";
 import ENV from "src/common/constants/ENV";
 import {
   API,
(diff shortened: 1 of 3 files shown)


mmmmmmaya (approved):

phoenixAja:
thank you timmy!!!!!

phoenixAja (approved):

PR #649, opened 2021-08-27
fixing check for if public_identifier in request

phoenixAja:
### Description

fixing keyerror on check if public_identifier is included in request data

#### Issue
no issue, fixing sentry issue https://sentry.prod.si.czi.technology/genepi/aspen-backend/issues/120862/?referrer=slack

### Test plan
updated test

### DIFF
--- a/src/backend/aspen/app/views/sample.py
+++ b/src/backend/aspen/app/views/sample.py
@@ -353,10 +353,7 @@ def create_sample():
         )
         if data_ok:
             # GISAID Stuff
-            public_identifier = None
-            if "public_identifier" in data["sample"].keys():
-                public_identifier = data["sample"]["public_identifier"]
-            if public_identifier:
+            if data["sample"].get("public_identifier"):
                 # if they provided a public_id they marked true to "submitted to gisaid"
                 submitted_to_gisaid = True
                 public_identifier = data["sample"]["public_identifier"]
--- a/src/backend/aspen/app/views/sample.py
+++ b/src/backend/aspen/app/views/sample.py
@@ -353,7 +353,8 @@ def create_sample():
         )
         if data_ok:
             # GISAID Stuff
-            if data["sample"].get("public_identifier"):
+            public_identifier = data["sample"].get("public_identifier")
+            if public_identifier:
                 # if they provided a public_id they marked true to "submitted to gisaid"
                 submitted_to_gisaid = True
                 public_identifier = data["sample"]["public_identifier"]
--- a/src/backend/aspen/app/views/sample.py
+++ b/src/backend/aspen/app/views/sample.py
@@ -353,7 +353,10 @@ def create_sample():
         )
         if data_ok:
             # GISAID Stuff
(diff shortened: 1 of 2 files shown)


src/backend/aspen/app/views/sample.py
```diff
@@ -353,7 +353,10 @@ def create_sample():
         )
         if data_ok:
             # GISAID Stuff
-            if data["sample"]["public_identifier"]:
+            public_identifier = None
+            if "public_identifier" in data["sample"].keys():
+                public_identifier = data["sample"]["public_identifier"]
+            if public_identifier:
```
jgadling:
```suggestion
            if "public_identifier" in data["sample"]:
```

jgadling (changes requested):
I think this can be simpler

src/backend/aspen/app/views/sample.py
phoenixAja: i think this might not be what we want for the case when public_identifier="" . I think we want to protect against the case when public_identifier is not in data["sample"] and when public_identifier=""

vincent-czi, line 357: The intention is still to skip the check if `public_identifier` is false-y, right? Not just the None case of not having the key at all? That is the old test with `"public_identifier": ""` would still pass and _should_ still pass, right?

phoenixAja: yes i was thinking i would protect against both cases

vincent-czi: Yeah, makes sense

phoenixAja: good catch/ call out!

vincent-czi: How about 
```
public_identifier  = data["sample"].get("public_identifier")`
```
Then it will be one line but pull None when missing and still handle the empty string case?

phoenixAja: alright let's be fancy one-liners then 💃

phoenixAja: i totally forgot about using .get thanks for the reminder @vincent-czi

src/backend/aspen/app/views/sample.py
```diff
@@ -353,7 +353,7 @@ def create_sample():
         )
         if data_ok:
             # GISAID Stuff
-            if data["sample"]["public_identifier"]:
+            if data["sample"].get("public_identifier"):
```
vincent-czi: I'm still learning team style, but for what it's worth, I would default to do this in two lines.
```
public_identifier = data["sample"].get("public_identifier")
if public_identifier:
```
Even if it didn't mean you could avoid the assignment below in the if-block, it just feels clearer to me.

jgadling (approved):

src/backend/aspen/app/views/sample.py
phoenixAja: added

PR #650, opened 2021-08-30
Return failed and completed trees from API call

lvreynoso:
### Description

Adds `started_date`, `status`, and `workflow_id` to the `/api/phylo_trees` response. Also returns all trees regardless of completion.

Adds a small bit of frontend code to filter out trees that don't have status `COMPLETED`, so they don't show up in the tree table just yet.

#### Issue
[ch150143](https://app.clubhouse.io/genepi/story/150143)

### Test plan

Tests to be updated

### DIFF
--- a/src/backend/aspen/app/views/phylo_trees.py
+++ b/src/backend/aspen/app/views/phylo_trees.py
@@ -94,7 +94,7 @@ def phylo_trees():
             "completed_date": format_datetime(phylo_run.end_datetime),
             "status": phylo_run.workflow_status.value,
             "workflow_id": phylo_run.workflow_id,
-            "pathogen_genome_count": 0  # TODO: do we still need this?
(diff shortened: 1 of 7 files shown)


src/backend/aspen/app/views/phylo_trees.py
```diff
@@ -64,30 +70,33 @@ def phylo_trees():
...
+                    "phylo_tree_id": phylo_tree.entity_id,
+                    "name": phylo_tree.name or humanize_tree_name(phylo_tree.s3_key),
+                }
+        if not phylo_tree:
+            result = result | {
+                "phylo_tree_id": None,
+                "name": phylo_run.name or generate_tree_name_from_template(phylo_run),
             }
-        )
+        result = result | {
```
jgadling: nit: Can we make this the default result instead of {} on line 79 instead of merging it in here?

jgadling (approved):

PR #652, opened 2021-08-30
Standardize api exceptions

jgadling:
### Description

Creates a few standard exceptions that our API knows how to catch and apply a standard format for.
- All error responses will be JSON objects with at least an "error" key and optionally additional values from now on

### Test plan

CI tests


lvreynoso:
We have a stubbed `error` folder/module, could we add the code for exception classes to that folder? Then we could nicely import them like `from aspen.app.error import ServerException`.

phoenixAja (approved):
looks great, thanks for adding this

phoenixAja:
i like lucia's idea of adding exceptions to the error folder.

jgadling:
> i like lucia's idea of adding exceptions to the error folder.

Yep, I like it too. I'll move these!

jgadling:
@phoenixAja @lvreynoso 
I've made some changes, including incorporating feedback:
- Exceptions have been moved to the error module
- Use flask's built-in support for exception handling to return AspenError responses
- I removed some code blocks that were returning error Response objects instead of raising exceptions, and updated their callers to not worry about multiple possible types.

phoenixAja (approved):
looks good! I'll approve a second time

PR #656, opened 2021-09-01
Add tree type

lvreynoso:
### Description

Adds a column, `tree_type`, to both the `phylo_runs` and the `phylo_trees` table. The column was added to both tables because a phylo tree is not generated until a run completes. In order to provide the user information on what kind of tree is being generated for in progress or failed runs, we need to add the column to the run as well.

The column is a non-nullable enum (using enumtables) that is one of four string values: `OVERVIEW`, `TARGETED`, `NON_CONTEXTUALIZED`, and `UNKNOWN`. The database migration backfills existing runs and trees with their proper value, and uses `UNKNOWN` for runs and trees where that information cannot be extrapolated. This only seems to be the case for imported failed phylo runs.

API requests to `/api/phylo_runs` must now use one of the first three enum values for `tree_type` (`UNKNOWN` is not acceptable`).

For scheduled tree runs, the tree type is now an explicit argument to the WDL file.

#### Issue
[ch157534](https://app.clubhouse.io/genepi/story/157534)

### Test plan

Tests and test framework updated accordingly.


.happy/terraform/modules/sfn_config/nextstrain.wdl
```diff
@@ -65,6 +68,7 @@ task nextstrain_workflow {
                       --all-group-sequences                                                                                      \
                       --builds-template-file /usr/src/app/aspen/workflows/nextstrain_run/builds_templates/~{template_filename}   \
                       --builds-template-args '~{template_args}'
+                      --tree-type "~{tree_type}"
```
jgadling, line 71: does nextstrain-ondemand.wdl need to be modified as well?

src/backend/database_migrations/versions/20210901_134600_add_tree_type_column_to_run_and_tree.py
```diff
@@ -0,0 +1,110 @@
...
+    conn.execute(ancestors_to_overview_sql)
+    conn.execute(local_to_non_contextualized_sql)
+
+    # update the runs with phylo trees
+    copy_tree_tree_type_to_run_sql = sa.sql.text(
+        "UPDATE aspen.phylo_runs SET tree_type = subquery.tree_type FROM (SELECT producing_workflow_id, tree_type FROM aspen.entities INNER JOIN aspen.phylo_trees ON entities.id = phylo_trees.entity_id) AS subquery WHERE phylo_runs.workflow_id = subquery.producing_workflow_id"
+    )
+    # update remaining runs without phylo trees but with some information on tree type
+    assign_overview_to_group_plus_context_sql = sa.sql.text(
+        "UPDATE aspen.phylo_runs SET tree_type = 'OVERVIEW' WHERE phylo_runs.tree_type = 'UNKNOWN' AND phylo_runs.template_file_path LIKE '%group_plus_context.yaml%' OR phylo_runs.template_file_path LIKE '%contextual.yaml%';"
```
jgadling: I think these OR clauses need to be in parens: `condition1 AND (condition2 OR condition3)`

src/backend/database_migrations/versions/20210901_134600_add_tree_type_column_to_run_and_tree.py
```diff
@@ -0,0 +1,110 @@
...
+    )
+    assign_non_contextualized_to_group_sql = sa.sql.text(
+        "UPDATE aspen.phylo_runs SET tree_type = 'NON_CONTEXTUALIZED' WHERE phylo_runs.tree_type = 'UNKNOWN' AND phylo_runs.template_file_path LIKE '%group.yaml%' OR phylo_runs.template_file_path LIKE '%local.yaml%';"
```
jgadling: same - parens for OR

src/backend/aspen/database/models/phylo_tree.py
```diff
@@ -12,13 +15,34 @@
...
+    UNKNOWN = "UNKNOWN"  # For old data imported from COVIDHub. Should never show up otherwise.
+
+
+_TreeTypeTable = enumtables.EnumTable(
+    TreeType,
+    base,
+    tablename="tree_types",
+)
+
+TemplateTreeTypeMap = {
```
jgadling: I don't think this is used?

.happy/terraform/modules/sfn_config/nextstrain.wdl
jgadling: I think I answered my own question - this step is handled by our api endpoint for ondemand builds so we don't need to modify the other wdl.

jgadling (changes requested):
The OR conditions need to be fixed before we merge this.

jgadling (approved):

PR #658, opened 2021-09-02
Feat/filtering

phoenixAja:
### Description

Self serve tree building and interactive sample filtering

#### Issue

### Test plan

rdev: https://featfiltering-frontend.dev.genepi.czi.technology


Janeece:
@phoenixAja I can't get over how great the filtering is! Here's my comments:

### All Filters:
- [ ] [Should Have] When filter menu is open, filter dropdown component should be in active state instead of default (text color would be black, icon would be primary400)
- [ ] [Should Have] Text size for filter dropdowns should be updated to use font mixin fontCapsXxs (11px)
Filtering-text updates

### Upload Date:
- [ ] [Nice to Have] Upload date is only supposed to have buckets for “Today”, “Yesterday”, and “Last 7 Days”
- [ ] [Should Have] - When error message appears for incorrect date format, can we have the inputs remain aligned to the top? 
Filtering-error alignement
- [ ] [Must Have] - do not allow user to apply date range that is incorrectly formatted (disable button if format error is present)
Filtering-error applied
- [ ] [Nice to Have] - I think we should only show the error message after use clicks out of the date input field (so they dont get an error before they finish typing)
- [ ] [Must Have, not a blocker] - when user makes a selection from the menu items (non-date range), the filter tag text should match. This ticket already exists [here](https://app.shortcut.com/genepi/story/161475/update-filter-tag-label-to-reflect-selection)

### Collection Date:
- [ ] Same as Upload Date, except that the menu item options are correct

### Lineage:
beautiful!

### Genome Recovery:
- [ ] [Must Have, not a blocker]  When I make a selection on Genome Recovery and then reopen the menu, the current selection should be using the “selected” state (text becomes semibold). Note: this may relate to [this ticket](https://app.shortcut.com/genepi/story/161482/remove-autofill-date-range-from-upload-date-collection-date-menu) for upload/collection date as well
Filtering- active state

PR #659, opened 2021-09-07
Phoenix/fix sample count button

phoenixAja:
### Description
czifui's chip component has changed a bit, material ui's chip component seems to work better for sample counter (need to add less stylings to have it fit the design)

#### Issue
no issue

### Test plan

locally and [rdev](https://chip-fix-frontend.dev.genepi.czi.technology/data/samples)
to check out what it looked like using czifui component check [here](https://featfiltering-frontend.dev.genepi.czi.technology/):

### DIFF
--- a/src/frontend/src/common/components/library/data_subview/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/index.tsx
@@ -260,7 +260,7 @@ const DataSubview: FunctionComponent<Props> = ({
     if (viewName === VIEWNAME.SAMPLES && tableData !== undefined) {
       downloadButton = (
         <DownloadWrapper>
-          <StyledChip isRounded size="large" label={checkedSamples.length} status="info" />
+          <StyledChip isRounded label={checkedSamples.length} status="info" />
           <StyledDiv>Selected </StyledDiv>
           <Divider />
           <IconButton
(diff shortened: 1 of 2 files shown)


phoenixAja:
@tihuan let me know if this approach is ok or if we should revert back to using czifui chip component (i tried messing around with it's styling but couldn't _quite_ get it to look right)

vincent-czi (approved):
Looks good and checks out on the rdev.

vincent-czi:
I'm still getting a sense of how we do things here, but philosophically I feel like we should probably strive to use the czifui components long-term. The material-ui chip certainly works here and the code looks good, and I don't know what's involved in making changes to the czifui library, so this definitely looks reasonable. But I am worried about divergence over time if changes necessitating use of external libs happen often enough.

phoenixAja:
@tihuan regarding vincent's question do you think this chip behavior is a potential bug with czifui's chip component? Do you think we should coerce the czifui chip component stylings a bit more so that we could use that component over the material ui one?

tihuan (approved):
LGTM! Sorry again for the breaking change and thank you for fixing it in Aspen 🙏 ! We will formalize version control soon, so hopefully it won't happen again!

phoenixAja:
@vincent-czi i worked with timmy to use the czifui chip instead so crisis averted. Thank you @tihuan for the help!!!

PR #660, opened 2021-09-07
info toast and tree modal updates

phoenixAja:
### Description

* added some info toast to inform the user that tree creation may take awhile. 
* added toast message for if tree creation has failed (using same component as download failed alert)
* adjusted tree modal with new design updates

#### Issue
[ch150458](https://app.clubhouse.io/genepi/story/150458)
[ch157597](https://app.clubhouse.io/genepi/story/157597)

### Test plan

tested locally and on https://info-toast-frontend.dev.genepi.czi.technology/data/samples

### DIFF
--- a/src/frontend/src/common/components/library/data_subview/components/AfterModalAlert/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/AfterModalAlert/index.tsx
@@ -1,0 +1,30 @@
+import React from "react";
+import { BoldText, DismissButton, StyledAlert, StyledDiv } from "./style";
+
+interface Props {
+  alertClassName: string;
+  alertSeverity: "error" | "info" | "success" | "warning";
+  boldText?: string;
+  lightText?: JSX.Element | string;
+  handleDismiss?: () => void;
+}
+
+export const AfterModalAlert = ({
+  alertClassName,
+  alertSeverity,
+  boldText,
+  lightText,
+  handleDismiss,
+}: Props): JSX.Element => {
+  return (
+    <StyledAlert className={alertClassName} severity={alertSeverity}>
+      <StyledDiv>
+        {boldText && <BoldText>{boldText}</BoldText>}
+        {lightText}
+      </StyledDiv>
+      {handleDismiss && (
+        <DismissButton onClick={handleDismiss}>DISMISS</DismissButton>
+      )}
+    </StyledAlert>
+  );
+};
--- a/src/frontend/src/common/components/library/data_subview/components/createTreeModal/components/RadioLabel/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/createTreeModal/components/RadioLabel/index.tsx
@@ -35,7 +35,7 @@ export const RadioLabelTargeted = ({ selected }: Props): JSX.Element => {
               <StyledIconCheckSmall />
             </StyledListItemIcon>
             <ListItemText>
-              <SmallText>Best for local outbreak investigation. </SmallText>
+              <SmallText>Best for local outbreak investigation.</SmallText>
             </ListItemText>
           </StyledListItem>
           <StyledListItem button={false as any}>
--- a/src/frontend/src/common/components/library/data_subview/components/createTreeModal/components/RadioLabel/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/createTreeModal/components/RadioLabel/index.tsx
@@ -17,13 +17,17 @@ interface Props {
   selected: boolean;
 }
 
-export const RadioLabelContextual = ({ selected }: Props): JSX.Element => {
+export const RadioLabelTargeted = ({ selected }: Props): JSX.Element => {
   return (
     <div>
       <Label>
-        <LabelMain>My county and other regions </LabelMain>
+        <LabelMain>Targeted </LabelMain>
         <LabelLight>— Recommended</LabelLight>
       </Label>
+      <SmallText>
+        Includes additional publicly-available samples on GISAID from your
+        jurisdiction and other regions.
+      </SmallText>
       {selected && (
         <List>
           <StyledListItem button={false as any}>
@@ -31,7 +35,7 @@ export const RadioLabelContextual = ({ selected }: Props): JSX.Element => {
               <StyledIconCheckSmall />
             </StyledListItemIcon>
             <ListItemText>
-              <SmallText>Best for local outbreak investigation.</SmallText>
+              <SmallText>Best for local outbreak investigation. </SmallText>
             </ListItemText>
           </StyledListItem>
           <StyledListItem button={false as any}>
@@ -40,7 +44,7 @@ export const RadioLabelContextual = ({ selected }: Props): JSX.Element => {
             </StyledListItemIcon>
             <ListItemText>
               <SmallText>
-                Best for seeing relationships between your selected samples,
+                Useful for seeing relationships between your selected samples,
                 other samples you have uploaded to Aspen, and publicly-available
                 samples on GISAID.
               </SmallText>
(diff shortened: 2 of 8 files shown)


src/frontend/src/common/components/library/data_subview/components/AfterModalAlert/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export const AfterModalAlert = ({
+  alertClassName,
+  alertSeverity,
+  boldText,
+  lightText,
+  handleDismiss,
+}: Props): JSX.Element => {
+  return (
+    <StyledAlert className={alertClassName} severity={alertSeverity}>
```
vincent-czi, line 20: This is more a question for my understanding: the `className={alertClassName}` is adding a `class` to the underlying HTML tag with value of whatever the passed `alertClassName` is, right? Do we have matching CSS rules that get imported elsewhere at a high-level to let that class be used (eg, `"elevated"` like I see get used later)? .... ahhhh, nevermind, just found it, it's happening via the "elevated" class inside of czifui. Hmmm, not sure I love that -- it seems to mix the styling approaches. In other czifui components (like `ComplexFilter`), I see styles as being put in because a prop gets passed which then conditionally triggers interpolating a chunk of CSS via emotion/styled. Having multiple approaches here feels like it muddies the water. I also don't see any type statements about `elevated` down in the czifui. The underlying MaterialUI has types of `'success' | 'info' | 'warning' | 'error'` for its Alert, but because of the way this is done through class name, we lose knowledge of the new `elevated` type. Hmmm. I'll try to remember this and chat with Timmy about it. I think new stuff should generally avoid using class name and stick with emotion/styled, but I could definitely be convinced otherwise. Regardless, losing the type of `elevated` seems like a problem.

Janeece:
@phoenixAja this looks fantastic! Only 3 comments:

**Tree Build Dialog**
- [ ] Update tooltip link to open doc in a new tab
- [x] Remove the second line of copy from below the create button "We will notify you when your tree is ready"
TreeBuildv1 - dialog copy update

**Toast Notification:**
- [x] When the notification is present, the selection count and sample action buttons (tree + download) are missing. They appear again once the notification starts to fade. I almost didnt notice, but the page content jumps down a little when they reappear. I think these should always be there, even if they are temporarily covered. 
Screen Shot 2021-09-10 at 10 39 13 AM

src/frontend/src/common/components/library/data_subview/components/createTreeModal/components/RadioLabel/index.tsx
```diff
@@ -17,21 +17,25 @@ interface Props {
...
+      </SmallText>
       {selected && (
         <List>
           <StyledListItem button={false as any}>
             <StyledListItemIcon>
               <StyledIconCheckSmall />
             </StyledListItemIcon>
             <ListItemText>
-              <SmallText>Best for local outbreak investigation.</SmallText>
+              <SmallText>Best for local outbreak investigation. </SmallText>
```
vincent-czi: Do we want the extra space at the end here? (incredibly minor as it is)

phoenixAja:
good catch on that shift janeece i didn't even notice that was happening

phoenixAja:
will hopefully be ready for re-review in the hour

src/frontend/src/common/components/library/data_subview/components/createTreeModal/index.tsx
```diff
@@ -40,33 +43,35 @@ interface Props {
...
 }: Props): JSX.Element => {
   const [treeName, setTreeName] = useState<string>("");
   const [isTreeNameTooLong, setTreeNameTooLong] = useState<boolean>(false);
   const [isTreeBuildDisabled, setTreeBuildDisabled] = useState<boolean>(false);
-  const [treeType, setTreeType] = useState<string>("");
-  const [isContextual, setContextual] = useState<boolean>(true);
-  const [isLocal, setLocal] = useState<boolean>(false);
+  const [treeType, setTreeType] = useState<string>("TARGETED");
+  const [isTargeted, setTargeted] = useState<boolean>(true);
+  const [isNonContextualized, setNoneContextualized] = useState<boolean>(false);
```
vincent-czi: I think this is a typo: `setNoneContextualized` vs `setNonContextualized` (and then match the `isNonContextualized`). Code will work fine as is, but probably nice to bring it in to alignment.

src/frontend/src/common/components/library/data_subview/components/createTreeModal/index.tsx
```diff
@@ -40,33 +43,35 @@ interface Props {
...
+  handleCreateTreeFailed,
+  handleSetCreateTreeStarted,
 }: Props): JSX.Element => {
   const [treeName, setTreeName] = useState<string>("");
   const [isTreeNameTooLong, setTreeNameTooLong] = useState<boolean>(false);
   const [isTreeBuildDisabled, setTreeBuildDisabled] = useState<boolean>(false);
-  const [treeType, setTreeType] = useState<string>("");
-  const [isContextual, setContextual] = useState<boolean>(true);
-  const [isLocal, setLocal] = useState<boolean>(false);
+  const [treeType, setTreeType] = useState<string>("TARGETED");
```
vincent-czi: I'm unclear on why we need a `treeType` and then also bools of `isTargeted` and `isNonContextualized` since those are just other ways to tell us the tree type, right? It's somewhat a matter of taste, but I would prefer to see tiny helper funcs get defined if we absolutely need the bools. Things like
```isTreeTargeted(treeType) --> true/false```

However, where I'm seeing those bools below, it's currently like
```checked={isTargeted}```
and I think that could just turn into
```checked={treeType === "TARGETED"} // or better yet, === TREE_TYPES.TARGETED, see my other comment```

I'm being nit-picky here, but I'm always worried when I see multiple sources of truth. It's not happening now, but if `treeType` gets out of sync with its various interpreted bools, we could get weird bugs.

vincent-czi: Also, this seems like a candidate for more fixed Types. Not just `string`, but a union of literal string types
`"TARGETED" | "NON_CONTEXTUALIZED" | ...etc...`
That way we have a clear spot in the code of what the allowed tree types are and we'll get TS type checking if there's ever a typo.

Could also go farther and define a fixed object of the allowed values, make a `TREE_TYPES` string enum (although I don't think I like enums generally [that said, I've only read them in the TS handbook], I think string enums are useful for locking in a fixed set of options that need to be referenced in various places -- I did similar stuff a lot in past work with constant objects when we needed fixed strings for complicated switch-case stuff).

vincent-czi (approved):
Looks good! I've got some minor suggestions in comments, but nothing terribly important.

src/frontend/src/common/components/library/data_subview/components/createTreeModal/index.tsx
phoenixAja: oh no that is a typo! good catch

src/frontend/src/common/components/library/data_subview/components/createTreeModal/index.tsx
phoenixAja: i agree, i'll change it to keep the single source of truth

phoenixAja: Yeah i was thinking about making an enum for this but i actually like the idea of creating a custom type instead. I'll add that

src/frontend/src/common/components/library/data_subview/components/createTreeModal/index.tsx
phoenixAja: i removed this as feedback to one of your other comments

src/frontend/src/common/components/library/data_subview/components/createTreeModal/index.tsx
phoenixAja: removed these vars

vincent-czi (approved):

Janeece:
The copy and toast notifications look great!

PR #667, opened 2021-09-14
Fix ondemand yamls

jgadling:
### Description

Fix the mapping of nextstrain templates to UI tree type names by renaming the files and using their new names. Also don't support on-demand builds of overview trees

#### Issue
[ch160863](https://app.clubhouse.io/genepi/story/160863)

### Test plan

CI


phoenixAja (approved):
TY JESSS!!

PR #669, opened 2021-09-15
[ch160837] filter state panel toggle

mmmmmmaya:
### Summary
- **What:** Don't remove filter state when toggling the filter panel
- **Why:** Preserve the work the user did to filter the samples
- **Ticket:** [[ch160837]](https://app.clubhouse.io/genepi/story/160837)
- **Env:** https://savefilterstate-frontend.dev.genepi.czi.technology

### Testing
1. Filter your samples
1. Choose some samples
2. Toggle filter panel
3. Verify the exact same set of samples are shown
4. Toggle panel again
5. Verify the same samples are still shown
6. Verify chips are still displayed under applied filters

### Demos
after

### Checklist
- [x] I merged latest `feat/filtering`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/components/FilterPanel/index.tsx
+++ b/src/frontend/src/components/FilterPanel/index.tsx
@@ -213,10 +213,8 @@ const FilterPanel: FC<Props> = ({
     }
   };
 
-  if (!isOpen) return null;
-
   return (
-    <StyledFilterPanel>
+    <StyledFilterPanel isOpen={isOpen}>
       <CollectionDateFilter
         updateCollectionDateFilter={updateCollectionDateFilter}
       />
--- a/src/frontend/src/components/FilterPanel/index.tsx
+++ b/src/frontend/src/components/FilterPanel/index.tsx
@@ -23,6 +23,7 @@ export interface DefaultMenuSelectOption {
 }
 
 interface Props {
+  isOpen: boolean;
   lineages: DefaultMenuSelectOption[];
   setActiveFilterCount: (count: number) => void;
   setDataFilterFunc: Dispatch<
@@ -122,6 +123,7 @@ const applyFilter = (data: TableItem[], dataFilter: FilterType) => {
 };
 
 const FilterPanel: FC<Props> = ({
+  isOpen,
   lineages,
   setActiveFilterCount,
   setDataFilterFunc,
@@ -210,6 +212,8 @@ const FilterPanel: FC<Props> = ({
       updateDataFilter("CZBFailedGenomeRecovery", { selected });
     }
   };
+
+  if (!isOpen) return null;
 
   return (
     <StyledFilterPanel>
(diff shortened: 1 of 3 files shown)


src/frontend/src/components/FilterPanel/style.ts
```diff
@@ -1,12 +1,23 @@
 import styled from "@emotion/styled";
-import { Chip, getColors, getSpacings, InputDropdown } from "czifui";
+import { Chip, getColors, getSpacings, InputDropdown, Props } from "czifui";
```
vincent-czi, line 2: I was taken aback for a moment by czifui exporting something just called `Props`, but I've been trying to work through it for a bit... Am I right in thinking that pretty much every single Props in Aspen has to have that `theme` from the czifui `Props` inside of it, because of how the theme provider works? And so it's reasonable for any/every Aspen component to extend off that czifui `Props`?

I'm really just talking out loud here, but please tell me if I'm wrong or right. I'm trying to get my head around when I should be extending off that czifui `Props` or just creating a new Props just for a component I'm working on and not worrying about extending.

src/frontend/src/components/FilterPanel/index.tsx
```diff
@@ -212,7 +214,7 @@ const FilterPanel: FC<Props> = ({
   };
 
   return (
-    <StyledFilterPanel>
+    <StyledFilterPanel isOpen={isOpen}>
```
vincent-czi, line 217: Doing this via the `display: ${isOpen ? "block" : "none"};` is totally reasonable, but I have a slight preference for instead just making the return short-circuit to `null` for the entire FilterPanel when `isOpen` is false. It's probably six-in-one, half-dozen in the other, but that way if there are any unexpected side-effects in a downstream child, they can't fire off while still being `display: none` hidden. That said, I didn't fully trace the filtering process while I was working on it: if those child components not "rendering" in a hidden way will drop the filter state, I think the display none approach is definitely the right way.

vincent-czi (approved):
I've got a minor comment, but you can take it or leave it if you disagree. Looks good!

src/frontend/src/components/FilterPanel/index.tsx
mmmmmmaya: I also actually have a preference for that, but then we just push the problem of state disappearing down a level. If I return `null` instead of hiding the panel, we lose the chip state even though the samples themselves stay filtered.

vincent-czi: Whoops, yeah. I had forgotten about the state in those lower levels, yup, definitely wouldn't work.

src/frontend/src/components/FilterPanel/style.ts
mmmmmmaya: TL;DR: Essentially, yes. Most components we style already have the correct set of types because of type inference on the part of TS. It is only when you need to pass a custom prop to a component you plan to style that you actually need to define the type which extends the basic `Prop` set. 

--

For example, if I try to omit this type definition for `StyledFilterPanel` altogether, I end up passing `isOpen` to a styled div. `isOpen` is not a known valid prop for a div, so I need to tell TS that I _meant_ to do that (so that I can use it for styling purposes), and then not forward the prop on to the div that gets rendered in the page. However, if I only define the type as having one prop `isOpen`, then TS complains because the props passed into the string interpolation in emotion's `styled` don't match what emotion provides (something of type `Props`, by default).

PR #670, opened 2021-09-15
Fix selections in multi-select for lineage filtering 

vincent-czi:
### Description

**Primary purpose** of PR is to fix issue with multi-select in the Lineage filter. Previously, selecting a set of lineages worked, but once the dropdown was closed, those selections would be "forgotten" by the dropdown on next opening. The filtering still worked, but the dropdown didn't know what was previously selected, causing it to not show prior selections and also allow repeated, duplicate selections (eg, selecting lineage `A.1` multiple times).

Additionally
- Trying to fix this at first, we noticed that czifui was on a different Material-UI version than aspen. It was thought that might be the source of the bug early on. It wasn't, but aspen should probably be using the same M-UI version as czifui regardless.
- Did a fresh install of node packages to get the M-UI versions matched.
- This also caused a new TS version bump. Seems to be no major changes, just one tiny tweak for things to be happy, which was...
- ...New TS version handles errors slightly different in terms of type. This caused an issue with one test, but tossing on some very reasonable Type narrowing fixed that issue.

If the above additional stuff causes any problems or is a bridge too far, this PR can be cut down to just cherry-picking the single commit `38501a9`: that's the one where the primary purpose of the PR happens. The rest of it is just the saga of doing an M-UI package update. But I think that's all fine to roll in here.

#### Issue
[ch<160830>](https://app.clubhouse.io/genepi/story/<160830>)

### Test plan

Tested locally, although I was only really focused on the lineage multi-select with my testing.
~Currently deploying an rdev~ [rdev deployed at `filter-fix-multi-select`](https://filter-fix-multi-select-frontend.dev.genepi.czi.technology/). Considering the changes to underlying TS and M-UI, it would be nice to do a bit more regression testing against the rdev. I'm not really sure how to do that since I'm not super familiar with the app still (and I'm about to drop off for next day and a half). **If someone else can prod the rdev and make sure it doesn't look weird, that would be appreciated greatly.**

### DIFF
--- a/src/frontend/src/components/FilterPanel/components/LineageFilter/index.tsx
+++ b/src/frontend/src/components/FilterPanel/components/LineageFilter/index.tsx
@@ -1,6 +1,5 @@
-import { ComplexFilter } from "czifui";
+import { ComplexFilter, ComplexFilterValue, DefaultMenuSelectOption } from "czifui";
 import React, { useEffect, useState } from "react";
-import { DefaultMenuSelectOption } from "../../index";
 
 interface Props {
   options: DefaultMenuSelectOption[];
@@ -15,11 +14,11 @@ const LineageFilter = (props: Props): JSX.Element => {
   // and null as possible input args. But if we do that, then `value` is implied to possibly
   // be those, which it can't be because of required `.map` below.
   // Not sure how to go about fixing this. Probably some forced assertion? But where?
-  const [value, setValue] = useState<DefaultMenuSelectOption[]>([]);
+  const [value, setValue] = useState<ComplexFilterValue>([]);
 
   useEffect(() => {
     if (value) {
-      updateLineageFilter(value.map((d: DefaultMenuSelectOption) => d.name));
+      updateLineageFilter((value as DefaultMenuSelectOption[]).map((d) => d.name));
     }
   }, [updateLineageFilter, value]);
 
@@ -45,7 +44,6 @@ const LineageFilter = (props: Props): JSX.Element => {
     <ComplexFilter
       label="Lineage"
       options={options}
-      /* @ts-expect-error -- TODO look into czifui ComplexFilter type for onChange*/
       onChange={setValue}
       multiple
       search
--- a/src/frontend/src/components/FilterPanel/components/LineageFilter/index.tsx
+++ b/src/frontend/src/components/FilterPanel/components/LineageFilter/index.tsx
@@ -37,10 +37,12 @@ const LineageFilter = (props: Props): JSX.Element => {
     }
   }, [updateLineageFilter, value]);
 
+  // TODO -- With czifui 0.0.55, the `InputDropdownComponent` was exposed so it
+  // can be directly styled and passed in as we do with others. Once we've upgraded
+  // to 0.0.55+, this should be swapped over so it matches the approach we do elsewhere.
   // (vince) To tweak the internal style of dropdown in ComplexFilter, need to
   // create a specialized set of props to insert CSS via raw `style` put into
-  // underlying HTML tag.
-  // TODO -- This approach is very different. Talk to SDS team about ways to avoid.
+  // underlying HTML tag. [As of czifui 0.0.53. See above for newer.]
   const PROPS_FOR_INPUT_DROPDOWN = {
     sdsStyle: "minimal", // Would be defaulted, but must set everything now.
     // This `style` gets directly put in as HTML style and interpolated to CSS.
--- a/src/frontend/src/components/FilterPanel/components/LineageFilter/index.tsx
+++ b/src/frontend/src/components/FilterPanel/components/LineageFilter/index.tsx
@@ -1,4 +1,8 @@
-import { ComplexFilter, ComplexFilterValue, DefaultMenuSelectOption } from "czifui";
+import {
+  ComplexFilter,
+  ComplexFilterValue,
+  DefaultMenuSelectOption,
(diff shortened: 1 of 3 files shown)


phoenixAja:
thanks vincent! I can do some regression testing with your rdev and check to see that existing functionality still works/behaves as expected 👍

phoenixAja:
I ran into some issues with viewing a tree in auspice/ downloading (seems to be permission errors in the backend, but this is not an issue on feat/filtering branch) let me cherry pick `38501a9` to see if that fixes the issue

phoenixAja:
verified with @jgadling that this is safe to merge from a regression testing standpoint

phoenixAja (approved):
this looks good to me, thanks for adding all the helpful comments/todos, and for patching some additional issues in this pr.

src/frontend/src/components/FilterPanel/components/LineageFilter/index.tsx
```diff
@@ -1,32 +1,48 @@
...
+// Instead, we must create a custom selection checker to check underlying value.
+const getOptionSelected = (
+  option: DefaultMenuSelectOption,
+  value: DefaultMenuSelectOption
+) => {
+  return option.name === value.name;
+};
+// ComplexFilter doesn't directly do the check, it's done by its child MenuSelect
+const optionCheckingMenuSelectProps = {
+  getOptionSelected: getOptionSelected,
```
tihuan: nit
```suggestion
  getOptionSelected,
```

src/frontend/src/components/FilterPanel/components/LineageFilter/index.tsx
```diff
@@ -1,32 +1,48 @@
...
+// equality is always false because they're not the same object, even if same content.
+// Instead, we must create a custom selection checker to check underlying value.
+const getOptionSelected = (
+  option: DefaultMenuSelectOption,
+  value: DefaultMenuSelectOption
+) => {
+  return option.name === value.name;
+};
+// ComplexFilter doesn't directly do the check, it's done by its child MenuSelect
+const optionCheckingMenuSelectProps = {
```
tihuan, line 24: nit: I was thinking the variable name could possibly be just `MenuSelectProps`, since that's what the object is, and also means if we add more props to the object we don't need to rename the variable from `optionCheckingMenuSelectProps` to `optionCheckingAndFooMenuSelectProps`, etc..

What do you think?

src/frontend/tests/features/utils/helpers.ts
```diff
@@ -52,9 +52,11 @@ export async function tryUntil(
 
       break;
     } catch (error) {
-      retry += 1;
-      savedError = error;
-      await page.waitForTimeout(WAIT_FOR_MS);
+      if (error instanceof Error) {
```
tihuan, line 55: for my own education, what prompted the need to add this if condition? Thank you!

tihuan (commented):
Looks super awesome! Love the detailed PR description too 🙏 Thanks so much, @vincent-czi 🎉 !!

src/frontend/src/components/FilterPanel/components/LineageFilter/index.tsx
vincent-czi: So I generally do that kind of naming, here's why I didn't do it here (although I actually was waffling on the name when I wrote this code):

- I wanted to be explicit about the purpose of passing these props. They're just there to target one aspect -- the option equality checking -- and I wanted that to be super clear. But your point is well taken: if we ever had to go back and add more MenuSelectProps, it would necessitate either changing the name to something absurdly specific or just generalizing it back to MenuSelectProps. Really, if someone is going to go changing it, they'll go read where it's defined and see what's up, so yeah, I've fully come around to your POV on this aspect.
- The other thing, and it's a little silly, but it gave me a way to avoid PascalCase naming. Since it's an object, I wanted to keep it in camelCase, and by tacking some words on the front explaining what the props were doing, it gave me a way to have my cake and eat it too.

Since I no longer agree with my first point after thinking about what you said, I'm on board with naming it `MenuSelectProps`, but I do kind of miss keeping it camelCase since it's an object.

src/frontend/tests/features/utils/helpers.ts
vincent-czi: So, with my new node package install, I wound up pulling in a more recent version of TS. They changed the implied Type of errors to `unknown`: https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables

Apparently -- I didn't know this before -- you can throw anything as an "error" in JavaScript. They used to catch them as `any`, but with the new version they changed it to `unknown`. Because `savedError` is explicitly an Error where it first gets created (above the try/catch), the `savedError = error` was failing because it was assigning an `unknown` now with the new TS version. So I put in a bit of type narrowing to guarantee to TS that it was in fact an Error.

More generally, this made TS happy here, but I think it's probably not The One True Pattern. I'm still thinking about it, but I feel like the way to go generally is
```
try {
  ...
} catch(err) {
  if (err instanceof Error) { // Or whatever kind of precise thing you should be catching
    ...
  } else { // Wait, it wasn't an Error we just caught? What the heck is it?
    throw err; // If it's not the Error we expected, we should probably blow up now.
  }
}
```

tihuan: Ahh thank you for the information! That makes sense 🎉

PR #673, opened 2021-09-15
Add tree type column to phylo trees view

jgadling:
### Description

Add a tree type column to the phylo trees data table

#### Issue
[ch158247](https://app.clubhouse.io/genepi/story/158247)

### Test plan

Tested in local dev and it looks good, I'll add a link to the rdev site once I have it.

### DIFF
--- a/src/frontend/src/common/components/library/data_table/style.ts
+++ b/src/frontend/src/common/components/library/data_table/style.ts
@@ -31,7 +31,7 @@ export const RowContent = styled("div", {
 
   ${(props: ExtraProps) => {
     const { header } = props;
-    const align = header && header.align !== undefined ? header.align : "left";
+    const align = header?.align ?? "left";
     const spacings = getSpacings(props);
 
     return `
--- a/src/frontend/src/common/components/library/data_table/style.ts
+++ b/src/frontend/src/common/components/library/data_table/style.ts
@@ -16,15 +16,26 @@ export const TableRow = styled.div`
   }}
 `;
 
-export const RowContent = styled.div`
+export interface ExtraProps extends Props {
+  header?: string;
+}
+
+const doNotForwardProps = ["header"];
+
+export const RowContent = styled("div", {
+  shouldForwardProp: (prop) => !doNotForwardProps.includes(prop as string),
+})`
   display: flex;
   align-items: center;
   width: 100%;
 
-  ${(props) => {
+  ${(props: ExtraProps) => {
+    const { header } = props;
+    const shouldCenter = header === "treeType";
     const spacings = getSpacings(props);
 
     return `
+      justify-content: ${shouldCenter ? "center" : "left"};
       padding: ${spacings?.l}px 0;
       margin-right: ${spacings?.m}px;
   `;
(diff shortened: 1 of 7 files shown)


src/frontend/src/common/components/library/data_table/style.ts
```diff
@@ -73,3 +84,20 @@ export const HeaderCheckbox = styled(Checkbox)`
...
+export interface AlignProps extends Props {
+  align?: string;
+}
+
+export const TableHeader = styled("div", {
+  shouldForwardProp: (prop) => !doNotForwardProps.includes(prop as string),
+})`
+  ${(props: AlignProps) => {
+    const { align } = props;
+    const justify = align !== undefined ? align : "left";
```
mmmmmmaya:
```suggestion
    const justify = align ?? "left";
```

mmmmmmaya: This is the TS way of saying "set the value to variable, but if it doesn't exist use this value instead"

src/frontend/src/common/components/library/data_table/style.ts
```diff
@@ -16,15 +16,26 @@ export const TableRow = styled.div`
...
+  shouldForwardProp: (prop) => !doNotForwardProps.includes(prop as string),
+})`
   display: flex;
   align-items: center;
   width: 100%;
 
-  ${(props) => {
+  ${(props: ExtraProps) => {
+    const { header } = props;
+    const align = header && header.align !== undefined ? header.align : "left";
```
mmmmmmaya:
```suggestion
    const align = header?.align ?? "left";
```

src/frontend/src/common/components/library/data_table/style.ts
```diff
@@ -73,3 +84,20 @@ export const HeaderCheckbox = styled(Checkbox)`
     `;
   }}
 `;
+
+export interface AlignProps extends Props {
+  align?: string;
+}
+
+export const TableHeader = styled("div", {
+  shouldForwardProp: (prop) => !doNotForwardProps.includes(prop as string),
```
phoenixAja: do you still need this line? it looks like TableHeader doesn't have a `header` prop?

phoenixAja:
@jgadling awesome work this is really exciting!!! do you have the rdev link yet? Also @mmmmmmaya does this PR fix the table alignment issue on narrow screens we were looking at earlier?

jgadling:
> @jgadling awesome work this is really exciting!!! do you have the rdev link yet? Also @mmmmmmaya does this PR fix the table alignment issue on narrow screens we were looking at earlier?

Rdev link: https://treetype-frontend.dev.genepi.czi.technology

mmmmmmaya (approved):

mmmmmmaya:
@phoenixAja Fortunately the trees table does not suffer from this as much as the sample table (in fact, I could not see the problem at all on the trees table, even on small screens). So I'm not sure if it is fixed or not, but it will give us some ideas for the future when we do attempt to address this problem on the samples table

PR #676, opened 2021-09-17
adding cli utility to update public_identifier

phoenixAja:
### Description

add cli utility to update public_identifiers from csv mapping of private -> new public id

#### Issue
[ch161375](https://app.clubhouse.io/genepi/story/161375)

### Test plan

wrote some tests

`python src/cli/aspencli.py --env local samples update_public_ids --group-id 1 --private-to-public-id-mapping ~/Downloads/test_rename_public_identifiers.csv ` (csv attached)

[test_rename_public_identifiers.csv](https://github.com/chanzuckerberg/aspen/files/7206074/test_rename_public_identifiers.csv)

```
make local-pgconsole
select private_identifier, public_identifier from samples limit 5;
```
public IDs should be updated:
```
 private_identifier |  public_identifier
--------------------+----------------------
 OCPHL118           | USA/CA-CZB-1880/2020/updated
 OCPHL117           | USA/CA-CZB-1879/2020/updated
 OCPHL116           | USA/CA-CZB-1878/2020/updated
 OCPHL115           | USA/CA-CZB-1877/2020/updated
 OCPHL113           | USA/CA-CZB-1875/2020/updated
```

### DIFF
--- a/src/backend/aspen/app/views/sample.py
+++ b/src/backend/aspen/app/views/sample.py
@@ -426,7 +426,7 @@ def update_sample_public_ids():
 
     if not user.system_admin:
         raise ex.BadRequestException(
-            f"user making update request must be a system admin",
+            "user making update request must be a system admin",
         )
 
     request_private_ids: list[str] = request_data["private_ids"]
--- a/src/backend/aspen/app/views/sample.py
+++ b/src/backend/aspen/app/views/sample.py
@@ -436,11 +436,7 @@ def update_sample_public_ids():
         for (private, public) in zip(request_private_ids, request_public_ids)
     }
 
-    group_id: int = request_data.get("group_id")
-
-    # group id not supplied (when admin is making request), use the user.group_id instead
-    if not group_id:
-        group_id = user.group_id
+    group_id: int = request_data["group_id"]
 
     # check that all private_identifiers exist
     existing_private_ids: list[str] = api_utils.get_existing_private_ids(
--- a/src/backend/aspen/app/views/sample.py
+++ b/src/backend/aspen/app/views/sample.py
@@ -424,6 +424,11 @@ def update_sample_public_ids():
     user: User = g.auth_user
     request_data = request.get_json()
 
+    if not user.system_admin:
+        raise ex.BadRequestException(
+            f"user making update request must be a system admin",
+        )
+
     request_private_ids: list[str] = request_data["private_ids"]
     request_public_ids: list[str] = request_data["public_ids"]
     private_to_public: Mapping[str, str] = {
--- a/src/backend/aspen/app/views/sample.py
+++ b/src/backend/aspen/app/views/sample.py
@@ -437,7 +437,8 @@ def update_sample_public_ids():
     }
 
     group_id: int = request_data.get("group_id")
-    # group id not supplied, use user group_id
+
+    # group id not supplied (when admin is making request), use the user.group_id instead
     if not group_id:
         group_id = user.group_id
 
--- a/src/backend/aspen/app/views/sample.py
+++ b/src/backend/aspen/app/views/sample.py
@@ -429,12 +429,9 @@ def update_sample_public_ids():
             "user making update request must be a system admin",
         )
 
-    request_private_ids: list[str] = request_data["private_ids"]
-    request_public_ids: list[str] = request_data["public_ids"]
-    private_to_public: Mapping[str, str] = {
-        private: public
-        for (private, public) in zip(request_private_ids, request_public_ids)
-    }
+    private_to_public: Mapping[str, str] = request_data["id_mapping"]
+    request_private_ids: list[str] = private_to_public.keys()
+    request_public_ids: list[str] = private_to_public.values()
 
     group_id: int = request_data["group_id"]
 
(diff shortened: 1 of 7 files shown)


phoenixAja:
drafting, making some quick changes

phoenixAja:
ready for review now (hopefully this implementation will make you not as upset with me @jgadling 😆 )

src/backend/aspen/app/views/sample.py
```diff
@@ -416,3 +416,60 @@ def create_sample():
     except Exception as e:
         raise ex.BadRequestException(f"Error encountered when saving data: {e}")
     return jsonify(success=True)
+
+
+@application.route("/api/samples/update/publicids", methods=["POST"])
+@requires_auth
+def update_sample_public_ids():
+    user: User = g.auth_user
+    request_data = request.get_json()
```
jgadling, line 425: Should we check that the user is a system admin here?

phoenixAja: yeah that's a good idea.. will add one sec

phoenixAja: added

src/backend/aspen/app/views/sample.py
```diff
@@ -416,3 +416,66 @@ def create_sample():
...
+
+    if not user.system_admin:
+        raise ex.BadRequestException(
+            "user making update request must be a system admin",
+        )
+
+    request_private_ids: list[str] = request_data["private_ids"]
```
jgadling: Why not just accept a map of private -> public ID's ?

src/backend/aspen/app/views/sample.py
```diff
@@ -416,3 +416,66 @@ def create_sample():
...
+    if existing_public_ids:
+        raise ex.BadRequestException(
+            f"Public Identifiers {existing_public_ids} are already in the database",
+        )
+
+    samples_to_update = (
+        g.db_session.query(Sample)
+        .filter(
+            and_(
+                Sample.submitting_group_id == request_data["group_id"],
```
jgadling:
```suggestion
                Sample.submitting_group_id == group_id,
```

jgadling (changes requested):

src/backend/aspen/app/views/sample.py
```diff
@@ -416,3 +416,66 @@ def create_sample():
...
+    request_private_ids: list[str] = request_data["private_ids"]
+    request_public_ids: list[str] = request_data["public_ids"]
+    private_to_public: Mapping[str, str] = {
+        private: public
+        for (private, public) in zip(request_private_ids, request_public_ids)
+    }
+
+    group_id: int = request_data.get("group_id")
+
+    # group id not supplied (when admin is making request), use the user.group_id instead
```
vincent-czi: I might be totally wrong here -- I'm missing a lot of general knowledge and context, so grain of salt and all that -- but I'm worried about the group_id falling through to the user's group when it's not explicitly passed in the request.

Looking at the CLI command, the `--group-id` value is required. If we only expect this endpoint to be hit via the CLI, it seems to me like the endpoint should match and make it required.

On the other hand, if this should get hit outside the context of the CLI, does it make sense to default to the user's group_id? [That's a real question BTW, I just don't know enough] Only system admins should be able to call this end point, right? Does a system admin's group correspond with the samples we would be changing private/public IDs on? I would figure that the `submitting_group_id` would be a DPH group and that admins would not be in that group. And if I understand what comes next, that would then cause it to raise on `missing_private_identifiers`, but it wouldn't be that the IDs are missing, it's that the group_id fetch locked them out because it used the admin group unexpectedly.

src/backend/aspen/app/views/sample.py
```diff
@@ -416,3 +416,66 @@ def create_sample():
...
+        g.db_session.query(Sample)
+        .filter(
+            and_(
+                Sample.submitting_group_id == request_data["group_id"],
+                Sample.private_identifier.in_(private_to_public.keys()),
+            )
+        )
+        .all()
+    )
+    for s in samples_to_update:
```
vincent-czi, line 469: Considering my whole `group_id` thing above, would it be possible for that to cause this `samples_to_update` to wind up being an empty list due to the filter AND with group_id above? Then if this is an empty list, it's a no-op, but the response would still be `success=True`, even though the submitted private/public IDs wound up not doing anything.

src/backend/aspen/app/views/sample.py
phoenixAja: so initially i was thinking that i would future proof this by allowing a user to update their own public ids, and would therefore not need to provide a group id. I was thinking that the only time when the group_id would be passed would be when a system admin is updating a groups public_ids. I realize that this is confusing now that we're restricting this to only system admins.. I'll take defaulting to the group_id out, since this isn't an existing use case and would cause the unclear missing_private_identifiers error like what you were mentioning. That's a good call out.

src/backend/aspen/app/views/sample.py
phoenixAja: yeah another good reason to take that out

src/backend/aspen/app/views/sample.py
phoenixAja: i don't know why i didn't do that.. lol

jgadling: I think we should update it to accept a map -- less risk of weird off-by-1 issues with two lists

phoenixAja: added

vincent-czi (approved):
Looks good! Nice work!

PR #681, opened 2021-09-22
Date Filtering fixes

vincent-czi:
### Summary:
- **What:** Fix Date Filtering to handle menu options more like specifications. Also covers a note from Janeece in feat/filtering PR regarding Genome Recovery filter.
- **Tickets:** [sc<161475>](https://app.shortcut.com/genepi/story/<161475>), [sc<161482>](https://app.shortcut.com/genepi/story/<161482>)
- **Env:**  https://filter-fixups-frontend.dev.genepi.czi.technology

### Demos:
image
Upload Date now has its own menu options instead of matching Creation Date's options.

---

image
Bad date format prevents user "Apply"ing the filter. Need to fix filter.

---

image
Selecting a menu option clears out field filters, they remain blank until a user types into them again. Menu option that is selected is highlighted.

---

image
Selected Genome Recovery filter shows "selected" state (bold).

---

### Checklist:
- [x] I merged latest `feat/filtering`
- [ ] I added labels to my PR

### DIFF
--- a/src/frontend/src/components/FilterPanel/components/DateFilter/index.tsx
+++ b/src/frontend/src/components/FilterPanel/components/DateFilter/index.tsx
@@ -1,6 +1,6 @@
 import { Menu, MenuItem } from "czifui";
 import { useFormik } from "formik";
-import { noop } from "lodash";
+import { noop, isEmpty } from "lodash";
 import React, { FC, useEffect, useState } from "react";
 import DateField, { FormattedDateType } from "src/components/DateField";
 import {
@@ -48,6 +48,9 @@ const DateFilter: FC<Props> = ({
   // `startDate` and `endDate` represent the active filter dates. Update on filter change.
   const [startDate, setStartDate] = useState<FormattedDateType>();
   const [endDate, setEndDate] = useState<FormattedDateType>();
+  // Are fields valid and user should be allowed to hit the "Apply" button?
+  // Related to some async form validation weirdness. See `validateForm` call below.
+  const [areFieldsValid, setAreFieldsValid] = useState<Boolean>(false);
   // What menu option is chosen. If none chosen, `null`.
   const [selectedDateMenuOption, setSelectedDateMenuOption] = useState<DateMenuOption | null>(null);
   const [anchorEl, setAnchorEl] = useState<HTMLElement>();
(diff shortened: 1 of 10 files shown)


src/frontend/src/components/FilterPanel/components/GenomeRecoveryFilter/index.tsx
```diff
@@ -12,6 +12,15 @@ interface Props {
   updateGenomeRecoveryFilter: (selected?: string) => void;
 }
 
+const GENOME_RECOVERY_OPTIONS = [
```
vincent-czi, line 9: This is kind of an interesting one. The `MenuSelect` component `value` wasn't able to identify that a previously selected option was previously selected because it did comparison with `===`, and since the OPTIONS were created inside component render, it was a new batch of objects each time, so they failed equality. Just pushing it outside of component render so it only gets made once is enough to get the value/option equality checking working so MenuSelect can now if it's looking at something it has already selected.

vincent-czi:
@Janeece This is the list of changes you posted to the feat/filtering PR: I've gone through and marked the ones that we have completed, and for those that aren't, I've got some explanation of where we are now or what's going on.

> ### All Filters:
> * [ ]  [Should Have] When filter menu is open, filter dropdown component should be in active state instead of default (text color would be black, icon would be primary400)

This is being worked on, but requires changes at the czifui level as well. Maya has a PR she's working on that handles [this](https://github.com/chanzuckerberg/sci-components/pull/63) but we may need to put a pause on this until this is resolved on the sds side

> * [ ]  [Should Have] Text size for filter dropdowns should be updated to use font mixin fontCapsXxs (11px)

Phoenix is working on this currently.

>   Filtering-text updates

---

> ### Upload Date:
> * [x]  [Nice to Have] Upload date is only supposed to have buckets for “Today”, “Yesterday”, and “Last 7 Days”
> * [ ]  [Should Have] - When error message appears for incorrect date format, can we have the inputs remain aligned to the top?

Phoenix is working on this currently.

>   Filtering-error alignement
> * [x]  [Must Have] - do not allow user to apply date range that is incorrectly formatted (disable button if format error is present)
>   Filtering-error applied
> * [ ]  [Nice to Have] - I think we should only show the error message after use clicks out of the date input field (so they dont get an error before they finish typing)

It turns out this is surprisingly difficult because of how the underlying validation/form library works. Also would necessitate working through some UX edge cases, I think. I made a ticket for it: https://app.shortcut.com/genepi/story/163399/clean-up-errors-for-date-field-in-filtering-panel

> * [x]  [Must Have, not a blocker] - when user makes a selection from the menu items (non-date range), the filter tag text should match. This ticket already exists [here](https://app.shortcut.com/genepi/story/161475/update-filter-tag-label-to-reflect-selection)
> 
> ### Collection Date:
> * [x]  Same as Upload Date, except that the menu item options are correct
> 
> ### Lineage:
> beautiful!
> 
> ### Genome Recovery:
> * [x]  [Must Have, not a blocker]  When I make a selection on Genome Recovery and then reopen the menu, the current selection should be using the “selected” state (text becomes semibold). Note: this may relate to [this ticket](https://app.shortcut.com/genepi/story/161482/remove-autofill-date-range-from-upload-date-collection-date-menu) for upload/collection date as well
>   Filtering- active state

src/frontend/src/components/FilterPanel/components/DateFilter/index.tsx
```diff
@@ -64,6 +68,23 @@ const DateFilter: FC<Props> = ({
     setAnchorEl(undefined);
   };
 
+  const validationSchema = useMemo(
```
phoenixAja, line 73: nice, we should be using useMemo in a lot of other places too.. lol

src/frontend/src/components/FilterPanel/components/GenomeRecoveryFilter/index.tsx
phoenixAja: cool glad it was a relatively easy fix

src/frontend/src/components/FilterPanel/components/DateFilter/index.tsx
```diff
@@ -22,40 +22,44 @@ import {
...
+  numDaysEndOffset?: number; // How far back end of date interval is
+  numDaysStartOffset: number; // How far back start of date interval is
+}
+
 interface Props {
   fieldKeyEnd: string;
   fieldKeyStart: string;
   inputLabel: string;
   updateDateFilter: (start: FormattedDateType, end: FormattedDateType) => void;
+  menuOptions: DateMenuOption[];
```
phoenixAja, line 47: this is great

phoenixAja (approved):
looks awesome 👍

src/frontend/src/components/FilterPanel/components/DateFilter/index.tsx
vincent-czi: Credit where it's due, the useMemo is all Timmy. He pushed up a branch with an alternative, better fix to the `isValid` visual flicker problems I was having, and this was one of his other improvements.

Janeece:
@vincent-czi Amazing, thank you! My only comment is on the Genome Recovery update. 

- [x] An active menu item should have a white background (unless hovering), like on the other menu items. Active Menu item font weight is correct!
- [x] The spacing within Genome recover dropmenu/menu items also seems much larger than the other menus -- can we reduce it to match? (let me know if this is an SDS rooted inconsistency & i will file a ticket there too)

Screen Shot 2021-09-22 at 1 48 17 PM

src/frontend/src/components/FilterPanel/components/DateFilter/index.tsx
phoenixAja: @tihuan 💯 😸 🚀

src/frontend/src/components/FilterPanel/components/DateFilter/index.tsx
phoenixAja: like that you added this prop and now i feel like this is all much more organized ❤️

mmmmmmaya (approved):

PR #685, opened 2021-09-22
Trigger Pangolin job on sample upload

lvreynoso:
### Summary:
- **What:** `Triggers a pangolin job on sample upload. Job processes uploaded samples.`
- **Ticket:** [sc150871](https://app.shortcut.com/genepi/story/150871)
- **Env:** `https://pangolin-frontend.dev.genepi.czi.technology/`

### Demos:
Sample uploaded and lineage assigned in ~6 minutes!
Screen Shot 2021-09-22 at 15 24 00

[SFN execution](https://us-west-2.console.aws.amazon.com/states/home?region=us-west-2#/executions/details/arn:aws:states:us-west-2:473004499091:execution:aspen-rdev-pangolin-swipe-spot-sfn:ADMIN-ondemand-pangolin-2021-09-22-22-13-18-670020)

### Notes:
* Pangolin now installed during Docker container build. This is roughly 3.5 minutes, which is painful at first but should be cached from then on.
* Nightly pangolin job still runs on all samples with out-of-date pangolin versions. Pulls latest trunk from pangolin repo, then installs.

### Checklist:
- [X] I merged latest `<base branch>`
- [X] I manually verified the change
- [X] I added labels to my PR
- ~~[ ] I tested in multiple browsers~~


src/backend/aspen/config/config.py
```diff
@@ -303,3 +310,38 @@ def NEXTSTRAIN_SPOT_MEMORY(self) -> str:
...
+    @property
+    def PANGOLIN_OUTPUT_PREFIX(self) -> str:
+        return self.AWS_PANGOLIN_SFN_PARAMETER["OutputPrefix"]
+
+    @property
+    def PANGOLIN_WDL_URI(self) -> str:
+        return self.AWS_PANGOLIN_SFN_PARAMETER["RUN_WDL_URI"]
+
+    @property
+    def PANGOLIN_EC2_MEMORY(self) -> str:
```
jgadling: Any thoughts on whether we should break all these parameters into individual settings vs just having general "pangolin_job_params" and "nextstrain_job_params" type settings that contain the necessary info?

jgadling (approved):

src/backend/aspen/config/config.py
lvreynoso: I can't really think of a great approach here right now, but I'm certainly not attached to this one. It makes it easy to see what's available at a glance but anyone tinkering with this could easily look it up. There's no reason they couldn't just be used by accessing config.AWS_PANGOLIN_SFN_PARAMETER with the right key

jgadling: Alright, just curious. I think we can revisit this in the future. Excited to see this feature working!

lvreynoso: I ended up getting rid of them, I realized that where I am using them I am constructing a JSON string with the same keys as the ones I'm accessing, so no memorization even needed - it's pretty natural.

PR #687, opened 2021-09-23
allow all tree statuses

phoenixAja:
### Summary:
- **What:** remove stub for passing on only completed trees
- **Ticket:** no ticket
- **Env:** https://featfiltering-frontend.dev.genepi.czi.technology/

### Checklist:
- [x] I merged latest `self serve tree v1`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/views/Data/index.tsx
+++ b/src/frontend/src/views/Data/index.tsx
@@ -49,10 +49,7 @@ const Data: FunctionComponent = () => {
       setIsDataLoading(false);
 
       const apiSamples = sampleResponse["samples"];
-      // TODO: Support in-progress and failed trees
-      const apiTrees = treeResponse["phylo_trees"].filter(
-        (tree) => tree.status === "COMPLETED"
-      );
+      const apiTrees = treeResponse["phylo_trees"];
 
       setSamples(apiSamples);
       setTrees(apiTrees);


src/frontend/src/views/Data/index.tsx
```diff
@@ -49,10 +49,7 @@ const Data: FunctionComponent = () => {
       setIsDataLoading(false);
 
       const apiSamples = sampleResponse["samples"];
-      // TODO: Support in-progress and failed trees
-      const apiTrees = treeResponse["phylo_trees"].filter(
-        (tree) => tree.status === "COMPLETED"
```
vincent-czi, line 54: This doesn't affect this PR, but if the frontend needs to know the different possible statuses (so it can match up coloring appropriately, etc), we should probably have constants for them somewhere and then import those constants for this kind of check.

vincent-czi (approved):

src/frontend/src/views/Data/index.tsx
phoenixAja: @vincent-czi @mmmmmmaya maya added that in [here](https://github.com/chanzuckerberg/aspen/blob/9037ead4f396e3b2c933864211589400be603d71/src/frontend/src/common/types/bioinformatics.d.ts#L18)

PR #688, opened 2021-09-23
Update policy copy

jgadling:
### Summary:
- **Ticket:** [sc163254](https://app.shortcut.com/genepi/story/163254) and [sc163251](https://app.shortcut.com/genepi/story/163251) and [sc161453](https://app.shortcut.com/genepi/story/161453)
- **Env:** https://chicagoui-frontend.dev.genepi.czi.technology

### Checklist:
- [ ] I added labels to my PR


mmmmmmaya (approved):

jgadling:
https://chicagoui-frontend.dev.genepi.czi.technology

src/frontend/src/views/AgreeTerms/index.tsx
```diff
@@ -11,7 +11,7 @@ import ENV from "src/common/constants/ENV";
 import { useUpdateUserInfo } from "src/common/queries/auth";
 import { ROUTES } from "src/common/routes";
 import { PageContent } from "../../common/styles/mixins/global";
-import { Details, Title } from "./style";
+import { BoldText, Details, Title } from "./style";
```
vincent-czi: Looking at the `Privacy` and `Terms` component pages, they're importing from a common style for some of these. (see https://github.com/chanzuckerberg/aspen/blob/9572eccaf89903c36e5ec7acfa6eabe1ba789c78/src/frontend/src/views/Terms/index.tsx#L15 )

I'm not sure if we should be trying to standardize to that and use the same here as well. It might be sensible to have it break off -- your `BoldText` is slightly different, for example -- but I just don't know which direction we're aiming.

vincent-czi (approved):
Looks good! My one note is super minor and kind of goes against how the page was already built, anyway.

src/frontend/src/views/Terms/index.tsx
```diff
@@ -195,31 +182,7 @@ export default function Terms(): JSX.Element {
...
-                  community as soon as possible.{" "}
-                  <B>
-                    For this reason, unless you choose to mark a Sample as
-                    ‘private’ within 2 weeks of upload, we will automatically
-                    share Pathogen Consensus Genomes with public repositories.
-                  </B>{" "}
-                  However, we recognize that this is not always your preference.
-                  Thus, you may choose to label your data as Private, so that it
-                  cannot be seen by anyone outside of your group.
+		  The pathogen genomes created from your Samples and the analytical outputs you create using Aspen are visible to other Users at your organization (ex: your DPH). You, along with other members of your Group, control whether you permit us to share this information with Users outside your organization. In certain circumstances, we may also share your data with third party entities, through the Aspen tool, in line with your organization’s policies or in line with applicable law.  Samples marked &quot;private&quot; will never be shared with any 3rd parties unless you choose to mark them &quot;public&quot; later on.
```
vincent-czi: I'd like to have this get some newlines just to make it easier to read at a glance in code editor

src/frontend/src/views/Privacy/index.tsx
```diff
@@ -31,17 +31,17 @@ const PrivacyPolicy = (): JSX.Element => {
...
-        &quot;our&quot;), provides the Aspen product (&quot;Services&quot; or
-        &quot;Aspen&quot;) in close collaboration with the Chan Zuckerberg
-        Biohub (&quot;CZB&quot;), and the Chan Zuckerberg Initiative, LLC
-        (&quot;CZI LLC&quot;). This Privacy Policy describes the types of
-        information we collect or that is uploaded by Aspen Users (collectively
-        &quot;Users&quot; or &quot;you&quot;, ex: registered public health
-        officials at state and/or county level Departments of Public Health
-        (&quot;DPH&quot;), other public health researchers), and how we use,
-        share, and protect that information.
+        The Chan Zuckerberg Initiative private foundation (&quot;<B>CZIF</B>,
```
vincent-czi: In the AgreeTerms, you're using `&rdquo;` and `&ldquo;`, while everywhere else seems to be using `&quot;`. Is there a reason why it changes? For this kind of copy, I'd assume they should just all the same.

And if you're using plain old `&quot;`, can we just use `"` symbols instead? Maybe the linter will scream its head off, but I certainly find that more readable if it's not going to screw up some escaping somewhere.

src/frontend/src/views/Privacy/index.tsx
```diff
@@ -31,17 +31,17 @@ const PrivacyPolicy = (): JSX.Element => {
         <H4>Last Updated: May 17, 2021. </H4>
       </Title>
       <P>
-        The Chan Zuckerberg Initiative Foundation, a 501(c)(3) nonprofit private
```
vincent-czi, line 34: Are you looking at a different doc than I am? The version I'm looking at (from the Shortcut story) still has the part about "501(c)(3) nonprofit private foundation". Are we looking at different Privacy Policy docs?

src/frontend/src/views/Privacy/index.tsx
```diff
@@ -31,17 +31,17 @@ const PrivacyPolicy = (): JSX.Element => {
         <H4>Last Updated: May 17, 2021. </H4>
```
vincent-czi: Probably an unnecessary reminder, but we'll need to remember to update this date when we publish.

src/frontend/src/views/AgreeTerms/index.tsx
```diff
@@ -68,7 +68,7 @@ export default function AgreeTerms(): JSX.Element {
                 key things to know:
               </ListSubheader>
               <ListItem fontSize="s">
-                You {" "}<B> always </B>{" "} own and control the data you
+                You &nbsp;<B>always</B>&nbsp; own and control the data you
```
vincent-czi: I'm wary of using non-breaking spaces for anything other than exactly that, and even then I'm usually worried they'll bite me. If the primary reason is to increase padding on bold font, then I probably made the wrong suggestion with switching over to the general `B` component instead of a custom BoldText. You could toss an additional style on the `B` in this file:
```

...

const spacedBold = styled(B)`
  ${(props) => {
    const spacings = getSpacings(props);
    return `
      padding: 0 ${spacings?.s}px;
    `;
  }}
`;
```

Would probably be the most matched to how things are done elsewhere. Although I think it would also be pretty reasonable to just have
```
styled(B)`
  padding: 0 0.5em; // adjust 0.5em size to taste, but because em, will handle later font size changes well
`;
```

PR #693, opened 2021-09-23
Phylo runs >12h old without a phylo tree are marked as failed.

jgadling:
### Summary:
- **What:** Any phylo runs started >12h ago that don't have a phylo tree will have a FAILED status
- **Env:** `<rdev link>`

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/src/backend/aspen/app/views/phylo_trees.py
+++ b/src/backend/aspen/app/views/phylo_trees.py
@@ -16,6 +16,7 @@ from aspen.app.views.api_utils import (
     format_date,
     format_datetime,
 )
+from datetime import timedelta, datetime
 from aspen.database.models import (
     DataType,
     PhyloRun,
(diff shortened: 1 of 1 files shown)


src/backend/aspen/app/views/phylo_trees.py
```diff
@@ -105,6 +106,10 @@ def phylo_trees():
                 "phylo_tree_id": None,
                 "name": phylo_run.name or generate_tree_name_from_template(phylo_run),
             }
+            if phylo_run.start_datetime and phylo_run.start_datetime < (
```
vincent-czi, line 109: The one thing I'm unclear on is if we get to this part of the code, we are guaranteed that the phylo tree is not "completed" right? I think that's due to the `if isinstance(output, PhyloTree):` section above if I'm reading it correctly? We check each thing for a PhyloTree output, and if it is one, then it's implicitly completed because we have the tree -- am I correct?

If so, then this all looks good. If I'm misunderstanding the above, then my one worry is something that should somehow be complete still getting marked as FAILED just because it ran super slow.

mmmmmmaya (approved):

vincent-czi (approved):
Looks good, assuming I'm reading what happens above the change correctly

src/backend/aspen/app/views/phylo_trees.py
jgadling: Yep, this is exactly what's happening.

PR #694, opened 2021-09-24
Samples table checkbox alignment

vincent-czi:
### Summary:
- **What:** Fix checkbox alignment
- **Ticket:** No ticket, from comments in Google Doc
- **Env:** No rdev, just see photos below

### Demos:

Fixes checkbox issue on Samples page
image

No impact on Trees page, spacing is same as before PR
image

### Notes:
I looked a little bit into getting the alignment for the row content (eg, collection date in the row being offset from its header), but I think that might be quite a bit more challenging. I'm not really sure how the alignment is currently happening, and I don't know if the header talks to the rows at all in terms of spacing and alignment right now. Not sure how to approach.

### DIFF
--- a/src/frontend/src/common/components/library/data_table/index.tsx
+++ b/src/frontend/src/common/components/library/data_table/index.tsx
@@ -1,3 +1,4 @@
+import classNames from "classnames";
 import { get, isEqual } from "lodash/fp";
 import React, {
   Fragment,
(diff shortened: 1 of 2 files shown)


phoenixAja (approved):
screenshots look great! thanks for taking this on! 🚀

src/frontend/src/common/components/library/data_table/index.tsx
```diff
@@ -334,9 +334,17 @@ export const DataTable: FunctionComponent<Props> = ({
       );
     }
 
+    // All headers use the base style.header
```
mmmmmmaya: You can do this more cleanly with classNames.

```
import classNames from "classnames";

// All headers use the base style.header
// However, if header needs to show checkbox, we add additional style
// to override certain aspects of header. See style file for more.
const headerClasses = classNames(style.header, {
  [style.headerWithCheckbox]: showCheckboxes,
});
```

mmmmmmaya (approved):

src/frontend/src/common/components/library/data_table/index.tsx
phoenixAja: neat

vincent-czi: Cool, thanks for the suggestion!

PR #700, opened 2021-09-27
[ch163990] Maya/invalidate tree cache

mmmmmmaya:
### Summary
- **What:** Invalidate FE cache when new tree created
- **Why:** So when users click the link to the trees page after creating a new tree, their tree is actually shown
- **Ticket:** [[sc-163990]](https://app.shortcut.com/genepi/story/163990)
- **Env:** https://phylotreecache-frontend.dev.genepi.czi.technology/data/phylogenetic_trees
- **Discussion:** https://czi-sci.slack.com/archives/C02F8V4LMNJ/p1632518947052000

### Testing
1. loading trees page directly doesnt yield a blank table
2. creating a new tree from samples page, then click “View my trees” link in toast message tkes you to the tree page AND your new tree is shown.
3. creating a new tree from samples page, then click “phylogenetic trees” tab in nav bar takes you to the tree page AND your new tree is shown.

#### Before: 
before

#### After: 
(if the cache is refreshed after the page change occurs)
after2

(if the cache is refreshed before the page change occurs)
after

### Notes
- I created a [ticket](https://app.shortcut.com/genepi/story/164215/refactor-data-subview) to refactor the data subview component. It's starting to be responsible for too many things, and there are a lot of low hanging fruit for making this component easier to understand. 

### Checklist
- [x] I merged latest `self-serve-tree-v1`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/views/Data/index.tsx
+++ b/src/frontend/src/views/Data/index.tsx
@@ -5,8 +5,9 @@ import Link from "next/link";
 import { useRouter } from "next/router";
 import React, { FunctionComponent, useEffect, useState } from "react";
 import { Menu } from "semantic-ui-react";
-import { fetchSamples, fetchTrees } from "src/common/api";
+import { fetchSamples } from "src/common/api";
 import { useProtectedRoute } from "src/common/queries/auth";
+import { useTreeInfo } from "src/common/queries/trees";
 import { FilterPanel } from "src/components/FilterPanel";
 import { DataSubview } from "../../common/components";
 import { EMPTY_OBJECT } from "../../common/constants/empty";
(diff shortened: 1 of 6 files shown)


phoenixAja (approved):
looks great!! THANK YOU MAYA!!

vincent-czi (approved):
To the extent that I don't really understand what the react-query stuff is doing, looks good!

Janeece (approved):
Works great, thanks for making this fix!

src/frontend/src/views/Data/index.tsx
```diff
@@ -37,26 +38,25 @@ const Data: FunctionComponent = () => {
 
   const router = useRouter();
 
+  const treeResponse = useTreeInfo();
+  const { data, isLoading } = treeResponse;
```
phoenixAja, line 42: it's a little weird to me that this is outside of the useEffect hook below, but that fetchSamples is not, do you think we could remove the fetchSamples from the useEffect hook as well? Or is there a reason why that still has to be there?

phoenixAja: can we take useTreeInfo out because of the react query queryKey parameter, so it only changes when the key hash is updated?

PR #702, opened 2021-09-28
strip off time from createdDate

phoenixAja:
### Summary:
- **What:** remove time from startedDate (display CreationDate) column until we add in date conversion based on user's timezone
- **Ticket:** no ticket
- **Env:** added screenshot instead

### Screenshot:
Screen Shot 2021-09-28 at 8 41 31 AM

### Checklist:
- [x] I merged latest `self-serve-tree-v1`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/views/Data/cellRenderers.tsx
+++ b/src/frontend/src/views/Data/cellRenderers.tsx
@@ -13,6 +13,7 @@ import TreeTableDownloadMenu from "src/components/TreeTableDownloadMenu";
 import { Lineage, LineageTooltip } from "./components/LineageTooltip";
 import TreeTableNameCell from "./components/TreeTableNameCell";
 import { TreeTypeTooltip } from "./components/TreeTypeTooltip";
+import style from "./index.module.scss";
 import {
   GISAIDCell,
   PrivacyIcon,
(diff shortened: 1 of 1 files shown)


phoenixAja:
@mmmmmmaya or @lvreynoso let me know if there is a better way to do this, i really only needed to change the value by stripping off the time info, the display elements are identical to the `deafaultCellRenderder` implementation.

vincent-czi (approved):
Looks good! The only thing that confuses me is that there is no corresponding removal of code? I don't think that's indicative of anything wrong, I'm just surprised: the `startedDate` part looks totally new, but I figured it would already have been in there and this would just be modification of an existing setup? Like I said, I don't think anything is wrong, I'm just curious to understand where it was coming from (with the time as well, etc) before the addition of this code.

mmmmmmaya (approved):

mmmmmmaya:
@vincent-czi all cells are rendered using a default cell rendering function _unless_ there is a corresponding custom one (keyed by column key). That's why you aren't seeing code removal here.

You can search for `defaultCellRenderer`.

phoenixAja:
yes you can see [here](https://github.com/chanzuckerberg/aspen/blob/4abf3c762d1799255737184ab869f416f519d6b8/src/frontend/src/views/Data/cellRenderers.tsx#L135:14) that if no custom `TREE_CUSTOM_RENDERERS` is supplied it will default to the `defaultCellRenderer` display (which was what was happening with the `creationDate` table cells before i added it).

vincent-czi:
Got it! Thank you both for the explanation!

PR #706, opened 2021-09-28
Add filter for private samples.

jgadling:
### Summary:
- **What:** `Filter private samples for tree builds`

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/src/backend/aspen/app/views/api_utils.py
+++ b/src/backend/aspen/app/views/api_utils.py
@@ -189,6 +189,9 @@ def authz_sample_filters(query: Query, sample_ids: Set[str], user: User) -> Quer
             ),
         )
     )
+    query = query.filter(
+        and_(or_(~Sample.private, Sample.submitting_group_id == user.group_id))
+    )
     return query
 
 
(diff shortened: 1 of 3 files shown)


lvreynoso (approved):
lgtm

src/backend/aspen/app/views/api_utils.py
```diff
@@ -189,6 +189,9 @@ def authz_sample_filters(query: Query, sample_ids: Set[str], user: User) -> Quer
             ),
         )
     )
+    query = query.filter(
+        and_(or_(~Sample.private, Sample.submitting_group_id == user.group_id))
+    )
```
phoenixAja, line 194: so this is different level of privacy aside from just being able to see private_identifiers? I guess i'm a little confused why we need this. i thought the `private` field was an indicator for if a sample didn't have an associated gisaid id?

jgadling: According to the privacy policy, it's a flag to prevent sharing samples with outside groups that normally have access to your group's samples (e.g: cdph)

PR #707, opened 2021-09-29
Maya/feature flags

mmmmmmaya:
### Summary
- **What:** Add FE-only feature flags to aspen
- **Why:** eliminate the need for giant feature branches in development
- **Env:**
  - https://featureflags-frontend.dev.genepi.czi.technology/data/samples?mayasFlag=true
  - https://featureflags-frontend.dev.genepi.czi.technology/data/samples?mayasFlag=false

### Demos
Screen Shot 2021-09-29 at 3 16 15 PM
Screen Shot 2021-09-29 at 3 22 25 PM

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/utils/featureFlags.tsx
+++ b/src/frontend/src/common/utils/featureFlags.tsx
@@ -1,0 +1,15 @@
+import { hasQueryParam } from ".";
+
+// we can add more keys to each flag as needed
+export const FEATURE_FLAGS = {
+  mayasFlag: {
+    isActive: true,
+    param: "mayasFlag",
+  },
+};
+
+export const usesFeatureFlag = (flag: string): boolean => {
+  const { isActive, param } = flag;
+  if (isActive && hasQueryParam(param)) return true;
+  return false;
+};
--- a/src/frontend/src/common/utils/featureFlags.tsx
+++ b/src/frontend/src/common/utils/featureFlags.tsx
@@ -8,7 +8,7 @@ const FEATURE_FLAG_PREFIX = "genepi-ff-";
 
 interface FeatureFlag {
   isActive: boolean;
-  param: string;
+  key: string;
 }
 
 interface FlagsObj {
@@ -25,7 +25,8 @@ export const FEATURE_FLAGS: FlagsObj = {
 
 const allowedKeys = Object.keys(FEATURE_FLAGS) as string[];
 
-const isStrTrue = (str: string): boolean => {
+const isStrTrue = (str: string | null): boolean => {
+  if (!str) return false;
   return str === "true";
 };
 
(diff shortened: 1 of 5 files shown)


phoenixAja (approved):
Looks good! We may want to include some instructions on this in the Frontend docs, but that can be done later

src/frontend/src/common/utils/featureFlags.tsx
```diff
@@ -0,0 +1,24 @@
+import { hasQueryParam } from ".";
```
vincent-czi: Can you explain how this works to me? I've never seen this usage for importing from something other than an `index.js` file. Is it effectively searching all the exports in the directory for that named export? I had been under the impression that you always had to specify the file being imported from OR specify a directory and it would default to importing from `index` in that directory. But... obviously I'm wrong. So what's the deal with `"."`?

src/frontend/src/common/utils/featureFlags.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+interface FeatureFlag {
+  isActive: boolean;
```
vincent-czi: This is a quibble, but I think `isActive` leads to the assumption that the flag is globally on, which is not the case. It's more like "enabled, assuming the user also has passed the matching query param". I think it might be less prone to misinterpretation to go the opposite direction and instead call it `disabled`, then flip the `if isActive` to `if !disabled`.

vincent-czi (approved):
Looks good! Very clear API for when it actually gets used, nice. A couple lines of usage documentation for `usesFeatureFlag` would be helpful to explain the usage intent, but it's also fine without since this is still in flux.

vincent-czi:
> * Feature flag in url is not preserved as you navigate through pages. May be a nice-to-have feature for future updates.

I remember this being a hassle at my previous job as well. If we put in a global state store, we can have the query param checker huck the flag into global state to persist it. If we don't, there's [this Stack Overflow](https://stackoverflow.com/questions/54560660/persist-query-string-parameters-in-nextjs-when-navigating-client-side): one approach in top comment, another in the accepted answer. Neither feels particularly elegant, but we could make it work in a pinch.

src/frontend/src/common/utils/featureFlags.tsx
mmmmmmaya: Oops, this was an autocompleted line. But the answer here is that there is an index.tsx file in the same directory that imports all the util files and re-exports them. That makes each function available from the directory level (like shown here), or from the file level (as if I were going to do `... from "./urlUtils";`

vincent-czi: Ah, thank you!

vincent-czi (approved):
This is unabashedly great! I'm not sure, I think I might have looked at a stale version when I did my previous review, but the way local storage is being used works great with the flags. Really like it.

My only nice-to-have would be adding a tiny usage example at `setFeatureFlagsFromQueryParams` and explicitly mentioning that a query param can be unset by putting it in as false in the URL (or whatever, I guess, just not "true"). It took me a bit to reading back and forth to be sure that's how it worked, so calling it out seems useful: since it persists from session to session, turning off flags will be almost as important as being able to turn them on.

PR #709, opened 2021-09-29
Sc163454 acknowledge policy changes banner

vincent-czi:
### Summary:
- **What:** Adds a banner that notifies user of policy changes, along with a way to track user acknowledgment of those changes.
- **Ticket:** [sc<163454>](https://app.shortcut.com/genepi/story/<163454>) & [sc<164302>](https://app.shortcut.com/genepi/story/<164302>)
- **Env:** https://chicagoui-frontend.dev.genepi.czi.technology

### Demos:

(Edit: whoops, took screenshots and pushed rdev before I remembered to move from "September XX" to "September 30". Will update rdev right now, should be in place by 11:30am)

How it looks in wide (normal) window
image

---

How it looks in narrow window (with text wrapping)
image

### Notes:

Some conditionality on when the banner displays:
- User must be logged in. Discussed with Olivia on Slack: we won't show the banner to anonymous visitors since they aren't necessarily app users who would need to agree to terms, etc or be made aware of changes.
- With above in mind, even if a user is logged in, we only display the banner if they have accepted the Terms of Service (that way we don't simultaneously display the Terms of Service pop-up _and_ the "new policies!" banner running along the top)
- Closing the banner is considered acknowledgment of the changes and will prevent the banner from appearing from there on out (or at least until the next time we need to change policies).

Other things:
- Agreeing to the ToS is _also_ considered an acknowledgment of the current set of policies. (This is fairly explicit in the ToS pop-up that you're agreeing to the policies). So when the agreement sets the `agreed_to_tos` to true, it also marks down the `acknowledged_policy_version` as the current version. That way new users aren't immediately spammed with the "new policies!" banner right after they sign up and agree to terms.
- Banner is set up so it's easy to bring it back if we change policies in the future. Just update `CURRENT_POLICY_VERSION`  when making changes to any policy pages, and the banner will re-appear. (It's arguable that this constant shouldn't live inside the banner component, but it seemed reasonable enough to me. If anyone has strong feelings, let me know where would be a better place to locate it.)
- Sets the **official date of policy change to Sep 30, 2021**, when this stuff is expected to go to Prod. Also sets that date on both Privacy Policy and Terms pages.
- Involves a DB migration: adds new column `acknowledged_policy_version` of type date  to the `users` table. If user has never done an acknowledgment, should be NULL.

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/backend/aspen/database/models/usergroup.py
```diff
@@ -47,6 +47,9 @@ class User(idbase, DictMixin):  # type: ignore
     group_admin = Column(Boolean, nullable=False)
     system_admin = Column(Boolean, nullable=False)
     agreed_to_tos = Column(Boolean, nullable=False, default=False)
+    # Date of policies (any of Privacy Policy, Terms of Service, etc, etc) the user
+    # has last acknowledged. Used to display notication to user when policies change.
```
jgadling:
```suggestion
    # has last acknowledged. Used to display notification to user when policies change.
```

src/backend/database_migrations/versions/20210928_203749_add_users_column_of_acknowledged_policy_.py
```diff
@@ -0,0 +1,28 @@
+"""Add users column of acknowledged_policy_version
+
+Create Date: 2021-09-28 20:37:51.381909
+
+"""
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
```
phoenixAja: you can remove these autogenerated comments from alembic
```suggestion
```

src/frontend/src/common/types/user.d.ts
```diff
@@ -18,4 +18,5 @@ interface User {
   name: string;
   systemAdmin: boolean;
   agreedToTos: boolean;
+  acknowledgedPolicyVersion: string; // Date column in DB, will be ISO 8601: YYYY-MM-DD
```
jgadling: This will just be serialized as '' if the version in the db is null?

phoenixAja: nice comment

jgadling (changes requested):
When I click the X in the banner, the banner disappears for a moment and then comes right back.

src/frontend/src/common/types/user.d.ts
vincent-czi: It comes through as JS `null` when DB NULL, so good point, I'll go back and add a union type

vincent-czi:
> When I click the X in the banner, the banner disappears for a moment and then comes right back.

Well that's bad.

src/frontend/src/components/AcknowledgePolicyChanges/style.ts
```diff
@@ -0,0 +1,84 @@
...
+export const DummyCenteringSpacer = styled.div`
+  width: 0px;
+`;
+
+// Both the margin-block and margin are here just to override defaults
+// (margin-block is coming in from user agent, plain margin is an upstream CSS)
+export const MainText = styled.p`
+  margin-block: 0;
+  margin: 0;
+  ${fontBodyS}
```
phoenixAja: usually we like to move the czifui mixins to the top
```suggestion
  ${fontBodyS}
  margin-block: 0;
  margin: 0;
  
```

phoenixAja (approved):
looks really solid, gj! The comments are excellent, inspires me to write better comments on what i'm doing 🤩

phoenixAja:
> When I click the X in the banner, the banner disappears for a moment and then comes right back.

wierd that didn't happen to me, i even refreshed the page and it didn't come back 😕

jgadling:
> > When I click the X in the banner, the banner disappears for a moment and then comes right back.
> 
> Well that's bad.

We debugged this together - the rdev db has some weird bugs in it -- this PR works for me. I'll fix the rdev snapshot.

src/frontend/src/common/queries/auth.ts
```diff
@@ -24,6 +24,7 @@ export const USE_USER_INFO = {
   id: "userInfo",
 };
 
+// FIXME (vince): I can't find this used anywhere? If that's true, should delete.
```
mmmmmmaya: going back to [where this was added](https://github.com/chanzuckerberg/aspen/pull/475/files), it doesn't seem to ever have been used anywhere. My guess is that fact got lost in the shuffle of the huge PR

src/frontend/src/components/AcknowledgePolicyChanges/style.ts
```diff
@@ -0,0 +1,84 @@
+import styled from "@emotion/styled";
+import CloseIcon from "@material-ui/icons/Close";
+import { fontBodyS, getColors, getIconSizes, getSpacings, Link } from "czifui";
+import IconInfo from "src/common/icons/IconInfo.svg";
+
+// Note the use of `DummyCenteringSpacer`. Explanation with it.
```
mmmmmmaya: seems like this has been satisfied and can be removed

mmmmmmaya (approved):

jgadling (approved):

src/frontend/src/common/queries/auth.ts
vincent-czi: Thanks for confirming. I've removed it.

src/frontend/src/components/AcknowledgePolicyChanges/style.ts
vincent-czi: This was not a very clear comment, but I wanted to point out the weird flex usage and how it depended on `DummyCenteringSpacer`. I've improved the comment, should be clear now.

PR #710, opened 2021-09-29
Run fastapi side by side with flask

jgadling:
### Summary:
- **What:** `Run fastapi alongside flask in docker`
- **Ticket:** [sc161311](https://app.shortcut.com/genepi/story/161311)
- **Env:** https://fastapi-frontend.dev.genepi.czi.technology  // https://fastapi-backend.dev.genepi.czi.technology/v2/health/

### Notes:
This PR runs fastapi AND flask inside our backend docker container, with [Nginx](https://www.nginx.com/) routing requests between the two applications. It uses [supervisord](http://supervisord.org/) to manage the different services inside the container. 

- By default, all requests will be forwarded from nginx to the flask app
- Any requests for endpoint paths beginning with '/v2/' will be sent to the fastapi app
- This is intended to be as close to a no-op as possible for the flask app, but I'd like to switch from using Pipenv as a python env manager, to Poetry. So for the moment we have both tools in the backend container fighting with each other. Updates to python libraries will have to be made to both poetry.lock and Pipfile for now :( 
- There's a little bit of (functional!) scaffolding and configuration to allow the fastapi app to connect to our DB and return serialized User objects, but for now that code isn't enabled. Fastapi is only exposing a single endpoint: `/v2/health/`

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/error/http_exceptions.py
+++ b/src/backend/aspen/api/error/http_exceptions.py
@@ -1,0 +1,46 @@
+from typing import Any, Mapping, Optional
+
+from fastapi.exceptions import RequestValidationError
+from fastapi.responses import JSONResponse
+from flask import jsonify, make_response, Response
+from pydantic.error_wrappers import ErrorWrapper
+from starlette.requests import Request
+
+
+class AspenException(Exception):
+    status_code = 500
+
+    def __init__(self, msg, extra: Optional[Mapping[str, Any]] = None):
+        super().__init__(msg)
+        self.extra = extra
+
+    def make_response(self) -> Response:
+        err = str(self)
+        respdata = {"error": err}
+        if self.extra:
+            respdata.update(self.extra)
+        return JSONResponse(
+            status_code=self.status_code,
+            content=respdata,
+            headers={"Content-Type": "application/json"},
+        )
+
+
+async def exception_handler(request: Request, exc: AspenException) -> JSONResponse:
+    return err.make_response()
+
+
+class UnauthorizedException(AspenException):
+    status_code = 401
+
+
+class BadRequestException(AspenException):
+    status_code = 400
+
+
+class NotFoundException(AspenException):
+    status_code = 404
+
+
+class ServerException(AspenException):
+    status_code = 500
--- a/src/backend/aspen/api/error/http_exceptions.py
+++ b/src/backend/aspen/api/error/http_exceptions.py
@@ -2,7 +2,6 @@ from typing import Any, Mapping, Optional
 
 from fastapi.exceptions import RequestValidationError
 from fastapi.responses import JSONResponse
-from flask import jsonify, make_response, Response
 from pydantic.error_wrappers import ErrorWrapper
 from starlette.requests import Request
 
@@ -14,7 +13,7 @@ class AspenException(Exception):
         super().__init__(msg)
         self.extra = extra
 
-    def make_response(self) -> Response:
+    def make_response(self) -> JSONResponse:
         err = str(self)
         respdata = {"error": err}
         if self.extra:
--- a/src/backend/aspen/api/error/http_exceptions.py
+++ b/src/backend/aspen/api/error/http_exceptions.py
@@ -1,8 +1,6 @@
 from typing import Any, Mapping, Optional
 
-from fastapi.exceptions import RequestValidationError
 from fastapi.responses import JSONResponse
-from pydantic.error_wrappers import ErrorWrapper
 from starlette.requests import Request
 
 
@@ -26,7 +24,7 @@ class AspenException(Exception):
 
 
 async def exception_handler(request: Request, exc: AspenException) -> JSONResponse:
-    return err.make_response()
+    return exc.make_response()
 
 
 class UnauthorizedException(AspenException):
--- a/src/backend/aspen/api/main.py
+++ b/src/backend/aspen/api/main.py
@@ -1,11 +1,9 @@
 from fastapi import Depends, FastAPI
-from fastapi.security import OAuth2PasswordBearer, OAuth2PasswordRequestForm
 
 from aspen.api.config.config import settings
 from aspen.api.deps import set_db
 from aspen.api.error.http_exceptions import AspenException, exception_handler
 from aspen.api.views.health import router as health_router
-from aspen.api.views.users import router as user_router
 
 
 def get_app() -> FastAPI:
--- a/src/backend/aspen/api/main.py
+++ b/src/backend/aspen/api/main.py
@@ -1,0 +1,31 @@
+from fastapi import Depends, FastAPI
+from fastapi.security import OAuth2PasswordBearer, OAuth2PasswordRequestForm
+
+from aspen.api.config.config import settings
+from aspen.api.deps import set_db
+from aspen.api.error.http_exceptions import AspenException, exception_handler
+from aspen.api.views.health import router as health_router
+from aspen.api.views.users import router as user_router
+
+
+def get_app() -> FastAPI:
+    _app = FastAPI(
+        title=settings.SERVICE_NAME,
+        debug=settings.DEBUG,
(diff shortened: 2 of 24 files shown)


src/backend/etc/supervisord.conf
```diff
@@ -0,0 +1,56 @@
...
+[supervisord]
+logfile=/dev/stdout                             ; supervisord log file
+logfile_maxbytes=0                              ; maximum size of logfile before rotation
+logfile_backups=10                              ; number of backed up logfiles
+loglevel=error                                  ; info, debug, warn, trace
+pidfile=/var/run/supervisord.pid                ; pidfile location
+nodaemon=true                                   ; don't run supervisord as a daemon
+minfds=1024                                     ; number of startup file descriptors
+minprocs=200                                    ; number of process descriptors
+user=root                                       ; default user
```
vincent-czi, line 13: It would be nice to eventually have the container/process not running as root.

src/backend/aspen/api/error/http_exceptions.py
```diff
@@ -0,0 +1,46 @@
+from typing import Any, Mapping, Optional
+
+from fastapi.exceptions import RequestValidationError
+from fastapi.responses import JSONResponse
+from flask import jsonify, make_response, Response
```
vincent-czi: Looks like these are generally unused, with exception of `Response` for typing?

src/backend/aspen/api/error/http_exceptions.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class AspenException(Exception):
+    status_code = 500
+
+    def __init__(self, msg, extra: Optional[Mapping[str, Any]] = None):
+        super().__init__(msg)
+        self.extra = extra
+
+    def make_response(self) -> Response:
```
vincent-czi: Should this actually be typed as returning `Response`? It's returning `JSONResponse` -- is there some type weirdness elsewhere that necessitates it "looking like" the Flask response?

src/backend/aspen/api/main.py
```diff
@@ -0,0 +1,31 @@
...
+        openapi_url="/v2/openapi.json",
+        docs_url="/v2/docs",
+        dependencies=[Depends(set_db)],
+    )
+
+    # _app.include_router(user_router, prefix="/v2/users")
+    _app.include_router(health_router, prefix="/v2/health")
+
+    _app.add_exception_handler(
+        AspenException,
```
vincent-czi, line 25: What happens when it's something other than an `AspenException` that bubbles up? Are we happy with the generic exception response for other errors?

src/backend/aspen/api/main.py
```diff
@@ -0,0 +1,31 @@
...
+        debug=settings.DEBUG,
+        openapi_url="/v2/openapi.json",
+        docs_url="/v2/docs",
+        dependencies=[Depends(set_db)],
+    )
+
+    # _app.include_router(user_router, prefix="/v2/users")
+    _app.include_router(health_router, prefix="/v2/health")
+
+    _app.add_exception_handler(
```
vincent-czi, line 24: When I was trying to figure out the answer to what happens with other exceptions, I think this might not be the ordinary way to add custom exception handlers in FastAPI? They don't mention it in their docs and the [docs section on that](https://fastapi.tiangolo.com/tutorial/handling-errors/#install-custom-exception-handlers) aspect instead talks about doing it via a decorator. It looks like `add_exception_handler` is inherited from Starlette, but the FastAPI codebase never has that anywhere. Probably not a big deal, but might be a hair more future-proof to use the decorator instead.

src/backend/aspen/api/views/health.py
```diff
@@ -0,0 +1,15 @@
+from typing import List
```
vincent-czi: Various imports here are unused

src/backend/aspen/api/views/health.py
```diff
@@ -0,0 +1,15 @@
...
+
+import sqlalchemy as sa
+from fastapi import APIRouter, HTTPException
+from pydantic import parse_obj_as
+
+from aspen.api.deps import get_db
```
vincent-czi: I figure `get_db` is an unused import, but I'm interested if the health check is intended to also do a check on the DB connection (like reading a static record or something to verify DB is healthy). I've only seen health checks for the container itself, so if the intent is broader, tell me more for my own learning, please.

src/backend/aspen/api/views/users.py
```diff
@@ -0,0 +1,26 @@
...
+from aspen.api.schemas.users import User as userschema
+from aspen.api.schemas.users import Users as userlistschema
+from aspen.database.models.usergroup import User
+
+router = APIRouter()
+
+
+@router.get("/", response_model=userlistschema)
+async def list_users() -> userlistschema:
+    objects = await User.all()
```
vincent-czi, line 12: If I'm reading the rest of the code correctly, this wouldn't actually expose anything beyond a list of `agreed_to_tos`, but it does make me nervous that we could launch this, later add actual User data being pulled, but have forgotten to implement auth and accidentally expose a listing of all users until we circle back to implement auth.

src/backend/aspen/api/views/users.py
```diff
@@ -0,0 +1,26 @@
...
+    return userlistschema.parse_obj({"items": objects})
+
+
+@router.get("/{user_id}/", response_model=userschema)
+async def get_user(user_id: int) -> userschema:
+    instance = await User.get_by_id(user_id)
+    if not instance:
+        raise HTTPException(status_code=404, detail="Deck is not found")
```
vincent-czi: "Deck is not found"?

src/backend/aspen/database/models/mixins.py
```diff
@@ -1,19 +1,58 @@
 import datetime
 import enum
+from typing import Any, Dict, List, NoReturn, Optional, Tuple, Type, TypeVar
 
+import sqlalchemy as sa
 from sqlalchemy import Column, Integer
 
+from aspen.api.deps import get_db
```
vincent-czi, line 9: Makes me kind of nervous to have code that the Flask app depends be dependent on the FastAPI side of things. The `get_db` call doesn't actually do anything with FastAPI **and** it looks like it only gets used in async stuff that I assume won't be used by Flask. But just `get_db` coming from that side of the divide makes me worried. If we see issues during the switchover and need to quickly revert to a pre-FastAPI integration I'm worried about this getting hairy. That said, I don't know if it's really worth extracting those deps out of the FastAPI app folder and into somewhere else...

vincent-czi (approved):
Overall, the integration looks good. Nice work, this is a major switchover. I do think it still needs a bit more cleaning up, but considering that this is the first step in an ongoing process, I think it's not a blocker and something that can be polished up over the next few iterations.

src/backend/aspen/api/error/http_exceptions.py
jgadling: Good catch! patching...

src/backend/aspen/api/views/health.py
jgadling: I'd like to have health checks do a quick db connectivity check in the future, but for now this is just an unused import.

src/backend/etc/supervisord.conf
jgadling: Yeah I agree, though that's going to take a little bit of thinking. I'll add a ticket for it.

src/backend/aspen/database/models/mixins.py
jgadling: Yeah it's not ideal, but it's just an import dependency (none of the fastapi code will be executed by flask) -- I think the risk of problems is fairly low here.

jgadling: @jgadling todo -- doublecheck cli/db.py works with this add.

lvreynoso (approved):

PR #715, opened 2021-10-01
Switch from pipenv to poetry

jgadling:
### Summary:
- **What:** `Remove any references to pipenv and use poetry intead`
- **Env:** `https://poetry2-frontend.dev.genepi.czi.technology/`

### Notes:
This is an initial change preparing for a fastapi migration. Poetry is significantly better supported by the python community than Pipenv, so this change allows us to use a single python package management system for both flask and fastapi.

This PR also installs several fastapi-only dependencies before they're technically needed. It's a bit overkill, but this lets us get early feedback about whether the new or changed deps will have a negative impact on the flask app.

### Checklist:
- [ ] I added labels to my PR


lvreynoso (approved):
🎉 

Awesome job!

PR #717, opened 2021-10-04
[sc161837] newtablink component

mmmmmmaya:
### Summary:
- **What:** Small refactor to replace links that open in new tabs with a single component
- **Why:** DRY
- **Ticket:** [sc-161837](https://app.shortcut.com/genepi/story/161837)
- **Env:** [pending]

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/frontend/src/common/components/library/data_subview/components/TreeCreateHelpLink/index.tsx
```diff
@@ -8,7 +8,7 @@ const TreeCreateHelpLink = (): JSX.Element => {
     "https://docs.google.com/document/d/1_iQgwl3hn_pjlZLX-n0alUbbhgSPZvpW_0620Hk_kB4/edit?usp=sharing";
   return (
     <StyledDiv>
-      <StyledLink target="_blank" rel="noopener" href={HREF}>
+      <StyledLink href={HREF}>
```
vincent-czi: I feel like for clarity, new tab opening `StyledLink`s should get renamed to `StyledNewTabLink`s. When I first thought this change, I thought there was an accident, and it wasn't until I read the matched `style.ts` that I realized that was where the component change-out had occurred.

src/frontend/src/views/Upload/components/Metadata/index.tsx
```diff
@@ -92,9 +92,7 @@ export default function Metadata({
             `Check local requirements for reporting to public health authorities. These may not be met by uploading samples to Aspen.`,
             <p key="1">
               Please read our{" "}
-              <NewTabLink href={ROUTES.PRIVACY}>
-                Privacy Policy
-              </NewTabLink>{" "}
```
vincent-czi: The space got lost here. I'm never totally sure when spaces get auto-inserted or lost, but it looks like without the forced one here the spacing gets lost when rendered:
image

vincent-czi (approved):
Looks good! Much cleaner to read. I think the two comments I put in would be good to address, but they're super minor tweaks.

PR #722, opened 2021-10-06
Sc159865 terms of service lockup

vincent-czi:
### Summary:
- **What:** Fixes issue with Terms of Service page locking up 
- **Ticket:** [sc<159865>](https://app.shortcut.com/genepi/story/<159865>)
- **Env:** COMING SOONISH! `<rdev link>`

### Notes:

The issue was caused by two different bugs in concert:
1. Pages that should only be seen by logged-in users who have agreed to ToS are behind `useProtectedRoute` calls. ThisThat func fires off a redirect if they aren't that. If not logged-in, redirects to homepage, if logged-in but not ToS-agreed, redirects to ToS page. However, the `useProtectedRoute` was doing this with two different effect hooks -- one for each redirect -- and there was a bug that let them both fire simultaneously. For some reason, the ToS redirect took precedence (I _think_ because it was later in the code, but I'm not sure). This caused a not logged-in user who went to a URL like `/data/samples` to wind up getting redirected to the ToS page. This wouldn't have caused the lockup on its own, though...
2. The ToS page had no concept of the user's login status. It did not actually try to fetch and read the user's data, it only tried to write to the user's data. That meant when a not logged-in user from (1) showed up on the ToS page and tried to click "Agree", the backend would flip out because the frontend was requesting to write user data to a non-user, and the page would then lock up.

Fixing (1) probably would have handled 99% of cases, since the only other way a not logged-in user could get to the ToS page would be to directly type its URL into the browser. However, I also fixed (2) so that ToS only renders for logged-in users who have not agreed to ToS yet. If user isn't in that overlap, ToS now redirects them elsewhere. [The redirect part was actually an open product question when I started, so I had a Slack conversation with Olivia and Janeece and they were happy with that approach.]

Also, not related to the bug, but while I was working on (2), I did some refactoring on the ToS page component. It had a lot of state calls that were effectively duplicating the info we could get directly from the `useUpdateUserInfo` mutation call, so I consolidated to using that by itself.

---

Additionally, and not directly related to this issue, we have another sorta-kinda bug with `fetch` requests for info when a user is not logged in (user info, sample table, etc) kicking up errors. Phoenix and I traced this for awhile -- seems to come down to the backend is trying to 302 redirect when the request is not authed. I'm unclear on how it should be working. In one way, it actually kind of makes sense to break: it was expecting JSON, but instead got a 302 redirect? But ... couldn't it also follow that redirect? ... But would it be the AJAX fetch following the redirect, or the user's entire browser? It's arguable that the backend should actually be throwing up 401 or 403 or 404 if not authed instead, but a 302 redirect also makes sense in its own way, I just don't know if it can work in the context of AJAX calls. I'm curious to investigate `fetch` a bit more. I don't think I fully understand what's going on here and I'm still surprised at how hard it was erroring and what the message was (`TypeError: Failed to fetch` [_not_ a TS Type, btw]).

Because all the data fetch calls happen async and kick off together, it means that visiting `/data/samples` when not logged-in throws up a bunch of errors in the console before the user gets redirected to the homepage to go do a login. However, while it's freaking out, none of it is actually visible to an ordinary user? The errors all die pretty quietly (assuming they don't open the console), so I think we can ignore this problem for now. But it might be nice to adjust our backend responses for non-authed requests and put in some handling on the frontend for how we handle those cases.

Regardless, none of that affects the specific bug this PR is about. Just an interesting thing that got noticed while working on the issue.

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR


mmmmmmaya (approved):
Love the write up and all the comments 👏

src/frontend/src/common/queries/auth.ts
```diff
@@ -70,24 +75,47 @@ export function useUserInfo(): UseQueryResult<UserResponse, unknown> {
...
+ * it will not prevent render, it just can then push the user away from that page.
+ *
+ * Returns: (obj) result of `useUserInfo` call
+ *   [Could be refactored to have no return, but right now some components use that userInfo.
+ *    Might be better to be more explicit though, have components directly call useUserInfo?]
+ *
+ * IMPORTANT NOTE:
+ * This function does not provide "real" security. Think of it as a picket fence.
+ * It's only intended to redirect users to where they should be. It does not prevent
+ * protected routes/components from firing and would be easy for an attacker to circumvent.
```
phoenixAja, line 99: we should make some follow up tickets to address the issues raised in this note. It's interesting that the components are still rendering even when the user is not authed, do you think this could `be` due to the backend sending a 302 response instead of a 40[1,3,4] response?

phoenixAja (approved):
thanks for adding the details on this. We should create some tickets or brainstorm how best to address the security issues you brought up.

src/frontend/src/common/queries/auth.ts
vincent-czi: It's not crazy that they render -- the components don't know the redirect is coming. I changed how the ToS component/page works to avoid rendering before the userInfo gets returned. But whether or not the component renders, it doesn't really change anything from a security perspective, it's just a question of if the component tries to (briefly, hopefully) render in a broken state.

The render aspect isn't due to the 302 vs 40x response from backend. It's just that the components render before the backend calls resolve. Those fetches are async, so they start fetching once the component mounts. It goes through a couple render cycles potentially before the backend calls resolve. It's a question of if we want our components to avoid rendering (well, kinda-sorta render by returning `null`) until those backend calls resolve, or if we want to render something in an intermediate state then finish populating it with data once the response comes in. There's nothing wrong with the second approach -- it's probably preferable -- it's just that right now our stuff isn't really built to gracefully handle the case where the API call gets made and then rejected.

vincent-czi:
@phoenixAja I don't actually think these rank as security issues. At the end of the day, any security in the frontend app (at least the React portion) is beatable. The user has full control over their side of things, so we can't actually secure anything if it's only on that side. From what I've seen, our backend is well-secured, and that's all that really matters. I do think the fetch 302 vs 40x aspect is worth discussing, but I wouldn't call that security, per se.

My warnings above `useProtectedRoute` were more about just trying to make it absolutely clear that it shouldn't be trusted for security on its own. I'm not actually worried about any of us making a mistake along those lines, but I try to make anything security-related crystal clear since the costs can be so sky-high if something gets misunderstood.

vincent-czi:
Forgot to post up the rdev link. Whoops!
It's https://bugfix-tos-lockup-frontend.dev.genepi.czi.technology
I just verified everything is working as expected on rdev, going to merge when auto-checks complete

vincent-czi:
Had some failing jobs/builds that might need rollback, holding off on merging until that's resolved.

PR #723, opened 2021-10-06
FastAPI auth

jgadling:
### Summary:
- **What:** This adds authentication to the fastapi backend, and provides a simple way to require authentication for endpoints.
- **Ticket:** [sc161314](https://app.shortcut.com/genepi/story/161314)
- **Env:** `n/a`

### Notes:
This updates the fastapi app to support 2 different authentication methods, and one login flow, identical to what our Flask application supports:
- Authentication via signed cookie
- Authentication via JWT (primarily for Device Authorization flow)
- Login via OAuth callbacks

In addition, it also adds a single auth-required endpoint: /v2/users/me that queries the db for information about the current user.

### Checklist:
- [ ] I added labels to my PR


oauth/clients-config.json
```diff
@@ -6,7 +6,9 @@
     "AllowedGrantTypes": ["authorization_code", "urn:ietf:params:oauth:grant-type:device_code"],
     "RedirectUris": [
       "https://backend.genepinet.local:3000/callback",
-      "http://backend.genepinet.local:3000/callback"
+      "http://backend.genepinet.local:3000/callback",
+      "http://backend.genepinet.local:3000/v2/auth/callback",
+      "https://backend.genepinet.local:3000/v2/auth/callback"
```
jgadling, line 11: Tell our oauth service (oidc-server-mock) that we have some additional valid login callback URL's.

src/backend/aspen/api/auth.py
```diff
@@ -0,0 +1,72 @@
...
+from sqlalchemy.ext.asyncio import AsyncSession
+from sqlalchemy.orm import joinedload
+from sqlalchemy.orm.exc import NoResultFound
+from sqlalchemy.orm.query import Query
+from starlette.requests import Request
+
+from aspen.api.config.config import settings
+from aspen.api.deps import get_db
+from aspen.auth.device_auth import validate_auth_header
+from aspen.database.models import Group, User
```
jgadling, line 14: Most of the functionality in this file is ~identical to the flask app auth code in src/backend/aspen/app/app.py, but with some minor tweaks to support fastapi and async.

src/backend/aspen/api/auth.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def get_usergroup_query(session: AsyncSession, user_id: str) -> Query:
+    return (
+        sa.select(User)  # type: ignore
+        .options(joinedload(User.group).joinedload(Group.can_see))  # type: ignore
+        .filter(User.auth0_user_id == user_id)  # type: ignore
+    )
```
jgadling, line 22: This is SQLAlchemy's 2.0 preferred query syntax. I haven't figured out how to make it play nice with mypy yet, but that will need to happen soon since fastapi will need to start executing more queries!

src/backend/aspen/api/auth.py
```diff
@@ -0,0 +1,72 @@
...
+    found_auth_user = await setup_userinfo(auth0_user_id)
+    if not found_auth_user:
+        # login attempt from user not in DB
+        # TODO - redirect to login.
+        raise Exception("No Userid Found!")
+    # There's a starlette-context module that allows us to manage
+    # per-request data without depending on having a request object
+    # available. For now we seem to have `request` when we need it,
+    # but we can change this if necessary.
+    request.state.auth_user = found_auth_user
```
jgadling, line 72: For now we're using `request.state` as an alternative to the `flask.g` request-global functionality. There are some limitations with this that the starlette-context module aims to solve, but I haven't run into those limitations yet, so let's keep things simple until we do.

src/backend/aspen/api/main.py
```diff
@@ -1,11 +1,29 @@
...
+    deployment = os.getenv("DEPLOYMENT_STAGE")
+    frontend_url = os.getenv("FRONTEND_URL")
+
+    if deployment not in ["staging", "prod"]:
+        allowed_origins.extend(
+            [r"http://.*\.genepinet\.local:\d+", r"^http://localhost:\d+"]
+        )
+    if frontend_url:
+        allowed_origins.append(frontend_url)
+    return allowed_origins
```
jgadling, line 26: Get ready to hook up CORS

src/backend/aspen/api/main.py
```diff
@@ -17,9 +35,23 @@ def get_app() -> FastAPI:
...
+        allow_credentials=True,
+        max_age=600,
+        allow_methods=["*"],
+    )
+    _app.add_middleware(SessionMiddleware, secret_key=settings.SECRET_KEY)
+
     # Warning - do not enable this route!
-    # _app.include_router(users.router, prefix="/v2/users")
+    _app.include_router(
+        users.router, prefix="/v2/users", dependencies=[Depends(get_auth_user)]
```
jgadling, line 51: The dependencies=[] block means that all listed dependencies need to be satisfied before executing the endpoints included in a given router. In this case, auth is required for all `/v2/users` endpoints.

More info: https://fastapi.tiangolo.com/tutorial/dependencies/dependencies-in-path-operation-decorators/

src/backend/aspen/api/middleware/session.py
```diff
@@ -0,0 +1,85 @@
...
+
+from flask.json.tag import TaggedJSONSerializer
+from itsdangerous import URLSafeTimedSerializer
+from itsdangerous.exc import BadSignature
+from starlette.datastructures import MutableHeaders
+from starlette.requests import HTTPConnection
+from starlette.types import ASGIApp, Message, Receive, Scope, Send
+
+
+class SessionMiddleware:
```
jgadling, line 11: This middleware is heavily based on [starlette's session middleware](https://github.com/encode/starlette/blob/master/starlette/middleware/sessions.py), but with some small modifications to support [Flask's secure cookie implementation](https://github.com/pallets/flask/blob/main/src/flask/sessions.py#L316). This means that session cookies created by the flask app will be readable by the fastapi app, and vice versa.

src/backend/aspen/api/views/auth.py
```diff
@@ -0,0 +1,68 @@
...
+auth0 = oauth.register(
+    "auth0",
+    client_id=settings.AUTH0_CLIENT_ID,
+    client_secret=settings.AUTH0_CLIENT_SECRET,
+    api_base_url=settings.AUTH0_BASE_URL,
+    access_token_url=settings.AUTH0_ACCESS_TOKEN_URL,
+    authorize_url=settings.AUTH0_AUTHORIZE_URL,
+    client_kwargs=settings.AUTH0_CLIENT_KWARGS,
+)
+
```
jgadling, line 27: These endpoints implement a basic oauth login flow, and are intended to be functionally identical to their counterparts in the flask app. This implementation borrows heavily from:
https://github.com/authlib/demo-oauth-client/tree/master/fastapi-google-login

src/backend/aspen/api/views/users.py
```diff
@@ -1,21 +1,15 @@
...
-    objects = await User.all()
-    return userlistschema.parse_obj({"items": objects})
-
-
-@router.get("/{user_id}", response_model=userschema)
-async def get_user(user_id: int) -> userschema:
-    instance = await User.get_by_id(user_id)
+@router.get("/me", response_model=userschema)
+async def get_current_user(request: Request) -> userschema:
+    instance = await User.get_by_id(request.state.auth_user.id)
```
jgadling, line 12: A minimum-viable endpoint that uses authentication context and returns whether or not a user has agreed to the TOS.

src/backend/aspen/app/app.py
```diff
@@ -91,61 +85,6 @@ def handle_bad_request(err):
 )
 
 
-class InsecureJwksFetcher(JwksFetcher):
```
jgadling, line 94: I moved the token validation code to a place where both the flask and fastapi apps can use it.

src/backend/pyproject.toml
```diff
@@ -48,6 +48,7 @@ asyncpg = "^0.24.0"
 fastapi = "^0.68.1"
 uvicorn = {extras = ["standard"], version = "^0.15.0"}
 gunicorn = "^20.1.0"
+httpx = "^0.19.0"
```
jgadling, line 51: Not sure why this wasn't already installed, but it's needed.

src/backend/aspen/api/auth.py
lvreynoso: `starlette-context` seems easy enough to add, but I'll defer to you - we can always add it later.

lvreynoso (approved):
lgtm

src/backend/aspen/api/main.py
vincent-czi: Reading over the FastAPI docs, it seems like they generally [talk about putting the security dependencies into the actual path function](https://fastapi.tiangolo.com/tutorial/security/first-steps/#security-first-steps), rather than handling it at a higher level.

```
@app.get("/items/")
async def read_items(token: str = Depends(oauth2_scheme)):
    return {"token": token}
```

If we want the auth_user aspect to totally global, this is a bit of a hassle. But -- I _think_ -- by doing it this way we can actually use the returned value from the functions we are `Depends`-ing on. So if we have `get_auth_user` return the `found_auth_user` instead, we can effectively make it both param and security declaration on each path function that needs security.

It also looks like you can chain these `Depends` parts way deep. The path functions `Depends` on some callable, but that callable also `Depends` on another thing, etc, etc. So for other paths that rely on auth but are using it to dig up some other set of info than the user id, we could potentially abstract that out. I'm on the fence on whether I like that or not, but it is interesting.

jgadling: Per-path deps are normal for simple/small demo apps, but they've got a doc that talks about more interesting applications, with some examples of adding deps at the routing level:

https://fastapi.tiangolo.com/tutorial/bigger-applications/#include-the-apirouters-for-users-and-items

I agree that being able to use the values directly is nicer though -- I'll look into whether we can actually access dependencies from further up in the call stack.

vincent-czi: Also, I don't know if we would ever need scopes / permissions, but if we think it's likely we might want to look into the `Security` call instead. It's probably not worth it -- the only real value I can think of in the near-term would be having an `admin` scope -- but if we think it would be useful, it might be nice to learn how to use it and establish the pattern now during the transition.

It took me quite awhile to make sense of how the use the scopes, but I finally made sense of it when I read the code sample under this section: https://fastapi.tiangolo.com/advanced/security/oauth2-scopes/ -- Focus on the parts near the bottom of that giant code-block: `get_current_user`, then `get_current_active_user` and `read_own_items`. I think we wouldn't get our scopes from the authentication process, but when we pull the user's record in the DB, we could effectively get scope info based on `group_admin` and `system_admin`.

Still, I think it's probably not worth it. But if permissioning like that is actually an upcoming need, could be worth the effort (and I'd be happy to work on it as needed).

vincent-czi: Finally, I was pretty surprised that the way to deny auth was explicitly and only to `raise` an exception. It's not crazy-town or anything,  but my first instinct was to assume that the auth process could also deny by something else coming out of the auth Depends (like `None` or a falsey or something). Now that I see their design is not that, I get it, but I think this warrants some very explicit documentation in `auth.py` saying: "If you need to deny access, you MUST raise an exception". I could easily see myself having taken the wrong thing for granted if I was rushed and had to implement a new auth func.

src/backend/aspen/api/middleware/session.py
vincent-czi: Excellent! That's great news!

Will we eventually pull this code out and convert to something FastAPI only when we leave Flask behind? Or will we want to keep this in place even after we're divested of Flask?

src/backend/aspen/api/auth.py
```diff
@@ -0,0 +1,72 @@
...
+    )
+
+
+async def setup_userinfo(user_id):
+    # sentry_sdk.set_user( { "requested_user_id": user_id, })
+    session = get_db()
+    try:
+        userquery = get_usergroup_query(session, user_id)
+        userwait = await session.execute(userquery)
+        user = userwait.unique().scalars().first()
```
phoenixAja, line 31: why is `.scalars `necessary?

src/backend/aspen/api/auth.py
```diff
@@ -0,0 +1,72 @@
...
+            )
+            auth0_user_id = payload["sub"]
+        except TokenValidationError as err:
+            logging.warn(f"Token validation error: {err}")
+    elif "profile" in request.session:
+        auth0_user_id = request.session["profile"]["user_id"]
+    # Redirect to Login page
+    if not auth0_user_id:
+        # TODO - redirect to login.
+        raise Exception("No Userid Found!")
```
phoenixAja: i think you might run into a key error here? 
```suggestion
        auth0_user_id = request.session["profile"].get("user_id")
    # Redirect to Login page
    if not auth0_user_id:
        # TODO - redirect to login.
        raise Exception("No Userid Found!")
```

src/backend/aspen/api/config/config.py
```diff
@@ -70,9 +70,21 @@ def SECRET_KEY(self) -> str:
...
+            return f"{self.AUTH0_BASE_URL}/v2/logout"
+        else:
+            return f"https://{self.AUTH0_DOMAIN}/Account/Logout"
+
     @property
     def AUTH0_CALLBACK_URL(self) -> str:
-        raise NotImplementedError()
+        flask_env = os.environ.get("FLASK_ENV")
+        if flask_env == "production":
+            return f"{os.getenv('API_URL', '')}/v2/auth/callback"
```
phoenixAja: why do we have the default as an empty string? will this error if that env variable is not present and no default was provided?

phoenixAja: if so i would rather we have it error, it seems more clear that way?

jgadling: It's a bad hack to make mypy happy.... I'll do something smarter. 😅

src/backend/aspen/api/middleware/session.py
```diff
@@ -0,0 +1,85 @@
...
+    def __init__(
+        self,
+        app: ASGIApp,
+        secret_key: str,
+        session_cookie: str = "session",
```
phoenixAja, line 16: is this supposed to be something more intense/ not easily guessable?

src/backend/aspen/api/middleware/session.py
```diff
@@ -0,0 +1,85 @@
...
+from itsdangerous.exc import BadSignature
+from starlette.datastructures import MutableHeaders
+from starlette.requests import HTTPConnection
+from starlette.types import ASGIApp, Message, Receive, Scope, Send
+
+
+class SessionMiddleware:
+    def __init__(
+        self,
+        app: ASGIApp,
```
phoenixAja, line 14: why `ASGIApp`? (just curious about the naming)

src/backend/aspen/api/middleware/session.py
jgadling: Nah, it's just the name of the cookie. It's supposed to match the name of the cookie that the flask app uses (which is just "session" by default)

src/backend/aspen/api/middleware/session.py
```diff
@@ -0,0 +1,85 @@
...
+                    # We have session data to persist.
+                    data = self.encode_flask_cookie(scope["session"])
+                    headers = MutableHeaders(scope=message)
+                    header_value = "%s=%s; path=%s; Max-Age=%d; %s" % (
+                        self.session_cookie,
+                        data,
+                        path,
+                        self.max_age,
+                        self.security_flags,
+                    )
```
phoenixAja: small nit, but i would rather use a format string here

src/backend/aspen/api/auth.py
jgadling: Just makes working with SQLAlchemy's result objects a little easier:
https://stackoverflow.com/questions/66974718/q-sqlalchemy-1-4-2-0-problem-converting-results-from-session-executestatemen

By default SA it will return a tuple-like object for each row, and scalars() gives us the objects directly.

src/backend/aspen/api/views/auth.py
```diff
@@ -0,0 +1,68 @@
...
+    try:
+        token = await auth0.authorize_access_token(request)
+    except OAuthError:
+        raise ex.UnauthorizedException("Invalid token")
+    userinfo = token.get("userinfo")
+    if userinfo:
+        # Store the user information in flask session.
+        request.session["jwt_payload"] = userinfo
+        request.session["profile"] = {
+            "user_id": userinfo["sub"],
```
phoenixAja, line 45: why is sub the key for user_id?

src/backend/aspen/api/views/auth.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+@router.get("/logout")
+async def logout(request: Request):
+    # Clear session stored data
+    request.session.pop("jwt_payload", None)
+    request.session.pop("profile", None)
+    # Redirect user to logout endpoint
+    params = {
+        "returnTo": os.getenv("FRONTEND_URL"),
```
phoenixAja, line 65: should this be `return_to` to match the other camel case keys?

src/backend/aspen/api/config/config.py
phoenixAja: mypy 😠

src/backend/aspen/api/views/auth.py
jgadling: It's just part of the oauth spec, it's a `sub`ject claim.

src/backend/aspen/api/views/auth.py
jgadling: It's part of the auth0 api:
https://auth0.com/docs/login/logout/redirect-users-after-logout

src/backend/aspen/api/middleware/session.py
jgadling: Changing our cookie format will force a mass logout event -- I need to check on what Flask is currently using for cookie expiration, maybe this is already basically happening already. So I think we'd try deprecate the explicit dependency on flask libs if we can, but the flask secure cookie implementation is solid, so there's no particular need to deprecate that format.

vincent-czi: Looking at my browser's cookie for Staging and over the [Flask docs](https://flask.palletsprojects.com/en/2.0.x/api/?highlight=session#flask.session.permanent) for `session`, it looks like the auth cookie expires on browser `Session`. So it probably depends on our users usage, but I would assume for a good chunk of our users they're effectively permanent (I hadn't known until a few minutes ago that the "continue where I left off" options in browsers restores those Session cookies too. Now I know!).

I'm totally happy to keep the Flask session cookie approach, just hopefully we can move away from needing to keep the underlying Flask lib.

PR #728, opened 2021-10-08
little bit of state clean up

mmmmmmaya:
### Summary
- **What:** A little bit of state clean up for data subview. 
- **Why:** I noticed some of the state being maintained by this component was actually state for the download modal. I chose to move it into DownloadModal to make the component more self contained and easier to understand. It also works as part of the data subview clean up.
- **Ticket:** [[sc-164215]](https://app.shortcut.com/genepi/story/164215)
- **Env:** https://mrclean-frontend.dev.genepi.czi.technology

### Testing
Basically check that the download modal functionality hasn't changed at all

### Demos
Should be no visual or functional changes.

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
@@ -28,12 +28,6 @@ import {
 interface Props {
   sampleIds: string[];
   failedSamples: any[];
-  isMetadataSelected: boolean;
-  setMetadataSelected: (prevState: boolean) => void;
-  isFastaSelected: boolean;
-  setFastaSelected: (prevState: boolean) => void;
-  isFastaDisabled: boolean;
-  setFastaDisabled: (prevState: boolean) => void;
   tsvData: [string[], string[][]] | undefined;
   open: boolean;
   onClose: () => void;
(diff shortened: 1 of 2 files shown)


src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
```diff
@@ -93,6 +85,12 @@ const DownloadModal = ({
     setFastaSelected(!isFastaSelected);
   };
 
+  const handleCloseModal = () => {
+    setFastaSelected(false);
```
vincent-czi, line 89: I might not be understanding how the DownloadModal works, but does the DownloadModal unmount with the `onClose` call? If so, I'm unsure about the benefit of calling these set_Selected(false) here. Seems like the unmount will inherently set them false by just knocking the whole component into oblivion?

That said, I don't actually know the larger use, if you think it makes sense, disregard what I'm saying. Also, heck, I can even argue against myself as this being a form of future-proofing if the onClose unmount stops happening but we would still want the selections to false out.

vincent-czi (approved):
Looks good! One super minor comment that I am not invested in, but if you think it's reasonable, yay.

src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
mmmmmmaya: I think the component does not actually unmount. The parent just tells the component whether or not to display itself. So, we show the component only when parent says `open=true`, but the component itself does not get destroyed, so it maintains the state.

Would be a different story if it were like
```
// ParentComponent/index.tsx

...
if (someParentState) return null;
return <DownloadModal />
```

PR #731, opened 2021-10-11
public id double dash bug fix

phoenixAja:
### Summary:
- **What:** fix double dash issue in automatically generated public ids
- **Ticket:** [sc165788](https://app.shortcut.com/genepi/story/165788)
- **Env:** no rdev

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/database/models/sample.py
+++ b/src/backend/aspen/database/models/sample.py
@@ -61,7 +61,7 @@ def create_public_ids(
     session :
         An open DB session object
     number_needed :
-        The number of czb_ids to generate
+        The number of aspen public ids to generate
 
     Returns
     --------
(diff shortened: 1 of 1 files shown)


src/backend/aspen/database/models/sample.py
```diff
@@ -76,10 +76,12 @@ def create_public_ids(
         next_id = 0
     next_id += 1
     ids: List[str] = []
+    # some group prefixes already have a dash, but if they don't add a dash to the end
```
vincent-czi, line 79: Will there be any issues elsewhere with an ID-->group parser or something? I'm thinking if we _always_ assume that the group prefix would be its literal prefix (including dash), and that would cause problems with something where the "true" prefix is, say `abc-`, but we parse an ID that is `hCoV-19/usa/abc-SomeNextID/2021`, something else would barf because it would read the group's prefix as `abc` instead of its true value of `abc-`? It's definitely an edge case, but I just don't know if the code base ever reads the ID the other direction to extract info.

vincent-czi: Similarly, would we ever have to worry about two groups where one is `abc` and the other is `abc-`? If that happens, then we could get identical IDs for different groups. Don't know if that's a problem, or if we're sure we won't have that kind of overlap in group prefixes.

vincent-czi (approved):
Looks good. I've got two extreme edge case comments. If you don't think those apply, all good!

src/backend/aspen/database/models/sample.py
phoenixAja: i don't think there would ever be a case where there would be 2 groups (one with a dash and one without a dash), aspen admins are supplying the group prefixes directly in the DB, so this would not be an issue (but maybe something to consider if groups are allowed to decide on their own prefixes in the future).

PR #732, opened 2021-10-11
Endpoint and data for UShER options

vincent-czi:
### Summary:
- **What:** Migration plus data population for providing UShER options, along with API endpoint to pull them
- **Ticket:** [sc<161393>](https://app.shortcut.com/genepi/story/<161393>), [sc<161377>](https://app.shortcut.com/genepi/story/<161377>)
- **Env:** No rdev

### Demos:
image

### Notes:
Getting the options requires auth

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/app/views/usher.py
+++ b/src/backend/aspen/app/views/usher.py
@@ -1,6 +1,8 @@
 """Views for handling anything related to UShER"""
-from typing import Iterable, Sequence, Mapping, Any
+from typing import Any, Iterable, Mapping, Sequence
+
 from flask import g, jsonify
+
 from aspen.app.app import application, requires_auth
 from aspen.database.models.usher import UsherOption
 
@@ -9,11 +11,14 @@ from aspen.database.models.usher import UsherOption
 @requires_auth
 def get_usher_tree_options():
     """DOCME"""
-    print('this works, wacky')
-
-    usher_options: Iterable[UsherOption] = g.db_session.query(UsherOption).order_by(
-        # Lowest priority is most important, lessens as priority ascends
-        UsherOption.priority.asc()).all()
+    usher_options: Iterable[UsherOption] = (
+        g.db_session.query(UsherOption)
+        .order_by(
+            # Lowest priority is most important, lessens as priority ascends
+            UsherOption.priority.asc()
+        )
+        .all()
+    )
 
     serializable_options: Sequence[Mapping[str, Any]]
     serializable_options = [option.to_dict() for option in usher_options]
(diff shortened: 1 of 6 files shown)


src/backend/aspen/database/models/usher.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class UsherOption(idbase, DictMixin):  # type: ignore
+    """A possible tree creation option when using UShER."""
+
+    __tablename__ = "usher_options"
+
+    description = Column(String, unique=True, nullable=False)
+    value = Column(String, unique=True, nullable=False)
+    # `priority` is order we display options to user. LOWEST number is max priority.
```
jgadling, line 15: 👍 for the comment!

jgadling (approved):

src/backend/aspen/app/views/usher.py
```diff
@@ -0,0 +1,30 @@
...
+        .order_by(
+            # Lowest priority is most important, lessens as priority ascends
+            UsherOption.priority.asc()
+        )
+        .all()
+    )
+
+    serializable_options: Sequence[Mapping[str, Any]]
+    serializable_options = [option.to_dict() for option in usher_options]
+    return jsonify({"usher_options": serializable_options})
```
jgadling: Very small nit -- it's nice to define response serializers for new endpoints where possible: https://github.com/chanzuckerberg/aspen/blob/trunk/src/backend/aspen/app/serializers.py

vincent-czi: Oh, that's great! I was trying to follow conventions I saw in other endpoints I was referencing, but I hadn't seen this approach. I actually prefer doing it this way. I'll add one.

phoenixAja (approved):
looks good pending response serializer!

src/backend/aspen/app/views/usher.py
```diff
@@ -0,0 +1,30 @@
...
+        g.db_session.query(UsherOption)
+        .order_by(
+            # Lowest priority is most important, lessens as priority ascends
+            UsherOption.priority.asc()
+        )
+        .all()
+    )
+
+    options_schema = UsherOptionResponseSchema(many=True)
+    return jsonify({"usher_options": options_schema.dump(usher_options)})
```
jgadling, line 30: I'm not super excited about nesting the response schema like this -- We can certainly define a second response schema that returns a list of another type of schemas.

vincent-czi: I spent awhile fighting this, and this felt like the least boiler-plate way to do it. Would you mind outlining what you're thinking? Doing this with pre-dump feels worse to me. I can create a second schema of `UsherOptionsResponseSchema` that has a single `fields.List` in it, but then I'd have to create a dictionary to put the query result into so that I can then dump it, which seemed overall less clear to me.

jgadling: Ahh and indeed that's the way [the marshmallow docs](https://marshmallow.readthedocs.io/en/stable/examples.html) represent this type of output. lgtm!

PR #733, opened 2021-10-11
Do a better job of caching config properties.

jgadling:
### Summary:
- **What:** Cache apiv2 config settings

### Notes:
I'd previously left a bunch of TODO's in the code for where we were handling settings poorly and not caching computed values as well as we should. This PR improves on that by:
- Explicitly caching ~all computed settings
- Not instantiating the Settings() module multiple times
- Using FastAPI's dependency injection system to manage the application configs in some cases. I think we can take this further, but I'm still getting used to it. 

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/src/backend/Makefile
+++ b/src/backend/Makefile
@@ -1,7 +1,6 @@
 STYLE_CHECK_PYTHON_CODE_DIRECTORIES = .
 STYLE_CHECK_PYTHON_CODE_SKIPPED_DIRECTORIES = third-party
 TYPE_CHECK_BASE_PYTHON_CODE_DIRECTORIES = aspen
-TYPE_CHECK_INDIVIDUAL_PYTHON_CODE_DIRECTORIES = database_migrations
 
 
 ### CHECK STYLE #############################################
@@ -16,15 +15,9 @@ black:
 isort:
 	isort --profile black --check $(STYLE_CHECK_PYTHON_CODE_DIRECTORIES) --skip $(STYLE_CHECK_PYTHON_CODE_SKIPPED_DIRECTORIES) 
 
-MYPY_TARGETS = $(foreach stylecheckdir, $(TYPE_CHECK_INDIVIDUAL_PYTHON_CODE_DIRECTORIES), mypy-$(stylecheckdir))
-
-mypy: mypy-base $(MYPY_TARGETS)
-
-mypy-base:
-	mypy --ignore-missing-imports $(TYPE_CHECK_BASE_PYTHON_CODE_DIRECTORIES)
-
-$(MYPY_TARGETS): mypy-%: mypy-base
-	mypy --ignore-missing-imports $(TYPE_CHECK_BASE_PYTHON_CODE_DIRECTORIES) $*
+mypy:
+	mypy --ignore-missing-imports aspen
+	mypy --ignore-missing-imports database_migrations
 
 
 ### RUN STYLE #############################################
--- a/src/backend/aspen/api/config/config.py
+++ b/src/backend/aspen/api/config/config.py
@@ -1,6 +1,6 @@
 import json
 import os
-from functools import lru_cache
+from functools import cached_property, lru_cache
 from typing import Any, Mapping
 
 from boto3 import Session
@@ -20,7 +20,7 @@ class Settings(BaseSettings):
 
     ####################################################################################
     # Stack name
-    @property
+    @cached_property
     def STACK_PREFIX(self) -> str:
         remote_prefix = os.environ.get("REMOTE_DEV_PREFIX")
         deployment_stage = os.environ.get("DEPLOYMENT_STAGE")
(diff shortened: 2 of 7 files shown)


src/backend/Makefile
```diff
@@ -16,15 +15,9 @@ black:
...
-mypy: mypy-base $(MYPY_TARGETS)
-
-mypy-base:
-	mypy --ignore-missing-imports $(TYPE_CHECK_BASE_PYTHON_CODE_DIRECTORIES)
-
-$(MYPY_TARGETS): mypy-%: mypy-base
-	mypy --ignore-missing-imports $(TYPE_CHECK_BASE_PYTHON_CODE_DIRECTORIES) $*
+mypy:
+	mypy --ignore-missing-imports aspen
+	mypy --ignore-missing-imports database_migrations
```
jgadling, line 20: This hopefully fixes mypy's blowups in local dev.

src/backend/aspen/api/config/config.py
```diff
@@ -205,45 +194,51 @@ def _AWS_SSM_PARAMETER(self, parameter_suffix) -> Mapping[str, Any]:
...
+    @cached_property
     def NEXTSTRAIN_SPOT_VCPU(self) -> str:
         return self.AWS_NEXTSTRAIN_SFN_PARAMETER["RunSPOTVcpu"]
 
     class Config:
         env_file = ".env"
         env_file_encoding = "utf-8"
+        keep_untouched = (
+            cached_property,
+        )  # https://github.com/samuelcolvin/pydantic/issues/1241
```
jgadling, line 240: This is necessary to make pydantic happy about all the `@cached_property` calls

src/backend/aspen/api/config/config.py
```diff
@@ -205,45 +194,51 @@ def _AWS_SSM_PARAMETER(self, parameter_suffix) -> Mapping[str, Any]:
...
 
 
-settings = Settings()
+@lru_cache()
```
lvreynoso: Even after googling I'm still unclear about the `@lru_cache` syntax - does it need to be called like a function, as you do here? Or are the parens not needed like in line 177?

src/backend/aspen/api/views/auth.py
```diff
@@ -1,38 +1,53 @@
...
-    client_kwargs=settings.AUTH0_CLIENT_KWARGS,
-)
+
+@lru_cache()
+def get_auth0_client() -> StarletteOAuth2App:
+    # TODO - settings is an unhashable type, so we can't make it a Dependency
+    # *and* use the lru_cache decorator on this getter. We should probably use
+    # fastapi's built-in dependency caching instead of lru_cache() here, but
+    # it needs a little more thought first.
+    # see: https://github.com/tiangolo/fastapi/issues/1985
```
lvreynoso: thanks for the comment! could we instantiate a `Settings` object in the scope directly above this function (or at the app entry point and attach it to a global object, like in flask?), then pass in a hashable object containing the needed settings?

src/backend/aspen/api/config/config.py
```diff
@@ -205,45 +194,51 @@ def _AWS_SSM_PARAMETER(self, parameter_suffix) -> Mapping[str, Any]:
...
+def get_settings() -> Settings:
+    settings = Settings()
+    return settings
```
lvreynoso, line 246: Should we call this file `settings.py` and folder `Settings` for clarity?

jgadling: good call, I'll move this!

src/backend/aspen/api/views/auth.py
jgadling: yep, that's definitely one option! I think this section of the code (thus the giant comment) needs a little more work for sure, but I haven't figured out which approach would be best just yet.

src/backend/aspen/api/config/config.py
jgadling: Used to be that we needed to call `lru_cache()` with parens to initialize a cache. Looks like as of python 3.8+ the parens are no longer required though! I'll remove the parens since I think we're using 3.9

src/backend/aspen/api/settings.py
```diff
@@ -205,45 +194,51 @@ def _AWS_SSM_PARAMETER(self, parameter_suffix) -> Mapping[str, Any]:
...
         env_file = ".env"
         env_file_encoding = "utf-8"
+        keep_untouched = (
+            cached_property,
+        )  # https://github.com/samuelcolvin/pydantic/issues/1241
 
 
-settings = Settings()
+@lru_cache
+def get_settings() -> Settings:
```
vincent-czi, line 244: I think it would be good to put a docstring on `Settings` saying that, generally, it shouldn't be called, and instead to use the `get_settings` helper below in the file to get caching benefits. I think it's possible for someone to overlook the helper if they aren't already aware of it.

src/backend/aspen/api/main.py
```diff
@@ -27,12 +26,13 @@ def get_allowed_origins():
...
 
 def get_app() -> FastAPI:
+    settings = get_settings()
     _app = FastAPI(
         title=settings.SERVICE_NAME,
         debug=settings.DEBUG,
         openapi_url="/v2/openapi.json",
         docs_url="/v2/docs",
-        dependencies=[Depends(partial(set_db, settings))],
+        dependencies=[Depends(set_db)],
```
vincent-czi, line 35: It weirds me out that this dependency gets called on every request to the app rather than just when the app is created, but I guess that's just a FastAPI thing.

vincent-czi (approved):
Looks good!

I'd make a GitHub comment on it, but it's not in the scope of this PR so I don't think I can comment on the location: there's an outdated comment in `src/backend/aspen/api/main.py` on line 49: "# Warning - do not enable this route!" for the users route, but that's now enabled and safe, so comment should get deleted.

PR #736, opened 2021-10-14
sample identifier validation endpoint

phoenixAja:
### Summary:
- **What:** adding endpoint to check if a list of samples exist as either public/private identifiers or gisaid strain names in the aspen DB, return a list of missing samples if not found. 
- **Ticket:** [sc166014](https://app.shortcut.com/genepi/story/166014)
- **Env:** no rdev

### Checklist:
- [x] I added labels to my PR


src/backend/aspen/app/views/validation.py
```diff
@@ -0,0 +1,62 @@
...
+    except ValidationError as verr:
+        sentry_sdk.capture_message("Invalid API request to /api/validate/ids", "info")
+        raise ex.BadRequestException(str(verr))
+
+    sample_ids: Iterable[str] = request_data["sample_ids"]
+
+    all_samples: Iterable[Sample] = g.db_session.query(Sample)
+
+    # get all samples from request that the user has permission to use
+    all_samples: Query = authz_sample_filters(all_samples, sample_ids, user)
```
jgadling: Should we move this query into get_missing_and_found_sample_ids() ?

phoenixAja: i thought about doing that but then decided against it because the phylo_run endpoint uses the results from authz_sample_filters to find associated pathogen_genomes.

src/backend/aspen/app/views/api_utils.py
```diff
@@ -220,3 +230,63 @@ def authz_phylo_tree_filters(query: Query, tree_ids: Set[int], user: User) -> Qu
         ),
     )
     return query
+
+
+def get_missing_and_found_sample_ids(
+    sample_ids: Iterable[str], all_samples: Query
+) -> Tuple[Set[str], Set[Any]]:
+    """
+    Check a list of sample identifiers against Sample table public and private identifiers
```
vincent-czi, line 239: Love seeing the docstrings throughout!

src/backend/aspen/app/views/phylo_runs.py
```diff
@@ -56,35 +59,19 @@ def start_phylo_run():
...
-        pathogen_genomes.append(sample.uploaded_pathogen_genome)
-        found_sample_ids.add(sample.private_identifier)
-        found_sample_ids.add(sample.public_identifier)
-    if len(pathogen_genomes) == 0:
-        sentry_sdk.capture_message(
-            f"No sequences selected for run from {sample_ids}.", "error"
-        )
-        raise ex.BadRequestException("No sequences selected for run")
+    # Step 3 - Enforce AuthZ (check if user has permission to see private identifiers)
+    all_samples = authz_sample_filters(all_samples, sample_ids, user)
```
vincent-czi: I think it might be good to rename `all_samples` -- it's really `user_visible_samples`, right? I know you're following the previous convention, but it seems a little confusing that `all_samples` goes from meaning _actually all the samples_ a few lines above, to meaning all the samples _visible to the user_ making the request once the authz func is applied.

src/backend/aspen/app/views/validation.py
```diff
@@ -0,0 +1,62 @@
...
+from aspen.app.views.api_utils import (
+    authz_sample_filters,
+    get_matching_gisaid_ids,
+    get_missing_and_found_sample_ids,
+)
+from aspen.database.models import Sample
+from aspen.error import http_exceptions as ex
+
+
+@application.route("/api/validate/ids")
```
vincent-czi: Reasonable people can have differing opinions on this, but I think this endpoint falls under the scope of `samples` since it's about validating the existence of samples generally. I would pitch for something under that previous URI umbrella, like `/api/samples/validate_ids` instead

vincent-czi (approved):
Looks good!

The one thing I wonder about is the interaction with frontend. It seems to me like the frontend can basically do half of this endpoint already on its own when validating IDs -- the frontend already knows all the Aspen samples the user can view, since that's what's in the Samples table -- so would it be worth it to have an additional endpoint that is only for validating the GISAID aspect? This could definitely be premature optimization on my part, though.

src/backend/aspen/app/views/validation.py
phoenixAja: good idea, i'll move it over to samples

src/backend/aspen/app/views/phylo_runs.py
phoenixAja: agreed all_samples is a little confusing. I'll rename!

jgadling:
> The one thing I wonder about is the interaction with frontend. It seems to me like the frontend can basically do half of this endpoint already on its own when validating IDs -- the frontend already knows all the Aspen samples the user can view, since that's what's in the Samples table -- so would it be worth it to have an additional endpoint that is only for validating the GISAID aspect? This could definitely be premature optimization on my part, though.

I disagree here - the backend has a lot more tools available for data access/filtering/etc and as our datasets grow, it may not be practical for the frontend to have a complete copy in memory all the time.

phoenixAja:
@vincent-czi and @jgadling moved the validation endpoint over to samples, re requested your review since this was a fairly large change

src/backend/aspen/app/views/phylo_runs.py
```diff
@@ -56,35 +59,19 @@ def start_phylo_run():
...
-    for sample in all_samples:
-        pathogen_genomes.append(sample.uploaded_pathogen_genome)
-        found_sample_ids.add(sample.private_identifier)
-        found_sample_ids.add(sample.public_identifier)
-    if len(pathogen_genomes) == 0:
-        sentry_sdk.capture_message(
-            f"No sequences selected for run from {sample_ids}.", "error"
-        )
-        raise ex.BadRequestException("No sequences selected for run")
+    # Step 3 - Enforce AuthZ (check if user has permission to see private identifiers)
```
jgadling: Note - This does more that check for permissions for private ID's -- it also scopes down the search for matching ID's to groups that the user has read access to.

jgadling (approved):

src/backend/aspen/app/views/phylo_runs.py
phoenixAja: i'll update the comment

PR #737, opened 2021-10-15
Add endpoint tests w/db integration to fastapi

jgadling:
### Summary:
- **What:** Refactor our fastapi app and tests to be able to set up test fixtures and run endpoint tests against them

### Local changes necessary:
Since this PR includes changes to pyproject.toml / poetry.lock, it'll be necessary to run `make local-sync` to rebuild the backend docker container to make sure pytest has its dependencies up to date.

### Notes:
This PR adds a lot of type-hints, and a new set of (async-compatible) test fixture code that parallels the flask setup code in `src/backend/aspen/test-infra`

It also deprecates the pseudo-global db session handling we had set up using contextvars in favor of relying on FastAPI's dependency injection for generating db sessions, since FastAPI gives us a pretty reasonable way to override dependencies in tests.

I also took the opportunity to add a bunch of type-hints where they were missing before.

### Checklist:
- [x] I added labels to my PR


src/backend/aspen/api/conftest.py
```diff
@@ -0,0 +1,122 @@
+import random
```
jgadling, line 1: It actually took me awhile to figure out how pytest was loading fixtures from `aspen/test_infra` and it turns out that `conftest.py` is just pytest *magic* -- conftest.py files will apply to tests in all child directories of a directory containing a conftest.py file:

https://docs.pytest.org/en/6.2.x/fixture.html#conftest-py-sharing-fixtures-across-multiple-files

src/backend/aspen/api/conftest.py
```diff
@@ -0,0 +1,122 @@
...
+from aspen.database import connection as aspen_connection
+from aspen.database import schema
+from aspen.database.connection import init_async_db
+
+USERNAME = "user_rw"
+PASSWORD = "password_rw"
+INITIAL_DATABASE = "postgres"
+
+pytestmark = pytest.mark.anyio
+
```
jgadling, line 24: Most of the code in this file is ~identical to the code in `aspen/test_infra` except that it's been updated to work with an async db interface

src/backend/aspen/api/deps.py
```diff
@@ -1,32 +1,18 @@
...
 
 from sqlalchemy.ext.asyncio import AsyncSession
+from starlette.requests import Request
 
 from aspen.api.settings import get_settings
 from aspen.database.connection import init_async_db
 
-session_context_var: ContextVar[Optional[AsyncSession]] = ContextVar(
-    "_session", default=None
-)
```
jgadling, line 11: This stuff was basically creating a global -- it was a means to an end to get the first version of this app working, but (no surprise) it made testing difficult. So we're now using a context manager to instantiate db sessions and letting fastapi manage dependency de-duplication.

src/backend/aspen/api/views/users.py
```diff
@@ -1,15 +1,19 @@
...
 router = APIRouter()
 
 
 @router.get("/me", response_model=userschema)
-async def get_current_user(request: Request) -> userschema:
-    instance = await User.get_by_id(request.state.auth_user.id)
+async def get_current_user(
+    request: Request, db: AsyncSession = Depends(get_db)
+) -> userschema:
+    instance = await User.get_by_id(db, request.state.auth_user.id)
```
vincent-czi, line 16: Isn't `request.state.auth_user` already the whole User object? Do we need to get it by its ID again, don't we already have it?

src/backend/aspen/api/conftest.py
```diff
@@ -0,0 +1,122 @@
...
+@dataclass
+class AsyncPostgresDatabase:
+    database_name: str
+    port: int
+
+    def as_uri(self) -> str:
+        return f"postgresql+asyncpg://{USERNAME}:{PASSWORD}@database:{self.port}/{self.database_name}"
+
+
+@pytest.fixture()
```
vincent-czi, line 35: I could be misunderstanding something, but it seems to me like creating a new DB would be a pretty heavy operation and one we would want to avoid doing for every test func, but doesn't the pytest fixture default to a function scope? https://docs.pytest.org/en/6.2.x/fixture.html#fixture-scopes

Is the intention to create a test DB for every test func, or would it be better to do it less frequently than that? I also could be wrong about my understanding of when the fixture is happening here, I've definitely beaten my head against pytest fixtures in the past.

src/backend/aspen/api/views/tests/test_health.py
```diff
@@ -0,0 +1,11 @@
+from fastapi.testclient import TestClient
+
+from aspen.api.main import app
+
+client = TestClient(app)
```
vincent-czi: I'm unclear on when we want to use `TestClient` from FastAPI vs the fixture `http_client` (which looks like a "real" client) -- is it that the first is used for synchronous stuff, while the second is for async stuff? Or is it for when we need to test DB interactions?

src/backend/aspen/api/conftest.py
```diff
@@ -0,0 +1,122 @@
...
+async def async_db() -> AsyncGenerator[AsyncPostgresDatabase, None]:
+    """Creates a postgres test database named a random string with username/password user_rw/password_rw, yields it, and then drops it."""
+
+    # create admin sql connection to create test database
+    admin_uri = "postgresql+asyncpg://postgres:password_postgres@database:5432/aspen_db"
```
vincent-czi, line 40: This feels like something that should be using env vars rather than hardcoded (not that it's actually likely to change anytime soon).

vincent-czi (approved):
Looks good! All my comments are pretty minor. I would definitely like to get a better understanding of when we should be using the TestClient vs the "real" app client, but that's hopefully something we can talk about down the road.

src/backend/aspen/api/views/users.py
jgadling: Yeah, though this endpoint is intended as an example of how we'd fetch <some resource> by an ID in the path.

src/backend/aspen/api/views/tests/test_health.py
jgadling: Oops, good catch. we probably ~never want to use TestClient. AsyncClient is async friendly and works more reliably.

src/backend/aspen/api/conftest.py
jgadling: For the moment I'm just mimic'ing our existing test implementation, which creates db's with a function scope, though I think it's important to be able to share a db between tests -- we need to make sure our endpoints are using DB isolation properly.

PR #738, opened 2021-10-15
Group level unique ids

vincent-czi:
### Summary:
- **What:** Restricts uniqueness of Private/Public IDs to within the user-group "namespace"
- **Ticket:** [sc<166263>](https://app.shortcut.com/genepi/story/<166263>)
- **Env:** https://group-unique-ids-frontend.dev.genepi.czi.technology/

### Notes:
Motivation here is that two different groups should be able to use the same sample IDs without impacting the other.

- Changes Private ID (`private_identifier`) and Public ID (`public_identifier`) ingestion process to only checking against the user's group for duplicated IDs, rather than checking globally.
- Tweaks uniqueness constraint for `public_identifier` to be unique only in context of the group ID, not globally unique.
- Removes uniqueness entirely for EPI ISL Accession ID (`gisaid_accessions` table, `public_identifier` column [NB: this is (almost always) **not** the same as the samples `public_identifier`, they just use the same name with a teeny bit of possible overlap in meaning.])
  - There is a downside here. Preferably, we would have the EPI ISL Accession be unique within a group, like we do with Private/Public ID in samples. But because of how the accessions table works, that would be really hard, and we have an [upcoming ticket](https://app.shortcut.com/genepi/story/150108) that will re-work this entirely, so it seems like an acceptable compromise.

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/app/views/api_utils.py
+++ b/src/backend/aspen/app/views/api_utils.py
@@ -72,7 +72,7 @@ def get_existing_private_ids(
 ) -> list[str]:
     samples = session.query(Sample).filter(Sample.private_identifier.in_(private_ids))
 
-    if group_id:
+    if group_id is not None:
         samples = samples.filter(Sample.submitting_group_id == group_id)
 
     return [i.private_identifier for i in samples.all()]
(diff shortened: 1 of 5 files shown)


src/backend/aspen/database/models/sample.py
```diff
@@ -90,7 +90,16 @@ class Sample(idbase, DictMixin):  # type: ignore
     """A physical sample.  Multiple sequences can be taken of each physical sample."""
 
     __tablename__ = "samples"
-    __table_args__ = (UniqueConstraint("submitting_group_id", "private_identifier"),)
+    __table_args__ = (
+        # Unique index uses default name format of `uq_samples_submitting_group_id`
```
vincent-czi, line 94: Vaguely interesting SQLAlchemy thing here:
The way that SQLAlchemy auto names (`naming_convention` boilerplate) unique indices only uses the table name and first column called out in unique index, just ignores later columns when creating the name. But because we already have a group-private unique index, that took the same name that the group-public unique index would use. So I had to manually give the new one a `name` for the alembic auto-generation to work (and for SQlAlchemy to be able to understand the uniqueness mapping, but I don't know if that really gets used or not...)

src/backend/aspen/app/views/api_utils.py
```diff
@@ -83,19 +83,32 @@ def get_existing_public_ids(
 ) -> list[str]:
     samples = session.query(Sample).filter(Sample.public_identifier.in_(public_ids))
 
-    if group_id:
+    if group_id is not None:
```
jgadling, line 86: any reason to explicitly check for None-ness here?

jgadling (approved):
Looks good!

src/backend/aspen/app/views/api_utils.py
vincent-czi: It's pretty minor, but I wanted to avoid ever having a false-y misstep from `if group_id`. Shouldn't ever impact because they start at 1 and go up, but I figured it was better to be explicit since the default `None` and deviating to using anything else really is saying, "hey, go use that additional filter on group id clause"

PR #739, opened 2021-10-16
Usher placement modal and tie it together

mmmmmmaya:
### Summary
- **What:** 
  - Add placement settings modal
  - Add interaction flow logic for usher tree creation
- **Ticket:** [[sc-161388]](https://app.shortcut.com/genepi/story/161388)
- **Env:** https://ushertwo-frontend.dev.genepi.czi.technology?usher=true

### Demos
cancel
confirm

### Notes
- This is mostly working as intended, logic wise (I hope)
- Styles to come for some things. Going to work on this Monday.
- This definitely should have happened over multiple PRs. My bad.

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/frontend/package.json
```diff
@@ -33,7 +33,7 @@
     "@typescript-eslint/parser": "^4.5.0",
     "babel-eslint": "^10.1.0",
     "classnames": "^2.2.6",
-    "czifui": "1.0.2",
+    "czifui": "^1.1.0",
```
vincent-czi, line 36: Did you get the chance to do any light poking around on the rest of the site, just to make sure nothing went funky with the version change? No big deal if not, but would be good to call out when there's the general demo so people know to keep an eye out for any regressions in rest of app.

src/frontend/src/common/components/library/data_subview/components/UsherTreeFlow/components/UsherPlacementModal/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+  const [isUsherDisabled, setUsherDisabled] = useState<boolean>(false);
+  const [shouldShowWarning, setShouldShowWarning] = useState<boolean>(false);
+
+  const defaultNumSamples = getDefaultNumSamplesPerSubtree(sampleIds?.length);
+
+  useEffect(() => {
+    const fetchUsherOpts = async () => {
+      const resp = (await getUsherOptions()) as UsherDataType;
+      const options = resp.usher_options;
+      forEach(options, (opt) => (opt.name = opt.description));
```
vincent-czi, line 85: I would prefer for this to not modify the `resp.usher_options` in-place. I'd rather see it done as a `.map` to generate new dropdownOption objects, just to keep things more pure.

src/frontend/src/common/components/library/data_subview/components/UsherTreeFlow/components/UsherPlacementModal/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+      options.sort((a: DropdownOptionProps, b: DropdownOptionProps) => {
```
vincent-czi, line 87: In theory, the backend should be sending over the options list already sorted in the same way. But, there's also nothing wrong with ensuring it's sorted, so I'm ambivalent.

src/frontend/src/common/components/library/data_subview/components/UsherTreeFlow/components/UsherPlacementModal/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+    fetchUsherOpts();
+  }, []);
+
+  useEffect(() => {
+    const hasValidSamplesSelected = sampleIds?.length > failedSamples?.length;
+    setUsherDisabled(!hasValidSamplesSelected);
+  }, [sampleIds, failedSamples]);
+
+  const mutation = useMutation(getFastaURL, {
```
vincent-czi, line 104: I think it might be worth pulling this out and moving it to somewhere in `src/common/queries` like some of the other API calls? My thinking is mainly to keep our API calls consistent in structure (like using `DEFAULT_POST_OPTIONS` unless they have a specific reason to do otherwise) and co-located so we know where to look for already written API interactions as opposed to accidentally writing repeats over time.

That said, this seems like the only place we'd really make this API call, so I get keeping located here. I dunno. If you feel like this is the right way to do it, just disregard what my comment.

src/frontend/src/common/components/library/data_subview/components/UsherTreeFlow/components/UsherPlacementModal/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+  );
+
+  const SAMPLES_PER_SUBTREE_TOOLTIP_TEXT = (
+    <div>
+      We recommend setting this number to at least 5x the number of selected
+      samples, and no less than 50.
+    </div>
+  );
+
+  const onOptionChange = (opt: DefaultMenuSelectOption | null) => {
```
vincent-czi, line 145: Are we not currently using the actual `.value` of the chosen option, ie, what kind of usher tree is being targeted?

src/frontend/src/common/utils/strUtils.ts
```diff
@@ -0,0 +1,3 @@
+export const pluralize = (str: string, count: number): string => {
```
vincent-czi, line 1: I like this, but English sure is a hassle to work with. If we wind up having to deal with something like "mouse" or "property" down the road, my vote is to just make a bunch of one-off funcs as we run into those words:
```pluralizeMouse(count) => count === 1 ? "mouse" : "mice"```

vincent-czi (approved):
Looks good! Nice work, this is a lot of stuff. I don't feel strongly about any of my comments, all the stuff I'm suggesting is merely in the nice-to-have category for me.

src/frontend/src/common/components/library/data_subview/components/UsherTreeFlow/components/UsherPlacementModal/index.tsx
vincent-czi: This is probably the only _actually important_ comment of the bunch. But if this is already on your roadmap for the next PR, it don't matter.

mmmmmmaya:
Addressing feedback in a follow up PR 🚀

src/frontend/src/common/components/library/data_subview/components/UsherTreeFlow/components/UsherPlacementModal/index.tsx
mmmmmmaya: oops 🤦‍♀️  I completely forgot about this. I need to ask Jess how to specify the option when sending the tree build request.

src/frontend/package.json
mmmmmmaya: I didn't poke around much. I will add a note for the demo! Thanks!

PR #741, opened 2021-10-18
Add autoflake to `make backend-run-style`

jgadling:
### Summary:
- **What:** This updates `make backend-run-style` to clean up unused imports & variables. This is handy because `make backend-check-style` which is part of our CI pipeline checks for these errors and we don't otherwise have a way to automatically clean them up
- **Env:** `<rdev link>`

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/src/backend/Makefile
+++ b/src/backend/Makefile
@@ -30,7 +30,7 @@ run-isort:
 	isort --profile black $(STYLE_CHECK_PYTHON_CODE_DIRECTORIES) --skip $(STYLE_CHECK_PYTHON_CODE_SKIPPED_DIRECTORIES) 
 
 run-autoflake:
-	autoflake -i --remove-all-unused-imports --remove-unused-variables -r $(TYPE_CHECK_BASE_PYTHON_CODE_DIRECTORIES)
+	autoflake -v -i --remove-all-unused-imports --remove-unused-variables -r $(TYPE_CHECK_BASE_PYTHON_CODE_DIRECTORIES)
 
 
 ### TESTING #############################################
--- a/src/backend/Makefile
+++ b/src/backend/Makefile
@@ -21,13 +21,16 @@ mypy:
 
 
 ### RUN STYLE #############################################
-run-style: run-black run-isort
+run-style: run-autoflake run-black run-isort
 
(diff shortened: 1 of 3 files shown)


vincent-czi:
How noisy is it when `autoflake` runs? Does it print out what changes it's making, or does it just make the changes silently and in-place? I'm ambivalent on this kind of tool. It's not common, but I know I have made a couple mistakes working in the past where seeing that I had an unused var made me realize I had screwed up (like accidentally leaving a code block commented out), and seeing a flake8 warning saved me an experience of "wait, what the heck is going on?". Auto-removing this sort of stuff without any notification of the change feels like a double-edged sword.

Counter-point: if the dev adds code, then runs tool, then diffs all the above is still visible. Similarly, if dev runs the check-style instead, the visibility is still available. Like I said, I'm ambivalent.

vincent-czi (approved):
I've got a separate comment with my philosophical waffling, but code-wise, this is all good

lvreynoso (approved):

lvreynoso:
> How noisy is it when `autoflake` runs? Does it print out what changes it's making, or does it just make the changes silently and in-place? I'm ambivalent on this kind of tool. It's not common, but I know I have made a couple mistakes working in the past where seeing that I had an unused var made me realize I had screwed up (like accidentally leaving a code block commented out), and seeing a flake8 warning saved me an experience of "wait, what the heck is going on?". Auto-removing this sort of stuff without any notification of the change feels like a double-edged sword.
> 
> Counter-point: if the dev adds code, then runs tool, then diffs all the above is still visible. Similarly, if dev runs the check-style instead, the visibility is still available. Like I said, I'm ambivalent.

You can always run `make backend-check-style` first :)

jgadling:
> How noisy is it when `autoflake` runs? Does it print out what changes it's making, or does it just make the changes silently and in-place? 

It was silently fixing things, but I added the `-v` flag so it'll print the names of files that it changes.

vincent-czi:
> It was silently fixing things, but I added the `-v` flag so it'll print the names of files that it changes.

Appreciate it, thanks!

PR #743, opened 2021-10-18
allow users to click through preparing modal if they want to

mmmmmmaya:
### Summary
- **What:** Allow the user to click the backdrop of the preparing modal in order to not wait 5 seconds to open their usher link.
- **Why:** The message may be useful for the first few times, but also blocks the user from getting to their data quickly
- **Env:** https://usherthree-frontend.dev.genepi.czi.technology?usher=true [pending]

### Testing
1. The modal should automatically open the link to usher and close itself after 5 seconds
1. If you click the back drop before 5 seconds are up, the link should open immediately

### Demos
after

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/components/library/data_subview/components/UsherTreeFlow/components/UsherPreparingModal/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/UsherTreeFlow/components/UsherPreparingModal/index.tsx
@@ -1,5 +1,5 @@
 import { LoadingIndicator } from "czifui";
-import React, { useEffect, useState } from "react";
+import React, { useCallback, useEffect, useState } from "react";
 import { StyledDialog, StyledP, StyledTitle } from "./style";
 
 interface Props {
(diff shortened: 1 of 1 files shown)


src/frontend/src/common/components/library/data_subview/components/UsherTreeFlow/components/UsherPreparingModal/index.tsx
```diff
@@ -18,32 +20,53 @@ const UsherPreparingModal = ({
...
+
+    const link = document.createElement("a");
+    link.href = `${USHER_URL + fastaUrl}`;
+    link.target = "_blank";
+    link.rel = "noopener";
+    link.click();
+    link.remove();
+
+    onClose();
+    setTimer(undefined);
```
vincent-czi: Nit: for clarity, I think 36 and 37 should be flip-flopped in order, just so it can't be misread as timer being set to undefined _then_ cleared

vincent-czi (approved):
Looks good!

PR #744, opened 2021-10-18
Add Sentry integration to FastAPI

lvreynoso:
### Summary:
- **What:** `Replicates our Sentry setup in FastAPI`
- **Ticket:** [sc161330](https://app.shortcut.com/genepi/story/161330)
- **Env:** `https://sentry-asgi-backend.dev.genepi.czi.technology/`

### Notes:
The key in our AWS secret store storing the Sentry DSN is changed from `SENTRY_URL` to `SENTRY_BACKEND_DSN` to avoid confusion with the `SENTRY_URL` environment variable that is referenced in `sentry-cli` documentation.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/main.py
+++ b/src/backend/aspen/api/main.py
@@ -1,6 +1,8 @@
 import os
 from typing import List
 
+import sentry_sdk
+from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
 from fastapi import Depends, FastAPI
 from starlette.middleware.cors import CORSMiddleware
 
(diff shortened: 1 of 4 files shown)


src/backend/aspen/api/main.py
```diff
@@ -1,7 +1,9 @@
 import os
 from typing import List
 
+import sentry_sdk
 from fastapi import Depends, FastAPI
+from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
```
jgadling, line 6: Do we have all of the necessary libs installed in the container already?

src/backend/aspen/api/settings.py
```diff
@@ -166,11 +166,11 @@ def EXTERNAL_AUSPICE_BUCKET(self) -> str:
     ####################################################################################
     # sentry properties
     @cached_property
-    def SENTRY_URL(self) -> str:
-        if os.getenv("SENTRY_URL"):
-            return os.environ["SENTRY_URL"]
+    def SENTRY_BACKEND_DSN(self) -> str:
+        if os.getenv("SENTRY_BACKEND_DSN"):
```
jgadling: Where does this env var get set?

jgadling (changes requested):

src/backend/aspen/api/settings.py
lvreynoso: It doesn't, it's left over from when I was trying to make Sentry play nice on a local setup. I don't expect this to ever work - so I will remove this check so no one gets confused!

jgadling (approved):

PR #746, opened 2021-10-18
Maya/usher feedback

mmmmmmaya:
### Summary
- **What:** Use the specified tree type when generating link to usher
- **Why:** So the user can decide the context they want to use
- **Env:** https://usherfive-frontend.dev.genepi.czi.technology?usher=true

### Testing
I am 99% sure this is correct. I generated two trees, one of each type. When I used the GISAID option, the page included some text that read `Nearest neighboring GISAID/public sequence ...`. When I used the type without GISAID, the same line of text read `Nearest neighboring public sequence ...`. Maybe there is someone on the team that can confirm the different options generate trees within the correct context 😬 

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/components/library/data_subview/components/UsherTreeFlow/components/UsherPreparingModal/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/UsherTreeFlow/components/UsherPreparingModal/index.tsx
@@ -12,9 +12,15 @@ interface Props {
 type TimeoutType = ReturnType<typeof setTimeout> | undefined;
 
 const TIME_MODAL_SHOWN = 5000; // 5 seconds
-const USHER_URL =
-  "https://genome.ucsc.edu/cgi-bin/hgPhyloPlace?db=wuhCor1&remoteFile=";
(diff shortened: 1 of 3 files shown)


src/frontend/src/common/components/library/data_subview/components/UsherTreeFlow/components/UsherPreparingModal/index.tsx
```diff
@@ -26,7 +29,9 @@ const UsherPreparingModal = ({
     if (!fastaUrl) return;
 
     const link = document.createElement("a");
-    link.href = `${USHER_URL + fastaUrl}`;
+    link.href = `${USHER_URL}${encodeURIComponent(
```
vincent-czi: The process of generating the redirect URL is getting complicated enough it feels like it should be extracted. Maybe pulled to a standalone Usher "API" file, maybe just a function that's defined in this file, but it seems complicated enough to me that it's better as its own process.

vincent-czi (approved):
Looks good, and certainly seems plausible that's how they would take the option value? I think you'll have to get Dan to verify, but looks good code-wise

PR #747, opened 2021-10-19
Change the domain for local-development

jgadling:
### Summary:
- **What:** This changes the domain we use for local development from `<app>.genepinet.local` to `<app>.genepinet.localdev`.  The `.local` domain is reserved for multicast DNS, which was causing a 5s delay for most requests to our dev environment. This `.localdev` domain isn't reserved for any special function as far as I can tell, and it isn't [on the list of real TLD's](https://data.iana.org/TLD/tlds-alpha-by-domain.txt) so it shouldn't cause any conflicts with the outside world.

To switch to the new domain, engineers will need to recreate their local-dev environment:
```
$ make local-clean
$ make local-init
```
And everything should be back up and running at the new url.

### Checklist:
- [ ] I added labels to my PR


vincent-czi (approved):

vincent-czi:
Not particularly important, but there are two unit tests that use the old `.local` convention. It's just a dummy string as far as I can see (since it's just used for pulling a mock), but if you want to really enforce the new convention, they are
- `src/cli/tests/test_api_client.py:`
- `src/cli/tests/test_token_handler.py:`

PR #751, opened 2021-10-20
Convert the ondemand phylo tree endpoint to fastapi

jgadling:
### Summary:
- **What:** Convert the ondemand phylo tree endpoint to fastapi
- **Env:** https://phylov2-frontend.dev.genepi.czi.technology/

### Notes:
This converts our first endpoint from flask to fastapi.

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/src/backend/Makefile
+++ b/src/backend/Makefile
@@ -30,7 +30,7 @@ run-isort:
 	isort --profile black $(STYLE_CHECK_PYTHON_CODE_DIRECTORIES) --skip $(STYLE_CHECK_PYTHON_CODE_SKIPPED_DIRECTORIES) 
 
 run-autoflake:
-	autoflake -v -i --remove-all-unused-imports --remove-unused-variables -r $(TYPE_CHECK_BASE_PYTHON_CODE_DIRECTORIES)
+	autoflake -v -i --ignore-init-module-imports --remove-all-unused-imports --remove-unused-variables -r $(TYPE_CHECK_BASE_PYTHON_CODE_DIRECTORIES)
 
 
 ### TESTING #############################################
--- a/src/backend/aspen/api/schemas/base.py
+++ b/src/backend/aspen/api/schemas/base.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timezone
 
 from pydantic import BaseModel
 
@@ -7,11 +7,12 @@ class BaseRequest(BaseModel):
     pass
 
 
-def convert_datetime_to_iso_8601_with_z_suffix(dt: datetime) -> str:
-    return dt.strftime("%Y-%m-%dT%H:%M:%SZ")
+def convert_datetime_to_iso_8601(dt: datetime) -> str:
+    dt = dt.replace(tzinfo=timezone.utc)
+    return dt.isoformat(timespec="seconds")
 
 
 class BaseResponse(BaseModel):
     class Config:
         orm_mode = True
-        json_encoders = {datetime: convert_datetime_to_iso_8601_with_z_suffix}
+        json_encoders = {datetime: convert_datetime_to_iso_8601}
--- a/src/backend/aspen/api/schemas/base.py
+++ b/src/backend/aspen/api/schemas/base.py
@@ -1,6 +1,9 @@
 from pydantic import BaseModel
 
 
-class Base(BaseModel):
+class BaseRequest(BaseModel):
+    pass
+
+class BaseResponse(BaseModel):
     class Config:
         orm_mode = True
--- a/src/backend/aspen/api/schemas/base.py
+++ b/src/backend/aspen/api/schemas/base.py
@@ -4,6 +4,7 @@ from pydantic import BaseModel
 class BaseRequest(BaseModel):
     pass
 
+
 class BaseResponse(BaseModel):
     class Config:
         orm_mode = True
--- a/src/backend/aspen/api/schemas/base.py
+++ b/src/backend/aspen/api/schemas/base.py
@@ -8,7 +8,8 @@ class BaseRequest(BaseModel):
 
 
 def convert_datetime_to_iso_8601(dt: datetime) -> str:
-    dt = dt.replace(tzinfo=timezone.utc)
+    if not dt.tzinfo:
+        dt = dt.replace(tzinfo=timezone.utc)
     return dt.isoformat(timespec="seconds")
 
 
--- a/src/backend/aspen/api/schemas/base.py
+++ b/src/backend/aspen/api/schemas/base.py
@@ -1,3 +1,5 @@
+from datetime import datetime
+
 from pydantic import BaseModel
 
 
@@ -5,6 +7,11 @@ class BaseRequest(BaseModel):
     pass
 
 
+def convert_datetime_to_iso_8601_with_z_suffix(dt: datetime) -> str:
+    return dt.strftime("%Y-%m-%dT%H:%M:%SZ")
+
+
 class BaseResponse(BaseModel):
     class Config:
         orm_mode = True
+        json_encoders = {datetime: convert_datetime_to_iso_8601_with_z_suffix}
--- a/src/backend/aspen/api/schemas/phylo_runs.py
+++ b/src/backend/aspen/api/schemas/phylo_runs.py
@@ -1,5 +1,5 @@
 import datetime
-from typing import List
+from typing import List, Union
 
 from pydantic import constr, StrictStr, validator
 
@@ -14,7 +14,8 @@ PHYLO_TREE_TYPES = {
 
 
 class PhyloRunRequestSchema(BaseRequest):
-    name: constr(min_length=1, max_length=128, strict=True)
+    # mpypy + pydantic is a work in progress: https://github.com/samuelcolvin/pydantic/issues/156
+    name: constr(min_length=1, max_length=128, strict=True)  # type: ignore
     samples: List[StrictStr]
     tree_type: StrictStr
 
@@ -39,7 +40,7 @@ class PhyloRunResponseSchema(BaseResponse):
 
     id: int
     start_datetime: datetime.datetime
-    end_datetime: datetime.datetime = None
+    end_datetime: Union[datetime.datetime, None] = None
     workflow_status: WorkflowStatusType  # TODO maybe we can keep SqlAlchemy out of this
     group: GroupResponseSchema
     template_file_path: StrictStr
--- a/src/backend/aspen/api/schemas/phylo_runs.py
+++ b/src/backend/aspen/api/schemas/phylo_runs.py
@@ -1,12 +1,10 @@
-from aspen.api.schemas.base import BaseRequest, BaseResponse
 import datetime
-
-from pydantic import BaseModel, ValidationError, validator, constr, StrictStr
 from typing import List
-from aspen.database.models import (
-    WorkflowStatusType, 
-    TreeType,
-)
+
+from pydantic import constr, StrictStr, validator
+
+from aspen.api.schemas.base import BaseRequest, BaseResponse
+from aspen.database.models import TreeType, WorkflowStatusType
 
 # What kinds of ondemand nextstrain builds do we support?
 PHYLO_TREE_TYPES = {
@@ -14,12 +12,13 @@ PHYLO_TREE_TYPES = {
     TreeType.TARGETED.value: "targeted.yaml",
 }
 
+
 class PhyloRunRequestSchema(BaseRequest):
     name: constr(min_length=1, max_length=128, strict=True)
     samples: List[StrictStr]
     tree_type: StrictStr
 
-    @validator('tree_type')
+    @validator("tree_type")
     def tree_type_must_be_supported(cls, value):
         uppercase_tree_type = value.upper()
         assert PHYLO_TREE_TYPES.get(uppercase_tree_type)
@@ -29,12 +28,15 @@ class PhyloRunRequestSchema(BaseRequest):
 class GroupResponseSchema(BaseResponse):
     class Config:
         orm_mode = True
+
     id: int
     name: StrictStr
+
 
 class PhyloRunResponseSchema(BaseResponse):
     class Config:
         orm_mode = True
+
     id: int
     start_datetime: datetime.datetime
     end_datetime: datetime.datetime = None
--- a/src/backend/aspen/api/schemas/phylo_runs.py
+++ b/src/backend/aspen/api/schemas/phylo_runs.py
@@ -1,7 +1,7 @@
 from aspen.api.schemas.base import BaseRequest, BaseResponse
 import datetime
 
-from pydantic import BaseModel, ValidationError, validator, constr
+from pydantic import BaseModel, ValidationError, validator, constr, StrictStr
 from typing import List
 from aspen.database.models import (
     WorkflowStatusType, 
@@ -15,9 +15,9 @@ PHYLO_TREE_TYPES = {
 }
 
 class PhyloRunRequestSchema(BaseRequest):
-    name: constr(min_length=1, max_length=128)
-    samples: List[str]
-    tree_type: str
+    name: constr(min_length=1, max_length=128, strict=True)
+    samples: List[StrictStr]
+    tree_type: StrictStr
 
     @validator('tree_type')
     def tree_type_must_be_supported(cls, value):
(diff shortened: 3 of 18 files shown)


src/backend/aspen/api/schemas/base.py
```diff
@@ -1,6 +1,17 @@
...
 from pydantic import BaseModel
 
 
-class Base(BaseModel):
+class BaseRequest(BaseModel):
+    pass
+
+
+def convert_datetime_to_iso_8601_with_z_suffix(dt: datetime) -> str:
+    return dt.strftime("%Y-%m-%dT%H:%M:%SZ")
```
jgadling: Pydantic wasn't formatting true 8601 dates by default. Not sure whether this is overkill.

src/backend/aspen/api/utils/get_matching_gisaid_ids.py
```diff
@@ -0,0 +1,44 @@
+from typing import Iterable, Mapping, Set
+
+import sqlalchemy as sa
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from aspen.database.models import GisaidMetadata
+
+
+async def get_matching_gisaid_ids(
```
jgadling, line 9: This is the ~same code we have in `src/backend/aspen/app/utils.py` but updated for async db sessions and SqlAlchemy 2.0 query syntax

src/backend/aspen/api/schemas/base.py
vincent-czi: What was it formatting as?

src/backend/Makefile
```diff
@@ -30,7 +30,7 @@ run-isort:
 	isort --profile black $(STYLE_CHECK_PYTHON_CODE_DIRECTORIES) --skip $(STYLE_CHECK_PYTHON_CODE_SKIPPED_DIRECTORIES) 
 
 run-autoflake:
-	autoflake -v -i --remove-all-unused-imports --remove-unused-variables -r $(TYPE_CHECK_BASE_PYTHON_CODE_DIRECTORIES)
+	autoflake -v -i --ignore-init-module-imports --remove-all-unused-imports --remove-unused-variables -r $(TYPE_CHECK_BASE_PYTHON_CODE_DIRECTORIES)
```
jgadling, line 33: Autoflake was eating our `__init__.py` imports and this tells it not to.

src/backend/aspen/api/schemas/base.py
jgadling: An empty file! those imports were technically unused since they weren't being referenced in `__init__.py`.

vincent-czi: No, I meant the pydantic time formatting

jgadling: Err -- https://github.com/tiangolo/fastapi/issues/1178

The strings looked like this: `2021-10-20T23:17:29.466173`

vincent-czi: I have this vague recollection that using `Z` for UTC can cause headaches, but I forget exactly where. In that issue you linked, the OP says

> So, first of all, I know that pydantic is smart and accepts strings ending in Z for datetimes. I am thankful that pydantic follows ISO guidelines even though python's standard datetime library rejects them.

... That sounds familiar. Like Python plays nicer with `+00:00` format for timezones generally. ... yeah, just checked:

```
>>> x = datetime.datetime.fromisoformat('2021-10-20T23:17:29Z')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: Invalid isoformat string: '2021-10-20T23:17:29Z'
>>> x = datetime.datetime.fromisoformat('2021-10-20T23:17:29+00:00')
# ^^^ works when using +00:00 instead of Z
```
So regardless, I'd recommend using `+00:00` instead of `Z` for indicating UTC, just because Python likes it more. (I know, `Z` is valid ISO8601, but whatcha gonna do)

---

I could be wrong, but I think the reason Pydantic isn't formatting with the timezone at the end is because (flipping through some DB tables on my local right now...) we don't generally have time zones in our database. Our timestamp columns lack timezone, so pydantic is seeing it as a naive datetime (which is correct from its POV), and thus avoids saying a timezone when it formats to a string. (Aside: is our implicit contract that we only work in UTC in the backend? Please let me know if that's not the case.)

Generally, I think we should be including the timezone in our timestamp columns just to be explicit. That would solve this, but would require a migration.

I think this could also be solved by setting the `tzinfo` as we read in datetimes from the DB. Or we could have the serialization process set the tzinfo on the datetime before pydantic serializes to a string. I don't know where the best place to do that is. I figured pydantic would have an easy way to customize serializers, but [this issue seems to indicate otherwise](https://github.com/samuelcolvin/pydantic/issues/951).

In any case, I'm thinking something like this:
```
timezone_aware_datetime = timezone_unaware_datetime.replace(tzinfo=datetime.timezone.utc)
return timezone_aware_datetime.isoformat(timespec='seconds')
```
NB: I thought the `.replace` method would do it in place, but it returns a new object. You can't just swap the timezone in place in native datetime from what I've seen so far, but I didn't look really carefully.

jgadling: Updated!

lvreynoso: Oh wow, TIL.

src/backend/aspen/api/schemas/phylo_runs.py
```diff
@@ -0,0 +1,47 @@
...
+
+# What kinds of ondemand nextstrain builds do we support?
+PHYLO_TREE_TYPES = {
+    TreeType.NON_CONTEXTUALIZED.value: "non_contextualized.yaml",
+    TreeType.TARGETED.value: "targeted.yaml",
+}
+
+
+class PhyloRunRequestSchema(BaseRequest):
+    # mpypy + pydantic is a work in progress: https://github.com/samuelcolvin/pydantic/issues/156
```
phoenixAja:
```suggestion
    # mypy + pydantic is a work in progress: https://github.com/samuelcolvin/pydantic/issues/156
```

src/backend/aspen/api/views/phylo_runs.py
```diff
@@ -0,0 +1,190 @@
...
+    TreeType,
+    Workflow,
+    WorkflowStatusType,
+)
+
+# What kinds of ondemand nextstrain builds do we support?
+PHYLO_TREE_TYPES = {
+    TreeType.NON_CONTEXTUALIZED.value: "non_contextualized.yaml",
+    TreeType.TARGETED.value: "targeted.yaml",
+}
```
phoenixAja: can we import PHYLO_TREE_TYPES from src/backend/aspen/api/schemas/phylo_runs.py?

src/backend/aspen/api/views/phylo_runs.py
```diff
@@ -0,0 +1,190 @@
...
+
+# route
+router = APIRouter()
```
phoenixAja: remove comment
```suggestion
router = APIRouter()
```

src/backend/aspen/api/views/phylo_runs.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+@router.post("/")
```
phoenixAja, line 42: we want this to be routing to "/"?

src/backend/aspen/api/views/phylo_runs.py
```diff
@@ -0,0 +1,190 @@
...
+    group = user.group
+
+    # validation happens in input schema
+
+    sample_ids = phylo_run_request.samples
+
+    # Step 2 - prepare big sample query per the old db cli
+    # all_samples: Iterable[Sample] = db.query(Sample).options(
+    #     joinedload(Sample.uploaded_pathogen_genome, innerjoin=True),
+    # )
```
phoenixAja:
```suggestion
```

src/backend/aspen/api/views/phylo_runs.py
jgadling: Yes, because this route gets appended to the route prefix defined here:
https://github.com/chanzuckerberg/aspen/blob/ab3c77b15ce7bf4235cbf78812f49ee536ae8e1d/src/backend/aspen/api/main.py#L62-L66

 If we had (for example) a set of user management endpoints that we wanted to be able to share with another application, leaving the route prefixes out of the endpoints adds some flexibility.

src/backend/aspen/uvicorn_worker.py
```diff
@@ -0,0 +1,38 @@
...
+import sys
+import threading
+import time
+from typing import Any
+
+from uvicorn.workers import UvicornWorker
+
+
+# Code from https://github.com/encode/uvicorn/pull/1193/files
+class ReloaderThread(threading.Thread):
```
phoenixAja, line 18: can you add a comment on why we need this?

src/backend/aspen/uvicorn_worker.py
```diff
@@ -0,0 +1,38 @@
...
+        self._interval = sleep_interval
+
+    def run(self) -> None:
+        while True:
+            if not self._worker.alive:
+                os.kill(os.getpid(), signal.SIGINT)
+            time.sleep(self._interval)
+
+
+class AspenUvicornWorker(UvicornWorker):
```
phoenixAja, line 32: same here (adding a comment would be helpful)

src/backend/aspen/api/views/phylo_runs.py
phoenixAja: oh got it thanks

phoenixAja (approved):
looks good! Very exciting 🎉 good work!!

src/backend/aspen/api/schemas/base.py
```diff
@@ -1,6 +1,18 @@
...
+
 from pydantic import BaseModel
 
 
-class Base(BaseModel):
+class BaseRequest(BaseModel):
+    pass
+
+
+def convert_datetime_to_iso_8601(dt: datetime) -> str:
```
vincent-czi, line 10: Probably would be good to rename this to `convert_naive_datetime_to_iso_8601` or `convert_utc_datetime_to_iso_8601` or just add some comments mentioning that it will stomp out any TZ info if it already exists and overwrite it as UTC (it's not a "real" conversion to UTC, it just overwrites the TZ and does nothing else). Since all our datetimes are naive right now, that's fine, but if we eventually start using TZ-ed datetimes in the DB and anything other than UTC in there (which we won't, but hey, JIC), this would actually output the wrong thing because of the `.replace` step.

src/backend/aspen/api/settings.py
```diff
@@ -174,7 +174,7 @@ def SENTRY_BACKEND_DSN(self) -> str:
 
     ####################################################################################
     # SSM Parameter properties
-    @lru_cache
+    # @lru_cache
```
vincent-czi, line 177: We're commenting out the lru_cache? Is this temporary, or should it be removed entirely? I only vaguely recall the caching implementation from the past PR, but I thought the LRU part was key to how it worked.

src/backend/aspen/api/views/tests/test_phylo_runs.py
```diff
@@ -0,0 +1,220 @@
...
+    async_session.add(gisaid_dump)
+    await async_session.commit()
+
+    auth_headers = {"user_id": user.auth0_user_id}
+    data = {
+        "name": "test phylorun",
+        "tree_type": "targeted",
+        "samples": [sample.public_identifier],
+    }
+    res = await http_client.post("/v2/phylo_runs/", json=data, headers=auth_headers)
```
vincent-czi, line 37: This is pretty minor, but it looks like the old, v1 endpoint was `/api/phylo_runs` (without trailing `/`), while the new v2 has a trailing `/`. I don't know enough about FastAPI routing -- it's possible that they'll both get treated the same. If they don't, it would be nice to change so the new v2 endpoint is identical (other than the `v2/` part, of course).

I didn't realize this earlier when I was just looking at the endpoint code because the router is broken up over two files.

src/backend/aspen/uvicorn_worker.py
jgadling: Comment added!

vincent-czi: Thank you for the explainer comment

src/cli/aspencli.py
```diff
@@ -357,5 +357,20 @@ def start_phylo_run(ctx, name, tree_type, sample_ids, show_headers):
         print(resp.headers)
     print(resp.text)
 
+@phylo_run.command(name="startv2")
+@click.option("-n","--name", required=True, type=str)
+@click.option("-t","--type", "tree_type", required=True, type=click.Choice(["OVERVIEW", "TARGETED", "NON_CONTEXTUALIZED"], case_sensitive=False))
+@click.option("-h", "--show-headers", is_flag=True)
+@click.argument("sample_ids", nargs=-1)
+@click.pass_context
+def start_phylo_run_v2(ctx, name, tree_type, sample_ids, show_headers):
```
vincent-czi, line 366: We should think about how we'll go about deprecating the old CLI commands for v1. Right now, if I needed to kick one off with the CLI and my eye first fell on `start_phylo_run`, I'd probably just use that and not realize there's a v2 that's now preferred.

src/frontend/src/common/components/library/data_subview/components/RedirectConfirmationModal/index.tsx
```diff
@@ -0,0 +1,52 @@
+import React from "react";
```
vincent-czi: It looks like one of the most recent commits added a bunch of touched FE stuff. Not sure what's going on.

vincent-czi (approved):
Looks good! One of the recent commits touched a bunch of FE stuff -- I don't think that was intentional -- that should probably get resolved before merging.

src/backend/aspen/api/settings.py
jgadling: It's temporary... one of the TODO's I'm looking at today 😱

src/backend/aspen/api/views/tests/test_phylo_runs.py
jgadling: This gets into a whole REST url philosophy debate. I don't think there's any particular reason to make the URL's the same (or even similar) between the old & new app, but I think we *should* agree on whether Create/List endpoints should require a trailing slash or not. I'm a little bit leaning towards making the trailing slash mandatory since it makes the context for the request (i.e. "a collection of resource X") clearer, but I'm not stuck on it.

ref: https://softwareengineering.stackexchange.com/questions/186959/trailing-slash-in-restful-api

src/frontend/src/common/components/library/data_subview/components/RedirectConfirmationModal/index.tsx
jgadling: yeah I'm not sure how that happened either -- fixed it though

src/backend/aspen/api/views/phylo_runs.py
jgadling: Good catch! updated.

src/cli/aspencli.py
lvreynoso: I think we could get rid of the old command when we get rid of the old endpoint - when this new one is up in prod.

src/backend/aspen/api/schemas/base.py
jgadling: I actually updated this so it only adds the UTC timezone if our db date is naive.

src/backend/aspen/api/views/tests/test_phylo_runs.py
vincent-czi: I don't really have an opinion one way or the other on the trailing slash aspect -- although I do think I've generally seen more APIs in the wild that lack the trailing slash -- all that really matters is consistency.

That said, I _do_ think that if we're trying to keep the interaction/functionality the same for a given endpoint, we should keep the URL the same (`/v2` prefix not withstanding). We certainly can change them -- we have to for the `/v2` prefix anyway -- but if it's primarily a backend refactor from the frontend's POV, I think URL naming should generally stay the same.

src/cli/aspencli.py
vincent-czi: Yeah, I agree: I think the right time to get rid of the old command is when we get rid of the old endpoint.

The motivation for my prior comment is just that I'm worried it's confusing which is the "right and proper" way to kick off a phylo run right now. This will be an issue whenever an endpoint/resource is transitioning from v1 to v2. Both commands should work right now, but if something comes up and we don't get around to removing the old endpoint soon or we forget or something, there could be drift between how v1 and v2 work, and then it matters which command version gets picked.

I'm probably being overly worried/fearful -- I think there's enough team awareness of these changes in the backend that everyone would implicitly understand the difference between what the commands will do -- but I just don't want someone to accidentally pick the wrong command if it ever starts mattering somehow down the road.

PR #752, opened 2021-10-21
Add sequencing date to Samples table

vincent-czi:
### Summary:
- **What:** Adds column of "Sequencing Date" to Samples table and corresponding info being sent from backend on call to the GET samples endpoint
- **Ticket:** [sc<148995>](https://app.shortcut.com/genepi/story/<148995>)
- **Env:** https://display-seq-date-frontend.dev.genepi.czi.technology [Note: when testing this, I'm not sure what groups actually have Sequencing Dates in their data? Both Fresno and Contra Costa counties in rdev DB don't seem to have any of seq date info in there. I think it's working as expected, it's just that it's not very interesting to look at with the users I've tried so far...]

### Demos:
image

---

Feels kinda crowded when on a thin browser window, but that's mostly pre-existing: this just exacerbates it by one more column's worth of squish.
image

### Notes:
- Some of the specific design call-outs in the ticket aren't here. There were width requirements in the design for the column, but I found it pretty challenging to work on the styling aspects because of how our tables are currently built in code.
- Similarly, the table still suffers from the column alignment issues it did previously. Certain long strings for Private ID or Public ID will cause their cell to grow larger than expected, which will then push the rest of the row out of alignment with the other rows.
    - Along those lines, I think the header row and the actual data rows aren't currently communicating width info to each other so they don't stay aligned when things get wonky because of the above. This also generally contributes to the column alignment issues, at least so far as the header row looks off from the below rows.

I discussed the above with Phoenix and Maya, and we generally think it would be worth it to refactor the tables throughout the FE. With that in mind, I'd like to tackle the refactoring instead of trying to fix the above in our current table implementation. And as part of the refactoring I'll work on handling the above.

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR
- ~[ ] I have notified others of changes they need to make locally~ [none needed!]

### DIFF
--- a/src/backend/aspen/app/views/api_utils.py
+++ b/src/backend/aspen/app/views/api_utils.py
@@ -33,7 +33,7 @@ def get_usergroup_query(session: Session, user_id: str) -> Query:
     )
 
 
-def format_date(dt: datetime.date, format="%Y-%m-%d") -> str:
+def format_date(dt: Optional[datetime.date], format="%Y-%m-%d") -> str:
     if dt is not None:
         return dt.strftime(format)
     else:
--- a/src/backend/aspen/app/views/sample.py
+++ b/src/backend/aspen/app/views/sample.py
@@ -99,6 +99,7 @@ def _format_created_date(sample: Sample) -> str:
 
 
 def _format_sequencing_date(sample: Sample) -> str:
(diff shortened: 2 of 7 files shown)


phoenixAja (approved):
looks good! excited for the sample table refactor

PR #754, opened 2021-10-22
Add slices to happy update/create.

jgadling:
### Summary:
- **What:** This adds a `--slice` argument to the happy CLI that bypasses building/pushing images for applications that haven't been modified
- **Ticket:** [sc168694](https://app.shortcut.com/genepi/story/168694)
- **Env:** `https://webslice2-frontend.dev.genepi.czi.technology`

### Notes:
Sometimes an engineer will only modify one or two components in our overall application stack, and then be forced to build & push all images in our stack in order to be able to launch a remote dev environment. This can be very time- and bandwidth-consuming and is often totally unnecessary.

This PR defines a slice as a set of images associated with an alias (eg: the `fullstack` slice includes the `frontend` and `backend` docker images) in `.happy/config.json` that can be referenced by the Happy CLI when creating or updating a stack. To create or update an rdev stack that uses the `branch-trunk` tag for all images not listed as part of our slice:

```
$ ./scripts/happy create --slice frontend testuichange
or
$ ./scripts/happy update --slice fullstack testapichange
```

The default tag can also be overridden:
```
./scripts/happy create --slice batch --slice-default-tag sha-1234567 mystack
```

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/.happy/config.json
+++ b/.happy/config.json
@@ -4,6 +4,30 @@
     "default_env": "rdev",
     "app": "aspen",
     "default_compose_env": ".env.ecr",
+    "slice_default_tag": "branch-trunk",
+    "slices": {
+      "frontend": {
+        "build_images": ["frontend"]
+      },
+      "backend": {
+        "build_images": ["backend"]
+      },
+      "fullstack": {
+        "build_images": ["frontend", "backend"]
+      },
+      "batch": {
+        "build_images": ["nextstrain", "pangolin", "gisaid"]
+      },
+      "nextstrain": {
+        "build_images": ["nextstrain"]
+      },
+      "pangolin": {
+        "build_images": ["pangolin"]
+      },
+      "gisaid": {
+        "build_images": ["gisaid"]
+      }
+    },
     "environments": {
         "rdev": {
             "aws_profile": "genepi-dev",
--- a/.happy/terraform/envs/prod/variables.tf
+++ b/.happy/terraform/envs/prod/variables.tf
@@ -13,6 +13,12 @@ variable image_tag {
   description = "Please provide an image tag"
 }
 
+variable image_tags {
+  type        = string
+  description = "Override the default image tags (json-encoded map)"
+  default     = "{}"
+}
+
 variable priority {
   type        = number
   description = "Listener rule priority number within the given listener"
--- a/.happy/terraform/envs/rdev/variables.tf
+++ b/.happy/terraform/envs/rdev/variables.tf
@@ -13,6 +13,12 @@ variable image_tag {
   description = "Please provide an image tag"
 }
 
+variable image_tags {
+  type        = string
+  description = "Override the default image tags (json-encoded map)"
+  default     = "{}"
+}
+
 variable priority {
   type        = number
   description = "Listener rule priority number within the given listener"
--- a/.happy/terraform/modules/ecs-stack/main.tf
+++ b/.happy/terraform/modules/ecs-stack/main.tf
@@ -10,7 +10,6 @@ locals {
   alb_key = var.require_okta ? "private_albs" : "public_albs"
 
   custom_stack_name     = var.stack_name
-  image_tag             = var.image_tag
   priority              = var.priority
   deployment_stage      = var.deployment_stage
   remote_dev_prefix     = var.stack_prefix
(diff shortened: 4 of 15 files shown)


.happy/config.json
```diff
@@ -4,6 +4,30 @@
     "default_env": "rdev",
     "app": "aspen",
     "default_compose_env": ".env.ecr",
+    "slice_default_tag": "branch-trunk",
```
jgadling, line 7: Defining some slices that seem relevant to aspen devs.

.happy/terraform/envs/prod/variables.tf
```diff
@@ -13,6 +13,12 @@ variable image_tag {
   description = "Please provide an image tag"
 }
 
+variable image_tags {
+  type        = string
+  description = "Override the default image tags (json-encoded map)"
+  default     = "{}"
+}
+
```
jgadling, line 21: This provides a way for the Happy CLI to pass in a map of `{image_name:tag_name}` overrides to the default `image_tag`

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -33,17 +32,16 @@ locals {
...
+  backend_image    = join(":", [local.secret["ecrs"]["backend"]["url"], lookup(var.image_tags, "backend", var.image_tag)])
 
   # Workflow images
-  pangolin_image_repo   = local.secret["ecrs"]["pangolin"]["url"]
-  nextstrain_image_repo = local.secret["ecrs"]["nextstrain"]["url"]
-  gisaid_image_repo     = local.secret["ecrs"]["gisaid"]["url"]
-  covidhub_import_image_repo  = local.secret["ecrs"]["covidhub-import"]["url"]
+  pangolin_image   = join(":", [local.secret["ecrs"]["pangolin"]["url"], lookup(var.image_tags, "pangolin", var.image_tag)])
+  nextstrain_image = join(":", [local.secret["ecrs"]["nextstrain"]["url"], lookup(var.image_tags, "nextstrain", var.image_tag)])
+  gisaid_image     = join(":", [local.secret["ecrs"]["gisaid"]["url"], lookup(var.image_tags, "gisaid", var.image_tag)])
```
jgadling, line 41: Simplifying the way we refer to these images in the rest of our tf code.

We look to see whether the image name is included in our list of tag overrides (`var.image_tags`) and if it's not overridden, we use the default tag.

scripts/happy
```diff
@@ -110,6 +112,7 @@ class StackMeta:
         "instance": "stack_name",
         "priority": "priority",
         "imagetag": "image_tag",
+        "imagetags": "image_tags",
```
jgadling, line 116: These are just metadata fields we track for help with debugging.

scripts/happy
```diff
@@ -253,7 +257,10 @@ class Stack:
             sensitive=False,
         )
         for k, v in self.meta.parameters.items():
-            self.workspace.set_var(k, str(v), "", sensitive=False)
+            if k == "image_tags":
+                self.workspace.set_var(k, json.dumps(v), "", sensitive=False)
+            else:
+                self.workspace.set_var(k, str(v), "", sensitive=False)
```
jgadling, line 265: This is where we populate the `image_tags` variable that terraform uses to identify overridden tag values. I tried making this an HCL value, but TF wasn't happy with it. It's just a json-encoded string for now 😢

scripts/happy
```diff
@@ -925,15 +932,32 @@ def logs(ctx, stack_name, service, since):
     orchestrator = ctx.obj["orchestrator"]
     orchestrator.logs(stack_name, service, since)
 
+def build_slice(ctx, slice_name, default_tag=None):
```
jgadling, line 937: If we recieve a `slice` argument for `create` or `update` we validate that argument here, then build&push the appropriate image[s]

scripts/happy
```diff
@@ -1076,12 +1111,15 @@ def list_command(ctx):
...
+    headings = ["Name", "Owner", "Tags", "Status", "URLs"]
     tp = TablePrinter(headings)
     for name, info in stackmgr.stacks.items():
         url = info.outputs.get("frontend_url", "")
         status = info.status
-        tp.add_row([name, info.meta.owner, info.meta.imagetag, status, url])
+        tag = info.meta.imagetag
+        if info.meta.imagetags:
+            tag = ", ".join(set(info.meta.imagetags.values()) | {tag})
+        tp.add_row([name, info.meta.owner, tag, status, url])
```
jgadling, line 1124: Include the default and overridden tag names in the output we display for `happy list`

Makefile
```diff
@@ -111,6 +111,8 @@ init-empty-db:
 .PHONY: local-init
 local-init: oauth/pkcs12/certificate.pfx .env.ecr local-ecr-login local-hostconfig ## Launch a new local dev env and populate it with test data.
 	docker-compose $(COMPOSE_OPTS) up -d database frontend backend localstack oidc utility
+	# Wait for psql to be up
+	while [ -z "$$(docker-compose exec -T database psql "postgresql://$(LOCAL_DB_ADMIN_USERNAME):$(LOCAL_DB_ADMIN_PASSWORD)@$(LOCAL_DB_SERVER)/$(LOCAL_DB_NAME)" -c 'select 1')" ]; do echo "waiting for db to start..."; sleep 1; done;
```
jgadling, line 115: This just makes `make local-init` more resilient so it waits for psql to be up before moving on to adding db users.

src/backend/scripts/setup_localdata.py
```diff
@@ -3,13 +3,19 @@
 from aspen.config.docker_compose import DockerComposeConfig
 from aspen.database.connection import get_db_uri, init_db
 from aspen.database.models import (
+    AlignedGisaidDump,
```
jgadling, line 6: The changes to this file just add an aligned gisaid dump to the db during local dev setup. This means that local dev envs will be able to run the 'create a new phylo run' endpoint without having to manually add rows to the db first.

.happy/terraform/envs/rdev/variables.tf
```diff
@@ -13,6 +13,12 @@ variable image_tag {
   description = "Please provide an image tag"
 }
 
+variable image_tags {
+  type        = string
+  description = "Override the default image tags (json-encoded map)"
```
phoenixAja, line 18: default is currently all containers right? or has the default changed?

jgadling: this is a map of which image tag to use for which service, for example:
```
{
  "frontend": "sha-abcd123",
}
``` 
This map provides tag values that override the default tag for a stack. For example if you're using the `--slice` option, the default tag is usually `branch-trunk`

scripts/happy
```diff
@@ -925,15 +934,32 @@ def logs(ctx, stack_name, service, since):
...
+
 
 @cli.command()
 @click.argument("stack_name")
 @click.option("--tag", help="Tag name for docker image. Leave empty to generate one automatically.", default=None)
 @click.option("--wait/--no-wait", is_flag=True, default=True, help="wait for this to complete")
 @click.option("--force", is_flag=True, default=False, help="Ignore already-exists errors")
 @click.option("--skip-check-tag", is_flag=True, default=False, help="Skip checking that the specified tag exists (requires --tag)")
+@click.option("--slice", "-s", "slice_name", help="If you only need to test a slice of the app, specify it here")
+@click.option("--slice-default-tag", help="For stacks using slices, override the default tag for any images that aren't being built & pushed by the slice")
```
phoenixAja, line 960: should this be this? 
```suggestion
@click.option("--slice-default-tag", help="For stacks using slices, override the default tag for any images that are being built & pushed by the slice")
```

phoenixAja: so if you add --tag but your updating a slice image would that not do anything? should we have some code to disable using --tag if you're also using --slice?

jgadling: Nope, it works the other way around. The default tag applies to any images that **aren't** part of the slice.

phoenixAja (approved):
thanks for also adding some gisaid data to the local db!

phoenixAja:
before merging will you update local dev readme to include some instructions on how to use this?

jgadling:
> before merging will you update local dev readme to include some instructions on how to use this?

Added that here!
https://github.com/chanzuckerberg/aspen/blob/63f7ed525785472e2c9cc923c46b7bbee4b9d80e/docs/REMOTE_DEV.md#slices

PR #757, opened 2021-10-26
Fix state management for settings & oauth

jgadling:
### Summary:
- **What:** This PR creates settings & oauth client objects once on application startup and then reuses them throughout the app

### Migration Notes:
Once pulling down this change, it's necessary to run `make local-start` to update env vars for the local dev backend.

### Notes:
This PR uses [pydantic's settings sources](https://pydantic-docs.helpmanual.io/usage/settings/#customise-settings-sources) to load AWS secrets & parameters when a Settings object is instantiated, and uses Fastapi's app.state semi-global parameter to keep track of a single Settings object, and a single OAuth client object for the lifetime of the application. This enables us to use the oauth/settings getters as dependencies properly without relying on complicated caching schemes.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -56,6 +56,8 @@ services:
     ports:
       - "3000:3000"
     environment:
+      - API_URL=http://backend.genepinet.localdev:3000
+      - FLASK_ENV=development
       - PYTHONUNBUFFERED=1
       - RESTART_ON_FAILURE=yes
       - BOTO_ENDPOINT_URL=http://localstack.genepinet.localdev:4566
--- a/src/backend/aspen/api/auth.py
+++ b/src/backend/aspen/api/auth.py
@@ -12,8 +12,8 @@ from sqlalchemy.orm.query import Query
 from starlette.requests import Request
 
 import aspen.api.error.http_exceptions as ex
-from aspen.api.deps import get_db
-from aspen.api.settings import get_settings
+from aspen.api.deps import get_db, get_settings
+from aspen.api.settings import Settings
 from aspen.auth.device_auth import validate_auth_header
 from aspen.database.models import Group, User
 
@@ -53,9 +53,10 @@ async def setup_userinfo(
 
 
 async def get_auth_user(
-    request: Request, session: AsyncSession = Depends(get_db)
+    request: Request,
+    session: AsyncSession = Depends(get_db),
+    settings: Settings = Depends(get_settings),
 ) -> User:
-    settings = get_settings()
     auth_header = request.headers.get("authorization")
     auth0_user_id = None
     if auth_header:
--- a/src/backend/aspen/api/deps.py
+++ b/src/backend/aspen/api/deps.py
@@ -14,7 +14,9 @@ def get_auth0_client(request: Request):
 
 
 def get_settings(request: Request):
-    # This parameter is added to the app when we instantiate it.
+    # We stashed our settings object in app.state when we loaded the app, and every
+    # request object has that app attached at request.app, so this dependency is just
+    # returning the settings object we created at startup.
     settings = request.app.state.aspen_settings
     return settings
 
--- a/src/backend/aspen/api/deps.py
+++ b/src/backend/aspen/api/deps.py
@@ -1,15 +1,28 @@
 from typing import AsyncGenerator
 
+from fastapi import Depends
 from sqlalchemy.ext.asyncio import AsyncSession
 from starlette.requests import Request
 
-from aspen.api.settings import get_settings
+from aspen.api.settings import Settings
 from aspen.database.connection import init_async_db
 
 
-async def get_db(request: Request) -> AsyncGenerator[AsyncSession, None]:
+def get_auth0_client(request: Request):
+    # This parameter is added to the app when we instantiate it.
+    return request.app.state.auth0_client
+
+
+def get_settings(request: Request):
+    # This parameter is added to the app when we instantiate it.
+    settings = request.app.state.aspen_settings
+    return settings
+
+
+async def get_db(
+    request: Request, settings: Settings = Depends(get_settings)
+) -> AsyncGenerator[AsyncSession, None]:
     """Store db session in the context var and reset it"""
-    settings = get_settings()
     db = init_async_db(settings.DB_DSN)
     session = db.make_session()
     try:
--- a/src/backend/aspen/api/main.py
+++ b/src/backend/aspen/api/main.py
@@ -2,6 +2,7 @@ import os
 from typing import List
 
 import sentry_sdk
+from authlib.integrations.starlette_client import OAuth
 from fastapi import Depends, FastAPI
 from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
 from starlette.middleware.cors import CORSMiddleware
@@ -9,7 +10,7 @@ from starlette.middleware.cors import CORSMiddleware
 from aspen.api.auth import get_auth_user
 from aspen.api.error.http_exceptions import AspenException, exception_handler
 from aspen.api.middleware.session import SessionMiddleware
-from aspen.api.settings import get_settings
+from aspen.api.settings import Settings
 from aspen.api.views import auth, health, phylo_runs, users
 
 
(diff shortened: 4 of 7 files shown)


docker-compose.yml
```diff
@@ -56,6 +56,8 @@ services:
     ports:
       - "3000:3000"
     environment:
+      - API_URL=http://backend.genepinet.localdev:3000
+      - FLASK_ENV=development
```
jgadling, line 60: These env vars are now required for the fastapi app to start. They're already configured for rdev/staging/prod.

src/backend/aspen/api/settings.py
```diff
@@ -1,139 +1,144 @@
...
     SERVICE_NAME: str = "Aspen"
-
     DB_DRIVER: str = "postgresql+asyncpg"
-
     DB_POOL_SIZE: int = 5
     DB_MAX_OVERFLOW: int = 0
     DB_ECHO: bool = False
+    DEBUG: bool = True
+
+    # Env vars usually read from env vars
```
jgadling, line 88: Pydantic automatically tries to load settings with matching names from the environment if available, and then goes down its list of "magic-settings-getters" to find more data to populate this settings object with. For some of these settings we're relying on env vars, others get loaded from AWS Secrets, and others are loaded from AWS SSM parameter store.

src/backend/aspen/api/settings.py
```diff
@@ -236,9 +194,40 @@ class Config:
...
+            init_settings,
+            env_settings,
+            file_secret_settings,
+        ):
+            return (
+                init_settings,
+                env_settings,
+                file_secret_settings,
+                aws_secret_settings,
+                aws_ssm_settings,
```
jgadling, line 251: ^^ Here's where we're telling pydantic to go ahead and load settings from its default loaders first, and then try reading them from the custom loaders we've defined at the top of this file.

src/backend/aspen/api/settings.py
```diff
@@ -1,139 +1,144 @@
...
-from typing import Any, Mapping
+from functools import cached_property
+from typing import Any, Dict
 
 from boto3 import Session
 from botocore.exceptions import ClientError
 from pydantic import BaseSettings
 
 
+def aws_secret_settings(settings) -> Dict[str, Any]:
```
jgadling, line 11: This loads settings from an aws secret whose value is JSON encoded.

src/backend/aspen/api/settings.py
```diff
@@ -1,139 +1,144 @@
...
+    )
+    try:
+        get_secret_value_response = client.get_secret_value(SecretId=secret_name)
+    except ClientError as e:
+        raise e
+    else:
+        secret = get_secret_value_response["SecretString"]
+    response: Dict = {}
+    secrets = json.loads(secret)
+    keys = settings.__config__.aws_secret_keys
```
jgadling, line 32: Pydantic will complain if we populate keys on the Settings object that it doesn't know about -- so if there are any keys in the AWS secret that don't have matching fields in our Settings class, our app will fail to start. Here's where we're only loading **specific** keys into our Settings object.

src/backend/aspen/api/settings.py
```diff
@@ -1,139 +1,144 @@
...
+    response = {key: secrets[key] for key in keys if secrets.get(key)}
+    remap_keys = settings.__config__.aws_secrets_remap
```
jgadling, line 38: Some of the keys in our AWS Secret don't match the names we're using in our Settings object. This maps the key name from the AWS secret to the key name in our Settings object.

src/backend/aspen/api/settings.py
```diff
@@ -1,139 +1,144 @@
...
+    response.update(
+        {val: secrets[key] for key, val in remap_keys.items() if secrets.get(key)}
+    )
+    return response
+
+
+def aws_ssm_settings(settings) -> Dict[str, Any]:
+    session = Session(region_name=os.environ["AWS_REGION"])
```
jgadling, line 47: This is pretty similar to the function above, but for SSM parameter store.

src/backend/aspen/api/settings.py
```diff
@@ -1,139 +1,144 @@
...
-        if flask_env != "production":
-            return "http://backend.genepinet.localdev:3000/v2/auth/callback"
-        api_url = os.environ.get("API_URL")
-        if not api_url:
-            raise Exception("Missing API_URL in config!")
+        api_url = self.API_URL
         return f"{api_url}/v2/auth/callback"
 
-    @cached_property
-    def AUTH0_CLIENT_SECRET(self) -> str:
```
jgadling, line 85: We're now populating these parameters directly, so there's no need for getters.

src/backend/aspen/api/main.py
```diff
@@ -28,14 +29,30 @@ def get_allowed_origins() -> List[str]:
...
+    settings = Settings()
     _app = FastAPI(
         title=settings.SERVICE_NAME,
         debug=settings.DEBUG,
         openapi_url="/v2/openapi.json",
         docs_url="/v2/docs",
     )
 
+    # Add a global settings object to the app that we can use as a dependency
+    _app.state.aspen_settings = settings
```
jgadling, line 41: Store the settings object somewhere we can easily fetch it later.

src/backend/aspen/api/auth.py
```diff
@@ -53,9 +53,10 @@ async def setup_userinfo(
 
 
 async def get_auth_user(
-    request: Request, session: AsyncSession = Depends(get_db)
+    request: Request,
+    session: AsyncSession = Depends(get_db),
+    settings: Settings = Depends(get_settings),
```
lvreynoso, line 58: Does this mean `get_auth_user` will not be using the global Settings object?

src/backend/aspen/api/settings.py
lvreynoso: Thanks for pointing this out! Was trying to follow the logic and not super-familiar with pydantic magic

src/backend/aspen/api/main.py
```diff
@@ -28,14 +29,30 @@ def get_allowed_origins() -> List[str]:
...
+    auth0 = oauth.register(
+        "auth0",
+        client_id=settings.AUTH0_CLIENT_ID,
+        client_secret=settings.AUTH0_CLIENT_SECRET,
+        api_base_url=settings.AUTH0_BASE_URL,
+        access_token_url=settings.AUTH0_ACCESS_TOKEN_URL,
+        authorize_url=settings.AUTH0_AUTHORIZE_URL,
+        client_kwargs=settings.AUTH0_CLIENT_KWARGS,
+    )
+    _app.state.auth0_client = auth0
```
lvreynoso, line 54: I see, this gets around the caching problem.

lvreynoso (approved):
Awesome job solving the caching problems!

src/backend/aspen/api/auth.py
jgadling: it will, because `get_settings` returns the settings object that we've stashed in app.state.aspen_settings!

src/backend/aspen/api/deps.py
```diff
@@ -1,15 +1,28 @@
...
 
-async def get_db(request: Request) -> AsyncGenerator[AsyncSession, None]:
+def get_auth0_client(request: Request):
+    # This parameter is added to the app when we instantiate it.
+    return request.app.state.auth0_client
+
+
+def get_settings(request: Request):
+    # This parameter is added to the app when we instantiate it.
+    settings = request.app.state.aspen_settings
```
jgadling, line 20: We stashed our settings object in `app.state` when we loaded the app, and every `request` object has that app attached at `request.app`, so this dependency is just returning the settings object we created at startup.

vincent-czi:
@jgadling There's a lot of useful comments in this PR, I think many of them would be useful as comments/docstrings in the code. I'm especially thinking about the parts explaining how Pydantic does its settings loading, but there are other explanations as well that I think would be quite helpful to keep alongside their code.

jgadling:
> @jgadling There's a lot of useful comments in this PR, I think many of them would be useful as comments/docstrings in the code. I'm especially thinking about the parts explaining how Pydantic does its settings loading, but there are other explanations as well that I think would be quite helpful to keep alongside their code.

Good call -- added comments to code!

PR #773, opened 2021-10-28
Make Happy CLI ask questions when updates take too long

jgadling:
### Summary:
- **What:** Update the Happy CLI to print some debug information / suggestions when updating stacks takes a long time
- **Ticket:** [sc166773](https://app.shortcut.com/genepi/story/166773)

### Notes:
If we update/create an rdev stack with a broken docker image (for example, with a typo in our code that puts the frontend/backend services in a crash loop) the Happy CLI doesn't provide any feedback about the failure. In fact, the update succeeds, and the running service hums along just fine **at the previous version of the image**. This can be very confusing for engineers who are expecting to see updated code running in rdev.

There's a flag in our terraform code `wait_for_steady_state` that forces terraform to wait until a new version of our ECS service is deployed and healthy before returning a success or failure. This flag was disabled in our remote-dev environment to make interacting with the Happy CLI faster. The confusion around this behavior is worse than the wait time though, so that flag has been re-enabled in this PR.

The normal ECS flow for updating a deployment looks something like:
1. Create a new deployment, and launch new tasks with the latest docker image
2. Wait for the new tasks to be healthy
3. If any of the new tasks is unhealthy after <X> time, restart it from scratch
4. Once the new tasks are healthy, drain the old tasks, deregister them, and stop them.

If the new tasks don't go healthy, Terraform will wait for up to 20 minutes before printing a failure. Our updates nearly always succeed in ~5 minutes, so I've chosen that (though it may need to be tweaked) as a cutoff for alerting the user that something might be going sideways. The CLI will print the event log for any services with incomplete deployments, and if the deployment appears to be incomplete due to a failing health check, it will print the command to show logs for that service.

### Demos:
When the `apply` phase of updating a stack takes > 5 minutes due to a crashing container, the output will look like so:

```
Running workspace rdev-editor
10:54:54 - plan_queued
10:54:59 - planning
10:55:26 - applying
This apply is taking an unusually long time. Are your containers crashing?
Incomplete deployment of service editor-frontend / Current status IN_PROGRESS:
  10:55 editor-frontend has started 2 tasks
  10:56 editor-frontend registered 2 targets in target-group
  10:56 editor-frontend deregistered 1 targets in target-group
  10:56 editor-frontend has begun draining connections on 1 tasks.
  10:56 editor-frontend deregistered 1 targets in target-group
  10:56 editor-frontend has started 1 tasks
  10:56 editor-frontend has started 1 tasks
  10:56 editor-frontend registered 1 targets in target-group
  10:56 editor-frontend deregistered 1 targets in target-group
  10:56 editor-frontend has begun draining connections on 1 tasks.
  10:56 editor-frontend registered 1 targets in target-group
  10:57 editor-frontend deregistered 1 targets in target-group
  10:57 editor-frontend has begun draining connections on 1 tasks.
  10:57 editor-frontend has started 1 tasks

Many "deregistered" events - please check to see whether your service is crashing:
  ./scripts/happy --env rdev logs editor frontend

```

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/scripts/happy
+++ b/scripts/happy
@@ -1,8 +1,9 @@
 #!/usr/bin/env python3
 import random
+from dateutil import tz
 from base64 import b64decode
 from contextlib import contextmanager
-from datetime import datetime
+from datetime import datetime, timedelta
 import io
 import json
 import os
@@ -249,7 +250,12 @@ class Stack:
         if not self.workspace:
             raise Exception(f"Could not find TFE workspace for stack {self.stack_name}")
 
-    def apply(self, wait):
+    def wait(self, stack_name, wait, orchestrator, services=None):
+        """Saves the variables and applies the workspace"""
+        self._ensure_workspace()
+        return self.workspace.wait(stack_name, orchestrator, services)
+
+    def apply(self, stack_name, wait, orchestrator, services=None):
         """Saves the variables and applies the workspace"""
         self._ensure_workspace()
         self.workspace.set_var(
@@ -270,7 +276,7 @@ class Stack:
 
         self.workspace.run_config_version(config_version_id)
         if wait:
-            return self.workspace.wait()
+            return self.workspace.wait(stack_name, orchestrator, services)
         return True
 
     def destroy(self):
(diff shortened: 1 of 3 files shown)


mmmmmmaya:
I can't verify the correctness of the code, but the output looks great!

scripts/happy
```diff
@@ -722,6 +736,44 @@ class Orchestrator:
...
+                if event_time < datetime.now().replace(tzinfo=tz.tzlocal()) - timedelta(seconds=600):
+                    continue
+                message = re.sub(r'^\(service ([^ ]+)\)', r"\1", event["message"])
+                message = re.sub(r'\(([^ ]+) .*?\)', r'\1', message)
+                message = re.sub(r':.*', '', message)
+                if "deregistered" in message:
+                    deregistered += 1
+                print(f'  {event_time:%H:%M} {message}')
+            # TODO -- this value really shouldn't be hardcoded, but it's a start.
+            if deregistered > 3:
```
vincent-czi, line 770: Nit: would be nice to set the `3` as a constant var, then use that instead, just to make it more visible as being a lever we can choose to turn up or down.

vincent-czi (approved):
This is a great feature add! I'm not super knowledgable with the AWS aspects, but everything Python-wise looks good to me.

PR #776, opened 2021-10-29
Allow Spaces in Private IDs

vincent-czi:
### Summary:
- **What:** Allow spaces in Private ID for samples
- **Ticket:** [sc<165800>](https://app.shortcut.com/genepi/story/<165800>)
- **Env:** https://private-id-spaces-frontend.dev.genepi.czi.technology

### Notes:
Various related things came up in the process of working on this:
- Looked into downstream impacts on the TSV metadata upload (if the new ID changes would affect anything), found some issues
  - => Wrote up a ticket and also documented a part of the TSV upload process that took me awhile to understand.
- Found out that UShER has pretty strict ID name requirements, lots of characters will cause it to completely break
  - => Took awhile to figure out what exactly those requirements are (I couldn't find it documented anywhere). Changed our backend fasta generator to produce different versions of the ID depending on the eventual downstream consumer, so now when a fasta is destined for usher, it will handle those naming requirements (eg, `my spaced id` in Aspen will become `my_spaced_id` when sent to UShER).
- Discovered that frontend was not enforcing most of the ID naming rules it stated in the upload instructions.
   - => At first this didn't seem like a big issue since it could be useful for international users to be flexible on characters. However, after understanding the restrictions when working with UShER, decided to just enforce our previously stated restrictions since that basically makes us usher-safe for IDs.

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR
- ~[ ] I have notified others of changes they need to make locally~ (none needed)

### DIFF
--- a/src/backend/aspen/fileio/fasta_streamer.py
+++ b/src/backend/aspen/fileio/fasta_streamer.py
@@ -80,11 +80,10 @@ class FastaStreamer:
         Certain downstream consumers (eg, UShER) restrict what characters can be
         used in the ID. Also handles any modifications that must be made to ID
         characters so they don't break the downstream consumer."""
+        output_id = identifier  # default, might get changed if specialty case
         if self.downstream_consumer == SpecialtyDownstreams.USHER.value:
-            output_ready_id = self._handle_usher_id(identifier)
-        else:
-            output_ready_id = identifier
-        return f">{output_ready_id}\n"
+            output_id = self._handle_usher_id(identifier)
+        return f">{output_id}\n"
 
     def _handle_usher_id(self, identifier) -> str:
         """Convert identifier into something that is UShER safe and roughly the same.
--- a/src/backend/aspen/fileio/fasta_streamer.py
+++ b/src/backend/aspen/fileio/fasta_streamer.py
@@ -1,4 +1,6 @@
-from typing import Set
+import re
+from enum import Enum
+from typing import Iterator, Optional, Set
 
 from sqlalchemy.orm import joinedload, Session
 from sqlalchemy.orm.query import Query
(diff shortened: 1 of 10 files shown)


src/backend/aspen/fileio/fasta_streamer.py
```diff
@@ -52,8 +68,37 @@ def stream(self):
...
+    def _output_id_line(self, identifier) -> str:
+        """Produces the ID line for current sequence in fasta.
+
+        Certain downstream consumers (eg, UShER) restrict what characters can be
+        used in the ID. Also handles any modifications that must be made to ID
+        characters so they don't break the downstream consumer."""
+        if self.downstream_consumer == SpecialtyDownstreams.USHER.value:
+            output_ready_id = self._handle_usher_id(identifier)
+        else:
+            output_ready_id = identifier
```
jgadling: Nit: I generally prefer to set a default, and then overwrite that default if there's a special case. This reduces the amount of nesting as our special cases get more complex, and makes the default case more obvious.

src/backend/aspen/fileio/fasta_streamer.py
```diff
@@ -52,8 +68,37 @@ def stream(self):
...
+        return f">{output_ready_id}\n"
+
+    def _handle_usher_id(self, identifier) -> str:
```
jgadling, line 88: Also nit: Ideally I'd prefer to break these filters out into their own classes and pass them into the FastaStreamer constructor. but since we only have one downstream atm, we can stress this later 😄

jgadling (approved):

src/backend/aspen/fileio/fasta_streamer.py
vincent-czi: Yeah, totally reasonable. Default case moved to above the `if` stuff

PR #778, opened 2021-11-01
Add sample delete endpoint

jgadling:
### Summary:
- **What:** Add an endpoint for deleting samples, and restrict it to be accessible only by the group that owns those samples.
- **Ticket:** [sc171771](https://app.shortcut.com/genepi/story/171771)

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/conftest.py
+++ b/src/backend/aspen/api/conftest.py
@@ -105,6 +105,7 @@ async def override_get_auth_user(
     if not found_auth_user:
         raise ex.UnauthorizedException("invalid user")
     request.state.auth_user = found_auth_user
+    return found_auth_user
 
 
 @pytest.fixture()
--- a/src/backend/aspen/api/conftest.py
+++ b/src/backend/aspen/api/conftest.py
@@ -16,6 +16,7 @@ from aspen.api.main import get_app
 from aspen.database import connection as aspen_connection
 from aspen.database import schema
 from aspen.database.connection import init_async_db
+from aspen.database.models import User
 
 USERNAME = "user_rw"
 PASSWORD = "password_rw"
@@ -100,7 +101,7 @@ async def override_get_db(
 
 async def override_get_auth_user(
     request: Request, session: AsyncSession = Depends(get_db)
-) -> None:
+) -> User:
     found_auth_user = await setup_userinfo(session, request.headers.get("user_id"))
     if not found_auth_user:
         raise ex.UnauthorizedException("invalid user")
--- a/src/backend/aspen/api/schemas/samples.py
+++ b/src/backend/aspen/api/schemas/samples.py
@@ -1,6 +1,6 @@
 from pydantic import constr
 
-from aspen.api.schemas.base import BaseRequest
+from aspen.api.schemas.base import BaseRequest, BaseResponse
 
 
 class SampleRequestSchema(BaseRequest):
@@ -8,5 +8,9 @@ class SampleRequestSchema(BaseRequest):
     name: constr(min_length=1, max_length=128, strict=True)  # type: ignore
 
 
-class SampleResponseSchema(BaseRequest):
+class SampleResponseSchema(BaseResponse):
     name: str
+
+
+class SampleDeleteResponse(BaseResponse):
+    id: int
--- a/src/backend/aspen/api/views/samples.py
+++ b/src/backend/aspen/api/views/samples.py
@@ -1,11 +1,15 @@
+import sqlalchemy as sa
 from fastapi import APIRouter, Depends
 from sqlalchemy.ext.asyncio import AsyncSession
+from sqlalchemy.orm.exc import NoResultFound
 from starlette.requests import Request
 
 from aspen.api.auth import get_auth_user
 from aspen.api.deps import get_db, get_settings
+from aspen.api.error import http_exceptions as ex
+from aspen.api.schemas.samples import SampleDeleteResponse
 from aspen.api.settings import Settings
-from aspen.database.models import User
+from aspen.database.models import Sample, User
 
 router = APIRouter()
 
@@ -20,12 +24,36 @@ async def list_samples(
     return False
 
 
-@router.delete("/{sample_id}")
+async def get_owned_sample_by_id(db, sample_id, user):
+    # TODO - We don't prevent a collision between public & private identifiers at the moment!!!
+    query = sa.select(Sample).filter(
+        sa.and_(
+            Sample.submitting_group == user.group,
+            sa.or_(
+                Sample.public_identifier == sample_id,
+                Sample.private_identifier == sample_id,
+            ),
+        )
+    )
+    results = await db.execute(query)
+    try:
+        return results.scalars().one()
+    except NoResultFound:
+        raise ex.NotFoundException("sample not found")
+
+
+@router.delete("/{sample_id:path}")
 async def delete_sample(
     sample_id: str,
     request: Request,
     db: AsyncSession = Depends(get_db),
     settings: Settings = Depends(get_settings),
     user: User = Depends(get_auth_user),
-) -> bool:
-    return False
+) -> SampleDeleteResponse:
+    # Make sure this sample exists and is delete-able by the current user.
+    sample = await get_owned_sample_by_id(db, sample_id, user)
+    sample_db_id = sample.id
+
+    await db.delete(sample)
+    await db.commit()
+    return SampleDeleteResponse(id=sample_db_id)
--- a/src/backend/aspen/api/views/samples.py
+++ b/src/backend/aspen/api/views/samples.py
@@ -24,15 +24,13 @@ async def list_samples(
     return False
 
 
-async def get_owned_sample_by_id(db, sample_id, user):
-    # TODO - We don't prevent a collision between public & private identifiers at the moment!!!
+async def get_owned_sample_by_id(db, group_id, sample_id, user):
     query = sa.select(Sample).filter(
         sa.and_(
-            Sample.submitting_group == user.group,
-            sa.or_(
-                Sample.public_identifier == sample_id,
-                Sample.private_identifier == sample_id,
-            ),
+            Sample.submitting_group == user.group,  # This is an access control check!
+            Sample.submitting_group_id
+            == group_id,  # This makes sure we included the correct group ID in our path.
+            Sample.public_identifier == sample_id,
         )
     )
     results = await db.execute(query)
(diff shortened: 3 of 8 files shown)


src/backend/aspen/api/conftest.py
```diff
@@ -105,6 +105,7 @@ async def override_get_auth_user(
     if not found_auth_user:
         raise ex.UnauthorizedException("invalid user")
     request.state.auth_user = found_auth_user
+    return found_auth_user
```
jgadling, line 109: This was just an oversight in the overridden method - the real method returns this value already.

src/backend/aspen/api/views/samples.py
```diff
@@ -20,12 +24,35 @@ async def list_samples(
...
+        )
+    )
+    results = await db.execute(query)
+    try:
+        return results.scalars().one()
+    except NoResultFound:
+        raise ex.NotFoundException("sample not found")
+
+
+@router.delete("/{group_id}/{sample_id:path}")
```
jgadling: I'm guessing that we may want to allow users that belong to groups other than the owner group (for example: admin users) to be able to delete these samples in the future. Since sample public id's are only required to be unique within a given group, we need to include the group id to uniquely identify a sample.

jgadling: Also! Since our public identifiers typically contain / characters, we need to tell fastapi/starlette that it's ok to use the entire path suffix as our resource ID, using the `:path` convertor:
https://www.starlette.io/routing/

src/backend/aspen/database/models/sequences.py
```diff
@@ -213,7 +213,7 @@ class UploadedPathogenGenome(PathogenGenome):
     sample_id = Column(Integer, ForeignKey(Sample.id), unique=True, nullable=False)
     sample = relationship(  # type: ignore
         Sample,
-        backref=backref("uploaded_pathogen_genome", uselist=False),
+        backref=backref("uploaded_pathogen_genome", uselist=False, cascade="delete"),
```
jgadling: This is where we tell SqlAlchemy that if we delete a sample, we should also delete any UploadedPathogenGenome rows associated with it:

https://docs.sqlalchemy.org/en/14/orm/cascades.html

src/backend/aspen/api/views/samples.py
vincent-czi: I think the above explanation about why we need `:path` should definitely be a comment in code.

Seems a shame we have to do it this way, but I agree that it's necessary. Hopefully we won't have any sub-attributes under a sample object we will eventually want to delete via CRUD, but that seems reasonably safe.

src/backend/aspen/database/models/sequences.py
```diff
@@ -211,9 +211,11 @@ class UploadedPathogenGenome(PathogenGenome):
         Integer, ForeignKey(PathogenGenome.entity_id), primary_key=True
     )
     sample_id = Column(Integer, ForeignKey(Sample.id), unique=True, nullable=False)
+    # The default value of cascade is "save-update, merge", so if we want to enable "delete", we 
+    # need to include the other options as well to maintain backwards compatibility.
     sample = relationship(  # type: ignore
         Sample,
-        backref=backref("uploaded_pathogen_genome", uselist=False),
+        backref=backref("uploaded_pathogen_genome", uselist=False, cascade="delete, merge, save-update"),
```
jgadling: This is where we tell SqlAlchemy to delete UploadedPathogenGenomes when we delete their associated samples:
https://docs.sqlalchemy.org/en/14/orm/cascades.html#save-update

src/backend/aspen/api/views/samples.py
jgadling: Yeah... I don't have a problem switching to using PK's instead of public ID's here -- in fact, we're going to *have* to do that for most resources since fairly few of them have an otherwise-unique identifier like samples do. But if we want to use public ID's I think we're stuck with reading a whole path suffix resource in URL's.

vincent-czi: On the whole, I think my vote is to eventually move over everything to PKs as the reference for API calls. It will definitely be a hassle, but the aspect of unambiguous references is so dang useful. That said, devil's advocate, the `/api/samples/validate-ids` route sure has a nice interface with just sending over a list of `sample_ids` and it being inclusive of private, public, AND gisaid. But if we ever have an ambiguous private ID that has been reused over multiple groups, things get messy quickly.

All that said, I do agree what's here in code right now with the whole path suffix resource is totally the right way to go about things if we want to stick with keeping public ID our primary way of referring to a sample in API calls. Even if we're sure we want to switch to PKs in the near future, I think this is totally the right way to go in the meantime.

... Hmmm, totally different alternative that just occurred to me. We _could_ go back to Public ID being fully unique. We can't really make Private ID unique since it should be confidential within a group, but I'd think that Public ID has less of a clear obligation to being totally confidential. Just being aware that _someone_ else took a public ID is an information leak, but it's pretty mild. Maybe we could revert that back to having a uniqueness requirement. Still have the issue with public IDs having `/` and other weird chars for URLs, but hey, we can at least trust it to be unambiguous again. And we could always handle the `/` aspect with URL encoding/decoding if we want to start sub-pathing the resource. ... Then again, if other resources will necessitate PKs anyway, probably not worth it to make samples the only one not using PKs. Just thinking out loud.

src/backend/aspen/api/schemas/samples.py
```diff
@@ -1,12 +1,16 @@
...
     name: constr(min_length=1, max_length=128, strict=True)  # type: ignore
 
 
-class SampleResponseSchema(BaseRequest):
+class SampleResponseSchema(BaseResponse):
     name: str
+
+
+class SampleDeleteResponse(BaseResponse):
+    id: int
```
vincent-czi, line 16: Minor: Since the sample deletion endpoint request is based on group_id and public_id, it seems odd that the response is sending back the id PK. Not a big deal, even less of a big deal if we'll be transitioning to PKs generally soon.

Related but not a comment about this PR: this might already be happening, I don't know the codebase well enough, but if we're shifting to using PKs more, we should try to enforce a convention of `id` always meaning the PK id of whatever is being worked with. If the var needs a more descriptive name, like `sample_id`, that would be assumed to be the sample's PK id, while we'd have to say `sample_public_id` to describe an id that wasn't the PK, etc. We've got so many id flavors running around with Samples, I'm just afraid of getting a non-PK id mixed up with a PK id. No issues with this PR -- it's adhering to this convention great -- I'm just thinking as we shift more stuff over to being PK focused.

src/backend/aspen/api/views/phylo_runs.py
```diff
@@ -182,3 +185,56 @@ async def kick_off_phylo_run(
     )
 
     return PhyloRunResponseSchema.from_orm(workflow)
+
+
+async def get_editable_phylorun_by_id(db, group_id, run_id, user):
+    aliased(PhyloRun)
```
vincent-czi: Does just applying `aliased` affect PhyloRun from there on out? I thought we'd have to assign a variable to the result then use that later in the query.

src/backend/aspen/api/views/phylo_runs.py
```diff
@@ -182,3 +185,56 @@ async def kick_off_phylo_run(
...
+async def list_runs(
+    request: Request,
+    db: AsyncSession = Depends(get_db),
+    settings: Settings = Depends(get_settings),
+    user: User = Depends(get_auth_user),
+) -> bool:
+    return False
+
+
+@router.delete("/{group_id}/{item_id}")
```
vincent-czi: If the `item_id` here is the phylo run PK -- I'm pretty sure it is -- I don't see the point of also putting the `group_id` in the route. We can identify the PhyloRun by the PK, and we know the `group_id` by inspecting the PhyloRun. It's structurally similar to the `delete_sample` route, but if we're going to move that over to PKs in the near future, I think we should drop the group_id aspect over there too when that happens.

src/backend/aspen/api/views/phylo_runs.py
```diff
@@ -182,3 +185,56 @@ async def kick_off_phylo_run(
...
+        )
+        .options(joinedload(PhyloRun.outputs))
+    )
+    results = await db.execute(query)
+    try:
+        run = results.scalars().unique().one()
+    except NoResultFound:
+        raise ex.NotFoundException("phylo run not found")
+    if run.workflow_status == WorkflowStatusType.STARTED:
+        raise ex.NotFoundException("Can't modify an in-progress phylo run")
```
vincent-czi: nit: I think this is more of a `BadRequestException` since we were able to find the resource

src/backend/aspen/api/views/samples.py
```diff
@@ -20,12 +24,38 @@ async def list_samples(
     return False
 
 
-@router.delete("/{sample_id}")
+async def get_owned_sample_by_public_id(db, group_id, public_id, user):
```
vincent-czi: nit: for phylo runs, it's `get_editable_phylorun_by_id`. I think we should either standardize on `editable` or `owned` if the result is "requesting user can delete this thing". But I also might be misunderstanding the eventual permission model and we do want to clearly distinguish for other purposes later.

src/backend/aspen/api/views/samples.py
jgadling: Ok, based on this feedback, and to reduce confusion. I'm going to update the **write** endpoints to be pk-only for now. I think we also make our read endpoints PK by default, but probably with some options to search via different fields (as needed).

src/cli/aspencli.py
```diff
@@ -317,6 +316,20 @@ def download_samples(ctx, sample_ids):
     print(resp.text)
 
 
+@samples.command(name="delete")
```
vincent-czi, line 320: I want to clarify: for both the sample deletion and the phylo run deletion, right now the user identity kicking off the CLI command has to be from the same group as the sample/run that is targeted for deletion for it to actually go through, right? At this moment, admins don't get super-user privilege for being able to delete from any group?

vincent-czi (approved):
Looks good! All my stuff is pretty minor, nothing that would need to block a merge certainly.

src/backend/aspen/api/schemas/samples.py
vincent-czi: You know, after thinking about it some more, as much as I like the idea of having `id` very consistently mean the PK, I don't know if it's actually doable what with all the model inheritance stuff we have going on.

jgadling: It is. Most/all of our endpoints are currently scoped to specific types of entities/workflows (in total misalignment with our db schema) so I think this will work fine.

src/cli/aspencli.py
jgadling: Right now there are no super-user privs. I'll check with Product as to what we want to do about that.

PR #784, opened 2021-11-04
Add V2 GET /samples/ endpoint

lvreynoso:
### Summary:
- **What:** Ports over the list samples endpoint to FastAPI.
- **Ticket:** [sc171825](https://app.shortcut.com/genepi/story/171825)
- **Env:** `https://list-samples-frontend.genepi.czi.technology/`

### Checklist:
- [x] I added labels to my PR


src/backend/aspen/api/schemas/samples.py
```diff
@@ -1,12 +1,45 @@
...
+class SampleResponseSchema(BaseResponse):
+    collection_date: str
+    collection_location: str
+    czb_failed_genome_recovery: bool
+    gisaid: SampleGisaidResponseSchema
+    lineage: SampleLineageResponseSchema
+    private: bool
+    private_identifier: str
+    public_identifier: str
+    sequencing_date: str = "N/A"
```
jgadling: Should we just leave these blank? I think the frontend should handle figuring out the best way to display an empty date.

src/backend/aspen/api/schemas/samples.py
```diff
@@ -1,12 +1,45 @@
...
+class SampleLineageResponseSchema(BaseResponse):
+    last_updated: Optional[str]
+    lineage: Optional[str]
+    probability: Optional[str]
+    version: Optional[str]
+
+
+class SampleGroupResponseSchema(BaseResponse):
+    group_id: int
+    group_name: str
```
jgadling: This is already under a "submitting_group" key. I think these should just be "id" and "name"

src/backend/aspen/api/schemas/samples.py
vincent-czi: I agree with this, but right now when dates are missing the convention is that it comes in as the string `N/A`. This was previously handled by the `format_date` and `format_datetime` util funcs.

We probably should break with that -- I agree with Jess, the FE should decide how to display null data -- but switching over to having date be `null` or just missing entirely from response will require corresponding changes to how FE handles and parses. I think it's best to just follow the convention for now and file a ticket to come back and convert the response for missing dates later.

vincent-czi: (To be explicit, there is currently code in the FE that is literally doing ` blah === "N/A"`, I don't think there's good handling for null/missing on the other hand)

jgadling: Yeah, let's fix the frontend as well.

lvreynoso: Null responses are working fine in my frontend, they just default to `---`.

vincent-czi: They work, but -- pretty sure -- they aren't working 100%? There's a general default that causes the `---` to appear for any table cell that is missing data (actually, looking at code right now, it's false-y, which is a different problem, sometimes you want to display `0`, but that's a tangent that has nothing to do with the BE). So while it won't break, I think the sorting and/or filtering might get mucked up?

There's definitely stuff in the FE that is looking for `"N/A"` as the canonical no-data for date. It's bad! Don't get me wrong! But I think for now we should just leave the date serializing for the `null` case as `"N/A"`. I think it will take more work on the FE to fix than the BE, so my vote is to plan to fix the FE in the near future, and as part of that work change over the BE serializing for dates (since that should just be a line or two of code).

My work on frontend table refactoring will be touching that code, so I could probably handle it as part of my work then. I can spawn a new ticket to track this so it doesn't get dropped if I wind up missing it during the refactor. Let me know if you're fine with that and I'll write the ticket. If you'd rather handle it now, just make sure to verify that filtering and column sorting are working correctly as well. Also, we should potentially check with Product/Design how they want the no-data case displayed for dates: on the Sequencing Date FE column the design was for it to be simply `-` instead of triple dash. [You can check out `src/frontend/src/views/Data/cellRenderers.tsx`, Line 118 to see how I had to make that happen. It's a bit ugly, but it's a limitation of how our table rendering currently works. I'm planning to make this easier in my refactor.]

jgadling: We need to modify the frontend code to point to the new endpoint anyway -- we can update the FE code to handle null dates at the same time.

src/backend/aspen/api/utils/authz.py
```diff
@@ -0,0 +1,64 @@
...
+        cansee.owner_group_id
+        for cansee in user.group.can_see
+        if cansee.data_type == DataType.PRIVATE_IDENTIFIERS
+    }
+    # add the user's own group
+    cansee_groups_private_identifiers.add(user.group_id)
+
+    public_identifier_filter = True
+    private_identifier_filter = True
+    if sample_ids:
```
jgadling, line 44: Nice additions!

src/backend/aspen/api/views/samples.py
```diff
@@ -1,27 +1,188 @@
...
+        sample_gisaid_tuple_map[sample_id] = SampleGisaidTuple(
+            origin_tuple.sample,
+            origin_tuple.uploaded_pathogen_genome,
+            gisaid_accession_workflow,
+            gisaid_accession,
+        )
+
+    # populate sample object according to response schema
+    results: MutableSequence[Mapping[str, Any]] = list()
+    for sample_gisaid_tuple in sample_gisaid_tuple_map.values():
```
jgadling, line 135: I think I learned how do do (most of) this with pydantic models - want to see if we can do that here too?

src/backend/aspen/app/views/sample.py
```diff
@@ -149,7 +149,7 @@ def _format_gisaid_accession(
         date_since_submitted = (
             datetime.date.today() - gisaid_accession_workflow.start_datetime.date()
         )
-        if date_since_submitted < GISIAD_REJECTION_TIME:
+        if date_since_submitted < GISAID_REJECTION_TIME:
```
jgadling, line 152: good catch!

src/backend/aspen/api/utils/date.py
```diff
@@ -0,0 +1,18 @@
+import datetime
+from typing import Optional
+
+
+def format_date(dt: Optional[datetime.date], format="%Y-%m-%d") -> Optional[str]:
```
jgadling: Can we just send datetimes to pydantic and let it figure this out?

src/frontend/src/common/api/index.tsx
```diff
@@ -3,7 +3,7 @@ import { jsonToType } from "src/common/utils";
 
 export enum API {
   USER_INFO = "/api/usergroup",
-  SAMPLES = "/api/samples",
+  SAMPLES = "/v2/samples/",
```
jgadling: I think we changed the response format -- there's more work to do before the frontend can just switch over to the new endpoint.

lvreynoso: changed back to old endpoint!

src/backend/aspen/api/views/samples.py
```diff
@@ -1,27 +1,156 @@
...
+        )
+    )
+    gisaid_accession_workflows_with_inputs_response = await db.execute(
+        gisaid_accession_workflows_inputs_query
+    )
+    gisaid_accession_workflows_with_inputs: List[
+        GisaidAccessionWorkflow
+    ] = gisaid_accession_workflows_with_inputs_response.scalars().all()
+
+    # get around circular references
```
jgadling, line 103: Is it possible to run the above query with an alias instead of breaking this block into separate queries?
https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.aliased

jgadling (approved):

PR #791, opened 2021-11-09
update isl accessions

phoenixAja:
### Summary:
- **What:** we also want to be able to update gisaid isl numbers through the update cli command. This will be used to fulfill an on call request 
- **Ticket:** [sc173210](https://app.shortcut.com/genepi/story/173210)
- **Env:** `<rdev link>`

### Demos:
usage: 
`python src/cli/aspencli.py --env staging samples update_public_ids --group-id 7 --private-to-public-id-mapping ~/Downloads/Aspen_Accessions_OC.csv --is-gisaid-isl
`
### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/app/views/sample.py
+++ b/src/backend/aspen/app/views/sample.py
@@ -517,6 +517,9 @@ def update_sample_public_ids():
     request_private_ids: list[str] = private_to_public.keys()
     request_public_ids: list[str] = private_to_public.values()
 
+    # check to see if public_identifier are gisaid isl accessions
+    public_ids_are_gisaid_isl: str = request_data.get("public_ids_are_gisaid_isl")
+
     group_id: int = request_data["group_id"]
 
     # check that all private_identifiers exist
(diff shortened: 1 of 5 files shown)


src/backend/aspen/app/views/sample.py
```diff
@@ -531,31 +534,52 @@ def update_sample_public_ids():
...
+            joinedload(Sample.uploaded_pathogen_genome),
         )
+        for s in samples_to_update:
+            isl_number = private_to_public[s.private_identifier]
+            accessions = s.uploaded_pathogen_genome.accessions()
+
+            # if the accessions length == 1 that means that currently no isl number exists on the uploaded_pathogen_genome
+            if not len(accessions) == 1:
+                # first accession public_identifier is 'gisaid_public_identifier', the second accession returned is the one that we want to update
+                accessions[1].public_identifier = isl_number
```
phoenixAja: I'm not quite sure why this behavior is like this.. @jgadling or @lvreynoso let me know if you know of a better way to do this

jgadling: We support (at least) GISAID and GENBANK accessions -- you're going to have to iterate over the accessions associated with the sample and find the one with the right type.

lvreynoso: That's the sample string used in the test's factory function for creating an `UploadedPathogenGenome`, that should be treated as an actual accession when you're testing things.  I think you would need something like 
```python
for accession in accessions:
    if isinstance(accession, GisaidAccession):
        ...
```
to find what you're looking for.

lvreynoso: Going by the currently existing database model this isn't strictly kosher, but since we're going to change how we handle GisaidAccessions very soon anyway I think it would be fine

phoenixAja: oh that was the test factory function accession.. that makes more sense. They were both of type `GisaidAccession` and that's why i didn't initially use this check.

phoenixAja: yeah this is a very temporary solution

phoenixAja: yeah they're both GisaidAccession objects, at least in the testing DB
Screen Shot 2021-11-10 at 9 26 18 AM

jgadling: I think (??) we only expect one gisaid accession associated with a sample. I'm not sure what's up with the testing db. Would it make sense to update both gisaid accession identifiers?

Either way, we *do* want to make sure we're only updating gisaid accessions, so blindly using the second accession isn't going to be safe.

phoenixAja: i can do the check and then update all of the associate gisaid accessions with that sample. I didn't _want_ to blindly use the second accession, it just seemed like that was the only way to grab the correct accession to update since they were both of the same type. Pushing a change now

phoenixAja: I think you're right that there should only be one gisaid accession associated with each sample, i think the reason for multiple accessions on a sample could be for the use case of a sample being uploaded to both gisaid and genbank? idk

phoenixAja:
thanks for feedback @jgadling and @lvreynoso!! I added your feedback

src/backend/aspen/app/views/sample.py
```diff
@@ -531,31 +534,52 @@ def update_sample_public_ids():
...
-    samples_to_update = (
-        g.db_session.query(Sample)
-        .filter(
-            and_(
-                Sample.submitting_group_id == group_id,
-                Sample.private_identifier.in_(private_to_public.keys()),
-            )
+        return jsonify(success=True)
+
+    else:
```
jgadling: This `else` is unnecessary -- we've already `return`ed in the `if` block above.

src/backend/aspen/app/views/sample.py
```diff
@@ -531,31 +534,52 @@ def update_sample_public_ids():
...
+                    if isinstance(accession, GisaidAccession):
+                        accession.public_identifier = isl_number
+            else:
+                # create a new accession if DNE
+                s.uploaded_pathogen_genome.add_accession(
+                    repository_type=PublicRepositoryType.GISAID,
+                    public_identifier=isl_number,
+                    workflow_start_datetime=datetime.datetime.now(),
+                    workflow_end_datetime=datetime.datetime.now(),
+                )
```
jgadling, line 563: are we missing some calls to `g.db_session.add()` and `g.db_session.commit()` here, or is that implicit?

phoenixAja: it was working without adding those calls in the test, but i added them now anyways

phoenixAja:
@jgadling or @lvreynoso would one of you mind approving so i can update these ids before i'm off for the next couple days?

jgadling (approved):

PR #800, opened 2021-11-16
Resources Page

phoenixAja:
### Summary:
- **What:** adding in aspen resources page
- **Ticket:** [sc150921](https://app.shortcut.com/genepi/story/150921)
- **Env:** `https://resources-frontend.dev.genepi.czi.technology `

### Demos:
use rdev

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/styles/support/components/List/index.tsx
+++ b/src/frontend/src/common/styles/support/components/List/index.tsx
@@ -1,28 +1,24 @@
-import { List as RawList } from "czifui";
+import { ListItem } from "czifui";
 import React from "react";
-import { StyledListItem } from "./style";
+import { StyledList } from "./style";
 
 interface Props {
   items: React.ReactNode[];
   ordered?: boolean;
-  listItemsShiftedLeft?: "s" | "m" | "l" | "xl"; // this prop will be used in styling for adding margin left
+  marginLeft?: "s" | "m" | "l" | "xl"; // this prop will be used in styling for adding margin left
 }
 
-const List = ({ items, ordered, listItemsShiftedLeft }: Props): JSX.Element => {
+const List = ({ items, ordered, marginLeft }: Props): JSX.Element => {
   return (
-    <RawList ordered={ordered}>
+    <StyledList ordered={ordered} marginLeft={marginLeft}>
       {items.map((item, index) => {
         return (
-          <StyledListItem
-            ordered={ordered}
-            key={index}
-            listItemsShiftedLeft={listItemsShiftedLeft}
-          >
+          <ListItem ordered={ordered} key={index}>
             {item}
-          </StyledListItem>
+          </ListItem>
         );
       })}
-    </RawList>
+    </StyledList>
   );
 };
 
--- a/src/frontend/src/common/styles/support/components/List/index.tsx
+++ b/src/frontend/src/common/styles/support/components/List/index.tsx
@@ -1,19 +1,25 @@
 import { List as RawList, ListItem } from "czifui";
 import React from "react";
+import { StyledListItem } from "./style";
 
 interface Props {
   items: React.ReactNode[];
   ordered?: boolean;
+  listItemsShiftedLeft?: "s" | "m" | "l" | "xl"; // this prob will be used in styling for adding margin left
 }
 
-const List = ({ items, ordered }: Props): JSX.Element => {
+const List = ({ items, ordered, listItemsShiftedLeft }: Props): JSX.Element => {
   return (
     <RawList ordered={ordered}>
       {items.map((item, index) => {
         return (
-          <ListItem ordered={ordered} key={index}>
+          <StyledListItem
+            ordered={ordered}
+            key={index}
+            listItemsShiftedLeft={listItemsShiftedLeft}
+          >
             {item}
-          </ListItem>
+          </StyledListItem>
         );
       })}
     </RawList>
--- a/src/frontend/src/common/styles/support/components/List/index.tsx
+++ b/src/frontend/src/common/styles/support/components/List/index.tsx
@@ -1,4 +1,4 @@
-import { List as RawList, ListItem } from "czifui";
+import { List as RawList } from "czifui";
 import React from "react";
 import { StyledListItem } from "./style";
 
--- a/src/frontend/src/common/styles/support/components/List/index.tsx
+++ b/src/frontend/src/common/styles/support/components/List/index.tsx
@@ -5,7 +5,7 @@ import { StyledListItem } from "./style";
 interface Props {
   items: React.ReactNode[];
   ordered?: boolean;
-  listItemsShiftedLeft?: "s" | "m" | "l" | "xl"; // this prob will be used in styling for adding margin left
+  listItemsShiftedLeft?: "s" | "m" | "l" | "xl"; // this prop will be used in styling for adding margin left
 }
 
 const List = ({ items, ordered, listItemsShiftedLeft }: Props): JSX.Element => {
--- a/src/frontend/src/common/styles/support/components/List/style.ts
+++ b/src/frontend/src/common/styles/support/components/List/style.ts
@@ -2,7 +2,7 @@ import styled from "@emotion/styled";
 import { getSpaces, ListItem, Props } from "czifui";
 
 interface ListItemProps extends Props {
-  // this prop is used for indenting list items
+  // this prop is used for indenting list items (undefined means list items will be indented 0px)
   listItemsShiftedLeft: "xs" | "s" | "m" | "l" | "xl" | undefined;
 }
 
--- a/src/frontend/src/common/styles/support/components/List/style.ts
+++ b/src/frontend/src/common/styles/support/components/List/style.ts
@@ -1,0 +1,27 @@
+import styled from "@emotion/styled";
+import { getSpaces } from "czifui";
+import { ListItem, Props } from "czifui";
+
+
+
+interface ListItemProps extends Props {
+  // this prop is used for indenting list items
+  listItemsShiftedLeft: "xs" | "s" | "m" | "l" | "xl" | undefined;
+};
+
+const doNotForwardProps = ["listItemsShiftedLeft"];
+
+export const StyledListItem = styled(ListItem, {
+  shouldForwardProp: (prop) => !doNotForwardProps.includes(prop as string),
+})`
+  ${(props: ListItemProps) => {
+    const { listItemsShiftedLeft } = props;
+    const spaces = getSpaces(props);
+
+    return `
+      margin-left: ${
+        spaces && listItemsShiftedLeft ? spaces[listItemsShiftedLeft] : 0
+      }px;
+    `;
+  }}
+`;
--- a/src/frontend/src/common/styles/support/components/List/style.ts
+++ b/src/frontend/src/common/styles/support/components/List/style.ts
@@ -1,13 +1,10 @@
 import styled from "@emotion/styled";
-import { getSpaces } from "czifui";
-import { ListItem, Props } from "czifui";
-
-
+import { getSpaces, ListItem, Props } from "czifui";
 
 interface ListItemProps extends Props {
   // this prop is used for indenting list items
   listItemsShiftedLeft: "xs" | "s" | "m" | "l" | "xl" | undefined;
-};
+}
 
 const doNotForwardProps = ["listItemsShiftedLeft"];
 
--- a/src/frontend/src/common/styles/support/components/List/style.ts
+++ b/src/frontend/src/common/styles/support/components/List/style.ts
@@ -1,24 +1,22 @@
 import styled from "@emotion/styled";
-import { getSpaces, ListItem, Props } from "czifui";
+import { getSpaces, List, Props } from "czifui";
 
-interface ListItemProps extends Props {
+interface ListProps extends Props {
   // this prop is used for indenting list items (undefined means list items will be indented 0px)
-  listItemsShiftedLeft: "xs" | "s" | "m" | "l" | "xl" | undefined;
+  marginLeft: "xs" | "s" | "m" | "l" | "xl" | undefined;
 }
 
-const doNotForwardProps = ["listItemsShiftedLeft"];
+const doNotForwardProps = ["marginLeft"];
 
-export const StyledListItem = styled(ListItem, {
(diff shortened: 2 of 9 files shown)


src/frontend/src/components/NavBar/components/RightNav/index.tsx
```diff
@@ -19,6 +19,13 @@ export default function RightNav(): JSX.Element {
       <div className={cx(style.item, style.link)}>Sign In</div>
     </a>
   );
+  const genEpiResourcesLink = (
+    <a href={ROUTES.RESOURCES} data-test-id="navbar-sign-in-link">
```
vincent-czi: I don't think this should have the same `data-test-id` since it's not actually the sign-in link

src/frontend/src/common/styles/support/components/List/style.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const StyledListItem = styled(ListItem, {
+  shouldForwardProp: (prop) => !doNotForwardProps.includes(prop as string),
+})`
+  ${(props: ListItemProps) => {
+    const { listItemsShiftedLeft } = props;
+    const spaces = getSpaces(props);
+
+    return `
+      margin-left: ${
```
vincent-czi: Unless I'm missing something, I think `listItemsShiftedLeft` is kinda named the opposite of what it's doing? Isn't it shifting items to the _right_ because it provides a `margin-left` CSS? Also, since it's a pretty direct way to use `margin-left` maybe also just a more tightly directed name would be reasonable? Like `marginLeft` or if you want to underline that it's using the `spaces` from czifui, `marginLeftSpace`?

src/frontend/src/common/styles/support/components/List/index.tsx
```diff
@@ -1,19 +1,25 @@
...
 }
 
-const List = ({ items, ordered }: Props): JSX.Element => {
+const List = ({ items, ordered, listItemsShiftedLeft }: Props): JSX.Element => {
   return (
     <RawList ordered={ordered}>
       {items.map((item, index) => {
         return (
-          <ListItem ordered={ordered} key={index}>
+          <StyledListItem
```
vincent-czi: I might be missing something, but is there a reason we can't just shift the parent List component rather than shifting each of its child items? I can totally believe that I'm wrong and it looks differently, but if it winds up being identical, I think it would be preferable to style the one parent rather than each child. When the rdev goes live, I'll pop on to it and try playing with the CSS in my dev tools and see if it is equivalent or if I'm wrong (unless you already tried it the other way and this really is the necessary approach).

src/frontend/src/views/Resources/index.tsx
```diff
@@ -0,0 +1,228 @@
...
+    </Title>
+  );
+
+  const renderGenomicEpiSeries = () => {
+    return (
+      <>
+        <H2>Genomic Epidemiology Seminar Series</H2>
+        <P>
+          <B>
+            Seminars occur every second Thursday from 11am to 12pm PST over
```
vincent-czi: This is style thing, not a code thing, but I think it's better to refer to time as "Pacific time" rather than "PST" since "PST" indicates it will actually shift around depending on if it's daylight savings time ("PDT"), but I'm being _suuuuper_ pedantic.

src/frontend/src/views/Resources/index.tsx
```diff
@@ -0,0 +1,228 @@
...
+                time.
+              </span>,
+              <span key={2}>
+                <NewTabLink href="https://www.youtube.com/watch?v=jYpjzP22HcM">
+                  Week 3
+                </NewTabLink>{" "}
+                - This week we discussed phylogeography, the technique in
+                genomic epidemiology of inferring spatial migration patterns of
+                a pathogen across the tree. We discussed the inferential
+                procedure that allows this, as well caveats and things to be
```
vincent-czi: Copy-editing: "as well caveats" --> "as well **as** caveats"

src/frontend/src/views/Resources/index.tsx
```diff
@@ -0,0 +1,228 @@
...
+                those different structures are useful for, and which
+                repositories each can be submitted to. She finished the talk off
+                with a discussion of handling GISAID rejections and looking at
+                calls in BAM files.
+              </span>,
+              <span key={6}>
+                <NewTabLink href="https://www.youtube.com/watch?v=HVorizRS4wk">
+                  Week 7
+                </NewTabLink>{" "}
+                - In todays seminar we talked about phylogenetic nomenclature
```
vincent-czi: Copy-editing: "In todays seminar" --> "In **today's** seminar"

src/frontend/src/views/Resources/index.tsx
```diff
@@ -0,0 +1,228 @@
...
+                systems (why we use them and types of systems). We finished off
+                with a more detailed discussion of how the Pango nomenclature
```
vincent-czi: Copy-editing: "We finished off with a more detailed discussion of how the Pango nomenclature system for SARS-CoV-2." --> "We finished off with a more detailed discussion of **~how~** the Pango nomenclature system for SARS-CoV-2." (Or the sentence needs to be restructured to include the target of what that _how_ was explaining, like "how the Pango nomenclature system for SARS-CoV-2 **works**.")

src/frontend/src/views/Resources/index.tsx
```diff
@@ -0,0 +1,228 @@
...
+              </span>,
+              <span key={9}>
+                <NewTabLink href="https://www.youtube.com/watch?v=cLLV2VpgXCU">
+                  Week 10
+                </NewTabLink>{" "}
+                - This week I gave a demo of Aspen, the new piece of software
+                that CZI is developing for managing genomic data, that
+                eventually will allow folks to build Nextstrain trees
+                independently with a GUI interface. Since that capability
+                isn&apos;t active quite yet, I also discussed phylogenetic
```
vincent-czi, line 146: Editing: I wonder if there should be an editorial aside put into this week's text to point out that Aspen does now actually have that capability. Perhaps something at the end of paragraph like "[Since the recording of this video, Aspen has added this capability.]"? @Janeece Thoughts?

src/frontend/src/views/Resources/index.tsx
```diff
@@ -0,0 +1,228 @@
...
+                examples of how you can use genomic epidemiology to investigate
+                the epidemiology of TB.
+              </span>,
+              <span key={12}>
+                <NewTabLink href="https://www.youtube.com/watch?v=lFJ_2G4u8w4">
+                  Week 14
+                </NewTabLink>{" "}
+                - This week the focus was on practical applications of genomic
+                epidemiology, and I walked through some case studies showing
+                different ways in which genomic epidemiology had supported epi
```
vincent-czi: Copy-editing: I'm not 100% sure on this one, but I _think_ "genomic epidemiology had supported epi investigations" --> "genomic epidemiology **has** supported epi investigations"? I'm pretty confident of the suggested change, I think "had" implies it used to but no longer does any more, while "has" implies past activity that is (probably) still ongoing, which is the one I think we'd want.

src/frontend/src/views/Resources/index.tsx
vincent-czi: Maybe specifically point out the "Aspen Tutorial!" section at the bottom of the list since it's showing those capabilities off? Or just have a direct link to the video in the editorial aside that points to the same place as "Aspen Tutorial!" does?

src/frontend/src/views/Resources/index.tsx
```diff
@@ -0,0 +1,228 @@
...
+    );
+  };
+
+  return (
+    <>
+      <Head>
+        <title>Aspen | Resources</title>
+      </Head>
+      <NarrowContainer>
+        {renderIntro()}
```
vincent-czi: I know this approach of `renderFooBar` matches the approach used in Privacy, Terms, and FAQ, but it seems kind of unnecessary to me?

For example, `renderQuestions` could just be written as a component of `Questions` with no changes to the func definition, then it would just get used as `<Questions />` instead of being called as `{renderQuestions()}`. It is purely a style thing, but that seems more in line with how we do the rest of our components. Seems odd that the big, text-heavy pages use a different convention.

vincent-czi (approved):
Looks good! All my stuff is really minor. I put on my English-editing hat and read through all the page text, but none of those are code things.

src/frontend/src/common/styles/support/components/List/index.tsx
vincent-czi: Okay, I just tried to remove it from the `li` level and put one single `margin-left` at the `ul` tag level and it looks the same to me, so I'd recommend to slot it over at the parent level instead (unless there is some kind of mutating margin-left we want depending on specific stuff regarding the ListItem in other places this gets used).

src/frontend/src/common/styles/support/components/List/style.ts
phoenixAja: that is totally fair, i'll change it to `marginLeft`

src/frontend/src/views/Resources/index.tsx
phoenixAja: yeah i was trying to follow with the stylings from the other pages, i can change it though, not a big dealio

src/frontend/src/common/styles/support/components/List/index.tsx
phoenixAja: i can shift the styling to the parent group!

phoenixAja:
@Janeece do you approve of @vincent-czi 's copy changes? 😄

src/frontend/src/views/Resources/index.tsx
Janeece: I agree, but we might need to check with Alli about her intent for this sentence.

src/frontend/src/views/Resources/index.tsx
Janeece: Hmmm thats a good call out. yeah, maybe an editorial like: "Since the recording of this video, Aspen has added self-serve tree building. See [Aspen Tutorial!] for more info." (We dont really have a style established for editorial notes. )

One thing i want to be be aware of is how much we want to support adding editorials to outdated videos/updated capabilities... I'm not sure who owns that.

src/frontend/src/views/Resources/index.tsx
Janeece: I agree, i think "has" is slightly better

Janeece (commented):
@phoenixAja @vincent-czi 
In general I agree with Vincent's comments. I wonder if we need to run them by Alli?
Some other comments:
- [x] Can we make seminar links 600 weight?
- [x] Link hovers should be primary500 (currently looks like default blue)
- [x] The links for week 12 - aspen tutorial don't go to the video indicated (not sure if this is just not having those links yet)
- [x] Increase the size of the "Questions" headline to match "Geonomi Epi Seminar Series" (fontHeaderM > fontHeaderXl)

mmmmmmaya:
My only question is: why did you choose to go with Aspen's `List` over SDS `List`? SDS has a completed `List` & `ListIem` we can use here. IMO, I prefer to use what's available in the design system rather than building out what is in Aspen further to prevent divergence.

src/frontend/src/views/Resources/index.tsx
```diff
@@ -0,0 +1,228 @@
...
+        {renderGenomicEpiSeries()}
+        {renderQuestions()}
+        <div style={{ height: "50px" }} />
```
mmmmmmaya, line 228: ok, actually, also what is this div about?

phoenixAja: this div is also in the faq, privacy, and terms pages, so i kept it to be consistent with those other pages

phoenixAja:
> My only question is: why did you choose to go with Aspen's `List` over SDS `List`? SDS has a completed `List` & `ListIem` we can use here. IMO, I prefer to use what's available in the design system rather than building out what is in Aspen further to prevent divergence.

This List component in aspen is a wrapper around the SDS List, ListItem components though?

src/frontend/src/views/Resources/index.tsx
mmmmmmaya: It seems like this is probably for spacing. We should update all the pages to use a styled component that has a margin-bottom instead of putting in the weird div with inline styling.

mmmmmmaya:
It doesn't seem like a big deal to me atm, but Aspen's List is only used on long text pages, and doesn't add any functionality or offer a large benefit over using sds list directly. Just my opinion, but it feels like an unnecessary level of indirection to avoid using ListItem. I guess I could go either way on it. Curious what @vincent-czi thinks.

vincent-czi:
> It doesn't seem like a big deal to me atm, but Aspen's List is only used on long text pages, and doesn't add any functionality or offer a large benefit over using sds list directly. Just my opinion, but it feels like an unnecessary level of indirection to avoid using ListItem.

I'm ambivalent. I haven't looked at the List component in czifui, but I agree that it's generally better to just use SDS purely since additional indirection to our own usage adds mental overhead. If we're not really doing something new with it, it's just another link to walk through when trying to trace how something works down the road. That said!, since all this is pretty much identical across Privacy, FAQ, Terms, it feels like doing it "correctly" in this component makes it confusing when comparing across its static page siblings because then it's the odd-duck out. So, ehhhhh?

Perfect world, unlimited time: Make the change to using SDS in a purer way, find every place using the old Aspen-version of List, and switch over to the purer approach.
Pressed for time: Leave it alone, slightly more tech debt but at least it's the same tech debt as elsewhere

In general, for this kind of choice where there's a clearly better way, but it's different from established convention, my first question is always, "which will be more confusing for a future dev to encounter in a couple months?" This seems pretty much like a six-in-one, half-dozen-in-the-other kind of situation if Resources is the only page using the more pure version, so ... ehhhhh? But if deadlines aren't breathing down the neck yet, seems like a switchover that would be good.

Janeece:
@phoenixAja this looks great. my only comment is it looks like Questions is still a size too small -- it should be an H2 and same size as "Gen Epi Seminar Series".

Janeece (approved):
Looks great, thanks!

phoenixAja:
@vincent-czi and @mmmmmmaya i think i'll merge this in and address the List component stuff in a follow up PR (along with div spacing at the bottom of the pages)

PR #801, opened 2021-11-16
Fix sample ID validation.

jgadling:
### Summary:
- **What:** Fix querying for gisaid sample ID's if the sample id begins/ends with 1 or more characters contained in the string "hCoV-19/"

### Notes:
`strip()` treats an input string as a **list of characters** that it should remove from the beginning/end of a string, rather than as a literal string to strip. For example:

```
>>> "hCoV-19/some-samplehCV1".strip("hCoV-19/")
'some-sample'
```

What we really want in this case is to just strip off a prefix string if it exists:
```
>>> mysample = "hCoV-19/some-samplehCV1"
>>> mysample.replace("hCoV-19/", "") if mysample.startswith("hCoV-19/") else mysample
'some-samplehCV1'
```

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/utils/gisaid.py
+++ b/src/backend/aspen/api/utils/gisaid.py
@@ -29,7 +29,10 @@ async def get_matching_gisaid_ids(
     # (Gisaid data is prepped by Nextstrain which strips off this prefix)
 
     # first create a mapping of ids that were stripped (so we can return unstripped id later)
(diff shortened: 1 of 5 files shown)


mmmmmmaya (approved):
👏

lvreynoso:
Lgtm!

PR #807, opened 2021-11-17
Maya/turn on gisaid

mmmmmmaya:
### Summary
- **What:** Turn on GISAID for the public
- **Why:** Feature release
- **Env:** https://gisaid-frontend.dev.genepi.czi.technology

### Testing
1. Clear cookies to turn off all local feature flags 
1. Log in
2. Verify GISAID ingest features are working

### Notes
This will not be merged until all [gisaid feedback](https://docs.google.com/document/d/1Fa4AT00U8CtS4nfTACJnKJtPr4XEu5r0syDJM0RfOBY/edit) changes are merged into the repo.

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/components/library/data_subview/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/index.tsx
@@ -265,6 +265,7 @@ const DataSubview: FunctionComponent<Props> = ({
 
   const numCheckedSamples = checkedSamples?.length;
   const hasCheckedSamples = numCheckedSamples > 0;
+  const hasTooManySamples = numCheckedSamples > 2000;
 
   const render = (tableData?: TableItem[]) => {
     let downloadButton: JSX.Element | null = null;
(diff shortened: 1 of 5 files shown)


src/frontend/src/common/components/library/data_subview/index.tsx
```diff
@@ -266,10 +265,6 @@ const DataSubview: FunctionComponent<Props> = ({
 
   const numCheckedSamples = checkedSamples?.length;
   const hasCheckedSamples = numCheckedSamples > 0;
-  const hasTooManyCheckedSamples = numCheckedSamples > 2000;
```
vincent-czi, line 269: Is there any checking for too many samples anywhere anymore? I just searched the whole FE code base and I can't find `2000` show up anywhere other than in English ("Depending on the Tree Type, add up to 2000 samples", `src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleIdInput/components/InputInstructions/index.tsx`). Are we actually enforcing this in code for the FE still?

If not, I wonder what happens if someone submits a million gisaid samples. Heck, even if we did enforce it in FE, I wonder what would happen if someone just manually hit the backend API with a million gisaid samples. Probably not something to worry about, though.

vincent-czi (approved):
Looks good. One comment isn't a blocker, but might be something we should address in near-term depending on why we had the "too many samples" restriction before.

src/frontend/src/common/components/library/data_subview/index.tsx
mmmmmmaya: oh crap, very good catch. I definitely don't want to disabled that upper limit

PR #809, opened 2021-11-17
Maya/gisaid tweaks

mmmmmmaya:
### Summary
- **What:** 
  - Update copy
  - Styling
  - Fixing types
  - Don't hide add button when user blurs input if sample IDs were ever submitted in the past
- **Why:** Team feedback
- **Env:** https://gisaid-frontend.dev.genepi.czi.technology
- **Discussion:** https://docs.google.com/document/d/1Fa4AT00U8CtS4nfTACJnKJtPr4XEu5r0syDJM0RfOBY/edit

### Notes
View with whitespace changes off to save 20 lines of review.

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleIdInput/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleIdInput/index.tsx
@@ -72,6 +72,8 @@ const SampleIdInput = ({
     return compact(trimmedTokens);
   }, [inputValue]);
 
+  // TODO (mlila): we don't actually surface this error to the user anywhere, but in the
+  // TODO          future we probably should if this happens with any frequency.
   const validateSampleIdentifiersMutation = useValidateSampleIds({
     onError: () => {
       setValidating(false);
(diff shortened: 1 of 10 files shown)


src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleIdInput/index.tsx
```diff
@@ -66,37 +72,35 @@ const SampleIdInput = ({
...
-
-        setIdsInFlight([]);
-        setFoundSampleIds(foundIds);
-        setMissingSampleIds(missingIds);
-        handleInputValidation(foundIds, missingIds);
-      },
-    }
-  );
+  const validateSampleIdentifiersMutation = useValidateSampleIds({
+    onError: () => {
```
vincent-czi, line 78: This is purely a network request error type thing or a 500, 404, etc, right? It looks like we're just resetting the process back to scratch, but we don't try to notify the user right now, yeah? I don't see anything getting triggered here that would then cause an error to be displayed to the user. Since those are pretty darn rare, I think that's fine, but maybe drop a comment or a TODO mentioning that we just reset and don't currently display an error to the user but that we might want to down the road? [Unless I'm wrong and the error message just gets made elsewhere somehow and I haven't gotten there yet.]

vincent-czi (approved):
Looks good! Grain of salt because I don't have a ton of context for what the larger code on this section looks like, but everything seems clear in this PR. My one comment is pretty minor.

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleIdInput/index.tsx
mmmmmmaya: Yea, I think this is a good point. I will add a comment!

PR #810, opened 2021-11-20
pages cleanup

phoenixAja:
### Summary:
- **What:** change pages to use SDS List/ListItem components directly, added padding to the bottom of NarrowContainer component (replacing bottom div that was previously used for spacing)
- **Ticket:** No Ticket, just feedback from [resources page PR](https://github.com/chanzuckerberg/aspen/pull/800)
- **Env:** [rdev](https://pages-cleanup-frontend.dev.genepi.czi.technology/)

### Demos:
see rdev

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/frontend/src/common/styles/support/style.ts
```diff
@@ -14,6 +14,7 @@ import {
 export const NarrowContainer = styled.div`
   max-width: 800px;
   margin: 90px auto;
+  padding-bottom: 50px;
```
vincent-czi: ~Lord this is minor, but I would probably put this as just more margin than padding since I think that's closer to the original intent of the spacing block. I know we already have a `margin`, but we could increase the bottom margin by altering the shorthand use like so:~
```
  margin: 90px auto 140px;
```

Huh! That's one I didn't know about, I tried this out before actually putting in the comment -- nope, doesn't work! Looks like `margin-bottom` at the very end of the page (at least in Chrome) just gets obliterated. No need to have a margin if it's just a margin between an element and the void of nothing-ness? I dunno what the reasoning there is, but looks like `padding-bottom` works like you'd expect and actually does space it out.

That said, it does mean that the shorthand `90px` for top-bottom is only half-true here! It provides a real 90px up top, but the bottom 90px just gets eaten by the void, so we only have the 50px from padding-bottom. Go figure.

So definitely keep the padding-bottom, but maybe adjust its amount and/or adjust the `margin` shorthand so it's not implying something that actually doesn't happen? Like
```suggestion
  margin: 90px auto 0;
  padding-bottom: 50px;
```

src/frontend/src/views/Faq/index.tsx
```diff
@@ -215,12 +228,20 @@ export default function Faq(): JSX.Element {
...
       </P>
-      <List
-        items={[
-          `After two weeks, we will automatically submit the sample to GISAID on your behalf. This upload includes the pathogen genome; collection date; sequencing lab (i.e., your group); and location (as always, at the county level or above). `,
-          `Consistent with the prior COVID Tracker program’s policies, the California Department of Public Health (CDPH) will be able to see non-private samples, but with your internal, private identifiers redacted.`,
-        ]}
-      />
+      <List>
+        <ListItem>
+          After two weeks, we will automatically submit the sample to GISAID on
```
vincent-czi: Not a code thing, but I wonder if we should still have this? We don't actually do this right now, right?

vincent-czi: (Not as part of this PR, but maybe just like a note we should take down and figure out if a change is needed next standup?)

src/frontend/src/views/Privacy/index.tsx
```diff
@@ -96,12 +95,27 @@ const PrivacyPolicy = (): JSX.Element => {
...
+          you share your data with can see your samples, but not your private,
+          internal identifiers.
+        </ListItem>
+        <ListItem>
+          Human genetic data with uploaded data is processed only so we can
+          filter out and permanently delete it. We do not keep this non-pathogen
+          genomic data and it’s not necessary to operate the tool.
+        </ListItem>
+        <ListItem>
           <span key="0">
```
vincent-czi: Same thing with `key` here

src/frontend/src/views/Privacy/index.tsx
```diff
@@ -367,8 +381,8 @@ const PrivacyPolicy = (): JSX.Element => {
         Upload Data includes Raw Sequence Data, Pathogen Consensus Genomes, and
         Sample Metadata.
       </P>
-      <List
-        items={[
+      <List>
+        <ListItem>
           <span key="0">
```
vincent-czi: Probably unnecessary `key` again here and below in this section

src/frontend/src/views/Privacy/index.tsx
```diff
@@ -540,15 +570,17 @@ const PrivacyPolicy = (): JSX.Element => {
 
       <UnderLineHeader>What We Collect</UnderLineHeader>
 
-      <List
-        items={[
+      <List>
+        <ListItem>
           <span key="0">
```
vincent-czi: More `key`

src/frontend/src/views/Privacy/index.tsx
```diff
@@ -690,8 +729,8 @@ const PrivacyPolicy = (): JSX.Element => {
       <P>
         We retain your personal data only as long as is reasonably necessary:
       </P>
-      <List
-        items={[
+      <List>
+        <ListItem>
           <span key="0">
```
vincent-czi: More key

src/frontend/src/views/Privacy/index.tsx
```diff
@@ -728,21 +775,34 @@ const PrivacyPolicy = (): JSX.Element => {
...
+          member of their organization group on the basis of the Pathogen
+          Consensus Genome.
+        </ListItem>
+        <ListItem>
+          Users are able to access and download analytical results relating to
+          Upload Data submitted by a member of their organization group within
+          Aspen.
+        </ListItem>
+        <ListItem>
           <span key="0">
```
vincent-czi: key

src/frontend/src/views/Resources/index.tsx
```diff
@@ -43,17 +43,18 @@ export default function Resources(): JSX.Element {
         </P>
         <H3>Seminar Recordings</H3>
         <div>
-          <List
-            marginLeft={"m"}
-            items={[
+          <StyledList>
+            <ListItem>
               <span key={0}>
```
vincent-czi: `key`s in this file as well

src/frontend/src/views/Terms/index.tsx
```diff
@@ -70,18 +69,24 @@ export default function Terms(): JSX.Element {
     return (
       <>
         <H3>Summary of Key Things to Know</H3>
-        <List
-          items={[
+        <List>
+          <ListItem>
             <span key="0">
```
vincent-czi: `key`s in this file

src/frontend/src/views/Terms/index.tsx
```diff
@@ -478,9 +577,9 @@ export default function Terms(): JSX.Element {
             </NewTabLink>{" "}
             (email does not constitute notice)
             <br />
-          </P>,
-        ]}
-      />
+          </P>
+        </ListItem>
+      </List>
       <div style={{ height: "50px" }} />
```
vincent-czi: Want to remove this now because it's done via NarrowContainer, right?

vincent-czi (approved):
Looks good! I've got some comments in there that would be nice to address, but I don't think it would need a re-review after.

**Important** to point out, Privacy page still has a 50px spacer, it weirdly happens in the `renderChangesToPrivacyNotice` part. There's also one in Terms page, but I called that out with a comment, but I can't comment on the one in Privacy because it's not next to any line changes.

Bunch of comments about the `key` attributes. I'm like 95% sure they can be removed because they are a React thing for handling array-map rendering, but maybe I'm wrong.

src/frontend/src/views/Faq/index.tsx
phoenixAja: i think we should just remove it for now.

src/frontend/src/views/Terms/index.tsx
phoenixAja: good catch

phoenixAja:
thanks for the thoughtful feedback @vincent-czi, i'm going to merge this in now

PR #813, opened 2021-11-23
Maya/137090 crud header tooltip

mmmmmmaya:
### Summary
- **What:** Add tooltips to sample & tree tables
- **Why:** create an interaction pattern & user clarity
- **Ticket:** [[sc-137090]](https://app.shortcut.com/genepi/story/137090)
- **Env:** https://tooltips-frontend.dev.genepi.czi.technology [pending]
- **Design:** https://czi.quip.com/MCCJANbGalkW/Aspen-Table-Tooltips

### Testing
1. All pages still load
1. Sample and tree tables all have a tooltip on hover
2. Tooltip links still work (lineage is an example)

### Notes
There's a lot of refactoring going on here, but it felt worth it to improve clarity/ease of reading and maintaining the table headers going forward.

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/frontend/src/common/components/library/data_table/components/HeaderRow/index.tsx
```diff
@@ -0,0 +1,60 @@
+import { isEqual } from "lodash";
```
phoenixAja, line 1: These file names are a little confusing, it's not immediately clear to me how `HeaderRow` is different from `TableHeader`.

phoenixAja: hmm i see so HeaderRow is an implementation of TableHeader, still a little confusing but not as confusing as i originally thought, no change needed

src/frontend/src/common/components/library/data_table/components/HeaderRow/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+interface Props {
+  handleHeaderCheckboxClick(): void;
+  handleSortClick(key: string[]): void;
+  headers: Header[];
+  isHeaderChecked: boolean;
+  isHeaderIndeterminant: boolean;
+  isSortedAscending: boolean;
+  shouldShowCheckboxes: boolean;
+  sortColKey: string[];
```
phoenixAja, line 14: i think renaming this to `sortColKeys` make more sense since we have multiple columns to sort on?

src/frontend/src/common/components/library/data_table/index.module.scss
```diff
@@ -8,49 +8,6 @@
...
-}
-
-// Checkboxes in a header need to match alignment of row checkboxes below, so
-// we override padding. Headers with checkboxes should use above `.header` class
-// AND this class together. Because it comes second, it overrides, but everything
-// else continues to work as usual.
-.headerWithCheckbox {
-  padding: 0;
-}
-
```
phoenixAja, line 53: yay! Get that scss out!

phoenixAja (approved):
this is great! Good refactoring, only small comments

src/frontend/src/common/components/library/data_table/components/HeaderRow/index.tsx
mmmmmmaya: Yea, I was mixed on this. The reason I kept it as singular was because the thing we're comparing it to (`sortKey`) is also an array specifying potentially multiple strings, but was named singular.

phoenixAja: i would also rename that to sortKeys 😆, but too late now

Janeece:
@mmmmmmaya So excited to see these! Overall they look great. Here's my comments:

- [ ] Theres a long delay on the tooltip visibility which causes them to overlap with each other. Can we reduce/remove the amount of time theses tooltips are visible after moving the cursor?

https://user-images.githubusercontent.com/63013360/145257254-16c282fb-4168-4151-b9b1-30023ea6e7c0.mov


- [ ] Update references to “Aspen” to “CZ GEN EPI”. See Public ID & Upload Date tooltips. *Note: this update is actually present in prod now, but “CZ GEN EPI” needs to be changed to caps*

PR #814, opened 2021-11-24
Import GISAID locations, use them for Samples

lvreynoso:
### Summary:
- **What:** Import GISAID locations into new `locations` table; give samples a `location_id` that gives it a relationship to a `Location`; allow users to search for GISAID locations in frontend and use them when creating samples.
- **Env:** [https://location-dropdown-frontend.dev.genepi.czi.technology/upload/1](https://location-dropdown-frontend.dev.genepi.czi.technology/upload/1)

### Demos:

Screen Shot 2021-11-24 at 16 10 50
Screen Shot 2021-11-24 at 16 11 31
Screen Shot 2021-11-23 at 16 32 55
Screen Shot 2021-11-23 at 16 37 03
Screen Shot 2021-11-23 at 16 48 49

### Checklist:
- [x] I added labels to my PR


src/frontend/src/common/types/location.d.ts
```diff
@@ -0,0 +1,15 @@
+interface GisaidLocation {
+  [index: string]: string | number;
```
mmmmmmaya: why did you add this to each type?

src/frontend/package.json
```diff
@@ -33,7 +33,7 @@
     "@typescript-eslint/parser": "^4.5.0",
     "babel-eslint": "^10.1.0",
     "classnames": "^2.2.6",
-    "czifui": "^1.2.1",
+    "czifui": "^1.4.0",
```
mmmmmmaya, line 36: did you consider pinning the latest version (1.5.0)?

src/frontend/src/views/Upload/components/Metadata/components/Table/components/Row/components/LocationField/index.tsx
```diff
@@ -1,75 +1,126 @@
...
-
-const SELECT_PROPS = {
-  displayEmpty: true,
-  renderValue,
-};
+import { StyledDiv } from "./style";
 
 interface Props {
   fieldKey: keyof Metadata;
+  accessoryKey: keyof Metadata;
```
mmmmmmaya: what does this prop represent, conceptually?

src/frontend/src/common/api/index.tsx
```diff
@@ -13,6 +13,7 @@ export enum API {
   GET_FASTA_URL = "/api/sequences/getfastaurl",
   USHER_TREE_OPTIONS = "/api/usher/tree_options",
   SAMPLES_VALIDATE_IDS = "/api/samples/validate-ids",
+  LOCATIONS = "/v2/locations/",
```
mmmmmmaya, line 16: I noticed this is the only endpoint with a trailing forward slash. Is that intentional?

mmmmmmaya:
As a general note, I wish this PR had been split up into at least 2-3 smaller PR. I had some trouble keeping all of the information about what was changed in my mind given the size of the changes here.

.happy/terraform/modules/sfn_config/gisaid.wdl
```diff
@@ -316,6 +325,39 @@ task ImportGISAID {
...
+    command <<<
+    set -Eeuo pipefail
+    aws configure set region ~{aws_region}
+
+    export ASPEN_CONFIG_SECRET_NAME=~{aspen_config_secret_name}
+    if [ "~{remote_dev_prefix}" != "" ]; then
+        export REMOTE_DEV_PREFIX="~{remote_dev_prefix}"
+    fi
+
+    export PYTHONUNBUFFERED=true
```
phoenixAja, line 359: small nit but i feel like setting this to `false` makes more sense since we want stdout and stderr streams to be unbuffered

danrlu:
I just tried this out, this is SO NICE TnT thank you guys!!

src/backend/aspen/workflows/import_locations/import_locations.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+# WDL inputs available through environmental vars:
+# AWS_REGION
+# ASPEN_CONFIG_SECRET_NAME
+# REMOTE_DEV_PREFIX (if set)
+# WORKFLOW_ID
+# S3_FILESTEM
```
phoenixAja: can we remove this?

src/backend/aspen/workflows/import_locations/save.py
```diff
@@ -0,0 +1,36 @@
...
+    if test:
+        print("Success!")
+        return
+
+    interface: SqlAlchemyInterface = init_db(get_db_uri(Config()))
+    with session_scope(interface) as session:
+        # Insert all locations from gisaid_metadata
+        insert_rows_from_gisaid = "INSERT INTO aspen.locations (region, country, division, location) SELECT DISTINCT region, country, division, location FROM aspen.gisaid_metadata ON CONFLICT (region, country, division, location) DO NOTHING"
+        # Make sure we have equivalent rows for region/country/division but with null locations
+        insert_rows_with_null_locations = "INSERT INTO aspen.locations (region, country, division) SELECT DISTINCT region, country, division FROM aspen.locations EXCEPT SELECT DISTINCT region, country, division FROM aspen.locations WHERE location IS NULL"
```
phoenixAja, line 26: why did you decide to use raw sql here instead of sqlalchemy api? just curious

src/backend/database_migrations/versions/20211117_213611_add_locations_table.py
```diff
@@ -0,0 +1,71 @@
...
+    # X    just the list of existing sample locations from gisaid
+    # 2a. Update the gisaid import job to keep this list up to date in the db.
+    # 4. Update Orange County and Rickettsial lab samples to point to a real location
+    # 5. Fix the test samples for the admin group
+    # 6. Update all the existing sample rows in the DB to match a real location
+    # 7. Create an API endpoint to return a list of locations
+    # 8. Update upload form to be searchable textbox for location api endpoint
+    # 9. Update upload form (or upload backend endpoint????) to find the appropriate location id to associate with a sample
+    # --- AFTER the above is deployed ---
+    # 10. remove sample, division, location fields from samples table, make sure the backend is 100% using associated location info.
```
phoenixAja: i think we can also remove this? I do enjoy seeing your process though lol

src/frontend/src/common/types/location.d.ts
lvreynoso: I did this because I was accessing object properties via string variables, e.g. [this function here](https://github.com/chanzuckerberg/aspen/blob/f144494f4e9b88f46052c3a15fa3f1125160c464/src/frontend/src/views/Upload/components/Metadata/index.tsx#L44-L59)

src/frontend/package.json
lvreynoso: I am happy to pin the latest version, `1.4.0` was just the latest when I changed it.

src/frontend/src/views/Upload/components/Metadata/components/Table/components/Row/components/LocationField/index.tsx
lvreynoso: That...I don't think we need any more. I'll fix it.

src/frontend/src/common/api/index.tsx
lvreynoso: Yes, our FastAPI setup is particular about having a trailing slash in the url.

.happy/terraform/modules/sfn_config/gisaid.wdl
lvreynoso: If we want the streams to be unbuffered, then we should keep this `true`. (The var is PYTHON + **UN**BUFFERED).

src/backend/aspen/workflows/import_locations/save.py
lvreynoso: @jgadling do you want to take this? I know when we first reasoned about how to do this we were thinking in SQL syntax

lvreynoso:
@mmmmmmaya I agree, we really should have split up this PR. Apologies for that, I'll make sure in the future PRs are more limited in scope.

.happy/terraform/modules/sfn_config/gisaid.wdl
phoenixAja: oh whoops i read that as `PYTHONBUFFERED`, my mistake

src/backend/aspen/database/models/locations.py
```diff
@@ -0,0 +1,15 @@
+from sqlalchemy import Column, String
+
+from aspen.database.models.base import idbase
+from aspen.database.models.mixins import DictMixin
+
+
+class Location(idbase, DictMixin):  # type: ignore
```
phoenixAja, line 7: do we need `DictMixin` if we're using the serializers now?

src/backend/database_migrations/versions/20211117_213611_add_locations_table.py
```diff
@@ -0,0 +1,58 @@
+"""Add Locations Table
+
+Create Date: 2021-11-17 21:36:11.102820
+
+"""
+import enumtables  # noqa: F401
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
```
phoenixAja, line 10:
```suggestion
```

phoenixAja (approved):
🤩 🤩 🤩 so awesome!! Great work @lvreynoso

src/backend/aspen/workflows/import_locations/save.py
jgadling: Know what? I just assumed SQLAlchemy didn't have a construct for `insert ... select` or `on conflict` but it turns out that it does:
https://docs.sqlalchemy.org/en/14/core/dml.html?highlight=from_select#sqlalchemy.sql.expression.Insert.from_select
https://docs.sqlalchemy.org/en/14/dialects/sqlite.html?highlight=on_conflict#sqlalchemy.dialects.sqlite.Insert.on_conflict_do_nothing

I even see an except clause:
https://docs.sqlalchemy.org/en/14/core/selectable.html?highlight=except#sqlalchemy.sql.expression.except_

I'll see about translating these queries to SA code in the morning.

src/backend/aspen/app/views/sample.py
```diff
@@ -447,6 +445,23 @@ def create_sample():
...
+                g.db_session.query(Location)
+                .filter(Location.id == data["sample"].get("location_id", None))
+                .one_or_none()
+            )
+            if valid_location:
+                sample_args["location_id"] = valid_location.id
+                sample_args["region"] = RegionType(valid_location.region)
+                sample_args["country"] = valid_location.country
+                sample_args["division"] = valid_location.division
+                sample_args["location"] = valid_location.location or ""
```
vincent-czi, line 458: There's currently a bug -- well, it could be intended, but I don't think it is -- regarding what happens when a user uploads a non-location-specific "location" for a sample's metadata. This is a screenshot of a Samples entry in table from the rdev when the user selects `North America/USA/California` as the Collection Location during its metadata upload:
image
The `---` is where the Current Location should be. I believe this line is the specific cause, but not the root issue. The cause is because, when submitting `North America/USA/California`, that has no `location` aspect (it's `division` level only), so when the Samples table pulls that sample, there is an empty string for the `location`, so that gets parsed as false-y, which triggers the table cell renderer in FE to show it as `---`. Previously, submitting "California" as the collection location actually set the samples table `location` column to `California`. But now since we're truly matching the correct GISAID location, the lowest level `location` has no value for a general division. We need some kind of fallback for cases like this.

More generally though, I think there's a root issue of conflicting sources of truth regarding location. While the `samples` table still has `region`, `country`, `division`, and `location`, the addition of now having a FK `location_id` means we can have conflict between the sample's location info and the canonical location info provided by looking it up in the `locations` table. I don't have an easy answer for how to fix this -- the migration for legacy data could be hairy -- but I think we probably don't want two ways to look up the location.

I think we should either:
1. Not have a `location_id` for the samples. While the available options for location data (when user is submitting the metadata via dropdown, TSV, etc) do come from the `samples` table, they are just used as a way to populate the `samples` table. **Pro:** Easily backwards compatible with our legacy approach; **Con:** If the GISAID canonical names for locations drift, this could cause us pain. Also adds a bit more data storage overhead since we're duplicating values across many samples that could be consolidated.
2. Not have any of `region`, `country`, `division`, and `location` in the samples table, only have a `location_id` that is a FK to pulling the canonical values from that table. **Pro:** Normalized data, keeps us using canonical terms from GISAID; **Con:** Lots more code updates required, probably a reasonably painful DB migration to move old sample locations to new format, we need to watch GISAID locations for drift -- I've only skimmed the PR, don't really know how we ingest the GISAID locations, but if we're using their IDs for locations, we have to make sure that never changes under our feet or it could screw up all our locations so far.

I think (2) is the more "right" way to do it, but it's definitely more of a hassle. Also ties us more heavily to GISAID than we might want long-term if we're ever needing to provide a different set of locations for a non-COVID pathogen in the future. But as it stands, I'm not really sure what should happen if there's ever a conflict between the location info in a sample row and the `locations` row referenced by `location_id`.

src/frontend/src/views/Upload/components/Metadata/components/Table/components/Row/components/LocationField/index.tsx
```diff
@@ -1,75 +1,126 @@
...
+    dispatch({ searching: true });
+
+    const regex = new RegExp(escapeRegExp(query), "i");
+    const filteredLocationOptions = locationOptions.filter((option) =>
+      regex.test(option.name)
+    );
+    // alphabetical sort
+    // this ensures partial locations (i.e. region, country and divison
+    // but no location) end up on top.
+    const sortedLocationOptions = filteredLocationOptions.sort((a, b) => {
```
mmmmmmaya, line 68: I think this is the default sort behavior (ie, you can just say filteredLocationOptions.sort() instead of defining the whole thing here).

Janeece:
Wow this is awesome! 

Here's design feedback:

- [x] **Should Have** - Can we use the dropdown use the InputDropdown style dropdown instead of the minimal dropdown?
Screen Shot 2021-11-29 at 2 49 36 PM

- [x] **Should Have** - When reopening the menu from a selected location, can we show the selected location in the menu? 
Screen Shot 2021-11-29 at 3 25 00 PM

- [ ] **Nice to Have** - It feels like we need a max width/design for content overflow for this input. I can think about that later in the week but this is not a blocker. 


- Given the feedback above, i thought a design prototype of how we want this to look would be helpful: https://www.figma.com/proto/pH5rRGCzcMkCKkn6UMQVUR/Upload?node-id=1265%3A62978&scaling=min-zoom&page-id=1265%3A61639
- **Question:** how does this info show up on the sample table? Just county or the full location name?

src/backend/aspen/app/views/sample.py
jgadling: @vincent-czi just chiming in here -- #2 is where we're going, but we still have a fair bit of code that depends on the country/division/location/etc fields in the samples table, and this PR is already massive.

Our YAML builder project will require us to touch most of the code that works with sample locations, so the plan is to deprecate these old fields as part of that project.

src/frontend/src/common/types/location.d.ts
lvreynoso: Ended up removing this.

src/frontend/src/views/Upload/components/Metadata/components/Table/components/Row/index.tsx
```diff
@@ -27,7 +30,16 @@ const validationSchema = yup.object({
     .min(10, DATE_ERROR_MESSAGE)
     .max(10, DATE_ERROR_MESSAGE)
     .required("Required"),
-  collectionLocation: yup.string().required("Required"),
+  collectionLocation: yup
+    .object({
+      id: yup.number().required(),
+      // region: yup.string(),
```
mmmmmmaya, line 36: why are these commented out? are they required?

src/backend/aspen/app/views/sample.py
vincent-czi: Okay, as long as there's a clear direction and it will be happening in the next couple months I think that's fine.

I do think we probably want some kind of helper func that can take either a Locations record or a dict of region, country, division, and location, then just return the first existing field of the most specific kind. That way we can solve for the "California" bug above.

src/frontend/src/views/Upload/components/Metadata/index.tsx
```diff
@@ -30,14 +33,67 @@ import Table from "./components/Table";
...
+  });
+  const candidateLocation = scoredLocations.reduce(
+    ([prevLocation, prevScore], [currLocation, currScore]) => {
+      if (currScore < prevScore) {
+        return [currLocation, currScore];
+      }
+      return [prevLocation, prevScore];
+    }
+  );
+  console.log(`Found match for ${locationString}:`);
```
mmmmmmaya, line 92: these logs should come out before we commit

src/frontend/src/views/Upload/components/Metadata/components/Table/components/Row/components/LocationField/index.tsx
```diff
@@ -1,75 +1,119 @@
...
+  };
+
+  return (
+    <StyledDiv onBlur={handleBlur}>
+      <Dropdown
+        label={value?.name || "Search For Location"}
+        value={value}
+        onChange={handleLocationChange}
+        options={state.results}
+        search={true}
```
mmmmmmaya, line 104:
```suggestion
        search
```

in React, `propName={true}` is equivalent to `propName` with no value specified

src/frontend/src/views/Upload/components/Metadata/components/Table/components/Row/components/LocationField/index.tsx
```diff
@@ -1,75 +1,119 @@
...
-            value={value}
-            handleClick={applyToAllColumn}
-          />
-        ))
+  const [state, dispatch] = useReducer(locationSearchReducer, {
+    results: [],
+    searching: false,
+  });
+
+  const searcher = async (query: string): Promise<void> => {
```
mmmmmmaya, line 54: why did you decide to manage the search state here instead of relying on the built-in search behavior in the Dropdown?

lvreynoso:
Closing to split this into 2 PRs. Will continue to work on feedback.

PR #817, opened 2021-11-26
Update ncov pin to label the new variant B.1.1.529/omicron

danrlu:
### Summary:
- **What:** Ncov made updates to label/color the newly discovered variant, so we're updating our ncov version. The situation is a bit fluid as we learn more about Omicron so there is continuous update on ncov side, we'll release the pin for now to keep things up to date.

### Demos: 
Testing tree run with ncov `dc8c7d0` released on Friday, there is 1 more update since then which shouldn't break things (the samples are biased, don't read anything into it other than the run worked and the new variant is properly labeled.)
image

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/Dockerfile.gisaid
+++ b/src/backend/Dockerfile.gisaid
@@ -57,7 +57,7 @@ RUN mkdir /ncov && \
     cd /ncov && \
     git init && \
     git remote add origin https://github.com/nextstrain/ncov.git && \
-    git fetch origin 6691370325e5980d4f6073f7672759711c305a05 && \
+    git fetch origin dc8c7d03774c1ed99dc8a2509494bb5dc9f05bbf && \
     git reset --hard FETCH_HEAD
 
 # Poetry: install app
--- a/src/backend/Dockerfile.gisaid
+++ b/src/backend/Dockerfile.gisaid
@@ -57,7 +57,7 @@ RUN mkdir /ncov && \
     cd /ncov && \
     git init && \
     git remote add origin https://github.com/nextstrain/ncov.git && \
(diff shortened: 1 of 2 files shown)


lvreynoso (commented):
I'm glad the tree run worked with the latest ncov version! Should we use the latest commit hash though (`dceb6bcd9e09209530a30ac78fe3cf04dee56f93`)? Or did you mean to unpin ncov versions and always use the latest `master` branch?

danrlu:
Sorry for the confusion! There were updates on Friday, which version I tested, then more updates on Saturday, which is the current master. I'm expecting more changes will continue to come in as we get more data about the new variants. So if we want to keep things up to date we can remove the pin and always use master. See what you think.

lvreynoso (approved):
I trust your judgement @danrlu!

src/backend/Dockerfile.gisaid
```diff
@@ -57,7 +57,7 @@ RUN mkdir /ncov && \
     cd /ncov && \
     git init && \
     git remote add origin https://github.com/nextstrain/ncov.git && \
-    git fetch origin 6691370325e5980d4f6073f7672759711c305a05 && \
+    git fetch origin master && \
```
jgadling, line 60: Can we still pin to a specific github rev, or do we just want to accept any risks of being pinned to latest so we can be more responsive to omicron?

jgadling (approved):

src/backend/Dockerfile.gisaid
jgadling: Nevermind, I see Dan's earlier comments now. We can pin to master for awhile!

danrlu: I'd love to see what you guys think. Punting this to discuss in product meeting

PR #820, opened 2021-11-30
CZ Gen Epi

jgadling:
### Summary:
- **What:** Rename most resources to "genepi" instead of aspen
- **Env:** `<rdev link>`

### Notes:
This PR changes the names of as many infra resources as possible from 'aspen' to 'genepi'. 

There are two resources that I don't think we can easily change right now -- our RDS database, and the S3 bucket that stores our trees. We'll have to work out a migration plan for those down the road.

We've launched a totally separate set of infra for these new environments - new ECS clusters, new batch clusters, some new S3 buckets, etc. These changes hook our app up to the new infrastructure. I had to use new environment names ("rdev" -> "dev", "staging" -> "gestaging", "prod" -> "geprod") in order to prevent conflicts with our existing infrastructure.

This PR currently *does not* enable scheduled jobs for these new environments because that would mean we'd have duplicate jobs running and writing to the same db! Once we're ready to tear down the old infrastructure, we'll have to change these conditionals in our TF code to match our new environment names:
https://github.com/chanzuckerberg/aspen/blob/trunk/.happy/terraform/modules/ecs-stack/main.tf#L173

### Checklist:
- [ ] I added labels to my PR


src/backend/aspen/api/settings.py
```diff
@@ -78,7 +78,7 @@ class Settings(BaseSettings):
     DB_POOL_SIZE: int = 5
     DB_MAX_OVERFLOW: int = 0
     DB_ECHO: bool = False
-    DEBUG: bool = True
+    DEBUG: bool = False
```
jgadling, line 81: Don't print tracebacks back to users 😨

phoenixAja (approved):
looks good!

PR #821, opened 2021-11-30
Add GISAID location support on backend

lvreynoso:
### Summary:
- **What:** Import GISAID locations into new locations table; give samples a location_id that gives it a relationship to a Location;
- **Env:** `https://locationb-backend.dev.genepi.czi.technology/`

### Demos:
Screen Shot 2021-11-30 at 14 32 10

### Notes:
The create samples API is backwards-compatible with the current location string we send in requests.

### Checklist:
- [x] I added labels to my PR


lvreynoso:
re: @vincent-czi 's comments:
You are right in that how we handle locations in the samples table will probably have to change in the future. I added code in the v2 list samples api endpoint to return the division if there is currently no location for a sample. (On that note, we are still using the old list samples endpoint - I'll add a PR to make a switch). For now this should ensure consistent behavior; this is also better than what's currently in prod, we have some unfortunate data in the samples table:
Screen Shot 2021-11-30 at 15 47 01

In the future we will probably want to return a location dict with each sample.

lvreynoso:
re @phoenixAja 's comments:
I translated the raw sql to SQLAlchemy code in the import locations script. I removed the TODO comments and the `DictMixin` class.

lvreynoso:
Currently running a test job in aws state machine!

src/backend/aspen/api/schemas/samples.py
```diff
@@ -61,6 +61,11 @@ def get(self, key: Any, default: Any = None) -> Any:
                 if self._obj.show_private_identifier
                 else None
             ),
+            "collection_location": (
+                self._obj.location
+                if self._obj.location != "" and self._obj.location != "NaN"
```
phoenixAja, line 66: will it be these strings or an actual null object? Should we also add a check for that or use .get() to avoid key errors?

jgadling: I think this refers to the `location` string field in the DB, which is `not null` ... though I'm definitely a little worried that we have `NaN` values in that field 😨

src/backend/aspen/app/views/sample.py
```diff
@@ -447,6 +445,23 @@ def create_sample():
                     user.group.name,
                 ]
 
+            valid_location: Optional[Location] = (
+                g.db_session.query(Location)
+                .filter(Location.id == data["sample"].get("location_id", None))
+                .one_or_none()
+            )
```
jgadling: NIT -- if they send us a location_id, it really is expected to be valid. Let's raise an exception if we can't find one?

src/backend/aspen/api/schemas/samples.py
```diff
@@ -61,6 +61,11 @@ def get(self, key: Any, default: Any = None) -> Any:
                 if self._obj.show_private_identifier
                 else None
             ),
+            "collection_location": (
```
jgadling, line 64: what if `collection_location` is an object like so:

```
"collection_location": {"region": "xxx", "country": "yyy", "division": "zzz", "location": null}
```

and we'll populate those fields with data from the locations table if we have a relationship to it, and we'll use the char fields otherwise?

src/backend/aspen/database/models/sample.py
```diff
@@ -183,6 +184,11 @@ class Sample(idbase, DictMixin):  # type: ignore
     )
 
     # location
+    location_id = Column(
```
jgadling, line 187: Can we set up `collection_location` as a relationship with the `locations` table so we don't wind up conflicting with the current `location` char field?

src/backend/aspen/api/schemas/samples.py
lvreynoso: Yeah as Jess said it's a non-nullable field.

src/backend/aspen/api/schemas/samples.py
lvreynoso: Well, the frontend wouldn't know how to handle it. It's a simple change, so I think that is best added as part of a future PR that updates how the samples table deals with locations.

src/backend/aspen/database/models/sample.py
lvreynoso: Yeah, I can do that

src/backend/aspen/api/schemas/samples.py
lvreynoso: I have no idea how `NaN` values got in there but they all seem to be CZB samples from earlier this year.

src/backend/aspen/api/schemas/samples.py
jgadling: ok, fair enough

src/backend/aspen/app/views/sample.py
```diff
@@ -447,6 +446,33 @@ def create_sample():
...
+                    valid_location = (
+                        g.db_session.query(Location)
+                        .filter(Location.id == location_id)
+                        .one()
+                    )
+                except NoResultFound:
+                    sentry_sdk.capture_message(
+                        f"No valid location for id {location_id}"
+                    )
+                    raise ex.ServerException("Invalid location id for sample")
```
vincent-czi: Minor, but I think this actually would be a bad request, not the server's fault? I mean, it certainly _could_ be the server's fault if the previous API call to get location options went screwy, but this feels more like it should be "hey, you sent us a bad location_id, that's on you!" This feels like a 400 category, not a 500 category.

src/backend/aspen/app/views/sample.py
vincent-czi: Long-term, when location_id becomes the primary way to talk about location (rather than current `location`), location_id will go from being optional to being required, right? I assume we'd also rip out accepting text `location` at all then, right?

lvreynoso: That's the idea!

vincent-czi:
@lvreynoso 
> In the future we will probably want to return a location dict with each sample.

Long-term, I'd be okay with the backend deciding what the frontend sees. Maybe we'd want to expose all that data to the frontend, but I think it's also totally okay for the backend to just fiat the `collectionLocation` name based on its internal logic. I can see a philosophical argument for sending the data to the FE and then it being the FE's responsibility to determine display, but it also doesn't seem crazy for the location-naming to be an abstraction that the BE takes care of on behalf of the FE. [Also, a different approach entirely: we could fetch entire `locations` set from BE when FE loads (we'll need it anyway if they ever upload), then just have each sample only contain its `location_id` which we use to cross-reference against `locations` to get the actual location info.]

What I'm saying is, I think it's six-in-one, half-dozen in the other whether we determine the `currentLocation`'s user-visible string on the FE or the BE. My big concern is just making sure we (eventually) have a clear, single source of location truth for each sample.

jgadling (approved):

PR #825, opened 2021-12-01
Maya/173806 uploader name

mmmmmmaya:
### Summary
- **What:** Add sample uploader name to samples table
- **Why:** Clarity around origin of data
- **Ticket:** [[sc-173806]](https://app.shortcut.com/genepi/story/173806)
- **Env:** https://sampleuploader-frontend.dev.genepi.czi.technology/data/samples
- **Design:** https://www.figma.com/proto/QuQtuWj8iU7nqI6AREd0RE/CRUD-V0?node-id=197%3A74360&scaling=min-zoom&page-id=88%3A64261&starting-point-node-id=111%3A80433

### Testing
1. View samples table
1. Ensure user name appear underneath the sample ID in grey text

### Demos
Screen Shot 2021-11-30 at 10 29 38 PM

### Notes
TODO: Add code for CZ Biohub label.

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/api/index.tsx
+++ b/src/frontend/src/common/api/index.tsx
@@ -3,7 +3,7 @@ import { jsonToType } from "src/common/utils";
 
 export enum API {
   USER_INFO = "/api/usergroup",
-  SAMPLES = "/api/samples",
+  SAMPLES = "/v2/samples",
   LOG_IN = "/login",
   LOG_OUT = "/logout",
   PHYLO_TREES = "/api/phylo_trees",
@@ -99,7 +99,9 @@ const SAMPLE_MAP = new Map<string, keyof Sample>([
   ["private_identifier", "privateId"],
   ["public_identifier", "publicId"],
   ["upload_date", "uploadDate"],
+  ["uploaded_by", "uploadedBy"],
   ["sequencing_date", "sequencingDate"],
+  ["submitting_group", "submittingGroup"],
   ["czb_failed_genome_recovery", "CZBFailedGenomeRecovery"],
 ]);
 
--- a/src/frontend/src/common/api/index.tsx
+++ b/src/frontend/src/common/api/index.tsx
@@ -3,7 +3,7 @@ import { jsonToType } from "src/common/utils";
 
 export enum API {
   USER_INFO = "/api/usergroup",
-  SAMPLES = "/v2/samples",
+  SAMPLES = "/v2/samples/",
   LOG_IN = "/login",
   LOG_OUT = "/logout",
   PHYLO_TREES = "/api/phylo_trees",
--- a/src/frontend/src/common/components/library/data_table/index.tsx
+++ b/src/frontend/src/common/components/library/data_table/index.tsx
@@ -60,8 +60,29 @@ function sortData(
   ascending: boolean
 ): TableItem[] {
   return data.sort((a, b): number => {
-    const order = defaultSorting(a, b, sortKey);
-    return ascending ? order : order * -1;
+    let order = 0;
+    // Keep failed samples at the bottom of the table
+    if (sortKey[0] === "uploadDate") {
+      const uploadDateAIsNull = a["uploadDate"] === null;
+      const uploadDateBIsNull = b["uploadDate"] === null;
+      if (uploadDateAIsNull && uploadDateBIsNull) {
+        order = 0;
+      } else if (uploadDateAIsNull && !uploadDateBIsNull) {
+        order = -1;
+      } else if (uploadDateBIsNull && !uploadDateAIsNull) {
+        order = 1;
+      } else {
+        order = defaultSorting(a, b, sortKey);
+      }
+    } else {
+      order = defaultSorting(a, b, sortKey);
+    }
+
+    if (!ascending) {
+      return order * -1;
+    } else {
+      return order;
+    }
   });
 }
 
--- a/src/frontend/src/common/components/library/data_table/index.tsx
+++ b/src/frontend/src/common/components/library/data_table/index.tsx
@@ -33,7 +33,7 @@ const ITEM_HEIGHT_PX = 60;
 
 const LOADING_STATE_ROW_COUNT = 10;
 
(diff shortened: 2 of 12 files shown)


src/frontend/src/common/components/library/data_table/index.tsx
```diff
@@ -61,14 +61,15 @@ function sortData(
 ): TableItem[] {
   return data.sort((a, b): number => {
     let order = 0;
+    // Keep failed samples at the bottom of the table
```
vincent-czi: I'm a bit confused by this comment. Could you expand on it a bit? Is it the case that if a sample was failed, its upload date in the table is `null` and those are the only samples with `null` upload dates, so we can use null-ness in date as a proxy for being a failed sample? I think that's gotta be what's going on, but it's a bit of a non-intuitive jump to understand. If I'm right, could you add another sentence that explains as much?

If I'm wrong, uh, what is going on then?

src/frontend/src/common/components/library/data_table/index.tsx
```diff
@@ -33,7 +33,7 @@ const ITEM_HEIGHT_PX = 60;
 
 const LOADING_STATE_ROW_COUNT = 10;
 
-const UNDEFINED_TEXT = "---";
+const UNDEFINED_TEXT = "-";
```
vincent-czi: Don't want to pile extra work on you, but would you mind taking a look at the specialty `sequencingDate` renderer (`src/frontend/src/views/Data/cellRenderers.tsx`, line 118 on my local)? I previously made that to A) handle the weirdness of missing date being sent as `"N/A"` and B) to use a (at the time) special no-value string of `"-"`. But I think all that is pointless now and it can just become a `defaultCellRenderer`?

(B) is definitely handled if we're generally converting to a false-y value being shown with `-`. I'm not sure about (A) though -- is the backend now sending missing dates as `null` instead of `"N/A"`? If so, then yeah, `sequencingDate` renderer is no longer special and should be culled. If you don't have the time to take care of it with this PR, that's fine, just let me know and I'll make a note to fix it in the next week. Thanks!

src/frontend/src/common/utils/timeUtils.tsx
```diff
@@ -0,0 +1,9 @@
+export const formatTZDate = (d: string): string => {
+  if (!d || d.startsWith("-")) return d;
```
vincent-czi: I'm assuming the `startsWith("-")` is related to the `UNDEFINED_TEXT` elsewhere, right? This feels kinda magic-value to me, not really clear why `-` should short-circuit the function if you don't know the convention of null being turned into `-`. I'd like to either see a comment that says, "the hyphen is because we use that for undefined/null values" or import `UNDEFINED_TEXT` from somewhere and use that var here (I think it would be clear at that point what's going on).

src/frontend/src/common/utils/timeUtils.tsx
```diff
@@ -0,0 +1,9 @@
+export const formatTZDate = (d: string): string => {
```
vincent-czi: I think this util should have a line or two of documentation to say what it's doing. If you just know the func name, I don't think you'd guess what it actually does (I would assume it formats a date object as a string with TZ info, probably).

If I'm reading it correctly, I think this is an accurate docstring if you want to use it:
> Takes ISO8601 datetime with TZ and returns corresponding date (only) on user's machine.
> Eg, if user is in Pacific TZ, formatTZDate("2021-12-02T01:07:48+00:00") => "2021-12-01"

Also, might be worth it to tweak the name to something a bit more explicit. Trying not to be overly verbose, but maybe `datetimeWithTzToLocalDate`?

src/frontend/src/common/api/index.tsx
```diff
@@ -99,7 +99,9 @@ const SAMPLE_MAP = new Map<string, keyof Sample>([
   ["private_identifier", "privateId"],
   ["public_identifier", "publicId"],
   ["upload_date", "uploadDate"],
+  ["uploaded_by", "uploadedBy"],
   ["sequencing_date", "sequencingDate"],
+  ["submitting_group", "submittingGroup"],
```
vincent-czi, line 104: To be clear, we now get this info from the API, but we still don't actually make use of it anywhere yet, right?

vincent-czi (approved):
Looks good logic-wise. All my comments are... comment-related. There's a couple little things that I don't think would be clear if you aren't already decently familiar with how other parts of the codebase work.

I also put in one special request comment for you to check something that really is more my responsibility than yours since I made the component, but if you have an extra moment to check it out, that's appreciated. No sweat if not though, I can look next week.

src/frontend/src/common/api/index.tsx
mmmmmmaya: correct. This definition basically says "for the keys listed in this map, covert the key from the left side (the key the BE sends) to the right side (the camel-cased equivalent, to keep things in line with JS standard)". Anything not shown in the map will still be a part of the sample objects returned from the backend. But when you print the objects or look at them in dev tools, any keys not mapped here will use the snake_case version that was sent from the server.

src/frontend/src/views/Data/cellRenderers.tsx
```diff
@@ -121,23 +121,8 @@ const SAMPLE_CUSTOM_RENDERERS: Record<string | number, CellRenderer> = {
     );
   },
 
-  // Preferably, we would just use defaultCellRenderer for this column, but we want
-  // to intercept its value and change it out in some cases before rendering.
-  sequencingDate: ({ header, value }): JSX.Element => {
```
vincent-czi, line 118: Thank you!!

vincent-czi (approved):
wooooo

Janeece:
@mmmmmmaya Text styling, placement and top margin looks good! My only concern i when a user has very long name, it seems like it will overlap into other rows, “making column unreadable. 

Can overflow on user names cut off with “…” so that it never wraps to 2 lines? *Ex: “Janeece Pour…”*

PR #827, opened 2021-12-01
only color users' counties

danrlu:
### Summary:
- **What:** With current setting, the coloring on the trees are not optimized for each user group, and as a result, for example, Orange's samples on Orange's tree often get assigned a grey color. 
image

- We could go back to use Nextstrain default, which assigns color based on prevalence. So the group with most samples gets the first color (dark blue), the 2nd most gets the 2nd color (slightly lighter dark blue), which is still not ideal.
image

- Before we come up with a way to automatically optimize this for each user group, for now let's only color counties our users come from to make things most distinguishable.

- **Ticket:** [sc<167081>](https://app.shortcut.com/genepi/story/167081/customize-colors-for-each-county)
- **Env:** `<rdev link>`

### Demos:
Assume Chicago and Cook County are 2 user groups:
image

### Notes:
- I threw in counties surrounding Chicago for experimenting to see how much it helps or not help. Shouldn't impact CA builds at all.
- Coloring based on `division` or `clade` etc won't be impacted and will still work as Nextstrain default

### Checklist:
- [x] I added labels to my PR


src/backend/aspen/workflows/nextstrain_run/nextstrain_profile/ordering.tsv
```diff
@@ -1,1485 +1,23 @@
...
-recency	One week ago
-recency	3-7 days ago
-recency	1-2 days ago
-recency	New
+location	Cook County IL
+location	DuPage County
+location	Kane County IL
+location	Lake County IL
+location	McHenry County
+location	Will County IL
```
jgadling, line 23: Is an extra blank line going to cause problems?

jgadling (approved):

src/backend/aspen/workflows/nextstrain_run/nextstrain_profile/ordering.tsv
danrlu: nope, I did a test run with this branch and all is fine :D

PR #833, opened 2021-12-03
Use new locations in upload flow

lvreynoso:
### Summary:
- **What:** Changes location drop down to one that searches for a location based on user input. Sends a location_id with create_sample API request.
- **Env:** [https://locationd-frontend.dev.genepi.czi.technology/data/samples](https://locationd-frontend.dev.genepi.czi.technology/data/samples)

### Demos:

https://user-images.githubusercontent.com/24234461/144658231-d7a00c53-65fd-4687-8b4e-d918e8d5de39.mov

### Notes:

* The concept of `Metadata` and `ParsedMetadata` types in the upload flow has been separated out more, to make it clear that the metadata parsed from a TSV is __not__ the same as the working metadata used by the app. This is necessary to translate a location string parsed from a TSV file to a `GisaidLocation` with an `id`.
* `NamedGisaidLocation` is simply a `GisaidLocation` that meets the requirements to be a `DefaultMenuSelectOption` by having a `name` property with a string value.
* Changed option filter sort behavior to use string `localeCompare()`.
* And why managed search state? Two reasons - the first is that importing locations from Gisaid means we have 21,750 of them as of writing this. Browsers grind to a halt trying to render all these options in the dropdown, so we want to pass in a limited selection of options to the dropdown based on user input. The other reason is that we want to keep the previous selected value, if there is one, at the top.
* The create_sample query still sends a `location` in the POST request payload, but this will be ignored by the backend since we are also sending a `location_id`. Continued work on locations will clean this up.
* The dropdown menu is a bit more stable now, hooray! There are still minor positioning issues, but for the sake of time I think those will need to be addressed later.

### Checklist:
- [x] I added labels to my PR


src/frontend/src/common/queries/samples.ts
```diff
@@ -118,9 +119,13 @@ export async function createSamples({
       },
       sample: {
         [METADATA_KEYS_TO_API_KEYS.collectionDate]: collectionDate,
-        [METADATA_KEYS_TO_API_KEYS.collectionLocation]: collectionLocation,
+        [METADATA_KEYS_TO_API_KEYS.collectionLocation]: collectionLocation!.id,
```
mmmmmmaya, line 122: I don't recognize this notation. Does the `!` have special TS significance? Or is this meant to be a question mark?

mmmmmmaya, line 122: Is it guaranteed that every sample will have a collection location?

lvreynoso: Yes, we use `formik` to guarantee it and disable the continue button until all samples have one. It's required in the validation schema.

src/frontend/src/views/Upload/components/Metadata/index.tsx
```diff
@@ -47,17 +90,46 @@ export default function Metadata({
...
+          ...location,
+        };
+      }
+    );
+    setLocations(namedLocations);
+  };
 
+  useEffect(() => {
+    loadLocations();
+  }, [samples]);
```
mmmmmmaya: why is this rerun whenever samples change?

mmmmmmaya (approved):

vincent-czi:
The "Notes" section on this PR is a fantastic overview, thanks for writing up the summary, very helpful.

src/frontend/src/views/Upload/components/Metadata/index.tsx
vincent-czi: Yeah, it feels like we probably could just run this once every time the user sees the Metadata component. I think an empty array so it only ever gets done once per visit to the Metadata section would be totally reasonable. That said, once the user gets to Metadata, `samples` should effectively be static -- I'm like 95% sure it never changes after upload samples step 1 -- so this implicitly does that, but could be clearer that's the intention.

Janeece:
@lvreynoso Overall this is looking really great. thank you so much for updating the visual implementation! I've added my prioritized feedback to this ticket: https://app.shortcut.com/genepi/story/175880/design-review-feedback-location-normalization. 

Please let me know if you have any questions or concerns! If feedback is rooted in some issue with the SDS component, I can kick it back to them instead. Little note: the only "must have" is not entirely a blocker, but it is more important than any of the other refinements for UX clarity.

src/frontend/src/views/Upload/components/Metadata/index.tsx
lvreynoso: Fixed!

lvreynoso:
@Janeece I was able to fix the "Must Have", and I created a ticket for updating the other components to SDS components. As far as your other comments go, I'm kind of at a loss at how to deal with them. I tried styling the Dropdown component and did not really succeed, nor is it clear from looking at the sci-component source or docs how to style it to get the effects you want. I'll see if Maya or Vincent can take a try at tackling those.

PR #836, opened 2021-12-06
Danrlu/remove adjustment

danrlu:
### Summary:
- **What:** 
    -  During bug bash I noticed that our trees sometimes label countries or divisions as continent (`region` in Nextstrain). It turned out to be a default Nextstrain set up. 

     - This PR will label country with country names (same for divisions) by setting `region` as `global` in builds.yaml and bypass [this step](https://github.com/nextstrain/ncov/blob/4ccc5faf82c246b4eebca1dc015ee2582d3ac066/workflow/snakemake_rules/main_workflow.smk#L650-L674) of country/division name conversion. We could also remove the `region` row in builds.yaml. Keep it for now to be explicit. 
     
    - Added GISAID logo along the way, which is a convention for all applications using GISAID data.
    
### Demos:
Before (all current trees):
`

After:
`
### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/workflows/nextstrain_run/builds_templates/group_plus_context.yaml
+++ b/src/backend/aspen/workflows/nextstrain_run/builds_templates/group_plus_context.yaml
@@ -11,7 +11,7 @@ builds:
     country: USA
     division: {division}
     location: {location}
-    region: North America
+    region: global
     subsampling_scheme: group_plus_context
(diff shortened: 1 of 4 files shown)


jgadling (approved):

danrlu:
This fix was only live for 2 weeks before it got lost. All trees before and after have this bug. Too comical 🤣 🤣 🤣

PR #843, opened 2021-12-07
Maya/173844 refactor

mmmmmmaya:
### Summary
- **What:** 
  - Convert data from an array of sample/tree ids into an object who's keys are ids and whose values are the full sample/tree objects
  - Rename related instances of `checkedSamples` and `failedSamples` to `checkedSampleIds` and `failedSampleIds`, respectively
- **Why:** Deleting samples requires the `id` instead of `publicId` from a sample, so we need to pass down more info about the sample than just a public id. The two tables are more or less inseparable in terms of typing/data structuring at the moment, so both were modified to hold data in the new way.
- **Ticket:** [[sc-173844]](https://app.shortcut.com/genepi/story/173844)
- **Env:** https://[url]-frontend.dev.genepi.czi.technology

### Testing
Check various functionality of the tables: sorting, filtering, checking/unchecking checkboxes, making trees, downloading fasta files, everything

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
@@ -33,8 +33,8 @@ import {
 } from "./style";
 
 interface Props {
-  sampleIds: string[];
-  failedSamples: string[];
+  checkedSampleIds: string[];
+  failedSampleIds: string[];
   open: boolean;
   onClose: () => void;
   handleCreateTreeFailed: () => void;
@@ -48,8 +48,8 @@ const TreeTypes = {
 type TreeType = typeof TreeTypes[keyof typeof TreeTypes];
 
 export const CreateNSTreeModal = ({
-  sampleIds,
-  failedSamples,
+  checkedSampleIds,
+  failedSampleIds,
   open,
   onClose,
   handleCreateTreeFailed,
@@ -116,8 +116,8 @@ export const CreateNSTreeModal = ({
     </div>
   );
 
-  const allPossibleTreeSamples = sampleIds.concat(validatedInputSamples);
-  const allFailedOrMissingSamples = failedSamples.concat(missingInputSamples);
+  const allPossibleTreeSamples = checkedSampleIds.concat(validatedInputSamples);
+  const allFailedOrMissingSamples = failedSampleIds.concat(missingInputSamples);
   const allValidSamplesForTreeCreation = allPossibleTreeSamples.filter(
     (id) => !allFailedOrMissingSamples.includes(id)
   );
@@ -202,7 +202,7 @@ export const CreateNSTreeModal = ({
         />
         <Separator marginSize="xl" />
         <MissingSampleAlert missingSamples={missingInputSamples} />
-        <FailedSampleAlert numFailedSamples={failedSamples?.length} />
+        <FailedSampleAlert numFailedSamples={failedSampleIds?.length} />
       </StyledDialogContent>
       <StyledFooter>
         <CreateTreeButton
--- a/src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
@@ -27,8 +27,8 @@ import {
 } from "./style";
 
 interface Props {
-  sampleIds: string[];
-  failedSamples: any[];
+  checkedSampleIds: string[];
+  failedSampleIds: string[];
   tsvData: [string[], string[][]] | undefined;
   open: boolean;
   onClose: () => void;
@@ -36,8 +36,8 @@ interface Props {
 }
 
 const DownloadModal = ({
-  sampleIds,
-  failedSamples,
+  checkedSampleIds,
+  failedSampleIds,
   tsvData,
   open,
   onClose,
@@ -71,12 +71,12 @@ const DownloadModal = ({
   }, [tsvData]);
 
   useEffect(() => {
-    if (JSON.stringify(sampleIds) === JSON.stringify(failedSamples)) {
+    if (JSON.stringify(checkedSampleIds) === JSON.stringify(failedSampleIds)) {
       setFastaDisabled(true);
     } else {
       setFastaDisabled(false);
     }
-  }, [sampleIds, failedSamples, setFastaDisabled]);
+  }, [checkedSampleIds, failedSampleIds, setFastaDisabled]);
 
   const handleMetadataClick = function () {
     setMetadataSelected(!isMetadataSelected);
@@ -130,7 +130,7 @@ const DownloadModal = ({
         </StyledIconButton>
         <Header>Select Download</Header>
         <Title>
-          {sampleIds.length} {pluralize("Sample", sampleIds.length)} Selected
+          {checkedSampleIds.length} {pluralize("Sample", checkedSampleIds.length)} Selected
         </Title>
       </DialogTitle>
       <DialogContent>
@@ -184,12 +184,12 @@ const DownloadModal = ({
               </CheckBoxInfo>
             </CheckBoxWrapper>
           </Container>
-          {failedSamples.length > 0 &&
+          {failedSampleIds.length > 0 &&
             !isFastaDisabled && ( //ignore alert if fasta is already disabled
               <Alert severity="warning">
                 <DownloadType>
-                  {failedSamples.length}
-                  {pluralize("sample", failedSamples.length)}
+                  {failedSampleIds.length}
+                  {pluralize("sample", failedSampleIds.length)}
                   will not be included in your Consensus Genome download
                 </DownloadType>
                 <DownloadTypeInfo>
@@ -260,7 +260,7 @@ const DownloadModal = ({
             variant="contained"
             isRounded
             onClick={() => {
-              mutation.mutate({ sampleIds });
+              mutation.mutate({ sampleIds: checkedSampleIds });
             }}
             disabled={false}
           >
@@ -277,7 +277,7 @@ const DownloadModal = ({
           variant="contained"
           isRounded
           onClick={() => {
-            mutation.mutate({ sampleIds });
+            mutation.mutate({ sampleIds: checkedSampleIds });
           }}
           disabled={false}
         >
@@ -293,7 +293,7 @@ const DownloadModal = ({
           variant="contained"
           isRounded
           onClick={() => {
-            mutation.mutate({ sampleIds });
+            mutation.mutate({ sampleIds: checkedSampleIds });
           }}
           disabled={true}
         >
--- a/src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
@@ -130,7 +130,8 @@ const DownloadModal = ({
         </StyledIconButton>
         <Header>Select Download</Header>
         <Title>
-          {checkedSampleIds.length} {pluralize("Sample", checkedSampleIds.length)} Selected
+          {checkedSampleIds.length}{" "}
+          {pluralize("Sample", checkedSampleIds.length)} Selected
         </Title>
       </DialogTitle>
       <DialogContent>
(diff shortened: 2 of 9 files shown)


src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
```diff
@@ -116,8 +116,8 @@ export const CreateNSTreeModal = ({
     </div>
   );
 
-  const allPossibleTreeSamples = sampleIds.concat(validatedInputSamples);
-  const allFailedOrMissingSamples = failedSamples.concat(missingInputSamples);
+  const allPossibleTreeSamples = checkedSampleIds.concat(validatedInputSamples);
```
vincent-czi, line 119: I'm confused about this stuff -- not really this PR's business though -- I thought that `foundSamples` and `missingSamples` from the API fetch could be trusted to have no overlap. Am I wrong? It seems like `allPossibleTreeSamples` is just being used with a `uniq` call below to create `allSamplesRequestedTableAndInput` when squished together with `allFailedOrMissingSamples`. I guess we might not be able to assume the backend is keeping them non-overlapping, but it feels like it should if one group is "validated" and the other is "missing/failed".

vincent-czi: Also -- once again, not really this PR's business -- `allPossibleTreeSamples` seems like a confusing var name. "Possible" makes me think that it could maybe be not allowed due to some reason, too overly inclusive. Seems like it should maybe be more `allValidatedTreeSamples`?

src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
```diff
@@ -71,12 +71,12 @@ const DownloadModal = ({
   }, [tsvData]);
 
   useEffect(() => {
-    if (JSON.stringify(sampleIds) === JSON.stringify(failedSamples)) {
+    if (JSON.stringify(checkedSampleIds) === JSON.stringify(failedSampleIds)) {
```
vincent-czi: Feels like this should probably use some other way to check equality, I assume that was originally just a workaround since Arrays don't do item comparison for equality. Maybe swap out to `_.isEqual`? (Once again, not _actually_ this PR's business.)

src/frontend/src/common/components/library/data_subview/components/UsherTreeFlow/components/UsherPlacementModal/index.tsx
```diff
@@ -120,9 +123,11 @@ export const UsherPlacementModal = ({
 
   const handleSubmit = (evt: SyntheticEvent) => {
     evt.preventDefault();
-    sampleIds = sampleIds.filter((id) => !failedSamples.includes(id));
+    checkedSampleIds = checkedSampleIds.filter(
```
vincent-czi: I think this should be a new variable -- `sampleIdsToSubmit`, maybe? -- instead of re-assigning existing `checkedSampleIds` var. (Also, I don't think I had ever tried to re-assign an input arg during a function or React component, I just assumed they came in as `const` and would not let you re-assign, so I was wrong about that!)

...

Thinking about it some more, I guess that actually means there's a meaningful difference between
```
const myComponent = ({x, y, z}) => {
    blah
    blah
    blah
};
```
VS
```
const myComponent = (props) => {
    const { x, y, z } = props;
    blah
    blah
    blah
};
```
I hadn't realized that before, I thought they were truly identical!

src/frontend/src/common/components/library/data_subview/index.tsx
```diff
@@ -229,14 +235,14 @@ const DataSubview: FunctionComponent<Props> = ({
...
-      dispatch({ results: data });
+      dispatch({ results: dataValues });
       return;
     }
 
     dispatch({ searching: true });
 
     const regex = new RegExp(escapeRegExp(query), "i");
-    const filteredData = data.filter((item) => recursiveTest(item, regex));
+    const filteredData = filter(data, (item) => recursiveTest(item, regex));
```
vincent-czi, line 241: Why the switch over to using the lodash filter instead of the native Array filter?

src/frontend/src/common/components/library/data_subview/index.tsx
```diff
@@ -87,7 +87,7 @@ function tsvDataMap(
   if (tableData) {
     const filteredTableData = [...tableData];
```
vincent-czi: Once again, not this PR's business, but... I wonder why this is here? Seems like there's no point to do this spread? ... I was thinking it was maybe helping avoid something blowing up if tableData was undefined, but there's an `if (tableData)` just above it, so unclear. I think the line should go, and `filteredTableDataForReals` should just become `filteredTableData` and have it be `tableData.filter`.

src/frontend/src/common/components/library/data_subview/index.tsx
```diff
@@ -134,14 +134,15 @@ const DataSubview: FunctionComponent<Props> = ({
 }: Props) => {
   // we are modifying state using hooks, so we need a reducer
   const [state, dispatch] = useReducer(searchReducer, {
-    results: data,
+    results: Object.values(data ?? {}),
     searching: false,
   });
 
-  const [checkedSamples, setCheckedSamples] = useState<string[]>([]);
+  const [dataValues, setDataValues] = useState<Sample[] | Tree[]>();
```
vincent-czi: I don't understand the need for keeping a copy of the `data` that makes up the table in the component state? It doesn't seem like we ever modify this copy of the data -- search filtering creates a new filtered version and is also using OG `data`, not the state dataValues -- so it seems like it's just a duplicate of what we fetched? If there's a reason we need to have `setDataValues` and track an internal copy of the table data in component state, that reason is unclear to me and I think some comments explaining what's going on would be helpful.

While there's probably some small performance gain to avoiding the state holding an additional copy, I'm really just worried about possible confusion between OG data and state `dataValues` in later development. If we can avoid keeping a copy of the data in state and instead just have ways of manipulating the original `data` (eg, for searching) when it's necessary to display, that seems preferable to me just because it's clearer where the source of truth lives.

src/frontend/src/views/Data/index.tsx
```diff
@@ -25,11 +25,22 @@ const TITLE: Record<string, string> = {
   [ROUTES.PHYLO_TREES]: "Phylogenetic Trees",
 };
 
+const mapObjectArrayToIdDict = (
```
vincent-czi: I know the intent of this func is to map to ID, but it's not actually coded to inherently do that. It's generalized, which is probably a good thing, but `keyString` could be anything: it could be run on `uploadDate` or whatever, right? I think either the name of the func should change to indicate what it really does (although I'm struggling with a good, not insanely verbose name... it's like a thing that "reduces an array of objects to a mapping between the keyString arg and the objects that make up the array" ... hmmm, maybe `reduceObjectArrayToLookupDict`? then have the keyString arg become `keyStringForLookups` or something?) OR make it less generalized and have the `keyString` not be an arg, but rather make it `dataType` or something which is one of "sample" or "tree" or whatever, then the func conditionally chooses the `keyString` based on if it's on the sample track or the tree track.

If you do decide to keep it general, I think a one line comment explaining that it's converting an array of objects into a lookup of those same objects based on the provided key string would be helpful, it definitely took me a couple reads to figure out that was what I was seeing.

src/frontend/src/views/Data/index.tsx
```diff
@@ -95,19 +111,24 @@ const Data: FunctionComponent = () => {
...
+          );
+          transformedDatum[transform.key] = transform.method(methodInputs);
+        });
+
+        return [key, transformedDatum];
+      }
+    );
 
-    category.data = transformedData as BioinformaticsDataArray;
+    category.data = Object.fromEntries(transformedData);
```
vincent-czi: Seeing the `Object.fromEntries` here makes me kinda wish the whole id extraction and conversion to lookup dict process (`mapObjectArrayToIdDict`) was smushed together with the transform process. Seems like it would be reasonable to just iterate through the raw data once, doing appropriate transforms to create a prepared data object, then placing that in the lookup dict based on the rules of that data type. Definitely feels like it would be easier to reason about as a single abstracted step that can be pulled into its own configuration function in a separate file. Not really worth it now, but if I do go forward with the table refactor I think that's something I'll play with.

src/frontend/src/views/Data/index.tsx
```diff
@@ -95,19 +111,24 @@ const Data: FunctionComponent = () => {
...
-
-      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- Asserted above
-      category.transforms!.forEach((transform) => {
-        const methodInputs = transform.inputs.map((key) => datum[key]);
-        transformedDatum[transform.key] = transform.method(methodInputs);
-      });
-
-      return transformedDatum;
-    });
+    const transformedData = map(
```
vincent-czi: Just curious, what does the lodash map get us over the native Array `.map` here?

vincent-czi (approved):
Generally looks good, I wouldn't be bothered if it was merged in now, but I do care a bit about the comments I have regarding the need for `setDataValues` and the context of `mapObjectArrayToIdDict`. Some adjustment on those before merging would be good.

Lots of other comments in here, but those are all pretty much not really the business of this PR, I'm just commenting on pre-existing stuff I happen to be seeing. If you have the time and want to mess with them now, awesome, but it's fine to let the other ones go too.

src/frontend/src/common/components/library/data_subview/index.tsx
mmmmmmaya: the standard map only works for arrays, but native map doesn't exist for obj. lodash's accepts any collection

src/frontend/src/views/Data/index.tsx
mmmmmmaya: same as above

src/frontend/src/common/components/library/data_subview/components/UsherTreeFlow/components/UsherPlacementModal/index.tsx
mmmmmmaya: 💡

src/frontend/src/common/components/library/data_subview/index.tsx
mmmmmmaya: I was thinking that if data was particularly large, it might be inefficient to be constantly calling Object.values(data) over and over again, and better to use state as a way of not having to recalculate it.

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
mmmmmmaya: I _was_ the one who wrote this, and I remember I did it for a reason, but now it's escaping me. I would like to follow up with you on this and explain it to you/discuss better options, but this is feeling more mental overhead than the other unrelated changes so I want to save it for later.

mmmmmmaya:
Ugh sorry about the force push, things started getting dicey with merging the big rebrand changes and base branch change.

src/frontend/src/common/components/library/data_subview/index.tsx
vincent-czi: ah, gotcha. Maybe `useMemo` instead? Just seeing it placed into state makes me assume we're going to do manipulation on it, so I got worried about possible confusion when it seemed like it was existing more to shadow the data.

src/frontend/src/views/Data/index.tsx
```diff
@@ -149,8 +167,8 @@ const Data: FunctionComponent = () => {
   // * using the data, but LineageFilter renders a child compnent that seems
   // * to reference the parent's props (?). Passing in only the lineages, or
   // * incomplete options causes the component to break
-  const sampleArr = viewName === "Samples" ? (category.data as Sample[]) : [];
-  const lineages = uniq(compact(sampleArr?.map((d) => d.lineage?.lineage)))
+  const sampleArr = viewName === "Samples" ? (samples as SampleMap) : {};
```
vincent-czi: Should this get renamed since it's now a lookup-dict / map?

vincent-czi (approved):
Looks great, thanks for incorporating all the feedback!

PR #844, opened 2021-12-07
add a cache for sample data

mmmmmmaya:
### Summary
- **What:** Add a data cache for samples
- **Why:** So they automatically disappear from the table when they are deleted, without a refresh
- **Ticket:** [[sc-173844]](https://app.shortcut.com/genepi/story/173844)

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/frontend/src/common/queries/samples.ts
```diff
@@ -166,18 +207,27 @@ interface SampleDeleteRequestType {
...
+  onSuccess,
+}: SampleDeleteCallbacks): UseMutationResult<
   SampleDeleteResponseType,
   unknown,
   SampleDeleteRequestType,
   unknown
 > {
-  return useMutation(deleteSamples, callbacks);
+  const queryClient = useQueryClient();
+  // TODO (mlila): pick less confusing name choices for callbacks/params
```
vincent-czi, line 229: Maybe `additionalOnSuccess` to indicate we already have something going on when there's a success? I think `onError` is fine since it just gets passed directly. Also a comment somewhere around the `additionalOnSuccess` / `onSuccess` params def for `useDeleteSamples` to explain that we have one callback already baked in for successful calls to re-fetch samples and thus show what the result is post-deletion.

src/frontend/src/common/queries/samples.ts
```diff
@@ -138,8 +160,27 @@ export async function createSamples({
   throw Error(`${response.statusText}: ${await response.text()}`);
 }
 
+/**
+ * sample cache
```
vincent-czi, line 168: I think this should be more generic. It does act as a cache, but that's -- to the extent I understand react-query, I never really read its docs thoroughly -- what react-query does with all of its fetch results. I think the main thing to underline here is that it exists to get the samples for a given user

src/frontend/src/common/queries/samples.ts
```diff
@@ -138,8 +160,27 @@ export async function createSamples({
...
+ */
+
+export const USE_SAMPLE_INFO = {
+  entities: [ENTITIES.SAMPLE_INFO],
+  id: "sampleInfo",
+};
+
+export function useSampleInfo(): UseQueryResult<SampleResponse, unknown> {
+  return useQuery([USE_SAMPLE_INFO], fetchSamples, {
+    retry: false,
```
vincent-czi, line 178: I know we generally have `retry` as false, but I wonder if we really ought to have like `2` or `3` so we do retries. Although I think we would need to fix up the responses we have when requests get made by not authed users. Right now things blow up because we aren't sending back 40x's for stuff that shouldn't get accessed when not authed, so we'd probably need to fix that up first so we don't spam ourselves.

To be clear, this is just me discussing -- I feel like having a small retry number would be good, curious to get your opinion -- but this isn't something that should be done with this PR, certainly, just stay the course as it was before.

src/frontend/src/common/queries/samples.ts
```diff
@@ -138,8 +160,27 @@ export async function createSamples({
...
+  });
+}
+
 // * Proceed with caution, you are entering the DANGER ZONE!
```
vincent-czi, line 182: This comment makes me so happy

src/frontend/src/views/Data/index.tsx
```diff
@@ -37,25 +37,26 @@ const Data: FunctionComponent = () => {
...
+  const sampleResponse = useSampleInfo();
   const treeResponse = useTreeInfo();
-  const { data, isLoading } = treeResponse;
+  const { data: sampleData, isLoading: isSampleInfoLoading } = sampleResponse;
+  const { data: treeData, isLoading: isTreeInfoLoading } = treeResponse;
 
   useEffect(() => {
     const setBioinformaticsData = async () => {
       setIsDataLoading(true);
-      if (isLoading) return;
```
vincent-czi, line 80: Ah, this is so much better, thank you for refactoring this

src/frontend/src/views/Data/index.tsx
```diff
@@ -37,25 +37,26 @@ const Data: FunctionComponent = () => {
...
-      const sampleResponse = await fetchSamples();
+      if (isTreeInfoLoading || isSampleInfoLoading) return;
       setIsDataLoading(false);
 
-      const apiSamples = sampleResponse["samples"];
-      const apiTrees = data?.phylo_trees;
-
+      const apiSamples = sampleData?.samples ?? [];
       setSamples(apiSamples);
```
vincent-czi: Do we really need this state stuff for `setSamples` or `setTrees` anymore? It seems unnecessary, to have state keeping a copy of what react-query fetched, feels like we could probably just use react-query `apiSamples` and `apiTrees` directly below rather than intermediating it with React component state. (Of course, this is me effectively asking you to do _more_ refactoring, so I'm totally fine with you not putting even more on your plate.)

vincent-czi (approved):
Looks great, the related stuff feels way cleaner, nice work

src/frontend/src/common/queries/samples.ts
mmmmmmaya: I'm not sure what the payoff would be. Do we often have failed requests that would succeed with a retry? I think a few retries could be ok for edge cases, but I think it probably (?) happens infrequently enough that we don't need to invest a lot of time into it.

mmmmmmaya:
This needs to wait for #843 to merge first.

src/frontend/src/common/queries/samples.ts
vincent-czi: I have no idea what the failure rate is. And our userbase is probably prepared to hit refresh when things go wrong, so yeah, probably pretty low value.

The effortful part is changing how the backend responds to requests when auth is bad. I think that's actually an outstanding issue that should be fixed, regardless of what we're doing here with retries. But if it does get fixed, the effort here is just changing a single line for the retry config which seems acceptable effort even if it's probably extremely low value.

mmmmmmaya:
So, after combining the map and the transform into one step, I feel like I do want to keep the state vars for samples and trees, so we don't need to redo on that on every render.

vincent-czi:
@mmmmmmaya Could we do it with `useMemo` instead? The motivation here is that the `transformData` call is reasonably heavy, not that we're going to need to modify the `samples` or `trees` state down the road, right? If that's the case, I think useMemo that fires on the same set of conditions (`[isTreeInfoLoading, isSampleInfoLoading, sampleData, treeData]`) will hit that goal of avoiding waste while keeping it a bit clearer that we do not expect `samples` or `trees` to be modified due to some later event or something.

PR #847, opened 2021-12-07
Update our automation to use new czgenepi infra.

jgadling:
### Summary:
- **What:** Update GitHub actions and terraform to use new staging/prod stack names (gestagingstack / geprodstack) instead of old ones.
- **Env:** `<rdev link>`

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/.github/workflows/deploy-happy-stack.yml
+++ b/.github/workflows/deploy-happy-stack.yml
@@ -44,7 +44,7 @@ jobs:
         env:
           TFE_TOKEN: ${{ secrets.TFE_TOKEN }}
         run: |
-          ./scripts/happy --profile="" --env ${{ github.event.deployment.environment }} update ${{ github.event.deployment.environment }}stack --tag ${{ github.event.deployment.payload.tag }}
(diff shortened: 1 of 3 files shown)


vincent-czi (approved):
Looks good!

Is the plan for it to remain `geprod` and `gestaging` forever, or are we going to eventually migrate back to the old name space of `prod` and `staging` once we tear down the old infra? I'm just thinking about since we tend to speak verbally about "prod" and "staging", we might accidentally get confused and target the old namespace if it's always `ge_____` from here on out.

jgadling:
> Looks good!
> 
> Is the plan for it to remain `geprod` and `gestaging` forever, or are we going to eventually migrate back to the old name space of `prod` and `staging` once we tear down the old infra? I'm just thinking about since we tend to speak verbally about "prod" and "staging", we might accidentally get confused and target the old namespace if it's always `ge_____` from here on out.

I don't expect to change these names back anytime soon - it basically requires standing *yet another* set of parallel infra and redeploying. I think we'll have to do this from time to time for other types of risky infra changes, so we can take one of those opportunities to change the name back. In hindsight, I probably should have versioned these names (`staging2`?) instead of adding the `ge` prefix.

PR #852, opened 2021-12-08
Fix import_location script so it is idempotent

lvreynoso:
* Import all distinct region, country, division, location combinations from gisaid_metadata where location is not an empty string
* Separately, insert rows with null location for every distinct region, country, division combination
* Use Python set math to avoid massive join

### Summary:
- **What:** This PR fixes two major issues: one, a syntax error in the `save.py` script meant that the compiled SQL statement did not properly check for null values, so multiple `Location` entries with the same `region`/`country`/`division` and a `null` in the `location` column were created. Two, many entries in the Gisaid metadata table had an empty string (`""`) in the `location` column instead of a null value; these were directly imported and duplicated the above created entries in the locations table. Instead, we are only directly importing locations from Gisaid metadata that do not have an entry string in the location column; and creating an entry for every distinct region/country/division present (including the entries with empty strings) with a null location, checking for existing entries using Python's set logic.
- **Ticket:** [sc176244](https://app.shortcut.com/genepi/story/176244)
- **Env:** `<rdev link>`

### Demos:

https://user-images.githubusercontent.com/24234461/145464416-abde795a-f737-4c0b-98d9-cac8aa941f9e.mov

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/workflows/import_locations/save.py
+++ b/src/backend/aspen/workflows/import_locations/save.py
@@ -45,26 +45,24 @@ def save():
             .where(Location.location == None)
             .distinct()
         )
-        existing_null_locations = session.execute(existing_null_location_select).all()
+        existing_null_locations = set(
+            session.execute(existing_null_location_select).all()
+        )
 
         gisaid_null_locations_select = sa.select(
             GisaidMetadata.region,
             GisaidMetadata.country,
             GisaidMetadata.division,
         ).distinct()
-        gisaid_null_locations = session.execute(gisaid_null_locations_select).all()
+        gisaid_null_locations = set(session.execute(gisaid_null_locations_select).all())
 
-        new_null_locations = {
-            region_country_division_tuple
-            for region_country_division_tuple in gisaid_null_locations
-            if region_country_division_tuple not in existing_null_locations
-        }
+        new_null_locations = gisaid_null_locations - existing_null_locations
         new_null_location_values = list(
             map(
-                lambda location: {
-                    "region": location[0],
-                    "country": location[1],
-                    "division": location[2],
+                lambda region_country_division_tuple: {
+                    "region": region_country_division_tuple[0],
+                    "country": region_country_division_tuple[1],
+                    "division": region_country_division_tuple[2],
                     "location": None,
                 },
                 new_null_locations,
--- a/src/backend/aspen/workflows/import_locations/save.py
+++ b/src/backend/aspen/workflows/import_locations/save.py
@@ -40,9 +40,11 @@ def save():
         session.execute(gisaid_locations_insert)
 
         # Insert an entry with a null location for every distinct Region/Country/Division combination
+        # Doing a double-equals comparison is critical for the statement to compile to the intended SQL,
+        # so we tell flake8 to ignore rule E711
         existing_null_location_select = (
             sa.select(Location.region, Location.country, Location.division)
-            .where(Location.location == None)
+            .where(Location.location == None)  # noqa: E711
             .distinct()
         )
         existing_null_locations = set(
--- a/src/backend/aspen/workflows/import_locations/save.py
+++ b/src/backend/aspen/workflows/import_locations/save.py
@@ -1,6 +1,12 @@
+# flake8: noqa: E711
+# Doing a double-equals comparison to None is critical for the statements
+# that use it to compile to the intended SQL, which is why tell flake8 to
+# ignore rule E711 at the top of this file
+
 import click
 import sqlalchemy as sa
 from sqlalchemy.dialects import postgresql
+from sqlalchemy.sql.expression import and_
 
 from aspen.config.config import Config
 from aspen.database.connection import (
@@ -25,7 +31,7 @@ def save():
                 GisaidMetadata.division,
                 GisaidMetadata.location,
             )
-            .where(GisaidMetadata.location != "")
+            .where(and_(GisaidMetadata.location != "", GisaidMetadata.location != None))
             .distinct()
         )
         gisaid_locations_insert = (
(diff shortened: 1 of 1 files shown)


src/backend/aspen/workflows/import_locations/save.py
```diff
@@ -35,21 +39,42 @@ def save():
         )
         session.execute(gisaid_locations_insert)
 
-        # Make sure we have equivalent rows for region/country/division but with null locations
-        except_select = (
+        # Insert an entry with a null location for every distinct Region/Country/Division combination
+        # Doing a double-equals comparison is critical for the statement to compile to the intended SQL,
```
vincent-czi: This is a wonderful comment, thank you

src/backend/aspen/workflows/import_locations/save.py
```diff
@@ -18,12 +18,16 @@ def save():
...
-            GisaidMetadata.location,
-        ).distinct()
+        gisaid_locations_select = (
+            sa.select(
+                GisaidMetadata.region,
+                GisaidMetadata.country,
+                GisaidMetadata.division,
+                GisaidMetadata.location,
+            )
+            .where(GisaidMetadata.location != "")
```
vincent-czi: My SQL is weak, so I really don't know -- do we have to worry about this still selecting on rows where location is NULL? Even if there are no NULLs in location right now, we might want to add an additional clause to protect against it just in case Gisaid changes how they represent nothingness later

src/backend/aspen/workflows/import_locations/save.py
```diff
@@ -35,21 +39,42 @@ def save():
         )
         session.execute(gisaid_locations_insert)
 
-        # Make sure we have equivalent rows for region/country/division but with null locations
-        except_select = (
+        # Insert an entry with a null location for every distinct Region/Country/Division combination
+        # Doing a double-equals comparison is critical for the statement to compile to the intended SQL,
+        # so we tell flake8 to ignore rule E711
+        existing_null_location_select = (
```
vincent-czi, line 49: So generally, region > country > division > location. This select is choosing all the places where we are missing the most specific piece of info, but what happens for places where we're missing `location` _and_ `division` (or for that matter, `country` as well)? I can't tell from the SQL if it would return pre-existing NULL spots for more general places (eg, if there was a row of `Africa | Kenya | NULL |  NULL`): would those be selected with this statement?

Zooming out, we want to create entries for every place that are more general than the `location` level. Every distinct Region/Country/Division combo gets a generalized entry where its Location is NULL. Do we want to do that at even broader levels as well? Should there be entries for every distinct Region/Country combo where Division and Location are both NULL also? Or is the broadest scope we want to allow for the level of unknown Location? (Revisiting example above, would we want to have a general "Kenya" that's just Africa/Kenya/NULL/NULL? And similarly, would we even want a general "Africa" that's only Africa/NULL/NULL/NULL?) I honestly don't know what the Product needs on this are, so it's quite possible null-Location is broad enough, but I wanted to bring it up if it hasn't been discussed already.

src/backend/aspen/workflows/import_locations/save.py
```diff
@@ -35,21 +39,42 @@ def save():
...
-            sa.select(Location.region, Location.country, Location.division)
-            .distinct()
-            .except_(except_select)
+        existing_null_locations = set(
+            session.execute(existing_null_location_select).all()
         )
-        null_locations_insert = Location.__table__.insert().from_select(
-            ["region", "country", "division"], null_locations_select
+
+        gisaid_null_locations_select = sa.select(
```
vincent-czi, line 58: Piggy-backing on my above comment, if we want broader areas than null-Location, we'd need to make an additional select for each "zoomed out" area level.

vincent-czi (approved):
Looks good! The one comment I think to really pay attention to before merging is the long one I have about what level of "broadest area" we need to be supporting. Right now, it looks like we're only supporting places that are null-Location at broadest. Would we ever want to support null-Division-null-Location as well? What about only-Region?

vincent-czi:
Not related directly to this PR, but I've got a general question about the stability of the `id`s in locations table if we ever had to re-build it from scratch. Right now, because of the upsert process, we can safely handle new locations coming in via update to the GisaidMetadata. But if we had a giant meltdown and everything exploded, if we re-build the `locations` table from scratch, is it possible the location ids might jump around (or some older locations might now be entirely missing)? Because the way the initial creation and later upserts build up the table over time, I think it implies a requirement of history. If we try to build `locations` from scratch on a new GisaidMetadata, I think that might produce a slightly different dataset than what we'd get building it up iteratively over many old->new GisaidMetadata versions.

To be clear, this is not a big deal -- if we have a giant, exploding meltdown, we're going to be reloading from DB backups anyway, so locations will also get reloaded -- but it might mean that we have to be careful about building our `locations` table locally if we need it to exactly match what's in Staging/Prod.

src/backend/aspen/workflows/import_locations/save.py
lvreynoso: Yes, the broadest scope we want allow is division level. There are some entries like `Asia/Afghanistan/Afghanistan` but none where the division is null. We don't support that, and it's not in the gisaid metadata. And a sample just narrowed down to `Africa` is not useful at all!

lvreynoso:
Yep, that's what database backups are for! This pattern, creating a relationship between a parent table and child table with a `child_id` column is known as the Entity-Relationship Model and is a very common database design. And if a meltdown clobbered our backups we wouldn't have any samples to worry about either.

And I really can't fathom any reason the ids would have to match what's in Staging/Prod. The point is the relationship between the two, not the arbitrary number connecting them.

src/backend/aspen/workflows/import_locations/save.py
lvreynoso: Added!

PR #856, opened 2021-12-11
HOTFIX: handle brittle TSV upload bug

vincent-czi:
### Summary:
- **What:** Hotfix for brittle TSV upload issue

### Notes:
Changes to currentLocation with the new location dropdown approach causes the app to crash hard if user sends up a malformed TSV. Because currentLocation is now an object, an error that used to be a simple string comes back as an object. But we inject that error directly into the React, and React is cool with rendering strings and will eat your heart if you try to render an object.

### Checklist:
- [ ] I added labels to my PR


phoenixAja (approved):
sweet, thanks for the explanation as well

PR #860, opened 2021-12-13
Maya/173849 crud notifs

mmmmmmaya:
### Summary
- **What:** 
  - Add notifications for deleting samples
  - Move notifications into the components that actually control them, instead of in data subview
  - I thought I did this before in the cache PR, but make a more sensible name for callbacks on mutations
- **Why:** 
  - CRUD v0
  - To remove a lot of unnecessary state/prop passing between data subview and the components that generate notifications
  - To avoid developer confusion between a components callback and the cache's callback
- **Ticket:** [[sc-173849]](https://app.shortcut.com/genepi/story/173849)
- **Env:** https://notifs-frontend.dev.czgenepi.org [pending]
- **Design:** https://www.figma.com/file/QuQtuWj8iU7nqI6AREd0RE/CRUD-V0?node-id=88%3A64261

### Testing
1. Delete a sample and ensure the notification appears
1. Turn Network throttling in dev tools to `Offline` and try to delete again
2. Error notification should appear
3. Test out other notifications, such as creating a NS tree, creating an Usher tree, downloading offline, etc.

### Notes
Turning off whitespace changes will save you a lot of reading here.

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/frontend/src/common/components/library/data_subview/components/ContactUsLink/index.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import { StyledNewTabLink } from "./style";
+
+const ContactUsLink = (): JSX.Element => (
+  <span>
+    Please try again later or{" "}
+    <StyledNewTabLink href="mailto:aspenprivacy@chanzuckerberg.com">
```
vincent-czi: I feel like this should probably be the helloaspen email, not privacy. Looks like it's being used for pretty generic "let us know" type things where I see the component show up elsewhere.

src/frontend/src/common/components/library/data_subview/components/DeleteSamplesConfirmationModal/index.tsx
```diff
@@ -39,14 +40,13 @@ const DeleteSamplesConfirmationModal = ({
   });
 
   const onDelete = () => {
+    setNumDeletedSamples(samplesToDelete.length);
     deleteSampleMutation.mutate({
       samplesToDelete,
     });
```
vincent-czi, line 46: This is a bit unnecessary wishlist-y, but if we're running `setNumDeletedSamples` in the block where the deletion actually takes place, can we just scoop up the delete API response and use its length? ... Looking at backend code now ... For bulk deletion, that's easy. Looks like deleting a single one just returns the id as an int though, so there might be some special-casing between bulk and singles, not sure it's worth the hassle. Although wait, do we ever call the single deletion endpoint from FE, or do we always use bulk and just pass it a 1-length array?

vincent-czi (approved):
Overall looks good.

Only thing that kinda weirds me out is the `componentOnError` and `componentOnSuccess` stuff. At first glance, I thought they were params for react-query I was unaware of. Eventually I realized I was wrong. Here's my current understanding: both are intended to be callback funcs passed from the component using the react-query hook for what happens on error and success respectively. That way if we need to set state or whatever upon error/success, we have that happen via those callbacks. Is that correct?

Assuming I'm right, that seems reasonable. That said, I definitely found it kinda confusing at first. I'm not sure how to make it less confusing, but I think the names made me think it was somehow the React component that was in the error/success state, not the API call. I think the old names of `onError`/`onSuccess` were a bit clearer, but they did shadow the internal react-query names, so that's a different flavor of confusing. Maybe `onErrorCallback`? I dunno. I'll get used to the convention, but something about it just feels a little confusing to me at the moment.

src/frontend/src/common/components/library/data_subview/components/DeleteSamplesConfirmationModal/index.tsx
vincent-czi: To be clear, the reason why I think this is a good idea is in case some deletions don't go through, we have a way of knowing. Otherwise we could say we deleted 47, but only 46 deletions went through or something. That said, looking at current BE implementation, it appears that if any of the requested samples for deletion won't get deleted, the whole thing just 404s instead of deleting at all, so my worry can't actually come to pass currently.

mmmmmmaya: Currently we don't call single deletes for samples so special casing wouldn't be necessary.

PR #862, opened 2021-12-14
strip hCoV-19 from gisaid ids for tree build

danrlu:
**I don't have a set up to verify this change. I could only eyeball the code TnT**

### Summary:
- **What:** Current the GISAID ID box will allow IDs with "hCoV-19" during front end check, but doesn't strip "hCoV-19" in the backend for GISAID sequences. So add that stripping in to make front-back end consistent.

- **Env:** `<rdev link>`

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/workflows/nextstrain_run/export.py
+++ b/src/backend/aspen/workflows/nextstrain_run/export.py
@@ -196,8 +196,8 @@ def write_includes_file(session, gisaid_ids, pathogen_genomes, selected_fh):
         selected_fh.write(f"{public_identifier}\n")
         num_includes += 1
     for gisaid_id in gisaid_ids:
-        if gisaid_id.lower().startswith("hcov-19"):
-            gisaid_id = gisaid_id[8:]
+        # remove leading hcov-19/ preceding characters, ignore case
+        gisaid_id = re.sub(r'^hcov-19\/', "", gisaid_id, flags=re.I)
         selected_fh.write(f"{gisaid_id}\n")
         num_includes += 1
     return num_includes
--- a/src/backend/aspen/workflows/nextstrain_run/export.py
+++ b/src/backend/aspen/workflows/nextstrain_run/export.py
@@ -1,6 +1,7 @@
 import csv
 import io
 import json
+import re
 from typing import Any, Iterable, List, Mapping, MutableMapping, Set, Tuple
 
 import click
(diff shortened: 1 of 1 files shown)


src/backend/aspen/workflows/nextstrain_run/export.py
```diff
@@ -196,6 +196,8 @@ def write_includes_file(session, gisaid_ids, pathogen_genomes, selected_fh):
         selected_fh.write(f"{public_identifier}\n")
         num_includes += 1
     for gisaid_id in gisaid_ids:
+        if gisaid_id.lower().startswith("hcov-19"):
+            gisaid_id = gisaid_id[8:]
         selected_fh.write(f"{gisaid_id}\n")
```
phoenixAja: remember to import re if using suggestion
```suggestion
        # remove leading hcov-19/ preceding characters, ignore case
        gisaid_id = re.sub(r'^hcov-19\/', "", gisaid_id, flags=re.I)
        selected_fh.write(f"{gisaid_id}\n")
```

phoenixAja (approved):

src/backend/aspen/workflows/nextstrain_run/export.py
```diff
@@ -191,11 +192,12 @@ def write_includes_file(session, gisaid_ids, pathogen_genomes, selected_fh):
...
     for sample in sample_query:
         public_identifier = sample.public_identifier
-        if public_identifier.lower().startswith("hcov-19"):
-            public_identifier = public_identifier[8:]
+        # remove leading hcov-19/ preceding characters, ignore case
+        public_identifier = re.sub(r'^hcov-19\/', "", public_identifier, flags=re.I)
         selected_fh.write(f"{public_identifier}\n")
         num_includes += 1
     for gisaid_id in gisaid_ids:
+        gisaid_id = re.sub(r'^hcov-19\/', "", gisaid_id, flags=re.I)
```
jgadling: I think we may want to do this *TOO*, but it's better to filter out the 'hcov-19' when we accept the list of gisaid ID's from the frontend... I'll add that in to this PR.

danrlu: down the road we may want to consider consolidate where we strip 'hcov-19', b/c once we move over to GenBank, it will be 'SARS-CoV-2/human' all over the place...... 
image

PR #869, opened 2021-12-15
upgrade next

phoenixAja:
### Summary:
- **What:** upgrade Next  (In versions of Next.js prior to 12.0.5 or 11.1.3, invalid or malformed URLs could lead to a server crash. In order to be affected by this issue, the deployment must use Next.js versions above 11.1.0 and below 12.0.5)
- **Ticket:** [sc177040](https://app.shortcut.com/genepi/story/177040)
- **Env:** [rdev link](https://next-upgrade-frontend.dev.czgenepi.org/data/samples)

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/next-env.d.ts
+++ b/src/frontend/next-env.d.ts
@@ -1,5 +1,4 @@
 /// <reference types="next" />
-/// <reference types="next/types/global" />
 /// <reference types="next/image-types/global" />
 
 // NOTE: This file should not be edited
--- a/src/frontend/package-lock.json
+++ b/src/frontend/package-lock.json
@@ -5,34 +5,34 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.14.5",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.14.5.tgz",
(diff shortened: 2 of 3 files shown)


src/frontend/next-env.d.ts
```diff
@@ -1,5 +1,4 @@
 /// <reference types="next" />
-/// <reference types="next/types/global" />
```
vincent-czi, line 2: What's this about?

vincent-czi (approved):
Seems reasonable! My only concern would be if there turns out to be some kind of version issue, but as long as you were able to do some QA and it was clean of any weird behavior, seems good to me.

src/frontend/next-env.d.ts
phoenixAja: a little confusing to tell from the docs but this [file is regenerated when you run `next build`](https://github.com/vercel/next.js/issues/26533), i think it removed that line because that next type is no longer relevant to the typescript compiler with the upgrade

vincent-czi: I just needed to docker-compose down then up the containers -- also ran a DB migration, but I don't think that matters -- and this line re-appeared in my `next-env.d.ts` file.
```
/// <reference types="next/types/global" />
```

Maybe we should re-commit it back in? Probably has no impact on how it runs since we build during the CI pipeline, but if it keeps showing up as a dirty file locally that will just be confusing to see.

PR #870, opened 2021-12-15
Return group id and name with phylo tree

lvreynoso:
### Summary:
- **What:** Returns a `group` key with each phylo tree entry, an object with the group's id and name.
- **Ticket:** [sc177022](https://app.shortcut.com/genepi/story/177022)
- **Env:** `<rdev link>`

### Demos:
Screen Shot 2021-12-15 at 11 16 10

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/app/views/phylo_trees.py
+++ b/src/backend/aspen/app/views/phylo_trees.py
@@ -73,6 +73,7 @@ def phylo_trees():
         .options(
             joinedload(phylo_run_alias.outputs.of_type(PhyloTree)),
             joinedload(phylo_run_alias.user),
+            joinedload(phylo_run_alias.group),
         )
         .filter(
             or_(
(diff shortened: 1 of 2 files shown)


mmmmmmaya (approved):
Thank you!

src/backend/aspen/app/views/phylo_trees.py
```diff
@@ -95,6 +96,10 @@ def phylo_trees():
             "pathogen_genome_count": 0,  # TODO: do we still need this?,
             "tree_type": phylo_run.tree_type.value,
             "user": {},
+            "group": {
+                "id": phylo_run.group.id,
+                "name": phylo_run.group.name,
+            },
```
jgadling, line 102: Are we still calling the old phylo trees endpoint?

lvreynoso: Yup. We will have to pick back up the migration when we can

lvreynoso: Oh yeah... @mmmmmmaya the v2 endpoint returns this information already, but the information is formatted differently. Do you want to just use the new endpoint?

mmmmmmaya:
It will require non trivial changes to the app to use the new endpoint. I can make a ticket to use the v2 endpoint in the future, but in the interest of finishing crud v0 before end of year, I'd like to use the existing endpoint.

PR #875, opened 2021-12-17
Sample route location migration

lvreynoso:
### Summary:
- **What:** This PR associates all Samples with a Location and drops the old `region`, `country`, `division`, and `location` columns in the `samples` table. The list samples route now returns a location dict for `collectionLocation`. The sample table is updated to properly display the sample's location (or division, if applicable). Tests are updated - creating a sample with `sample_factory()` now requires at least three args: `sample_factory(group, user, location)`. A test helper function `location_factory()` has been added, but no default args are provided since each created location must be unique.
- **Ticket:** [sc175531](https://app.shortcut.com/genepi/story/175531), [sc175114](https://app.shortcut.com/genepi/story/175114), [sc175747](https://app.shortcut.com/genepi/story/175747)
- **Env:** [https://sloc-frontend.dev.czgenepi.org](https://sloc-frontend.dev.czgenepi.org)

### Checklist:
- [x] I added labels to my PR


src/backend/aspen/app/views/tests/test_samples_view.py
```diff
@@ -20,8 +22,24 @@ def test_samples_create_view_pass_no_public_id(
...
+        .filter(
+            and_(
+                Location.region == "North America",
+                Location.country == "USA",
+                Location.division == "California",
+                Location.location == "Santa Barbara County",
+            )
+        )
+        .one()
+    )
```
jgadling, line 42: Why create the location and then query it back out of the db? Is this so we can get its id?

src/backend/scripts/setup_localdata.py
```diff
@@ -63,7 +64,47 @@ def create_test_user(session, group, user_id, name):
...
+        session.query(Location)
+        .filter(
+            and_(
+                Location.region == region,
+                Location.country == country,
+                Location.division == division,
+                Location.location == location,
+            )
+        )
+        .one_or_none()
```
jgadling, line 102: same question as above -- does SA not automatically populate the location.id column when it writes to the db?

src/backend/aspen/workflows/nextstrain_run/export.py
```diff
@@ -290,13 +292,13 @@ def write_sequences_files(session, pathogen_genomes, sequences_fh, metadata_fh):
...
-            "region_exposure": sample.region.value,
-            "country_exposure": sample.country,
-            "division_exposure": sample.division,
+            "region": sample.collection_location.region,
+            "country": sample.collection_location.country,
+            "division": sample.collection_location.division,
+            "location": sample.collection_location.location,
+            "region_exposure": sample.collection_location.region,
+            "country_exposure": sample.collection_location.country,
+            "division_exposure": sample.collection_location.division,
```
jgadling, line 301: ❤️

src/backend/aspen/app/views/tests/test_samples_view.py
lvreynoso: Yep

src/backend/scripts/setup_localdata.py
lvreynoso: I tried several different approaches to debug this...and I'm mystified why a User gets an id upon creation but a Location only gets one when it's been committed.

src/backend/aspen/app/views/tests/test_samples_view.py
jgadling: gross. Is there some sort of SA flag we can twiddle to make it smarter?

src/backend/database_migrations/versions/20211222_115458_migrate_sample_locations_to_location_id.py
```diff
@@ -0,0 +1,37 @@
...
+
+    set_broader_locations_stmt = sa.sql.text(
+        "UPDATE aspen.samples as s SET location_id = (SELECT id FROM aspen.locations as l WHERE l.region = s.region AND l.country = s.country AND l.division = s.division and l.location is NULL) WHERE s.location IN ('California', 'NaN', '', NULL)"
+    )
+    conn.execute(set_broader_locations_stmt)
+
+    deprecated_columns = ["region", "country", "division", "location"]
+    with op.batch_alter_table("samples", schema="aspen") as batch_op:
+        for column in deprecated_columns:
+            batch_op.drop_column(column)
```
jgadling, line 33: How much validation have you done on the UPDATE queries above? I'm wondering if we should wait to actually drop these columns until we've validated that we have 0 samples with null locations?

src/backend/aspen/app/views/tests/test_samples_view.py
jgadling: weird thing -- I checked out your branch and deleted this query and the tests still pass! I think the session.commit() in there takes care of updating the id!

src/backend/scripts/setup_localdata.py
jgadling: And same here -- this seems to work ok for me without the additional query.

jgadling (approved):
This looks good to me except that I'd feel better if we held off on dropping the old sample location columns until we're sure the collection_location column is 100% filled!

src/backend/aspen/app/views/tests/test_samples_view.py
lvreynoso: weirdddd

lvreynoso:
I see. The old-columns are non-null, so I would have to undo some core parts of this PR to re-accommodate them. I think I'll just add another migration to fill in the location_id and merge it in before this one - I think that's an easier approach with the least amount of dragging this out.

src/backend/database_migrations/versions/20211222_115458_migrate_sample_locations_to_location_id.py
lvreynoso: Screen Shot 2022-01-03 at 09 58 22

PR #879, opened 2021-12-18
State/Country level builds

jgadling:
### Summary:
- **What:** Add support for state & country-level builds
- **Ticket:** [sc173906](https://app.shortcut.com/genepi/story/173906)

### Notes:
This PR makes a few important changes:
- Adds a `default_tree_location` to each group as a reference to the locations table.
- If the `divsision` and/or `location` fields for a group's `default_tree_location` are empty, update the nextstrain YAML to do a division- or country-centered build
- Removes all references to `phylo_runs.template_file_path` - we're currently figuring out which template to use based on the `tree_type` field
- Empties out the `phylo_runs.template_args` field for all new runs. We're not currently using it, but I *think* we might have some uses for it in the near future, if we allow users to pass additional config to tree builds.
- I'm leaving the old `group.division` and `group.location` fields where they are for the time being. Those fields are being used to generate some s3 filenames, and I think we'll learn more about how to work with groups & locations soon (maybe we'll want separate tree build locations and default sample locations) so we can refine that later.

TODO - I still need to update our export_test.py script to handle state/country runs.

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/.happy/terraform/modules/ecs-stack/county_runs.tf
+++ b/.happy/terraform/modules/ecs-stack/county_runs.tf
@@ -25,11 +25,7 @@ module nextstrain_chicago_contextual_sfn_config {
     remote_dev_prefix        = local.remote_dev_prefix
     group_name               = "Chicago Department of Public Health"
     s3_filestem              = "Chicago Contextual"
-    template_filename        = "group_plus_context.yaml"
-    template_args            = {
-      division = "Illinois"
-      location = "Chicago"
-    }
+    template_args            = {}
     tree_type                = "OVERVIEW"
   }
 }
@@ -55,11 +51,7 @@ module nextstrain_scc_contextual_sfn_config {
     remote_dev_prefix        = local.remote_dev_prefix
     group_name               = "Santa Clara County Public Health"
     s3_filestem              = "Santa Clara Contextual"
-    template_filename        = "group_plus_context.yaml"
-    template_args            = {
-      division = "California"
-      location = "Santa Clara County"
-    }
+    template_args            = {}
     tree_type                = "OVERVIEW"
   }
 }
@@ -85,11 +77,7 @@ module nextstrain_alameda_contextual_sfn_config {
     remote_dev_prefix        = local.remote_dev_prefix
     group_name               = "Alameda County Public Health Department"
     s3_filestem              = "Alameda Contextual"
-    template_filename        = "group_plus_context.yaml"
-    template_args            = {
-      division = "California"
-      location = "Alameda County"
-    }
+    template_args            = {}
     tree_type                = "OVERVIEW"
   }
 }
@@ -115,11 +103,7 @@ module nextstrain_contra_costa_contextual_sfn_config {
     remote_dev_prefix        = local.remote_dev_prefix
     group_name               = "Contra Costa County Public Health Laboratories"
     s3_filestem              = "Contra Costa Contextual"
-    template_filename        = "group_plus_context.yaml"
-    template_args            = {
-      division = "California"
-      location = "Contra Costa County"
-    }
+    template_args            = {}
     tree_type                = "OVERVIEW"
   }
 }
@@ -145,11 +129,7 @@ module nextstrain_fresno_contextual_sfn_config {
     remote_dev_prefix        = local.remote_dev_prefix
     group_name               = "Fresno County Public Health"
     s3_filestem              = "Fresno Contextual"
-    template_filename        = "group_plus_context.yaml"
-    template_args            = {
-      division = "California"
-      location = "Fresno County"
-    }
+    template_args            = {}
     tree_type                = "OVERVIEW"
   }
 }
@@ -175,11 +155,7 @@ module nextstrain_humboldt_contextual_sfn_config {
     remote_dev_prefix        = local.remote_dev_prefix
     group_name               = "Humboldt County Dept Human and Health Sevices-Public Health"
     s3_filestem              = "Humboldt Contextual"
-    template_filename        = "group_plus_context.yaml"
-    template_args            = {
-      division = "California"
-      location = "Humboldt County"
-    }
+    template_args            = {}
     tree_type                = "OVERVIEW"
   }
 }
@@ -205,11 +181,7 @@ module nextstrain_marin_contextual_sfn_config {
     remote_dev_prefix        = local.remote_dev_prefix
     group_name               = "Marin County Department of Health & Human Services"
     s3_filestem              = "Marin Contextual"
-    template_filename        = "group_plus_context.yaml"
-    template_args            = {
-      division = "California"
-      location = "Marin County"
-    }
+    template_args            = {}
     tree_type                = "OVERVIEW"
   }
 }
@@ -235,11 +207,7 @@ module nextstrain_monterey_contextual_sfn_config {
     remote_dev_prefix        = local.remote_dev_prefix
     group_name               = "Monterey County Health Department"
     s3_filestem              = "Monterey Contextual"
-    template_filename        = "group_plus_context.yaml"
-    template_args            = {
-      division = "California"
-      location = "Monterey County"
-    }
+    template_args            = {}
     tree_type                = "OVERVIEW"
   }
 }
@@ -265,11 +233,7 @@ module nextstrain_orange_contextual_sfn_config {
     remote_dev_prefix        = local.remote_dev_prefix
     group_name               = "Orange County Public Health Laboratory"
     s3_filestem              = "Orange Contextual"
-    template_filename        = "group_plus_context.yaml"
-    template_args            = {
-      division = "California"
-      location = "Orange County"
-    }
+    template_args            = {}
     tree_type                = "OVERVIEW"
   }
 }
@@ -295,11 +259,7 @@ module nextstrain_san_bernardino_contextual_sfn_config {
     remote_dev_prefix        = local.remote_dev_prefix
     group_name               = "San Bernardino County Public Health"
     s3_filestem              = "San Bernardino Contextual"
-    template_filename        = "group_plus_context.yaml"
-    template_args            = {
-      division = "California"
-      location = "San Bernardino County"
-    }
+    template_args            = {}
     tree_type                = "OVERVIEW"
   }
 }
@@ -325,11 +285,7 @@ module nextstrain_del_norte_contextual_sfn_config {
     remote_dev_prefix        = local.remote_dev_prefix
     group_name               = "Del Norte Public Health Laboratory"
     s3_filestem              = "Del Norte Contextual"
-    template_filename        = "group_plus_context.yaml"
-    template_args            = {
-      division = "California"
-      location = "Del Norte County"
-    }
+    template_args            = {}
     tree_type                = "OVERVIEW"
   }
 }
@@ -355,11 +311,7 @@ module nextstrain_san_joaquin_contextual_sfn_config {
     remote_dev_prefix        = local.remote_dev_prefix
     group_name               = "San Joaquin County Public Health Services"
     s3_filestem              = "San Joaquin Contextual"
-    template_filename        = "group_plus_context.yaml"
-    template_args            = {
-      division = "California"
-      location = "San Joaquin County"
-    }
+    template_args            = {}
     tree_type                = "OVERVIEW"
   }
 }
@@ -385,11 +337,7 @@ module nextstrain_san_luis_obispo_contextual_sfn_config {
     remote_dev_prefix        = local.remote_dev_prefix
     group_name               = "San Luis Obispo County Health Agency, Public Health Laboratories"
     s3_filestem              = "San Luis Obispo Contextual"
-    template_filename        = "group_plus_context.yaml"
-    template_args            = {
-      division = "California"
-      location = "San Luis Obispo County"
-    }
+    template_args            = {}
     tree_type                = "OVERVIEW"
   }
 }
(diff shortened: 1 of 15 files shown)


.happy/terraform/modules/ecs-stack/county_runs.tf
```diff
@@ -506,11 +442,7 @@ module nextstrain_ventura_contextual_sfn_config {
     remote_dev_prefix        = local.remote_dev_prefix
     group_name               = "Ventura County Public Health Laboratory"
     s3_filestem              = "Ventura Contextual"
-    template_filename        = "group_plus_context.yaml"
-    template_args            = {
-      division = "California"
-      location = "Ventura County"
-    }
+    template_args            = {}
```
danrlu, line 445: Thanks for cleaning these up! I've been thinking, we can prob remove some of the counties that never logged in (low priority, can wait till analytics is out).

.happy/terraform/modules/sfn_config/nextstrain-dbg.wdl
```diff
@@ -66,7 +63,6 @@ task nextstrain_workflow {
     workflow_id=$(aspen-cli db create-phylo-run                                                                                  \
                       --group-name "~{group_name}"                                                                               \
                       --all-group-sequences                                                                                      \
```
danrlu: This line `--all-group-sequences` can be deleted. It gave an error and w/o it, it worked fine.

src/backend/aspen/test_infra/models/usergroup.py
```diff
@@ -1,22 +1,31 @@
...
+    tree_loc = Location(
+        region="North America", country="USA", location=location, division=division
+    )
     return Group(
         name=name,
         address=address,
         prefix=prefix,
         location=location,
         division=division,
+        default_tree_location=tree_loc,
```
danrlu, line 28: The testing scripts always throw me off. I was like, why did we do this twice but somewhat differently?? LOL

src/backend/aspen/database/models/usergroup.py
```diff
@@ -29,6 +30,14 @@ class Group(idbase, DictMixin):  # type: ignore
...
     location = Column(String, nullable=True)
 
+    # Default location context (int'l or division or location level)
+    default_tree_location_id = Column(
+        Integer,
+        ForeignKey(Location.id),
+        nullable=True,
+    )
+    default_tree_location = relationship("Location")  # type: ignore
+
```
danrlu, line 40: So this associates a group with a default tree location, which is pulled from Lucia's location table and later sent into tree building. In the yaml builder below, we use `if not location.division:` to tell it's a country build, but I was searching through the location table (on Aspen upload page), it looks like all entries have a division. Are those locations w/o division simply not displayed for sample upload location dropdown?

src/backend/aspen/workflows/nextstrain_run/build_config.py
```diff
@@ -29,7 +29,9 @@ def update_subsampling(self, config, subsampling):
         if self.group.name == "Chicago Department of Public Health":
             subsampling["group"][
                 "query"
-            ] = "--query \"((location == '{location}') & (division == '{division}')) | submitting_lab == 'RIPHL at Rush University Medical Center'\""
+            ] = '''--query "((location == '{location}') & (division == '{division}')) | submitting_lab == 'RIPHL at Rush University Medical Center'"'''
```
danrlu, line 29: what magic does `'''` do here? 🤩

src/backend/aspen/workflows/nextstrain_run/builds_templates/mega_template.yaml
```diff
@@ -15,7 +15,7 @@ builds:
     division: {division}
     location: {location}
     subsampling_scheme: {tree_type}
-    title: '{tree_type} tree for samples collected in {location}, {division}, {country}'  # not sure how this works w country-level builds
+    title: '{tree_type} tree for samples collected in {location}'  # not sure how this works w country-level builds
```
danrlu: will yaml builder confuse this `{location}` with the `{location}` 2 lines above? Love this set up to populate the title regardless!

.happy/terraform/modules/sfn_config/nextstrain-dbg.wdl
jgadling: great catch, thanks!

src/backend/aspen/workflows/nextstrain_run/build_config.py
jgadling: Most of our subsampling queries have *both* literal `'` and `"` characters in them, and we *also* need a way to tell python where to start and stop reading the string. If we use " or ' to wrap the outside of the string, we need to escape (`\"` or `\'`) those characters inside the string. This is fine/normal, but it makes our strings look different than the ones we copy out of your YAML files. Using triple quotes (or triple double quotes) lets us use those characters more naturally in our script, they won't need to be escaped since Python is only looking for `'''` to mark the end fo the string.
https://towardsdatascience.com/single-double-and-triple-quotes-in-python-7ceea990baf

src/backend/aspen/workflows/nextstrain_run/builds_templates/mega_template.yaml
jgadling: I can see how this looks really confusing. The short story is that it won't because we expand *this* string very early in our yaml generation process, and leave essentially all other `{somevar}` strings alone. But I'll try to figure out how to make this clearer.

src/backend/aspen/database/models/usergroup.py
jgadling: You're right! We currently don't have any locations without a division. We have a couple options for that:
- Manually create locations without divisions for each group that needs country-level runs (easy enough)
- Automatically generate country-level locations when we import gisaid data the same way we generate division-level locations.

We can do either or both, and they're both pretty straightforward

danrlu: Route 1 with a lighter lift sounds good to me. I'm slightly unsure where we'd end up w LMIC users b/c their tree building may run into various other problems that  in the end turned out to be less suitable for Aspen.

src/backend/aspen/workflows/nextstrain_run/builder_base.py
```diff
@@ -29,28 +29,31 @@ def load_template(self):
...
-        build["title"] = build["title"].format(
+
+        # Update the tree's title with build type & location.
+        title_template = "{tree_type} tree for samples collected in {location}"
+        build["title"] = title_template.format(
             tree_type=self.subsampling_scheme.title(),
-            division=division,
-            location=location,
-            country=country,
+            location=", ".join(location_values),
```
danrlu, line 62: I think I understand this now (why even before the change this worked). here we're specifically replacing the `{location}` in the `build["title"]` section, before writing out the `builds.yaml` for `ncov`

jgadling: yeah, but I changed it so now YAML builder is completely overwriting the title, and we don't have the {location} template in mega_template.yaml at all.

src/backend/aspen/database/models/usergroup.py
phoenixAja: @danrlu what do you mean by "various other problems that in the end turned out to be less suitable for Aspen", can you explain a bit more?

src/backend/aspen/workflows/nextstrain_run/builder_base.py
```diff
@@ -29,28 +30,36 @@ def load_template(self):
...
-        # build["title"] = build["title"].format(division=division, location=location, country=country, region=region)
-        country = "USA"
-        division = self.group.division
-        location = self.group.location
-        build["country"] = country
-        build["division"] = division
-        build["location"] = location
+
+        location = self.group.default_tree_location
+        # Make a shortcut to decide whether this is a location vs divsision vs country level build
```
phoenixAja:
```suggestion
        # Make a shortcut to decide whether this is a location vs division vs country level build
```

src/backend/aspen/workflows/nextstrain_run/builds_templates/mega_template.yaml
```diff
@@ -15,7 +15,7 @@ builds:
     division: {division}
     location: {location}
     subsampling_scheme: {tree_type}
-    title: '{tree_type} tree for samples collected in {location}, {division}, {country}'  # not sure how this works w country-level builds
+    title: Aspen Tree  # This gets replaced with a more specific title in builder_base.py
```
phoenixAja: 😝 
```suggestion
    title: CZ Gen Epi Tree  # This gets replaced with a more specific title in builder_base.py
```

src/backend/aspen/workflows/nextstrain_run/export_test.py
```diff
@@ -25,14 +25,48 @@
     type=click.Choice(["overview", "targeted", "non_contextualized"]),
     required=True,
 )
-@click.option("--sequences", type=int, required=False, default=10)
-@click.option("--selected", type=int, required=False, default=10)
-@click.option("--gisaid", type=int, required=False, default=10)
+@click.option(
+    "--sequences",
```
phoenixAja, line 30: small nit but i think `included-sequences` might be a better flag name

phoenixAja: or something to refer more intuitively that this means a number of sequences to include, idk maybe i'm the only one confused

phoenixAja: actually it's fine, ignore me

src/backend/aspen/workflows/nextstrain_run/export_test.py
```diff
@@ -68,12 +102,24 @@ def cli(
             "num_sequences": num_sequences,
             "num_included_samples": num_included_samples,
         }
-        group = session.query(Group).first()
+        if group_name:
+            group = session.query(Group).filter(Group.name == group_name).first()
+        else:
+            group = session.query(Group).first()
+        if not group:
```
phoenixAja: you're not using sqlalchemy 2.0 style querying?

phoenixAja: also would one() make more sense than first()?

src/backend/aspen/workflows/nextstrain_run/export_test.py
```diff
@@ -68,12 +102,24 @@ def cli(
...
+            tree_location = Location(
+                region=region, country=country, division=div, location=loc
+            )
+            group.default_tree_location = tree_location
         builder = builder_factory(build_type, group, template_args, **context)
         builder.write_file(builds_file_fh)
 
         print("Wrote output files!")
         print(json.dumps(context))
+        session.rollback()  # Don't save any changes to the DB.
```
phoenixAja, line 123: why arn't we saving changes to the DB?

src/backend/database_migrations/versions/20211214_003923_add_location_to_groups.py
```diff
@@ -0,0 +1,52 @@
+"""Add Location to Groups
+
+Create Date: 2021-12-14 00:39:25.682465
+
+"""
+import enumtables  # noqa: F401
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
```
phoenixAja: remove comment
```suggestion
```

src/backend/database_migrations/versions/20211214_003923_add_location_to_groups.py
```diff
@@ -0,0 +1,52 @@
...
+update aspen.groups g set default_tree_location_id = (select id from aspen.locations l where l.location is null and l.division = 'California' and l.country = 'USA' and l.region = 'North America') where g.location is null or g.location = 'NaN'
+            """
+    )
+    conn.execute(non_null_locations)
+    conn.execute(null_locations)
+
+
+def downgrade():
+    # Don't downgrade.
+    pass
```
phoenixAja, line 51: small nit: should we raise a NotImplemented exception for this? I feel like that's more clear what's going on if downgrading the db is not allowed

src/backend/aspen/workflows/nextstrain_run/export_test.py
jgadling: `one()` will freak out if the query *could* return multiple rows. And you're right that I should convert the query style.

src/backend/aspen/workflows/nextstrain_run/export_test.py
jgadling: This script is just meant to be a quick and dirty way to generate lots of different nextstrain build.yaml files without too much setup, and internally it can associate groups with different locations than they normally use -- we don't want to write those re-associations to the db when we're just trying to generate some test output.

phoenixAja (approved):
👏 👏 👏

src/backend/aspen/workflows/nextstrain_run/builder_base.py
```diff
@@ -29,28 +30,36 @@ def load_template(self):
...
-        build["country"] = country
-        build["division"] = division
-        build["location"] = location
+
+        location = self.group.default_tree_location
+        # Make a shortcut to decide whether this is a location vs division vs country level build
+        if not location.division:
+            self.tree_build_level = "country"
+        elif not location.location:
+            self.tree_build_level = "division"
```
jgadling, line 39: @danrlu here we're stashing a tree build level as `location` or `division` or `country` so we can use that in `build_config.py`

src/backend/aspen/workflows/nextstrain_run/build_config.py
```diff
@@ -29,15 +26,33 @@ def update_subsampling(self, config, subsampling):
...
+
+        # Update our sampling for state/country level builds if necessary
+        update_subsampling_for_location(self.tree_build_level, subsampling)
```
jgadling, line 32: @danrlu headsup that this file has changed a bit since you last looked at it

danrlu:
I added minor comments in https://github.com/chanzuckerberg/aspen/pull/887, but even without it all looks good to me! thanks Jess!

danrlu (approved):
thanks Jess! all looks good!

PR #882, opened 2021-12-20
rename crowding penalty

danrlu:
### Summary:
- **What:** Rename the parameter section for crowding penalty per https://github.com/nextstrain/ncov/pull/828
- This should work w or w/o the upstream PR.
- I'm not sure how the [current patch](https://github.com/chanzuckerberg/aspen/blob/trunk/src/backend/Dockerfile.nextstrain#L46) will behave once the upstream PR merges. Will it error out or just skip and keep going? @jgadling If it will error, I'll pin the ncov until the upstream PR merges.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/Dockerfile.nextstrain
+++ b/src/backend/Dockerfile.nextstrain
@@ -43,7 +43,7 @@ RUN mkdir -p /ncov/auspice
 RUN mkdir -p /ncov/logs
 
 ADD aspen/workflows/nextstrain_run/patches/crowding_penalty.patch /tmp/
-RUN patch /ncov/workflow/snakemake_rules/main_workflow.smk < /tmp/crowding_penalty.patch
+RUN (grep 'config\["priorities"\]\["crowding_penalty"\]' /ncov/workflow/snakemake_rules/main_workflow.smk || patch /ncov/workflow/snakemake_rules/main_workflow.smk < /tmp/crowding_penalty.patch)
 
 WORKDIR /usr/src/app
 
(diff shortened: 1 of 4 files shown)


jgadling:
>     * I'm not sure how the [current patch](https://github.com/chanzuckerberg/aspen/blob/trunk/src/backend/Dockerfile.nextstrain#L46) will behave once the upstream PR merges. Will it error out or just skip and keep going? @jgadling If it will error, I'll pin the ncov until the upstream PR merges.

The current patch will error once the upstream commit is merged. I have a workaround I'm working on testing though!

src/backend/Dockerfile.nextstrain
```diff
@@ -43,7 +43,7 @@ RUN mkdir -p /ncov/auspice
 RUN mkdir -p /ncov/logs
 
 ADD aspen/workflows/nextstrain_run/patches/crowding_penalty.patch /tmp/
-RUN patch /ncov/workflow/snakemake_rules/main_workflow.smk < /tmp/crowding_penalty.patch
+RUN (grep 'config\["priorities"\]\["crowding_penalty"\]' /ncov/workflow/snakemake_rules/main_workflow.smk || patch /ncov/workflow/snakemake_rules/main_workflow.smk < /tmp/crowding_penalty.patch)
```
jgadling, line 46: @danrlu this should only apply the patch if necessary so we won't need to scramble when ncov is merged. Though we should remove this line entirely when we get a chance.

danrlu:
Per progress here https://github.com/nextstrain/ncov/pull/828 we're good to go! I tested this branch together with Sidney's branch.

jgadling (approved):

PR #886, opened 2021-12-21
Phylo Run/Tree update endpoint

phoenixAja:
### Summary:
- **What:** add endpoint to update PhyloTree and PhyloRun name
- **Ticket:** [sc171784](https://app.shortcut.com/genepi/story/171784)
- **Env:** no rdev

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/backend/aspen/api/views/phylo_trees.py
```diff
@@ -0,0 +1,51 @@
...
+from aspen.api.deps import get_db
+from aspen.api.error import http_exceptions as ex
+from aspen.api.schemas.phylo_trees import PhyloTreeRequest, PhyloTreeResponse
+from aspen.database.models import Group, PhyloRun, PhyloTree, User
+
+router = APIRouter()
+
+
+@router.put("/")
+async def update_phylo_tree(
```
jgadling: The rest of our CRUD endpoints for phylo trees are actually set up at the phylo *run* level -- to be able to list in-progress phylo trees, we decided the best way to do that was to get a list of phylo runs (since trees don't get created till the end of a run). And once we made that decision, it didn't make sense to delete a tree without deleting its associated phylo run either.

This update endpoint is at the *tree* level, which has some tradeoffs:
- once a phylo run is complete, I can see how we *might* want to consider the run to be immutable, while the tree metadata can be updated... so from that angle only allowing the trees to be updated makes sense
- but this is a different API interface than the rest of the UI has been counting on, so the UI will need to keep track of the phylo tree ID in our phylo runs response to be able to update the right tree.

There are probably more pros/cons than I've had a chance to think through, but I'm not sure what your take on this is.

src/backend/aspen/api/views/phylo_trees.py
```diff
@@ -0,0 +1,51 @@
...
+from aspen.api.auth import get_auth_user
+from aspen.api.deps import get_db
+from aspen.api.error import http_exceptions as ex
+from aspen.api.schemas.phylo_trees import PhyloTreeRequest, PhyloTreeResponse
+from aspen.database.models import Group, PhyloRun, PhyloTree, User
+
+router = APIRouter()
+
+
+@router.put("/")
```
jgadling: I think the tree ID needs to be part of the route here, rather than part of the request body.

src/backend/aspen/api/views/phylo_trees.py
```diff
@@ -0,0 +1,51 @@
...
+    phylo_run_alias = aliased(PhyloRun)
+
+    # use associated PhyloRun to check that user has permission to update PhyloTree
+    phylo_runs_query: Select = (
+        sa.select(phylo_run_alias)  # type: ignore
+        .options(
+            joinedload(phylo_run_alias.outputs.of_type(PhyloTree)),
+            joinedload(phylo_run_alias.group),
+        )
+        .filter((PhyloTree.id == phylo_tree_request.id))
```
jgadling: I don't think this filter is right - the query is mostly trying to fetch phylo_runs and associated info, but then we're adding a PhyloTree filter without telling SqlAlchemy how it's related to the other tables in our query.

src/backend/aspen/api/views/phylo_trees.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+    phylo_run_results = await db.execute(phylo_runs_query)
+    phylo_run: PhyloRun = phylo_run_results.unique().scalars().one()
+
+    group: Group = user.group
+    if phylo_run.group_id != group.id:
+        if not user.system_admin:
+            not_sufficient_permission_msg: str = f"User {user.name} from group {group.name} does not have permission to update tree name"
+            sentry_sdk.capture_message(not_sufficient_permission_msg, "error")
+            raise ex.BadRequestException(not_sufficient_permission_msg)
```
jgadling: Should this be an UnauthorizedException?

src/backend/aspen/api/views/phylo_trees.py
```diff
@@ -0,0 +1,51 @@
...
+
+    await db.execute(
+        sa.update(PhyloTree)  # type: ignore
+        .where(PhyloTree.id == phylo_tree_request.id)
+        .values(name=phylo_tree_request.name)
+    )
```
jgadling: I usually prefer using the ORM interface to update small numbers of rows, something more like:

```
tree_obj = phylo_run.outputs[0]
tree_obj.name = phylo_tree_request.name
await db.commit()
```

src/backend/aspen/api/views/phylo_trees.py
phoenixAja: makes sense, i'll change it

src/backend/aspen/api/views/phylo_trees.py
phoenixAja: hmmm interesting, it seems to be returning the correct phylo trees in my test but maybe something weird is going on. I'll adjust

src/backend/aspen/api/views/phylo_trees.py
phoenixAja: oh really? I thought that this was the correct way to update in [sqlalchemy 2.0]( https://docs.sqlalchemy.org/en/14/orm/session_basics.html#orm-expression-update-delete)

src/backend/aspen/api/views/phylo_trees.py
phoenixAja: yeah that's a better exception to throw here

src/backend/aspen/api/views/phylo_trees.py
phoenixAja: hmm i'll move this over to phyloruns since the phylotree/phylorun models are closely entwined at the moment. Should we allow users to update started PhyloRuns or would that mess up the tree building pipeline?

src/backend/aspen/api/views/phylo_trees.py
jgadling: That's the correct way to issue an explicit `UPDATE` query - which is a capability we definitely need if we want to update more than a handful of rows at once, but if we already have ORM objects loaded and aren't trying to update many rows at once, I think this is the section of the docs we want to look at:
 https://docs.sqlalchemy.org/en/14/tutorial/orm_data_manipulation.html#updating-orm-objects

phoenixAja:
converting to draft until i address feedback

src/backend/aspen/api/views/phylo_runs.py
```diff
@@ -269,3 +271,28 @@ async def delete_run(
...
+async def update_phylo_tree_and_run(
+    item_id: int,
+    phylo_run_update_request: PhyloRunUpdateRequest,
+    db: AsyncSession = Depends(get_db),
+    user: User = Depends(get_auth_user),
+):
+    # get phylo_run and check that user has permission to update PhyloRun/PhyloTree
+    phylo_run = await get_editable_phylo_run_by_id(db, item_id, user)
+
+    async with db as session:
```
jgadling: `db` is already an `AsyncSession` -- do we need this `with` ?

src/backend/aspen/api/views/phylo_runs.py
```diff
@@ -269,3 +271,28 @@ async def delete_run(
...
+
+        # if there are any associated PhyloTrees update those names as well:
+        if phylo_run.outputs:
+            # there should only be one phylotree associated with the phylorun
+            phylo_tree = phylo_run.outputs[0]
+            phylo_tree.name = phylo_run_update_request.name
+
+        await session.commit()
+
+    return PhyloRunUpdateResponse(id=phylo_run.id)
```
jgadling: It's more common to return the same response body for an `update` request as a `get` request, but if the UI won't use the extra data 🤷‍♀️ I suppose this is a nit.

src/backend/aspen/api/views/phylo_runs.py
```diff
@@ -269,3 +271,28 @@ async def delete_run(
...
+    phylo_run = await get_editable_phylo_run_by_id(db, item_id, user)
+
+    async with db as session:
+        # update phylorun name
+        phylo_run.name = phylo_run_update_request.name
+
+        # if there are any associated PhyloTrees update those names as well:
+        if phylo_run.outputs:
+            # there should only be one phylotree associated with the phylorun
+            phylo_tree = phylo_run.outputs[0]
```
jgadling: It's theoretically possible for a run to have multiple outputs, can we iterate over the outputs looking for one with a PhyloTree type?

src/backend/aspen/api/views/phylo_runs.py
phoenixAja: yes because otherwise we get an io error when updating these objects.

src/backend/aspen/api/views/phylo_runs.py
phoenixAja: yes! when would a run have multiple outputs though? I'm just curious

src/backend/aspen/api/views/phylo_runs.py
phoenixAja: a little confused, so should i rename `PhyloRunUpdateResponse` to be more generic (like `PhyloRunResponse`)? Or should i use `PhyloRunsListResponse`?

src/backend/aspen/api/views/phylo_runs.py
jgadling: Right now this won't happen! It's just future-proof'ing

src/backend/aspen/api/views/phylo_runs.py
jgadling: It should probably return a `PhyloRunResponse`!

src/backend/aspen/api/views/phylo_runs.py
phoenixAja: ok i updated it

src/backend/aspen/api/views/phylo_runs.py
jgadling: I don't think we do ... this should work without the `with` and just running `await db.commit()` below.

phoenixAja: removed

phoenixAja:
ok ready to review again @jgadling

src/backend/aspen/api/views/phylo_runs.py
```diff
@@ -193,25 +193,42 @@ async def kick_off_phylo_run(
...
+            joinedload(PhyloRun.group),  # For Pydantic serialization
+        )
         .filter(
-            sa.and_(
-                PhyloRun.group == user.group,  # This is an access control check!
-                PhyloRun.id == run_id,
-            )
+            sa.or_(
+                PhyloRun.group == user.group,
+                user.system_admin,
```
jgadling, line 203: This function is `get_editable_phylo_runs_by_id` and system admins aren't currently allowed to edit/delete phylo runs for groups other than the one they belong to. I'm not sure if we need a separate `get_readable_phylo_runs` function, or to be able to switch between queries for editable vs readable runs.

jgadling (changes requested):
Looks good aside from the access control query!

src/backend/aspen/api/views/phylo_runs.py
phoenixAja: i updated it within this one function

jgadling (approved):

PR #887, opened 2021-12-22
Danrlu/yaml builder comments

danrlu:
Wondering whether this works. Don't have to use it, I was just trying it out. Not tested.

Also I think I finally understand the class/object/method in python lol re-wrote the yaml builder piece in tree wiki accordingly: https://czi.atlassian.net/wiki/spaces/SCI/pages/2390394398/Aspen+-+set+up+for+tree+run+backend

### DIFF
--- a/src/backend/aspen/workflows/nextstrain_run/build_config.py
+++ b/src/backend/aspen/workflows/nextstrain_run/build_config.py
@@ -44,15 +44,8 @@ class NonContextualizedBuilder(BaseNextstrainConfigBuilder):
     crowding_penalty = 0.1
 
     def update_subsampling(self, config, subsampling):
-        # Update our query to be division or country level.
-        if self.tree_build_level == "division":
-            subsampling["same_county"][
-                "query"
-            ] = '''--query "(division == '{division}')"'''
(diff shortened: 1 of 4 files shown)


src/backend/aspen/workflows/nextstrain_run/build_config.py
```diff
@@ -126,14 +119,14 @@ def update_subsampling_for_location(tree_build_level, subsampling):
 
 def update_subsampling_for_country(subsampling):
     # State and country aren't useful
-    del subsampling["state"]
-    del subsampling["country"]
+    if subsampling["state"] exist, then del subsampling["state"]        # fake code. I cannot python...
+    if subsampling["country"] exist, then del subsampling["country"]    # fake code
```
jgadling: be careful here -- this is almost real python! I'm going to start expecting unit tests with your PR's soon 😝

danrlu: hahahaha i was 50% ok with python, and after realizing i understood the yaml builder code invocation totally backward, i'm now 20% confident

PR #892, opened 2021-12-23
reconfigure auto build overview tree

danrlu:
### Summary:
- **What:** Make the overview trees to focus on samples collected in the past 3 months. 

   - Pass a min date (12 weeks from now) to the nextstrain yaml 

   - Change the title shown on the tree to include "3 months"

   - Append "Recency-Focused Build" to the tree names in Aspen tree table. Implemented in https://github.com/chanzuckerberg/aspen/pull/894 Thanks Jess!

   - All runs should work regardless of whether the county has samples in the past 3 months.

### Demos:
Only samples from the last 3 months from Marin are on the tree (blue), together w other contextual samples

image

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/workflows/nextstrain_run/builder_base.py
+++ b/src/backend/aspen/workflows/nextstrain_run/builder_base.py
@@ -57,9 +57,9 @@ class BaseNextstrainConfigBuilder:
 
         # Update the tree's title with build type & location.
         if self.subsampling_scheme == "OVERVIEW":
-            title_template = "{tree_type} tree for samples collected in {location}"
-        else:
             title_template = "Contextualized tree for samples collected in {location} in the last 3 months"
+        else:
+            title_template = "{tree_type} tree for samples collected in {location}"
         build["title"] = title_template.format(
             tree_type=self.subsampling_scheme.title(),
             location=", ".join(location_values),
--- a/src/backend/aspen/workflows/nextstrain_run/builder_base.py
+++ b/src/backend/aspen/workflows/nextstrain_run/builder_base.py
@@ -56,7 +56,10 @@ class BaseNextstrainConfigBuilder:
         build["subsampling_scheme"] = self.subsampling_scheme
 
         # Update the tree's title with build type & location.
-        title_template = "{tree_type} tree for samples collected in {location}"
+        if self.subsampling_scheme == "OVERVIEW":
+            title_template = "{tree_type} tree for samples collected in {location}"
+        else:
+            title_template = "Contextualized tree for samples collected in {location} in the last 3 months"
         build["title"] = title_template.format(
             tree_type=self.subsampling_scheme.title(),
             location=", ".join(location_values),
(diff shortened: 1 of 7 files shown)


src/backend/aspen/workflows/nextstrain_run/builder_base.py
```diff
@@ -56,7 +56,10 @@ def update_build(self, config):
         build["subsampling_scheme"] = self.subsampling_scheme
 
         # Update the tree's title with build type & location.
-        title_template = "{tree_type} tree for samples collected in {location}"
+        if self.subsampling_scheme == "OVERVIEW":
+            title_template = "Contextualized tree for samples collected in {location} in the last 3 months"
+        else:
+            title_template = "{tree_type} tree for samples collected in {location}"
```
danrlu, line 62: this could go to builder_config.py, but having the concatenated {location} here is too convenient XD

src/backend/aspen/workflows/nextstrain_run/builds_templates/mega_template.yaml
```diff
@@ -73,6 +73,12 @@ subsampling:
           type: "proximity"
           focus: "group"
 
+    international_serial_sampling:
+      group_by: "region year month"
+      seq_per_group: 2
+      query: --query "(country != '{country}')"
+
+
```
danrlu, line 81: this is added so we still have some samples across all time to support a normal-looking tree. it's the same section in targeted tree

jgadling (approved):

src/backend/aspen/workflows/nextstrain_run/run_nextstrain_scheduled.sh
```diff
@@ -59,6 +61,9 @@ aligned_gisaid_location=$(
            --builds-file /ncov/my_profiles/aspen/builds.yaml       \
 )
 
+# temporary fix for recency focused automatic build
+# even if the group has no samples in the past 3 months, the run will not fail
+sed -i '/include\.txt/d' /ncov/my_profiles/aspen/builds.yaml
```
danrlu, line 66: this will make the run use default `include.txt` from `ncov`. if the county has no samples from past 3 months, the tree will be built around sequences in the default `include.txt` (the Wuhan sequence, and will not be very useful XD). I put it here instead of yaml builder since on-demand overview tree will need us to provide the `include.txt`.

danrlu:
Just adding a note to off-load what's in my head: 
- It might be helpful to [re-instate](https://github.com/chanzuckerberg/aspen/blob/e840cfd38974b3587c6577045ab48427491e1cd7/.happy/terraform/modules/ecs-stack/county_runs.tf#L17) a daily automatic tree run in staging for routine testing (Marin County is better to use than Santa Clara w much less samples). 
- To-learn item once I'm back to work: how are tree names in tree table generated. We changed JSON raw name, folder structure, and now there are 2 overview tree types that use diff names to distinguish them.

src/backend/aspen/workflows/nextstrain_run/run_nextstrain_scheduled.sh
danrlu: @jgadling this should make all automatic overview tree work. it deletes the include.txt line in the builds.yaml [here](https://github.com/chanzuckerberg/czgenepi/blob/7c13ed736ec420cb8d6be3b91bfae3249204886d/src/backend/aspen/workflows/nextstrain_run/builds_templates/mega_template.yaml#L23) so it will use the default include.txt, which contains ref genomes.

PR #893, opened 2021-12-23
name the new overview tree differently on the tree table

danrlu:
I'm not confident in the solution here, just leaving this as a note. But not sure how robust it is to put a stopgap on a [stopgap](https://github.com/chanzuckerberg/aspen/pull/874) for tree names on the tree table.

Not sure what is the usage of this argument: https://github.com/chanzuckerberg/aspen/blob/4d75b11391f45ab09acdeb85d16bc4841bba5fce/src/backend/aspen/workflows/nextstrain_run/run_nextstrain_scheduled.sh#L104

### DIFF
--- a/src/backend/aspen/app/views/phylo_trees.py
+++ b/src/backend/aspen/app/views/phylo_trees.py
@@ -37,6 +37,8 @@ def humanize_tree_name(s3_key: str):
     basename = re.sub(r".json", "", json_filename)
     if basename == "ncov_aspen":
         return s3_key.split("/")[1]  # Return the directory name.
+    if basename == "ncov_aspen_3m":
+        return s3_key.split("/")[1] + " recency-focused build"  # Name it differently from overview tree.
     title_case = basename.replace("_", " ").title()
     if "Ancestors" in title_case:
         title_case = title_case.replace("Ancestors", "Contextual")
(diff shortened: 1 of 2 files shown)


src/backend/aspen/app/views/phylo_trees.py
```diff
@@ -37,6 +37,8 @@ def humanize_tree_name(s3_key: str):
     basename = re.sub(r".json", "", json_filename)
     if basename == "ncov_aspen":
         return s3_key.split("/")[1]  # Return the directory name.
+    if basename == "ncov_aspen_3m":
+        return s3_key.split("/")[1] + " recency-focused build"  # Name it differently from overview tree.
```
jgadling, line 41: I don't really like this hack, but it will work if we're in a super-hurry. I'll think of a cleaner option.

danrlu: We can go without re-naming the trees if this evolves into a full-on tree name tree table update XD Closing this PR and leaving it to experts.

jgadling: I think we just want to add a `--tree-name` flag to the create-phylo-runs script, and put the name you want here, and then we'll be all set:
https://github.com/chanzuckerberg/aspen/blob/trunk/src/backend/aspen/workflows/nextstrain_run/run_nextstrain_scheduled.sh#L35

I'll give this a shot.

PR #894, opened 2021-12-23
Set scheduled tree names.

jgadling:
### Summary:
- **Env:** `<rdev link>`

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/src/backend/aspen/cli/db.py
+++ b/src/backend/aspen/cli/db.py
@@ -182,6 +182,12 @@ def add_can_see(
 
 @db.command("create-phylo-run")
 @click.option(
+    "--tree-name",
+    type=str,
+    required=False,
+    help="Name of tree being created",
+)
+@click.option(
     "--group-name",
     type=str,
     required=True,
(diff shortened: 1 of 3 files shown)


src/backend/aspen/cli/db.py
```diff
@@ -242,6 +249,8 @@ def create_phylo_run(
         )
         workflow.inputs = [aligned_gisaid_dump]
         workflow.template_args = json.loads(builds_template_args)
+        if tree_name:
+            workflow.name = tree_name
```
danrlu, line 253: how is this passed to what the user see on tree table? (all the front end related stuff is very mysterious to me)

jgadling: This name gets copied over to the PhyloTrees table when the tree is complete, and that's the default name for the tree in the UI:
https://github.com/chanzuckerberg/aspen/blob/trunk/src/backend/aspen/app/views/phylo_trees.py#L113

(this is the same way that user-inputted names for on-demand trees work)

danrlu (approved):
Should we attach this to the re-configure PR so they go in together? 
Thanks Jess!! Lightening speed~~~~~~~~

PR #897, opened 2021-12-29
HOTFIX -- Fix missing JSON download link for trees

vincent-czi:
### Summary:
- **What:** Allison Black brought this up [in Slack](https://czi-sci.slack.com/archives/C01HYJYA50W/p1640722279417900): going to Phylo Trees table and trying to download the JSON -- "Tree file (.json)" -- wasn't working. When you clicked the link, it just brought you back to the Samples table, but nothing was downloaded. This PR fixes that. See Notes below for what happened.
- **Ticket:** [sc<178847>](https://app.shortcut.com/genepi/story/178847/phylo-tree-table-json-download-broken) -- _Edit:_ I filed a ticket after writing up the PR
- **Env:** https://hotfix-jsontrees-frontend.dev.czgenepi.org/data/samples ... I think? I'm on vacation and have bad internet here, so I'm trying to push an rdev by using the branch naming approach, but I've never done this before and not sure I haven't screwed it up... **IMPORTANT** You can't actually verify this change on the rdev! (Also probably why this bug got through) Because the rdev data is not complete, the phylo tree links always fail when you try to download them. You can kinda verify by looking at the fact that the URL the link is pointing to is "real" though -- see the Demos below.

### Demos:

Before (current picture from Staging)
image

Currently mousing over the "Tree file" link: note the `https://staging.czgenepi.org/data/-` URL it's pointing to in the bottom-left.

----
Now (picture from my local dev branch, what this PR is)
image

Now the link URL in bottom-left is pointing to a real thing.

### Notes:
The root cause of this bug was the restructuring of how the Phylo Tree table presents the various additional options for interaction as part of CRUD work -- the `TreeActionMenu` component. The table renderers assume a key name <--> table row item <--> cell/column identifier connection. This is used to pull the `value` prop for the specific cell renderer. With the restructuring to accomplish CRUD, the download aspect got moved a level down, and while it used to be keyed under the column id of `downloadLink`, that became `actionMenu`, so the `value` prop was now a pull of a non-existent table row value, so when that got sent down to the component actually making the link, it had nothing to target.

Additionally, because rdev and local aren't structured the same as Staging and Prod for the phylo run data, it was very hard to notice this during testing -- the download links already failed in rdev and local. It's just that after this bug, the JSON download started failing for a new reason, but that would have been a very subtle difference to see during testing. It wouldn't become clear it's a bug until after being merged and noticed in Staging/Prod.

That also means **it's very important to test this fix in Staging after it gets merged**, since that's the only place the fix can really be verified.

### Checklist:
- [x] I merged latest `trunk`
- [x] I manually verified the change -- Kinda, it's hard to verify because the phylo runs aren't the same in local and rdev. But the URL is being generated correctly now and is visible in the link, where it was not before, so I'm confident this will work. But has to be verified for realsies in Staging.
- [x] I added labels to my PR
- ~[ ] I have notified others of changes they need to make locally (migrations, jobs, package updates, etc)~ none needed

### DIFF
--- a/src/frontend/src/views/Data/components/TreeActionMenu/components/TreeTableDownloadMenu/index.tsx
+++ b/src/frontend/src/views/Data/components/TreeActionMenu/components/TreeTableDownloadMenu/index.tsx
@@ -8,10 +8,9 @@ import { StyledIcon, StyledIconWrapper } from "../../style";
 
 interface Props {
   item: TableItem;
-  value: string;
 }
 
-const TreeTableDownloadMenu = ({ item, value }: Props): JSX.Element => {
+const TreeTableDownloadMenu = ({ item }: Props): JSX.Element => {
   const [anchorEl, setAnchorEl] = useState<Element | null>(null);
 
   const handleClick: MouseEventHandler = (event) => {
@@ -22,7 +21,7 @@ const TreeTableDownloadMenu = ({ item, value }: Props): JSX.Element => {
     setAnchorEl(null);
   };
 
-  const jsonLink = stringGuard(value);
+  const jsonLink = stringGuard(item["downloadLink"]);
   const tsvDownloadLink = stringGuard(item["accessionsLink"]);
   const disabled = item?.status !== TREE_STATUS.Completed;
   // TODO (mlila): This is necessary due to an sds bug -- MUI tooltips should not display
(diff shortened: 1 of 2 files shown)


vincent-czi:
I just checked in the rdev -- yay, I got the branch-naming right -- and to the extent this thing can be verified in rdev, it checks out. The JSON download link now does point to a "real" URL as expected.

mmmmmmaya (approved):

PR #898, opened 2021-12-30
fix sample filters.. apply date filters from midnight until 115959 fo…

mmmmmmaya:
### Summary
- **What:** When user filters on a date field, apply the date filter in the user's tz instead of utc
- **Why:** This is intuitive to the user (for example, if user says "yesterday", they probably mean 12:00am - 11:59pm of the previous day in their own time zone).
- **Ticket:** [[sc-166337]](https://app.shortcut.com/genepi/story/166337)
- **Env:** https://tzbug-frontend.dev.czgenepi.org

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/components/FilterPanel/components/CollectionDateFilter/index.tsx
+++ b/src/frontend/src/components/FilterPanel/components/CollectionDateFilter/index.tsx
@@ -1,5 +1,5 @@
 import React, { FC } from "react";
-import { FormattedDateType } from "src/components/DateField";
+import { UpdateDateFilterType } from "../..";
 import { DateFilter, DateMenuOption } from "../DateFilter";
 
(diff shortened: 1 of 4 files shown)


src/frontend/src/components/FilterPanel/components/CollectionDateFilter/index.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { FC } from "react";
-import { FormattedDateType } from "src/components/DateField";
+import { UpdateDateFilterType } from "../..";
```
phoenixAja, line 2: i like when the import comes from an actual file, but that's a nit

phoenixAja (approved):

PR #910, opened 2022-01-12
Revert to older ncov versions

jgadling:
### Summary:
- **What:** Ncov broke something in this PR (https://github.com/nextstrain/ncov/pull/814/) that means it can't find our `ncov/results/filtered_gisaid.fasta.xz` file so we're reverting to an older version while we figure it out.

### Checklist:
- [ ] I added labels to my PR


danrlu (approved):
Thank you very much!!

PR #911, opened 2022-01-12
Add bulk samples update endpoint

jgadling:
### Summary:
- **What:** Add an API endpoint to update sample metadata.
- **Ticket:** [sc171783](https://app.shortcut.com/genepi/story/171783)

### Notes:
This adds an endpoint that accepts a list of samples with metadata fields to update. Any empty or missing metadata fields will be ignored, and only non-empty fields will be modified. 

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/views/phylo_runs.py
+++ b/src/backend/aspen/api/views/phylo_runs.py
@@ -181,8 +181,8 @@ async def kick_off_phylo_run(
     return PhyloRunResponse.from_orm(workflow)
 
 
-async def get_editable_phylo_runs_by_id(db, user, run_id=None, all_viewable=False):
-    # get list of all viewable phylo_runs viewable to user, otherwise returh phylo_run matching run_id if sufficient permissions
+async def _get_accessible_phylo_runs(db, user, run_id=None, editable=False):
+    # get phylo_runs viewable or editable by a user, optionally filtered by id
     cansee_owner_group_ids: Set[int] = {
         cansee.owner_group_id
         for cansee in user.group.can_see
@@ -195,7 +195,12 @@ async def get_editable_phylo_runs_by_id(db, user, run_id=None, all_viewable=Fals
     )
 
     # These are access control checks!
-    if all_viewable:
+    if editable:
+        # for update and delete views return only trees that are in the users group
+        query = query.filter(
+            PhyloRun.group == user.group,
+        )
+    else:
         # this is for list view, return all runs that are viewable
         query = query.filter(
             sa.or_(
@@ -203,11 +208,6 @@ async def get_editable_phylo_runs_by_id(db, user, run_id=None, all_viewable=Fals
                 user.system_admin,
                 PhyloRun.group_id.in_(cansee_owner_group_ids),
             ),
-        )
-    else:
-        # for update and delete views return only trees that are in the users group
-        query = query.filter(
-            PhyloRun.group == user.group,
         )
 
     if run_id:
@@ -224,6 +224,12 @@ async def get_editable_phylo_runs_by_id(db, user, run_id=None, all_viewable=Fals
     results = await db.execute(query)
     return results.unique().scalars().all()
 
+async def get_editable_phylo_runs(db, user, run_id=None):
+    return await _get_accessible_phylo_runs(db, user, run_id, editable=True)
+
+async def get_readable_phylo_runs(db, user, run_id=None):
+    return await _get_accessible_phylo_runs(db, user, run_id, editable=True)
+
 
 @router.get("/")
 async def list_runs(
@@ -233,8 +239,8 @@ async def list_runs(
     user: User = Depends(get_auth_user),
 ) -> PhyloRunsListResponse:
 
-    phylo_runs: Iterable[PhyloRun] = await get_editable_phylo_runs_by_id(
-        db, user, all_viewable=True
+    phylo_runs: Iterable[PhyloRun] = await get_readable_phylo_runs(
+        db, user
     )
 
     # filter for only information we need in sample table view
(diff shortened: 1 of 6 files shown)


src/backend/aspen/api/views/phylo_runs.py
```diff
@@ -225,6 +225,14 @@ async def get_editable_phylo_runs_by_id(db, user, run_id=None, all_viewable=Fals
     return results.unique().scalars().all()
 
 
+async def get_editable_phylo_runs(db, user, run_id=None):
+    return await _get_accessible_phylo_runs(db, user, run_id, editable=True)
+
+
+async def get_readable_phylo_runs(db, user, run_id=None):
+    return await _get_accessible_phylo_runs(db, user, run_id, editable=True)
```
lvreynoso: I believe `get_readable_phylo_runs()` should pass `editable=False`, and I think that's what is breaking the tests.

src/backend/aspen/api/views/samples.py
```diff
@@ -211,3 +224,63 @@ async def delete_sample(
...
+
+    # are there any samples that can't be updated?
+    uneditable_samples = [
+        s for s in sample_ids_to_update if s not in [i.id for i in editable_samples]
+    ]
+    if uneditable_samples:
+        raise ex.NotFoundException("some samples cannot be updated")
+    # Check that we don't have dupe public/private ID's
+    [sample.private_identifier for sample in editable_samples]
+    [sample.public_identifier for sample in editable_samples]
```
lvreynoso: I think you intended to flesh this out, yeah?

lvreynoso (changes requested):
A few minor things. Looks like you forgot to flesh out some code to check for duplicate sample identifiers? Otherwise looks good - thanks for porting the tests!

src/backend/aspen/api/views/phylo_runs.py
jgadling: Man, I had some git fails on this PR. 😞  Thanks for catching this!

src/backend/aspen/api/views/samples.py
jgadling: That's leftover from me trying to generate nicer duplicate-name error messages, which didn't go well and got nixed. Removed!

lvreynoso (approved):

src/backend/aspen/api/views/samples.py
```diff
@@ -211,3 +224,60 @@ async def delete_sample(
...
+
+@router.put("/")
+async def update_samples(
+    update_samples_request: UpdateSamplesRequest,
+    db: AsyncSession = Depends(get_db),
+    settings: Settings = Depends(get_settings),
+    user: User = Depends(get_auth_user),
+) -> SamplesResponseSchema:
+
+    # reorganize request data to make it easier to update
```
vincent-czi, line 237: This is interesting to me, because the reorganized form below is actually closer to how we internally track the FE's sample data/metadata during the Upload flow. It looks like the creation API format uses an incoming list, but it's odd because the FE is actually re-formatting its internal structure from object/dict to list to do that call. I don't know why that format got picked for creation, I wonder if it would be worth re-structuring the request format since it looks like that's the better way for both FE and BE to represent the data.

vincent-czi: Looking at it some more, I guess it's because in the original creation request, the pathogen_genome and the sample are each sub-objects that make up different aspects? Seems like it could have been bundled up under the sampleId, but it does make more sense in that context.

Not sure if it's worth changing the request format between creation and update, but it already changes somewhat since update doesn't have sub-dicts for pathogen_genome and sample. If you'd rather just have the incoming data fit what you're reorganizing it into, it's certainly no harder for the FE.

src/cli/aspencli.py
```diff
@@ -349,6 +355,41 @@ def delete_samples(ctx, sample_ids):
...
+@click.option("--public-id", required=False, type=str, help="Update the sample public id")
+@click.option("--collection-date", required=False, type=str, help="Update the sample collection date")
+@click.option("--sequencing-date", required=False, type=str, help="Update the sample sequencing date")
+@click.option("--private", required=False, type=bool, help="Set whether the sample is private")
+@click.option("--location", required=False, type=int, help="Set the sample's collection location")
+@click.pass_context
+def update_samples(ctx, sample_id, private_id, public_id, collection_date, sequencing_date, private, location):
+    api_client = ctx.obj["api_client"]
+    if collection_date:
+        collection_date = dateparser.parse(collection_date).strftime('%Y-%m-%d')
```
vincent-czi, line 370: Hadn't heard of `dateparser` before, that's a neat library. I'm not sure it's necessary to bring it in as a dep though since we require our users to strictly adhere to YYYY-MM-DD format -- making ourselves follow the same convention when we use the CLI tool seems reasonable. Seems like the tradeoff here is adding an additional dependency VS more forgiving ingestion of dates when we use the tool. But since the request will just get turned down if we screw up the date formatting, it doesn't seem that beneficial to have the extra flexibility for date formats.

That said, it's just a local-only dependency for running our CLI tool, so it's definitely not a big deal.

jgadling: Yeah, good callout -- I agree that YYYY-MM-DD is pretty standard but I'm curious to see if having the flexibility to say "3 days ago" or "January 14 2021" might come in handy. Let's try it out for awhile and we can drop it if it we never need it.

src/backend/aspen/api/views/samples.py
jgadling: I really want the API **responses** to send back lists as much as possible because eventually we'll need to add some sort of searching/pagination on the backend and we'll be returning *ordered* results.

~~As for sub-dicts... since we already require those structures for upload, we should probably keep them similar! I'll update the request format.~~
It looks like our samples list endpoint returns the sequencing date as part of the sample response - without nesting. We already have our structures out of sync between list/update and create. 😢 I think we should commit to the format in the list response for now.

vincent-czi: Thanks for looking into it. Seems like the most reasonable option then.

PR #914, opened 2022-01-14
Associate a bot user account with all automated tree builds.

jgadling:
### Summary:
- **What:** Associate all automatic builds in the DB with a bot user so the UI doesn't have to "figure out" what to do with empty user fields as much. After this migration is complete, we **will** still have some (though relatively few) tree builds in the db without user id's since we enabled on-demand runs **before** we started tracking which user initiated the run.
- **Ticket:** [sc177356](https://app.shortcut.com/genepi/story/177356)
- **Env:** https://botuser-frontend.dev.czgenepi.org/data/phylogenetic_trees

### Notes:

We're currently setting the user's name to **Automatic Build**, which is what will show up in the frontend. We can use some other name (or even update the name later) if there's a better option.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/database_migrations/versions/20220114_192608_set_autoubuild_tree_user.py
+++ b/src/backend/database_migrations/versions/20220114_192608_set_autoubuild_tree_user.py
@@ -19,13 +19,13 @@ def upgrade():
 
     # Create a new bot group and add our bot user to it.
     create_group_sql = sa.sql.text(
-        "INSERT INTO aspen.groups (name, address, prefix, default_tree_location_id) SELECT 'Bot Users', 'Bot Users', 'bot', id FROM locations WHERE division = 'California' AND location = 'San Mateo County'"
+        "INSERT INTO aspen.groups (name, address, prefix, default_tree_location_id) SELECT 'Bot Users', 'Bot Users', 'bot', id FROM aspen.locations WHERE division = 'California' AND location = 'San Mateo County'"
     )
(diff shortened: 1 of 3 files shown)


src/backend/database_migrations/versions/20220114_192608_set_autoubuild_tree_user.py
```diff
@@ -0,0 +1,45 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    conn = op.get_bind()
+
+    # Create a new bot group and add our bot user to it.
+    create_group_sql = sa.sql.text(
+        "INSERT INTO aspen.groups (name, address, prefix, default_tree_location_id) SELECT 'Bot Users', 'Bot Users', 'bot', id FROM aspen.locations WHERE division = 'California' AND location = 'San Mateo County'"
```
phoenixAja, line 22: id doesn't need to be in quotes?

phoenixAja (approved):

src/backend/database_migrations/versions/20220114_192608_set_autoubuild_tree_user.py
jgadling: the fields that we've quoted in this query are hard-coded strings - we want to use `Bot Users` as the group name, but we want to use the dynamic `id` value we select from the locations table to populate the group's location.

PR #916, opened 2022-01-15
Metadata TSV Upload Kitchen Sink

vincent-czi:
### Summary:
- **What:** Refactor of Metadata TSV, plus various fixes/improvements, along with some restructuring of how Locations get pulled and scaffolding general API usage.
- **Ticket:** Whole lotta tickets! 
[sc<171401>](https://app.shortcut.com/genepi/story/<171401>)
[sc<167222>](https://app.shortcut.com/genepi/story/<167222>)
[sc<167758>](https://app.shortcut.com/genepi/story/<167758>)
[sc<176832>](https://app.shortcut.com/genepi/story/<176832>)
- **Env:** [https://tsv-upload-frontend.dev.czgenepi.org/](https://tsv-upload-frontend.dev.czgenepi.org/)

### Notes:
There's a lot going on in this PR.

- Tweaked how `Locations` data get pulled in. Now uses react-query since it benefits from caching. I'm also just generally hoping to standardize our API calls to using react-query for consistency.
- Started to scaffold out a standardization for the `fetch` portion of API calls with some convenience helpers.
- Moved fetch of `Locations` data up in the component hierarchy to ensure it loads earlier. There's a possible race condition where if the user starts to enter a location before `Locations` has loaded, things break. This doesn't fix the race condition, but it at least gives us a big head start for the race.
- Abstracted a lot of the `Locations` stuff to utilities so it can be shared around the app more easily.
- Major refactoring of how the Metadata TSV gets ingested: instead of doing custom TSV file parsing, offloads the TSV/CSV parsing to parsing library, then builds up metadata and warnings from the various resulting row objects. This also allowed for us to drop the previously required row number column from the metadata template. [New approach is still backward compatible with old templates.]
- Break out warning notifications into more distinct components to make it easier to add in different kinds of warnings.
- Add in new warning notification to indicate when a data row from TSV was missing required data (ticket 167222)
- Ensure that missing required data also populates warnings on individual cells in table (ticket 167222)
- Add in new warning notification when metadata TSV is completely missing data for some of the samples (came about in discussions with Janeece)
- Fixes issue with touched status of fields constantly resetting (ticket 176832). The underlying cause of this was that the `formik` form objects were constantly re-initializing due to `enableReinitialize` being on as a way to control the form values. I was able to turn that off and change the approach for how we control form values from above, instead just relying on `setValues` from formik.
-  Restructure the metadata template and the example rows.
- Cleaned up and removed some unused interfaces/types

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR
- ~[ ] I have notified others of changes they need to make locally (migrations, jobs, package updates, etc)~ (no changes needed)


src/frontend/src/common/api/index.tsx
```diff
@@ -60,10 +78,24 @@ function convert<U extends APIResponse, T extends U[keyof U]>(
   const converted = jsonToType<T>(entry, keyMap);
   // key should always be a string anyways. no funky business please.
   converted.type = API_KEY_TO_TYPE[String(key)];
+  // ^^^ FIXME (Vince) Does this get used anywhere? What was original purpose?
```
lvreynoso, line 81: I wrote that, but I don't remember why!

src/frontend/src/common/queries/locations.ts
```diff
@@ -1,15 +1,65 @@
...
+  unknown
+> {
+  return useQuery([USE_LOCATIONS_INFO_QUERY_KEY], fetchLocations, {
+    retry: false,
+    // Using `select` allows it to share cache with other USE_LOCATIONS_INFO_QUERY_KEY,
+    // but give a different view on the same data after processed by `select` func.
+    select: foldInNamesToLocations,
+    // It's stable, avoid unnecessary re-fetches. More info in `useLocations`
+    staleTime: ONE_HOUR,
+  });
```
lvreynoso, line 64: very nice

src/frontend/src/views/Upload/index.tsx
```diff
@@ -29,8 +30,18 @@ export default function Upload(): JSX.Element | null {
...
 
+  // To avoid possible race condition in Metadata upload where we need locations,
+  // we fetch locations info early in parent container, so ready by that step.
+  // FIXME (Vince): Need to put in place error handling if fetch fails.
+  // Right now the use of empty array prevents a hard crash, but user will be
+  // confused because page will seem fine but location finding just won't work.
+  const { data: namedLocationsData } = useNamedLocations();
+  const namedLocations =
+    namedLocationsData?.namedLocations || ([] as NamedGisaidLocation[]);
+
```
lvreynoso, line 41: I think this is fine. If the fetch for locations fails, I doubt uploading samples is going to work!

lvreynoso (approved):

src/frontend/src/views/Upload/index.tsx
vincent-czi: For sure! I don't think we really need anything fancy, it would just be nice to have some kind of generic thing across the app for when one of our API calls fail, something like, "There has been an error, please refresh the page. If this persists, contact us." If any API call fails, I think pretty much everything is hosed, but right now failure is quiet and the app might get into weird places where it looks like it's functioning before glitching out later.

src/frontend/src/common/api/index.tsx
vincent-czi: Okay, I'll make a note to see about removing this in an upcoming PR. Thanks!

PR #919, opened 2022-01-19
Add accessions table and update Sample Gisaid ISL on Gisaid Ingest

lvreynoso:
### Summary:
- **What:** Adds an accessions table with a one-to-one relationship with a samples; populates table with rows for each sample; updates table gisaid_isl column from gisaid_metadata table; adds step in gisaid job to continue updating isl column;
- **Ticket:** [sc150108](https://app.shortcut.com/genepi/story/150108)
- **Env:** `<rdev link>`

### Notes:
This is the first step - this PR just adds an accession table, a migration that populates it, and adds code to keep the table updated. Changes in code to use the new accessions table will come next.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/database/models/accessions.py
+++ b/src/backend/aspen/database/models/accessions.py
@@ -6,14 +6,7 @@ from dataclasses import dataclass
 from typing import Optional, Type
 
 import enumtables
-from sqlalchemy import (
-    Column,
-    ForeignKey,
-    Integer,
-    PrimaryKeyConstraint,
-    String,
-    UniqueConstraint,
-)
+from sqlalchemy import Column, ForeignKey, Integer, String, UniqueConstraint
 
 from aspen.database.models.base import base
 from aspen.database.models.entity import Entity, EntityType
--- a/src/backend/aspen/database/models/accessions.py
+++ b/src/backend/aspen/database/models/accessions.py
@@ -9,7 +9,6 @@ import enumtables
 from sqlalchemy import (
     Column,
     ForeignKey,
-    ForeignKeyConstraint,
     Integer,
     PrimaryKeyConstraint,
     String,
--- a/src/backend/aspen/database/models/accessions.py
+++ b/src/backend/aspen/database/models/accessions.py
@@ -35,9 +35,8 @@ class Accessions(base):
     )
 
     sample_id = Column(
-        Integer, ForeignKey(Sample.id, ondelete="CASCADE"), primary_key=True
+        Integer, ForeignKey("samples.id", ondelete="CASCADE"), primary_key=True
     )
-    sample = relationship("Sample", back_populates="accessions")
 
     gisaid_isl = Column(String, nullable=True)
 
--- a/src/backend/aspen/database/models/accessions.py
+++ b/src/backend/aspen/database/models/accessions.py
@@ -40,16 +40,6 @@ class Accession(base):  # type: ignore
     __tablename__ = "accessions"
     __table_args__ = (
         PrimaryKeyConstraint("sample_id", name="pk_accessions_sample_id"),
-        ForeignKeyConstraint(
-            ["sample_id"],
-            ["aspen.samples.id"],
-            name="fk_accessions_sample_id_samples",
-        ),
-        ForeignKeyConstraint(
-            ["accession_type"],
-            ["aspen.accession_types.item_id"],
-            name="fk_accessions_accession_type_accession_types",
-        ),
         UniqueConstraint(
             "sample_id",
             "accession_type",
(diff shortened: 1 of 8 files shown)


src/backend/aspen/database/models/sample.py
```diff
@@ -242,6 +243,12 @@ class Sample(idbase, DictMixin):  # type: ignore
         ),
     )
 
+    accessions = relationship(
+        Accessions,
+        backref=backref("samples", cascade="all, delete", uselist=False),
```
jgadling: I think the backref should just be 'sample' here since an accession will only be associated with a single sample.

lvreynoso: `"samples"` as in the `samples` table. The `uselist=False` tells SQLAlchemy that this is a one-to-one relationship

lvreynoso: Or maybe you're right? The SQLAlchemy docs are so worthless, I can't tell what the hell that argument refers to

lvreynoso: nvm what I wrote earlier, changed it

src/backend/database_migrations/versions/20220119_153132_add_sample_accessions_table.py
```diff
@@ -0,0 +1,45 @@
...
+        sa.ForeignKeyConstraint(
+            ["sample_id"],
+            ["aspen.samples.id"],
+            name=op.f("fk_accessions_sample_id_samples"),
+        ),
+        schema="aspen",
+    )
+
+    populate_accessions_sql = sa.sql.text(
+        "INSERT INTO aspen.accessions (sample_id, gisaid_isl) SELECT id, NULL FROM aspen.samples"
```
jgadling: This means we'll have an accession row for every sample regardless of whether we have an ISL for it. Is that the intent here? If that's what we want, then we have to make sure we're also automatically creating a new Accession row for every new Sample we add to the db.

I think it's more typical to have a structure for the accessions table like:

```
accessions
--
sample_id
accession_type ENUM('gisaid', 'genbank') # we have a fancy enum library that makes this easy
accession
```

And then only create rows for samples where we find a match in the relevant database. This lets us add support for new databases easily (just update the enum) and the presence/absence of an accessions row lets us know whether we have matching accessions.

lvreynoso: Yeah, that's a good idea. I'll implement that

src/backend/aspen/workflows/import_gisaid_isls/save.py
```diff
@@ -0,0 +1,69 @@
...
+                    "accession_type"
+                ),
+                GisaidMetadata.gisaid_epi_isl,
+            )
+            .select_from(Sample)
+            .join(GisaidMetadata, Sample.public_identifier == GisaidMetadata.strain)
+            .subquery()
+        )
+
+        insert_accessions_stmt = insert(Accession.__table__).from_select(
```
jgadling, line 43: This query is identical to the one in your migration, ya?

src/backend/aspen/database/models/accessions.py
```diff
@@ -5,13 +5,61 @@
...
+    tablename="accession_types",
+)
+
+
+class Accession(base):  # type: ignore
+    """An accession for a single sample."""
+
+    __tablename__ = "accessions"
+    __table_args__ = (
+        PrimaryKeyConstraint("sample_id", name="pk_accessions_sample_id"),
```
jgadling: Sample ID isn't a primary key! We do need the unique key below but don't need this pk constraint.

src/backend/database_migrations/versions/20220119_153132_add_sample_accessions_table.py
```diff
@@ -0,0 +1,64 @@
...
+        ["GISAID_ISL", "GENBANK"],
+        schema="aspen",
+    )
+
+    op.create_table(
+        "accessions",
+        sa.Column("sample_id", sa.Integer(), nullable=False),
+        sa.Column("accession_type", enumtables.enum_column.EnumType(), nullable=False),
+        sa.Column("accession", sa.String(), nullable=False),
+        sa.PrimaryKeyConstraint("sample_id", name=op.f("pk_accessions_sample_id")),
```
jgadling: Again, no need for this primary key.

src/backend/aspen/workflows/import_gisaid_isls/save.py
lvreynoso: yep

src/backend/aspen/database/models/accessions.py
lvreynoso: yes, this was leftover from the previous way I did the table, thanks for catching this!

jgadling (approved):
looks good!

PR #920, opened 2022-01-20
use swipe module

morsecodist:
### Summary:
- **What:** Updates to use the swipe module instead of the duplicated swipe code
- **Ticket:** [sc139115](https://app.shortcut.com/idseq/story/139115/migrate-aspen-to-use-the-swipe-module)
- **Env:** [tmorse-frontend.rdev.](https://tmorse-frontend.dev.czgenepi.org/

### Checklist:
- [X] I merged latest `<base branch>`
- [X] I added labels to my PR
- [X] I have notified others of changes they need to make locally (migrations, jobs, package updates, etc)

### DIFF
--- a/.happy/terraform/modules/ecs-stack/main.tf
+++ b/.happy/terraform/modules/ecs-stack/main.tf
@@ -46,7 +46,7 @@ locals {
   batch_role_arn = local.secret["batch_queues"]["genepi"]["role_arn"]
   ec2_queue_arn  = local.secret["batch_envs"]["genepi"]["envs"]["EC2"]["queue_arn"]
   spot_queue_arn = local.secret["batch_envs"]["genepi"]["envs"]["SPOT"]["queue_arn"]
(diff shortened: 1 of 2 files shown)


jgadling:
Do we need to update any of our SFN invocations?

morsecodist:
@jgadling I don't THINK so but I am not 100% confident. Does my disabling in prod look reasonable? I am not sure that will work.

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -147,6 +148,9 @@ module backend_service {
 
 module swipe_sfn {
   source                     = "../swipe-sfn"
+
+  count = local.deployment_stage == "prod" ? 1 : 0
```
jgadling, line 152: I'm not 100% confident that this and the sfn_arn check will prevent prod from changing, but hopefully we remove this kill switch in the next day or two anyway and the whole thing goes to prod

jgadling (approved):

PR #922, opened 2022-01-21
Update APIs that use Sample accessions

lvreynoso:
### Summary:
- **What:** Updates API endpoints to use new accessions, deletes unused ones. Updates tests. Drops old tables `genbank_accession_workflows`, `genbank_accessions`, `gisaid_accession_workflows`, `gisaid_accessions`, and deprecated rows in `entities` `workflows`, and `workflow_inputs`. Removes GISAID ISL from upload form.
- **Ticket:** [sc150108](https://app.shortcut.com/genepi/story/150108)
- **Env:** `<rdev link>`

### Demos:
Screen Shot 2022-01-21 at 15 35 23

### Notes:
Removing the deprecated accession models broke the old migration to backfill gisaid isls, so I deleted the migration since it is now (1) useless and (2) redundant, the isls are in the new table. It's a good example of why we shouldn't rely on models in migrations.

### Checklist:
- [x] I added labels to my PR


vincent-czi:
I just merged in my TSV refactor changes. There's a bit of overlapping code because we both made changes that involved `islAccessionNumber`: let me know if there are any complicated conflicts you want to discuss.

jgadling (approved):

PR #925, opened 2022-01-24
Add subheaders for sample collection_location

lvreynoso:
### Summary:
- **What:** Adds proper parsing for a Sample's Collection Location in a user's metadata download.
- **Ticket:** [sc181535](https://app.shortcut.com/genepi/story/181535)
- **Env:** [https://mdta-fix-frontend.dev.czgenepi.org](https://mdta-fix-frontend.dev.czgenepi.org)

### Demos:
Screen Shot 2022-01-24 at 16 27 39

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/types/bioinformatics.d.ts
+++ b/src/frontend/src/common/types/bioinformatics.d.ts
@@ -32,7 +32,7 @@ interface Sample extends BioinformaticsType {
     name: string;
   };
   collectionDate: string;
-  collectionLocation: string;
+  collectionLocation: GisaidLocation;
   sequencingDate: string;
   submittingGroup: {
     id: number;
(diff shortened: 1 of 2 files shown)


danrlu (approved):
I can't read the code so I tried the dev env and it works. thanks for fixing it!!

vincent-czi:
Code-wise, it's good, but I wonder if our users would be better to receive the location in the format `North America/USA/California/Contra Costa County` rather than as individual items of "Region", "Country", "Division", "Location". I honestly don't know what's better for how they use it, but the single-string format is what we use during the upload process to unambiguously identify a location (`stringifyGisaidLocation`). If they're going to turn around and re-upload that location data to us, it might be more useful in that format. But if they're going on to directly interface with GISAID, I have no idea.

@danrlu Do you have any opinion between the two? If it's not a big deal, I think it's fine to leave it as it currently is -- breaking it up into the four categories -- but I really don't know how this gets consumed.

danrlu:
Good points! I was wondering about it, but thought breaking it up like currently is actually nicely formatted the metadata for Nextstrain, if they ever want to do their own tree build. So it's good either way! We can sit tight until someone complains.

PR #929, opened 2022-01-25
Use docker-compose profiles to specify which services to launch

jgadling:
### Summary:
- **What:** Use docker-compose profiles to choose which containers to build/start in local dev

### Notes:
I just learned that docker-compose has a [profiles feature](https://docs.docker.com/compose/profiles/) that allows us to specify which containers need to be launched in which contexts. This feature should simplify our makefiles a bit, and make it easier to run the correct sets of services in CI.

### Checklist:
- [ ] I added labels to my PR


lvreynoso (approved):
very cool!

PR #930, opened 2022-01-26
Fix bug where GISAID ISLs were not included in metadata downloads

lvreynoso:
### Summary:
- **What:** In parsing table headers for creating a metadata TSV file, an extra `CZBFailedGenomeRecovery` column is added. However instead of being pushed into the array it was added using subscript notation at the index where the `GISAID` header existed, thus overwriting that header.
- **Env:** `<rdev link>`

### Checklist:
- [x] I added labels to my PR


vincent-czi (approved):
Looks good!

Talking more generally, I feel like we should probably decouple the Metadata download from being so tied to the table's configuration. They're related, but this seems a little too tightly tied together. This is definitely not the PR to do it, though! Just something I'm thinking about when we refactor the tables since this will be impacted.

PR #932, opened 2022-01-27
add ondemand overview tree

phoenixAja:
### Summary:
- **What:** Add overview tree as a phylotree create option, also update other tree type copies.
- **Ticket:** [sc177170](https://app.shortcut.com/genepi/story/177170)
- **Env:** [rdev](https://overview-frontend.dev.czgenepi.org/)

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/RadioLabel/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/RadioLabel/index.tsx
@@ -35,8 +35,9 @@ export const RadioLabelOverview = ({ selected }: Props): JSX.Element => {
             <ListItemText>
               <SmallText>
                 Best for seeing an overall picture of viral diversity within
-                your jurisdiction, in the context of genetically similar GISAID
-                samples from outside of your jurisdiction.
+                your jurisdiction in the past 12 weeks, in the context of
+                genetically similar GISAID samples from outside of your
+                jurisdiction.
               </SmallText>
             </ListItemText>
           </StyledListItem>
--- a/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/RadioLabel/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/RadioLabel/index.tsx
@@ -4,7 +4,6 @@ import { List } from "czifui";
 import React from "react";
 import {
   Label,
-  LabelLight,
   LabelMain,
   SmallText,
   StyledIconCheckSmall,
@@ -17,16 +16,15 @@ interface Props {
   selected: boolean;
 }
 
-export const RadioLabelTargeted = ({ selected }: Props): JSX.Element => {
+export const RadioLabelOverview = ({ selected }: Props): JSX.Element => {
   return (
     <div>
       <Label>
-        <LabelMain>Targeted </LabelMain>
-        <LabelLight>— Recommended</LabelLight>
+        <LabelMain>Overview </LabelMain>
       </Label>
       <SmallText>
-        Includes additional publicly-available samples on GISAID from your
-        jurisdiction and other regions.
+        Includes samples from both within and outside of your jurisdiction, at a
+        ratio of roughly 2:1.
       </SmallText>
       {selected && (
         <List>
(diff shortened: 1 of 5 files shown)


Janeece:
@phoenixAja Looks great! Only update is to add in a small copy change to the first overview checkmark (this is the interim copy change until we add in filtering)

Copy: "…within your jurisdiction **in the past 12 weeks**, in the context…"
_Note: bolded here to show the change, only semi-bold "past 12 weeks" when you implement_

On-Demand Overview Trees Feedback

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/RadioLabel/index.tsx
```diff
@@ -17,16 +16,68 @@ interface Props {
...
+      <Label>
+        <LabelMain>Overview </LabelMain>
+      </Label>
+      <SmallText>
+        Includes samples from both within and outside of your jurisdiction, at a
+        ratio of roughly 2:1.
+      </SmallText>
+      {selected && (
+        <List>
+          <StyledListItem button={false as any}>
```
vincent-czi, line 31: I wish the `false as any` wasn't necessary -- it feels like there shouldn't be a need for a `button` prop declaration at all since MUI v4 docs say it defaults to false -- but I just wasted 30 minutes trying to alter it then understand why TS throws errors in response and I'm still unclear how it manages to go wrong. So yeah, definitely agree it should be here, but pretty annoying it's necessary!

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
```diff
@@ -71,15 +73,25 @@ export const CreateNSTreeModal = ({
   const [validatedInputSamples, setValidatedInputSamples] = useState<string[]>(
     []
   );
+  const [isValidTreeType, setValidTreeType] = useState<boolean>(false);
```
vincent-czi: This is super minor, but I think I'd prefer `setValidTreeType` to be `setIsValidTreeType`. It's cumbersome but I like the convention of a very clear state <--> setState relationship. Also, it's unlikely, but I think it could momentarily be confusing if you see `setValidTreeType`: you might think it's setting what the currently valid tree type is rather than setting whether a valid tree type has been selected.

src/frontend/src/common/queries/trees.ts
```diff
@@ -24,13 +24,13 @@ import { MutationCallbacks } from "./types";
 interface CreateTreePayload {
   name: string;
   samples: string[];
-  tree_type: string;
+  tree_type: string | undefined;
```
vincent-czi, line 27: I'm a bit torn on this. I think it's not really proper for the `CreateTreePayload` to allow `tree_type` to be undefined. That would be a bad payload to send, it's required on the BE side, right? But I'm torn because it could make for annoying TS errors above since the treeType can be undefined while on the modal, it's just that once it's valid for submission it will have gelled into a string by then. I think you'd have to put some kind of pointless type guard into the `handleSubmit` call so it makes it clear to TS that undefined will never get passed. Still, I guess it's an added layer of safety in case someone accidentally messes up the `isValidTreeType` in the modal.

I think similar notes apply to the other changes in this file.

vincent-czi (approved):
Overall, looks good. I've got some minor notes, but they're definitely in the nice-to-have category and not blockers.

src/frontend/src/common/queries/trees.ts
phoenixAja: yeah i was wondering if there was a way to have the `treeType` type be `undefined | string` and then have the `CreateTreePayload` `treeType` be a `string` since there will never be an occasion where we pass `undefined` in the payload.

phoenixAja: i couldn't find a clear way to do so 😞

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
phoenixAja: i'll change it

src/frontend/src/common/queries/trees.ts
vincent-czi: I think you'd have to make sure wherever the CreateTreePayload interface gets used -- or somewhere upstream of it where the payload gets generated and then also have any intermediate interfaces updated as well -- there's a type guard that's filtering out `undefined` cases.

very pseudocode
```
on handleSubmit
if (treeType === undefined) {
  // some kind of error handling or maybe just setShouldShowErrorNotification(true)
}
handle payload like usual here
```

Definitely feels like unnecessary boilerplate, but I'm pretty sure the type guard in that would let TS know it can trust it's not in the undefined case.

Alternatively, could have the interface let treeType be optional, but something in the mutation side raises an error that blows up the call before it actually POSTs to backend, but that's basically a matter of moving the boilerplate type guard.

phoenixAja: yeah that does seem like a lot of boilerplate for this.. i'll keep it as is for now

Janeece:
@phoenixAja Dan noticed a typo on Targeted trees that needs to be corrected -- the ratio is 1:2 instead of 2:1. 
Screen Shot 2022-02-01 at 3 17 55 PM

PR #935, opened 2022-01-31
add edit tree name to more actions menu

phoenixAja:
### Summary:
- **What:** adding editTrees feature flag, adding edit option to tree more actions menu
- **Ticket:** [sc179028](https://app.shortcut.com/genepi/story/179028)
- **Env:** no rdev, i'll wait for design review after adding edit tree menu

### Checklist:
- [x] I added labels to my PR


src/frontend/src/views/Data/components/TreeActionMenu/components/MoreActionsMenu/index.tsx
```diff
@@ -77,9 +78,15 @@ const MoreActionsMenu = ({
...
+          {usesFeatureFlag(FEATURE_FLAGS.editTrees) && (
+            <MenuItem onClick={() => undefined}>
+              <StyledEditIcon />
+              <StyledText>Edit Tree Name</StyledText>
+            </MenuItem>
+          )}
           <MenuItem onClick={() => onDeleteTreeModalOpen(item)}>
             <StyledTrashIcon />
-            <StyledText>Delete Tree</StyledText>
+            <StyledText isRed>Delete Tree</StyledText>
```
vincent-czi: Definitely a nit, but I'd generally prefer to call this sort of thing `isWarning` rather than tie the prop to the implied styling. That way if our warnings eventually change to being orange or whatever, there's no disconnect between meaning and intent.

src/frontend/src/views/Data/components/TreeActionMenu/components/MoreActionsMenu/style.ts
```diff
@@ -1,13 +1,23 @@
...
+export const StyledText = styled("span", {
+  shouldForwardProp: (prop) => !doNotForwardProps.includes(prop as string),
+})`
+  ${(props: ExtraProps) => {
+    const { isRed } = props;
     const colors = getColors(props);
 
     return `
-      color: ${colors?.error[600]};
+      color: ${isRed ? colors?.error[600] : "black"};
```
vincent-czi: This is a pain, but I _think_ we might actually not be using pure black for most text? I don't know well enough to be sure, but a quick check on my local shows our `<body>` having a blanket color of `rgba(0,0,0,.87)`, so a not-quite-black color.

There's two ways I can think of remedying this, but I don't really have a complete understanding of how we do theming with czifui. The easier option would be to just use the general text theme color instead of "black", but... I don't see `rgba(0,0,0,.87)` anywhere in our theme... _looking_... hooo, I think it might actually come from some other package called "semantic"? Yeah, looks like the CSS is pulling from that, and I see an import of `import "semantic-ui-css/semantic.min.css";` But that means we don't really directly set that color ourselves, so this option is kinda out...

The other option -- and I think pretty much the only reasonable one now -- is to have the CSS completely disappear when `isRed` is not present. That way it will inherit whatever the usual color would be without us needing to explicitly set it. Basically something like this:
```
  ${(props: ExtraProps) => {
    const { isRed } = props;
    if ( !isRed ) {
      return ''; // This might actually be better as null than empty string? I forget the nuances of the styled API
    }
    const colors = getColors(props);
    return `color: ${colors?.error[600]};`;
  }}
```

src/frontend/src/views/Data/components/TreeActionMenu/components/MoreActionsMenu/style.ts
```diff
@@ -26,3 +36,16 @@ export const StyledTrashIcon = styled(TrashIcon)`
     `;
   }}
 `;
+
+export const StyledEditIcon = styled(EditIcon)`
+  fill: black;
```
vincent-czi, line 41: I'm fine with this being black, but considering the above, it's interesting to think how our fill colors are ever so slightly different than our text colors. This might be something we want to bring up to Janeece, maybe talk to SDS team about.

vincent-czi (approved):
Looks good! I've got a really fine-grained CSS thing that should probably be addressed, but only the most eagle-eyed user would catch the issue. It does highlight a slight usage issue with how our styling approach works right now though, might be worth talking about more later.

src/frontend/src/views/Data/components/TreeActionMenu/components/MoreActionsMenu/style.ts
vincent-czi: It's challenging to see, but if you look at this screenshot, you can (just barely) make out the slight difference.
image

The "Selected" was explicitly set to `black`, while the text in the rows of the table (eg, "Orange County") is inheriting the general body value of `rgba(0,0,0,.87)`

vincent-czi: Considering we have other places where we're mixing in pure `black`, it's not a big deal here if you want to keep what you have, but it does speak to an underlying difficulty with how coloring and general app styles are being applied. Not sure if it's worth trying to figure out some months down the road or if we should just let it slide forever.

src/frontend/src/views/Data/components/TreeActionMenu/components/MoreActionsMenu/index.tsx
phoenixAja: yeah i see what you mean, it's not _really_ a warning though, it's more so just styling, i can change it though to make the wording more clear

src/frontend/src/views/Data/components/TreeActionMenu/components/MoreActionsMenu/style.ts
phoenixAja: Janeece already design approved this, and `black` is listed in the product spec, so I think we're ok here

src/frontend/src/views/Data/components/TreeActionMenu/components/MoreActionsMenu/style.ts
phoenixAja: ok resolved based on suggestion

PR #943, opened 2022-02-09
Crud v1 - edit tree modal

phoenixAja:
### Summary:
- **What:** Finish crud V1 work for updating tree names
- **Tickets:** 
  *  [sc179306](https://app.shortcut.com/genepi/story/179306)
  *  [sc179301](https://app.shortcut.com/genepi/story/179301)
- **Env:** [update trees rdev](https://update-trees-frontend.dev.czgenepi.org/data/phylogenetic_trees)

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR


phoenixAja:
@Janeece rdev is available for review now when you get a spare second!

src/frontend/src/common/components/library/data_subview/components/EditTreeConfirmationModal/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+  open,
+  tree,
+}: Props): JSX.Element | null => {
+  const [shouldShowErrorNotification, setShouldShowErrorNotification] =
+    useState<boolean>(false);
+  const [shouldShowSuccessNotification, setShouldShowSuccessNotification] =
+    useState<boolean>(false);
+  const [newTreeName, setNewTreeName] = useState<string>("");
+
+  const treeNameLength = newTreeName ? newTreeName.length : 0;
```
vincent-czi: Nit: since `newTreeName` inits to empty string and should only ever be set as a string, the ternary here isn't really necessary.

src/frontend/src/common/components/library/data_subview/components/EditTreeConfirmationModal/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+      setNewTreeName("");
+    },
+  });
+
+  const handleClose = function () {
+    onClose();
+    setNewTreeName("");
+  };
+
+  if (!tree) return null;
```
vincent-czi, line 62: I'm kinda torn on this line of having it return null considering how that quiet null return on another component confused us for 20 minutes yesterday. It's probably the cleanest API, but lifting up the determination of whether to render component or just pass `null` to the parent and cutting this line would make it more explicit and hopefully less confusing to troubleshoot.

... you know, thinking about it a bit more, if we were to have the parent do that determination but determine it with `open && tree ? <COMPONENT> : null`, that would allow this component to only mount when it's actually being used, then unmount ... ah shoot, right, that will unmount the Notifications. erhrhgh

src/frontend/src/common/components/library/data_subview/components/EditTreeConfirmationModal/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+export const EditTreeConfirmationModal = ({
+  onClose,
+  open,
+  tree,
+}: Props): JSX.Element | null => {
+  const [shouldShowErrorNotification, setShouldShowErrorNotification] =
+    useState<boolean>(false);
+  const [shouldShowSuccessNotification, setShouldShowSuccessNotification] =
+    useState<boolean>(false);
+  const [newTreeName, setNewTreeName] = useState<string>("");
```
vincent-czi, line 26: I don't think there's anything in this file that causes autofilling of the current tree name. Looking at rdev, it doesn't appear to be happening there.

src/frontend/src/common/components/library/data_subview/components/EditTreeConfirmationModal/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+      onClick={onEdit}
+    >
+      Update
+    </Button>
+  );
+
+  return (
+    <>
+      {shouldShowSuccessNotification && (
+        <Notification
```
vincent-czi, line 109: I kept thinking about these Notifications, and I've got a really clear implementation in mind for when we have a global state manager. Instead of displaying in the component that uses Notification, that component will just emit a notification to the global state, we capture them there, and then a component at the top level will subscribe to those notifications and can display/clear them at its leisure.

src/frontend/src/common/components/library/data_subview/index.tsx
```diff
@@ -147,6 +148,9 @@ const DataSubview: FunctionComponent<Props> = ({
   const [isDeleteTreeConfirmationOpen, setDeleteTreeConfirmationOpen] =
     useState<boolean>(false);
   const [treeToDelete, setTreeToDelete] = useState<Tree>();
+  const [isEditTreeConfirmationOpen, setEditTreeConfirmationOpen] =
+    useState<boolean>(false);
+  const [treeToEdit, setTreeToEdit] = useState<Tree>();
```
vincent-czi, line 153: I'm not sure it's worth it, but we could probably combine these two into a single state. Just have it be `treeToEdit` by itself: when it's false-y, that implies that the Edit Tree Confirmation should be closed, etc.

src/frontend/src/common/queries/trees.ts
```diff
@@ -24,13 +26,13 @@ import { MutationCallbacks } from "./types";
 interface CreateTreePayload {
   name: string;
   samples: string[];
-  tree_type: string | undefined;
+  tree_type: string | undefined; // treeType can be undefined when user first opens the NSTreeCreate modal
```
vincent-czi, line 29: Would we allow the user to emit `tree_type` to the BE API if it's undefined? I can see it starting that way when the FE presents the field to the user, but I'd think the payload would need it as string only.

src/frontend/src/common/queries/trees.ts
```diff
@@ -200,3 +202,58 @@ export function useDeleteTree({
...
+export async function editTree({
+  treeIdToEdit,
+  newTreeName,
+}: TreeEditRequestType): Promise<TreeEditResponseType> {
+  const payload: EditTreePayloadType = {
+    name: newTreeName,
+  };
+  const response = await fetch(API_URL + API.PHYLO_TREES_V2 + treeIdToEdit, {
+    ...DEFAULT_PUT_OPTIONS,
+    ...DEFAULT_HEADERS_MUTATION_OPTIONS,
```
vincent-czi, line 234: I strongly believe we should fold `DEFAULT_HEADERS_MUTATION_OPTIONS` into `DEFAULT_PUT_OPTIONS` (this is currently done with `DEFAULT_POST_OPTIONS`). The only prior existing use of `DEFAULT_PUT_OPTIONS` in the codebase already does the same `...DEFAULT_HEADERS_MUTATION_OPTIONS` spread as is being done here (this happens with `updateUserInfo` in auth.ts). I just can't really imagine us sending a PUT request that isn't ultimately sending JSON to the BE unless it's something so special and unusual that it can get a special case to handle its weirdo request headers.

src/frontend/src/common/queries/trees.ts
```diff
@@ -200,3 +202,58 @@ export function useDeleteTree({
...
+type TreeEditCallbacks = MutationCallbacks<TreeEditResponseType>;
+
+export async function editTree({
+  treeIdToEdit,
+  newTreeName,
+}: TreeEditRequestType): Promise<TreeEditResponseType> {
+  const payload: EditTreePayloadType = {
+    name: newTreeName,
+  };
+  const response = await fetch(API_URL + API.PHYLO_TREES_V2 + treeIdToEdit, {
```
vincent-czi, line 232: If you have the time, I think this would be a great candidate to make a generic `putBackendApiJson` ala my previous `getBackendApiJson`, then just use it here. You don't have to convert over anything else right now, but I want to standardize our request functions so we ensure consistency in our API requests and don't have so many bare `fetch` calls in the code.

src/frontend/src/common/utils/featureFlags.tsx
```diff
@@ -21,10 +21,6 @@ export const FEATURE_FLAGS: FlagsObj = {
     isDisabled: true,
     key: "mayasFlag",
   },
-  editTrees: {
```
vincent-czi: oh neat, I didn't realize it was ready to go live with this PR! Congrats!

src/frontend/src/components/ConfirmDialog/index.tsx
```diff
@@ -14,41 +21,67 @@ export interface ConfirmDialogProps {
...
+  withCloseIcon = false,
+  isConfirmButtonClickable = true,
 }: ConfirmDialogProps): JSX.Element {
   const confirmButton = customConfirmButton ?? (
     <Button color="primary" variant="contained" isRounded>
       Continue
     </Button>
   );
 
+  const confirmButtonwithDiv = isConfirmButtonClickable ? (
```
vincent-czi: If it's not clickable, it feels like there should be some styling change. This has it continue to present visually the same whether clickable or not, right? If that's what we'd want, I think I'd just dynamically change from `onConfirm` to `noop` depending on if it's clickable. But I think the `confirmButton` definition above should have the default case get some sort of styling for the `!isConfirmButtonClickable` case

src/frontend/src/components/EditDialog/index.tsx
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+import ConfirmDialog, {
+  ConfirmDialogProps,
+} from "src/components/ConfirmDialog";
+import { StyledSpan } from "src/components/DeleteDialog/style";
+
+interface Props extends Omit<ConfirmDialogProps, "onConfirm"> {
```
vincent-czi: Nit: Considering how there are now three components based around `ConfirmDialogProps`, this feels to me like we should life out the shared portion into `DialogProps`, then make each of them do their own extends to tack on their one additional prop.

src/frontend/src/components/EditDialog/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+  open,
+  title,
+  ...props
+}: Props): JSX.Element | null => {
+  if (!open) return null;
+
+  const styledTitle = <StyledSpan>{title}</StyledSpan>;
+
+  return (
+    <ConfirmDialog
```
vincent-czi: Nit, continued: Along the same lines, maybe create a shared `<InteractableDialog>` component, then `ConfirmDialog`, `EditDialog`, and `DeleteDialog` are all basically just convenience components that handle that setup aspect. They feel really tightly bundled together right now, but the way there written implies they're less tightly coupled than they really are. You couldn't currently change `ConfirmDialog` without majorly disrupting the other two, but I don't think that's obvious at a glance currently.

src/frontend/src/views/Data/components/TreeActionMenu/components/MoreActionsMenu/index.tsx
```diff
@@ -26,7 +27,8 @@ const MoreActionsMenu = ({
   const isAutoBuild = group?.name === "";
   const isTreeInUserOrg = userGroup?.name === group?.name;
   const canUserDeleteTree = isAutoBuild || isTreeInUserOrg;
-  const isDisabled = status === TREE_STATUS.Started || !canUserDeleteTree;
+  // Fix this to allow users to edit/delete FAILED runs once phylotrees V2 endpoint has been updated to better reflect tree status
```
vincent-czi: Can you put in the word FIXME in comment just to help it get noticed in the future?

vincent-czi (commented):
Overall, looks good. I've got a bunch of comments, but they're basically all in the nit or nice-to-have category. Nothing I see would block merging it in. If you have time to address some of the comments, great! but definitely don't feel like every one of them is seriously pressing. The only issue I want to verify before formally approving is that it's okay for the tree name in edit dialog to not default to currently set tree name (I believe that's what's happening rn).

Janeece (commented):
@phoenixAja this looks great! I added my feedback to this ticket: https://app.shortcut.com/genepi/story/185106/crud-v1-update-trees-design-feedback

src/frontend/src/common/components/library/data_subview/components/EditTreeConfirmationModal/index.tsx
phoenixAja: lol at "that quiet null return on another component confused us for 20 minutes yesterday"

phoenixAja: yeah i can't think of a good solution either 😩

src/frontend/src/common/components/library/data_subview/components/EditTreeConfirmationModal/index.tsx
phoenixAja: yeah i couldn't quite get the tree name population to work out, it's not triggering because of the component mounting before the tree prop is defined

phoenixAja: i did let janeece know

src/frontend/src/common/components/library/data_subview/components/EditTreeConfirmationModal/index.tsx
phoenixAja: i think that's a good idea, these notifications could definitely be handled a better way

src/frontend/src/components/ConfirmDialog/index.tsx
phoenixAja: i'm a little confused on what styling you think should be happening if it's not clickable, would you mind elaborating a bit more?

phoenixAja: the isConfirmButtonClickable prop name is confusing.. I can probably just change the delete tree modal instead to include onConfirm in the button itself instead of passing to the div

src/frontend/src/components/EditDialog/index.tsx
phoenixAja: so for Edit and Delete dialogs you think we should just have their props extend DialogProps instead of ConfirmDialogProps?

src/frontend/src/views/Data/components/TreeActionMenu/components/MoreActionsMenu/index.tsx
phoenixAja: oh yep

src/frontend/src/components/EditDialog/index.tsx
phoenixAja: we may need to sync on monday about what your suggesting so we're on the same page, not quite sure i'm following

src/frontend/src/common/components/library/data_subview/components/EditTreeConfirmationModal/index.tsx
phoenixAja: fixed!

src/frontend/src/components/ConfirmDialog/index.tsx
phoenixAja: removed this

src/frontend/src/components/EditDialog/index.tsx
phoenixAja: i create a basic dialog that should hopefully be a better solution

src/frontend/src/common/components/library/data_subview/components/EditTreeConfirmationModal/index.tsx
phoenixAja: removed

src/frontend/src/common/queries/trees.ts
phoenixAja: i'll do this when i add the update samples request, I sort of just want to get this PR merged in as soon as possible now 😓

src/frontend/src/common/components/library/data_subview/components/EditTreeConfirmationModal/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+    // this makes sure that the newTreeName defaults to the current tree name,
+    //  and that the newTreeName state variable resets when we edit a new tree
+    if (tree) {
+      setNewTreeName(tree.name);
+    }
+
+    // when we have a new tree name that also means that we should reset
+    // shouldShow notifications to false since the component is now focused on editing a new tree
+    setShouldShowSuccessNotification(false);
+    setShouldShowSuccessNotification(false);
```
vincent-czi: One of these should be `shouldShowErrorNotification`, it's currently two of a kind

src/frontend/src/common/components/library/data_subview/components/EditTreeConfirmationModal/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+  }, [
+    tree,
+    setNewTreeName,
+    setShouldShowSuccessNotification,
+    setShouldShowSuccessNotification,
```
vincent-czi: idem

vincent-czi (approved):
Looks good! There is one new comment that I'm pretty sure is pointing out something that needs fixing, but it's a two line fix so I'm just going to approve it now. I like the switch to the shared BaseDialog: good work!

src/frontend/src/common/components/library/data_subview/components/EditTreeConfirmationModal/index.tsx
phoenixAja: thank you!!!

PR #936, opened 2022-02-01
Clarify Public ID usage in metadata upload

vincent-czi:
### Summary:
- **What:** Tweaks to metadata upload flow to clarify how user should interact with Public (GISAID) ID. Also some other fixes.
- **Ticket:** [sc<177067>](https://app.shortcut.com/genepi/story/<177067>); [sc<181309>](https://app.shortcut.com/genepi/story/<181309>); [sc<182546>](https://app.shortcut.com/genepi/story/<182546>); [sc<143531>](https://app.shortcut.com/genepi/story/<143531>)
- **Env:** https://public-id-meta-tweaks-frontend.dev.czgenepi.org

### Demos:

Public ID column has moved and been renamed, toggle for "Previously Submitted to GISAID?" is gone
image

-----

Now has the date of last change to metadata template structure
image

-----

Additional explanation to user they should check current metadata template when they have incorrect TSV/CSV headers for metadata file [This will change down the road once Christine's usage instructions are complete, will become a link out to that then.]
image

### Notes:

A sample can now be both private (private within app, not exposed to "parent" orgs of a group) and have a public ID (submitted to GISAID or similar). Previously, we would auto-correct uploaded TSV data so that a sample could not be both of these at once, but we've changed our mind on how that should work. Since this was the only form of auto-correcting that was occurring, I've removed all the warning components and functions doing the auto-correcting. However, there is another flavor of auto-corrrect that is in upcoming product specifications for CRUD, so I left the warning type in the enum. 

### Checklist:
- [x] I merged latest `trunk`
- [x] I manually verified the change
- [x] I added labels to my PR
- [x] I tested in multiple browsers
- [ ] I added relevant unit tests
~[ ] I have notified others of changes they need to make locally (migrations, jobs, package updates, etc)~ (none needed)

PR #944, opened 2022-02-09
Upload flow changes

vincent-czi:
# Long-lived feature branch
Since there's a bunch of different changes coming in at once to the Upload flow but it's hard to feature-flag off due to the nature of the work, this is a long-lived branch to capture all those changes and make a single PR.

**Should be okay to rubber-stamp approve** this PR once it's ready:
- Initial creation of branch is from the approved PR of #936
- Occasionally gets `trunk` merges to stay up to date with what's current (so below PR diffs are only on material changes)
- Every other commit on it will come from a PR merge where that PR has been separately approved

`==>`As such, when this PR is ready and gets taken out of draft status, it should be okay to just rubber-stamp approval on it and merge it in.

-----

This PR should be complete and ready to merge in now. It basically has two large chunks of work:
- Removing the "Sample is Public" toggle column; clarifying intent of the Public ID column
- Adding a new "Private ID" column for users to specify an incoming samples' `privateId` rather than just relying on it coming from the FASTA sample names.


lvreynoso (approved):
rubber stamp, it's all you

danrlu:
Not a blocker, I just noticed this header on fasta upload page should probably be `Sample Name (from FASTA)`?
image
Thanks again!!

vincent-czi:
Excellent catch, @danrlu! Thanks, I'll make that change then merge it in.

PR #950, opened 2022-02-11
remove depreciated backdrop click prop

phoenixAja:
### Summary:
- **What:** removes
```
Failed prop type: The prop disableBackdropClick of ForwardRef(Dialog) is deprecated. Use the onClose prop with the reason argument to filter the backdropClick events.
```
warnings from the console

- **Ticket:** no ticket
- **Env:** no rdev

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
@@ -194,14 +194,11 @@ export const CreateNSTreeModal = ({
       </Notification>
       <StyledDialog
         disableEscapeKeyDown
+        disableBackdropClick
         open={open}
-        onClose={(_event, reason) => {
-          if (reason !== "backdropClick") {
-            handleClose();
-          }
-        }}
         fullWidth={true}
         maxWidth={"sm"}
+        onClose={handleClose}
       >
         <StyledDialogTitle>
           <StyledIconButton onClick={handleClose}>
--- a/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
@@ -193,10 +193,13 @@ export const CreateNSTreeModal = ({
         </NextLink>
       </Notification>
       <StyledDialog
-        disableBackdropClick
         disableEscapeKeyDown
         open={open}
-        onClose={handleClose}
+        onClose={(_event, reason) => {
+          if (reason !== "backdropClick") {
+            handleClose();
+          }
+        }}
         fullWidth={true}
         maxWidth={"sm"}
       >
--- a/src/frontend/src/components/DialogNoDepreciation/index.tsx
+++ b/src/frontend/src/components/DialogNoDepreciation/index.tsx
@@ -1,0 +1,24 @@
+// import Dialog, { DialogProps as MuiDialogProps } from "@mui/material/Dialog";
+import { Dialog, DialogProps as MuiDialogProps } from "@material-ui/core";
+import React from "react";
+
+interface Props extends MuiDialogProps {
+  disableBackdropClick?: boolean;
+  onClose(): void;
+}
+
+// This component wraps Dialog and handles the backdropClick prop which is now depreciated
+export default function DialogNoDepreciation({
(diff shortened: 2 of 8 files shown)


src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
```diff
@@ -193,10 +193,13 @@ export const CreateNSTreeModal = ({
         </NextLink>
       </Notification>
       <StyledDialog
-        disableBackdropClick
         disableEscapeKeyDown
         open={open}
-        onClose={handleClose}
+        onClose={(_event, reason) => {
```
vincent-czi: I don't love having to repeat the same boilerplate for every `onClose` prop. Especially if this version ever gets deprecated as well. Maybe we could instead create a helper func we stash in some shared util and they can all import it? Something like
```
export function noBackdropCloseHandler(handleClose: () => void) {
  return (_event, reason) => {
    if (reason !== "backdropClick") {
      handleClose();
    }
  };
}
```

Then it would get used as just
```
<Dialog
  ...other props ...
  onClose={noBackdropCloseHandler(handleClose)}
  ...
/>
```

Don't get me wrong, this doesn't feel super elegant either, but at least it abstracts it away to a shared location in code. Also not sure how good a name `noBackdropCloseHandler` is, but I'm straining to come up with a better name that takes no more than five words...

My real preference would be to have our own Dialog component that just handled this for us, but since we build off MUI and each of our dialogs are kinda different, I don't think that's going to be a slam dunk either...

vincent-czi (approved):
I will be so happy to see these warnings go away. I've got a slight preference comment about abstracting this specialty handler away, but I can see arguments against it too. If you disagree, it's fine as is.

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
phoenixAja: i think we could create our own dialog component and that would work.. one sec

src/frontend/src/components/DialogNoDepreciation/index.tsx
```diff
@@ -0,0 +1,23 @@
+import { Dialog, DialogProps as MuiDialogProps } from "@material-ui/core";
+import React from "react";
+
+interface Props extends MuiDialogProps {
+  disableBackdropClick?: boolean;
+  onClose(): void;
+}
+
+// This component wraps Dialog and handles the backdropClick prop which is now depreciated
+export default function DialogNoDepreciation({
```
vincent-czi: I apologize for being a pedant, but there's no `i`: Deprecation
> Failed prop type: The prop `disableBackdropClick` of `ForwardRef(Dialog)` is deprecated

vincent-czi: Regardless of spelling, I think I'd prefer to avoid calling out the deprecation-dodging aspect in the component name. I think `DialogNoDeprecation` sounds kind of mysterious if you aren't already steeped in the fact that we want to avoid this specific deprecation warning. It's not like it couldn't be affected by deprecation again in the future, or itself be deprecated by us in favor of some other, new Dialog (like if SDS makes one).

Instead, I think we can just call it `Dialog`. It's basically intended to be a drop-in replacement for the MUI version, we're just side-stepping the API change they made for one we prefer and that lets us avoid a deprecation we disagree with.

If you think `Dialog` is too generic a name -- or if we have other dialogs running around that would confuse the matter -- I think it would be okay to put something else on it. But I think pointing out the deprecation-dodge aspect is better in a brief comment explaining why it exists than following the component name around. Also that way if the component evolves down the road, the more generic name will be less confusing.

vincent-czi (approved):
Code wise, great, like the easy to work with API and going back to how we experienced the MUI component before.

I would request a change of the component name though, see comments

src/frontend/src/components/DialogNoDepreciation/index.tsx
phoenixAja: yeah i was thinking that we did have a fair amount of other Dialogs around that might make this confusing, really have no good ideas on what we should name..

phoenixAja: i renamed the component to `Dialog` once tests pass i'll merge

PR #955, opened 2022-02-14
fix: Accommodating both dictionary and json formatted tag values

alexlokshin-czi:


scripts/happy
```diff
@@ -59,6 +59,9 @@ class HappyConfig:
             raise CliError(error)
 
     def __getattr__(self, field):
+        print(f'Field: {field}')
```
jgadling: print!

alexlokshin-czi: Arrgh

.happy/config.json
```diff
@@ -3,7 +3,7 @@
     "terraform_version": "0.13.5",
     "default_env": "rdev",
     "app": "genepi",
-    "default_compose_env": ".env.ecr",
+    "default_compose_env_file": ".env.ecr",
```
jgadling, line 6: Why change this parameter name?

jgadling (approved):

.happy/config.json
alexlokshin-czi: default_env (environment name) and default_compose_env (an environment file name) have completely different semantics, we've added some clarity in the golang version.

PR #959, opened 2022-02-17
Specify Private ID during metadata Upload

vincent-czi:
### Summary:
- **What:** Adds way for user to set the Private ID for each sample during the Upload flow.
- **Ticket:** [sc<177066>](https://app.shortcut.com/genepi/story/<177066>), with a new warning box from [sc<181272>](https://app.shortcut.com/genepi/story/<181272>)
- **Env:** https://specify-private-id-frontend.dev.czgenepi.org/

### Demos:
image

During Metadata upload, user now is able to set the Private ID via the new column or they can upload TSV and specify it in that interface instead.

### Notes:
- For the duration of the Upload flow, we track a given sequence/sample with the ID of its name from the FASTA. This is different from the rest of the app where we (generally) track samples by an ID of their privateId.
- For the webform, the Private ID defaults to being the same as what was submitted for the sequence name in the FASTA.
- For the TSV upload, the Private ID defaults to blank and it's required for the user to enter it.
- Private ID follows naming requirements that match the pre-existing requirements for FASTA sequence name: maximum length and only certain characters allowed.

**Important:** **Dan needs to be notified when this gets merged** and told when it will go live. She's going to message to our users who regularly use the TSV Upload flow because this will be a significant change for them. I'm planning to handle this, but if this somehow gets stuck and takes awhile to get merged and I'm OOO, whoever takes charge of this PR needs to let her know.
Also There's a var `TEMPLATE_UPDATED_DATE` that has the expected date this would go live in Prod. Assuming we merge this in the next day or so, it's accurate, but if it gets held up, that needs to be adjusted.

### Checklist:
- [x] I merged latest `trunk`
  - This was a big ol' headache because of the way GitHub shows file diffs. Turns out that GitHub doesn't do a great job of seeing two branches (this working branch and the long-lived feature branch it's going to merge into) that both had the same merge (trunk). So if you look over the commits, you'll see two merges: one of `trunk` and also one of `vincent/upload-flow-changes`. That secondary one brought in no changes, but by doing it, it caused GitHub to display the right file diff for this PR. wooooo
- [x] I added labels to my PR


src/frontend/src/views/Upload/components/Metadata/components/ImportFile/components/Alerts/warnings.tsx
```diff
@@ -158,3 +162,69 @@ export function WarningMissingData({ missingData }: PropsMissingData) {
     />
   );
 }
+
+/**
+ * WARNING_CODE.BAD_FORMAT_DATA
+ */
```
phoenixAja, line 168: confused what your trying to say with this comment

vincent-czi: Oh, each one of the warnings in this file corresponds to one of the WARNING_CODE enums. It shows up when that WARNING_CODE is present in the overall warnings.

phoenixAja (approved):
looks good!!

src/frontend/src/views/Upload/components/Metadata/components/ImportFile/components/Alerts/warnings.tsx
phoenixAja: oh got it, thanks

PR #960, opened 2022-02-17
Frontend changes for split.io

jgadling:
### Summary:
- **What:** Frontend changes 
- **Ticket:** [sc166024](https://app.shortcut.com/genepi/story/166024)
- **Env:** https://splitfe-frontend.dev.czgenepi.org

### Notes:
This builds on top of changes made to the `/v2/users/me` endpoint in the backend PR to configure Split.IO for a particular user.

The split react hook can then be used to fetch treatments (flag values) like so:

```
  import { useTreatments } from '@splitsoftware/splitio-react';
  const MyComponent = (): JSX.Element => {
     const flag = useTreatments(["my_flag_name"]);
     return (
        <div>{flag.my_flag_name.treatment}</div>
     )
   }
```

### Checklist:
- [ ] I added labels to my PR


vincent-czi:
I don't love the API the Split SDK provides via `SplitTreatments`. It feels cumbersome to me and definitely looks different than the rest of our codebase (the whole passing a child that is a function rather than passing a child(ren) that is JSX React Component(s)). That said, it is a kind of a special case since it's how we set up a feature flag and those will be somewhat rare, so that's not a strong objection for me. I can get used to it.

Did you get the chance to try out the [SDK's `useTreatments` hook](https://github.com/splitio/react-client/blob/master/src/useTreatments.ts)? They mention having hooks in their SDK docs near the top, but sadly don't give any examples of using that hook. But if we could get it working, I think that would feel a lot more like the rest of our FE code as opposed to using `SplitTreatments` and it feeling like a special case.

This [comment in their `SplitTreatments`](https://github.com/splitio/react-client/blob/0cdc3a1c72d5b5a640fffed089d1bf57bca6ce7e/src/SplitTreatments.tsx#L48) ("SplitTreatments only accepts a function as a child, not a React Element (JSX)") freaks me out, but I _think_ we can still return JSX just fine in the function, so it's probably not a big deal. Did you get the chance to verify returning components conditionally via your code example? By which I mean,
```
          return isReady
            ? treatments[featureName]["treatment"] // prints the flag value.
            : "waiting...";
```
instead of having the flag value print or "waiting" print, putting in `<ComponentWhenFlagOn>` and `<ComponentWhenFlagOff>`, along with some props being passed to them? I can help you with this if what I'm talking about is not clear. I expect it will work! It's just that comment freaks me out and so I want to double-check.

In any case, I would like to see some documentation of how to use these feature flags either in the codebase or on an internal wiki. It doesn't need to be much, basically just that code snippet you put in your description with maybe a couple sentences of additional explanation. At a glance, their SDK docs don't really look the same as how we have most our components written (their primary example is using class-based React components), so having something that looks like our code and is an easy drop-in seems helpful to me.


Finally -- and I'm really burying the lede here -- I'm worried about the warning [near the top of their docs](https://help.split.io/hc/en-us/articles/360038825091-React-SDK) regarding server-side rendering:
> React SDK is not meant for server-side rendering. It uses the [JavaScript SDK API for browsers](https://help.split.io/hc/en-us/articles/360020448791-JavaScript-SDK). Thus, using it for SSR, with the [ReactDOMServer API](https://reactjs.org/docs/react-dom-server.html) or frameworks such as [Next.js](https://nextjs.org/) or [Gatsby](https://www.gatsbyjs.org/), might lead to unexpected behaviours. You can use [Node.js SDK](https://help.split.io/hc/en-us/articles/360020564931-Node-js-SDK) instead.

I don't really have a great grasp on how we're currently using SSR, but I do know we use the Next framework. I think we actually make very little use of the library or the SSR aspect, but it's quite possible it sets up defaults that will cause problems with this library. It's also possible everything would be totally fine? I really don't know without looking into it more. My gut says we're probably okay, but this would be a bad thing to be wrong about.  Maybe we could talk to Timmy, see if he has a better idea of what the problematic interactions could be?

src/frontend/pages/_app.tsx
```diff
@@ -27,10 +29,39 @@ function Nav(): JSX.Element {
   }
 }
 
+interface Props {
+  children: React.ReactElement;
+}
+const SplitInitializer = ({ children }: Props): JSX.Element | null => {
```
vincent-czi: Regarding my other comment about putting in some docs explaining how to use Split/feature flags as a dev, I think this component could be broken out into a separate file, then those docs could live there. Maybe moved to `frontend/src/common/featureFlags` and then any Split-based components or utils involving feature flags can get dumped in that dir.

.happy/terraform/modules/service/variables.tf
```diff
@@ -1,135 +1,139 @@
-variable stack_resource_prefix {
+variable "stack_resource_prefix" {
```
vincent-czi, line 1: Just curious, what changed and made all the variables start needing quotes? Is it just TF changed how configs are done?

src/frontend/src/common/types/user.d.ts
```diff
@@ -20,3 +20,16 @@ interface User {
   agreedToTos: boolean;
   acknowledgedPolicyVersion: string | null; // Date or null in DB. ISO 8601: "YYYY-MM-DD"
 }
+
+interface V2User {
```
vincent-czi, line 25: Maybe throw a "TODO -- Remove old `User` interface when fully converted to v2 API for user data" comment somewhere in this file?

vincent-czi (commented):
Overall, I'm happy with the changes here. If you haven't already tried out the `useTreatments` hook though, I'd like see that before merging this in. Also we should look into whether the Next/SSR aspect will actually cause issues or if we're safe because of how we use them.

.happy/terraform/modules/service/variables.tf
jgadling: The quotes aren't strictly required, but it's a lint thing.  I ran `terraform fmt` in this directory and terraform made everything "pretty"

jgadling:
Alright, there's a lot here, I'm breaking it down as best I can

> I don't love the API the Split SDK provides via `SplitTreatments`. 

Agreed. It's awful. We shouldn't use it.

> Did you get the chance to try out the [SDK's `useTreatments` hook](https://github.com/splitio/react-client/blob/master/src/useTreatments.ts)? They mention having hooks in their SDK docs near the top, but sadly don't give any examples of using that hook. But if we could get it working, I think that would feel a lot more like the rest of our FE code as opposed to using `SplitTreatments` and it feeling like a special case.

I didn't see this hook at all before, thanks for the pointer! I've tested it out and included a note in the code as to how to use it. Let's agree to use the hook by default. Docs are coming next week.
 
> Finally -- and I'm really burying the lede here -- I'm worried about the warning [near the top of their docs](https://help.split.io/hc/en-us/articles/360038825091-React-SDK) regarding server-side rendering:

Yeah, that note worried me as well, enough that I burned a fair amount of time on trying to integrate the Node.js client. In the end though, we have a 99.9% client-side app and I think the React client is the best option for us. I haven't seen any  strange behaviors in my (admittedly, limited) testing so far and Split support says that warning only applies if we're using server-side rendering (and we're not using server side rendering)

vincent-czi (approved):
Great! If you're reasonably confident about the SSR stuff not being an issue, then I think we'll be fine. It seemed unlikely to cause problems to me, but their warning was kind of vague about why it's dangerous so I was just worried in a knee-jerk way.

PR #963, opened 2022-02-18
[bug] Convert all tree sample ID's (or none)

jgadling:
### Summary:
- **What:** Update tree downloads and tree metadata downloads to convert *all* of a group's public sample ID's to private ID's instead of only converting selected sample id's.
- **Ticket:** [sc185836](https://app.shortcut.com/genepi/story/185836)
- **Env:** [`<rdev link>`](https://treeids-frontend.dev.czgenepi.org/)

### Notes:
This PR actually consists of 3 main changes:
- The unit tests for the metadata endpoint were horribly broken (my fault) and I fixed them.
- Update the endpoints to convert all ID's, per the shortcut ticket
- All of the "get me tree info" endpoints have been updated to accept a query string parameter called "id_style" with possible values "public" or "private" so they will optionally return public or private sample ID's

Example links to try downloading public/private identifiers in metadata and trees:
https://treeids-backend.dev.czgenepi.org/api/phylo_tree/21972?id_style=public
https://treeids-backend.dev.czgenepi.org/api/phylo_tree/sample_ids/21972?id_style=public

### Checklist:
- [x] I added labels to my PR


src/backend/aspen/app/views/phylo_trees.py
```diff
@@ -163,6 +140,37 @@ def sample_filter(sample: Sample):
     )
     json_data = json.loads(data)
 
+    if id_style == "public":
+        return json_data
```
jgadling, line 153: If the user only wanted public ID's, we can just return the raw tree data and skip the rest of the code in this function that converts ID's.

danrlu (approved):
I read the code as much as I could XD it looks like this is all the backend changes needed to support both public ID converted and non-converted trees?? 🤩 thanks very much!!

jgadling:
> 

Yep, this includes code to support sending back either public or private ID's.

PR #976, opened 2022-03-01
add security txt instructions for security researchers to report issu…

mmmmmmaya:
### Summary
- **What:** Add a `security.txt` to the website, following instructions from https://securitytxt.org/
- **Why:** So security researchers can report bugs easily and find jobs
- **Ticket:** [[sc-137545]](https://app.shortcut.com/genepi/story/137545)

### Demos
Screen Shot 2022-03-01 at 9 50 56 AM
Screen Shot 2022-03-01 at 9 51 07 AM

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/tests/staticFiles/security.test.ts
+++ b/src/frontend/tests/staticFiles/security.test.ts
@@ -1,8 +1,11 @@
 import { readFileSync } from "fs";
 
-const checkExpirationDate = (path) => {
+const checkExpirationDate = (path: string) => {
   const text = readFileSync(path).toString();
-  const expirationDateStr = text.match(/Expires: .*/g)[0].slice(9);
+  const expirationDateStr = text?.match(/Expires: .*/g)?.[0]?.slice(9);
+
+  if (!expirationDateStr) throw Error("security.txt missing expiry date");
+
   const expirationDate = new Date(expirationDateStr);
   const now = new Date();
 
@@ -17,4 +20,11 @@ test("ensure second security.txt file has not expired", () => {
   expect(() =>
     checkExpirationDate("public/.well-known/security.txt")
   ).not.toThrow();
+});
+
+test("ensure security files in sync", () => {
+  const file = readFileSync("public/security.txt").toString();
+  const file2 = readFileSync("public/.well-known/security.txt").toString();
+
+  expect(file == file2).toBeTruthy();
 });
--- a/src/frontend/tests/staticFiles/security.test.ts
+++ b/src/frontend/tests/staticFiles/security.test.ts
@@ -1,0 +1,20 @@
+import { readFileSync } from "fs";
+
+const checkExpirationDate = (path) => {
+  const text = readFileSync(path).toString();
+  const expirationDateStr = text.match(/Expires: .*/g)[0].slice(9);
+  const expirationDate = new Date(expirationDateStr);
+  const now = new Date();
+
+  if (expirationDate < now) throw Error("security.txt files need updating");
+};
+
+test("ensure first security.txt file has not expired", () => {
+  expect(() => checkExpirationDate("public/security.txt")).not.toThrow();
+});
+
+test("ensure second security.txt file has not expired", () => {
+  expect(() =>
+    checkExpirationDate("public/.well-known/security.txt")
+  ).not.toThrow();
+});
--- a/src/frontend/tests/staticFiles/security.test.ts
+++ b/src/frontend/tests/staticFiles/security.test.ts
@@ -12,7 +12,7 @@ const checkExpirationDate = (path: string) => {
   if (expirationDate < now) throw Error("security.txt files need updating");
 };
 
-test("ensure second security.txt file has not expired", () => {
+test("ensure security.txt file has not expired", () => {
   expect(() =>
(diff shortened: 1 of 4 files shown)


vincent-czi (commented):
I don't love having two copies of the same info, worried we'll remember to update one and forget about the other. That said, I tried flipping through some Next docs and googling for 15min about how to serve a static file with a different path and I didn't find anything, so... probably not worth a ton of effort? Maybe poke at it for 30 minutes then let it be?

kkusznir:
Looks good to me & thanks for getting this one done!
We may suggest some updates later as we update some of our VDP / Bounty guidance, but I'm thrilled to see you guys take this up.  You are officially the first team here to publish a security.txt!

@vincent-czi 
What about a simple test automation to check each file for expected content?  If you have a framework that you already use for automated tests this could be a quick item to add.

kkusznir (approved):

mmmmmmaya:
I do like the approach of using tests to make sure both are within the expiry window. it's a little bit more of a shove in case we miss the slackbot reminder to do this chore.

vincent-czi:
I like the failing test if they get out of date, but I'm not a fan of using tests to ensure that they maintain the same content / are actually identical files, it feels like a workaround to not being able to serve the same file from two routes.

@kkusznir, how important is it to be present under the `/.well-known/security.txt` route? Would it be okay to just have a single file live under the root (`/security.txt`), then have the "Canonical" field point at that URI? The info on ["Canonical" here](https://datatracker.ietf.org/doc/html/draft-foudil-securitytxt-12#section-3.5.2) makes it sound like it's common to live under the `.well-known` route, but not required? I'm used to `robots.txt` and `humans.txt` living under the root, so it seems reasonable to me for `security.txt` to be there as well, but maybe it would be really unusual or unexpected in a security context? What do you think?

If it's okay to just have it at root, let's drop the `/.well-known` version and then we just have one copy to update. Still keep the expiration unit test regardless, it's valuable no matter what.

If we really need the `/.well-known` route, could we get away with dropping the top-level one? Don't love it, but it's not like we have a `robots.txt` or anything right now.

And if we need both, then use a test to ensure they match.

kkusznir:
@vincent-czi & @mmmmmmaya 

Ultimately It's not **_critical_** that it goes in one directory or another;  It's more a matter of discoverability so it's about making it consistent and easy to find.  

Both are common. If you look around (https://www.google.com/search?q=inurl%3A+security.txt) you'll see some places use one or the other and often both.
Google => https://www.google.com/.well-known/security.txt
FB => https://www.facebook.com/.well-known/security.txt AND https://www.facebook.com/security.txt

From https://securitytxt.org/ :
> **Where should I put the security.txt file?**
> For websites, the security.txt file should be placed under the /.well-known/ path (/.well-known/security.txt) [[RFC8615](https://tools.ietf.org/html/rfc8615)]. It can also be placed in the root directory (/security.txt) of a website, especially if the /.well-known/ directory cannot be used for technical reasons, or simply as a fallback. The file can be placed in both locations of a website at the same time.

Security researchers aka hackers are typically familiar with both.   The spec evangelizes use of /.well-known but the authors and others acknowledge that it may not make sense and the root is ok, too.  If you ask my preference, I would prefer both but we have not yet set a standard at CZI.  I would like us to move towards a standard and intend to evangelize that this year.  It's important then that we try to be consistent, for ease and predictability.  Having said that, I would find the root, /.well-known, or both as a welcome improvement to where we are today.  If you feel strongly, I'd be happy to start where you're comfortable.  We can always adjust this later if we'd like to or need to.

vincent-czi:
Yeah, I would definitely prefer both places. This is all just because of an implementation annoyance with using Next's public folder that means we have to keep identical files in two places. My argument for just having one file comes down purely to maintainability/elegance, but I think practicality should win out anytime it really matters, so... I don't actually feel strongly?

A test that ensures the two files always have the same content will keep us safe, it's just too bad that we need such a test, but oh well.

vincent-czi:
I just realized, what about a redirect? What do you think of having `/.security.txt` redirect over to the `.well-known` path, then we only have one file in our public folder under `.well-known`?

https://nextjs.org/docs/api-reference/next.config.js/redirects

kkusznir:
That's a nice simple solution that works for me. I like redirecting to the /well-known directory version.
If I'm thinking about it right, I think it might affect indexing a little but the destination should still be indexed.

mmmmmmaya:
I like this solution best, thanks for the suggestion @vincent-czi

src/frontend/tests/staticFiles/security.test.ts
```diff
@@ -12,19 +12,8 @@ const checkExpirationDate = (path: string) => {
   if (expirationDate < now) throw Error("security.txt files need updating");
 };
 
-test("ensure first security.txt file has not expired", () => {
-  expect(() => checkExpirationDate("public/security.txt")).not.toThrow();
-});
-
 test("ensure second security.txt file has not expired", () => {
```
vincent-czi: so ridiculously minor, but should remove the word "second" now

vincent-czi (approved):

PR #978, opened 2022-03-01
chore: Use chanzuckerberg/github-actions/docker-build-push in Push Remote Dev action

edulop91:
### Summary:
- **What:** Starting to look into integrating the new happy cli with GitHub actions. As part of this, looking to release a set of common chanzuckerberg/github-actions. Chose the docker step first because it is fairly self-contained and common across many repos. Will follow up with Happy specific actions later.

See https://github.com/chanzuckerberg/github-actions/pull/17/ for corresponding code changes.

- **Ticket:** NA
- **Env:** `<rdev link>`

### Checklist:
- [ ] I added labels to my PR


edulop91:
hmm think i'm missing some build keys

.github/workflows/push-rdev.yml
```diff
@@ -1,9 +1,9 @@
 name: Push Remote Dev
 
 on:
-  push:
+  pull_request:
```
jgadling, line 4: I think we specifically want to use the `push` event here

jgadling (changes requested):
This looks good aside from the event hook!

edulop91:
see https://github.com/chanzuckerberg/czgenepi/pull/990 instead for the push event instead of pull_request

PR #979, opened 2022-03-02
Update Makefile to use docker's built-in compose v2

jgadling:
### Summary:
- **What:** Update our makefile to always use the docker CLI's built-in `compose` command

### Notes:
Docker desktop has a checkbox that allows us to switch between using compose v1 or v2 when we run the `docker-compose` command, however when we run `docker compose` we're always using compose v2.

These changes switch our Makefile to always use compose v2 via `docker compose` in alignment with our new `happy` cli. We also needed to update a few commands to make use of the profiles we have configured in `docker_compose.yml`. By default, any `docker compose up/down` commands will run any containers that *do not* have a profile specified. Since all of the containers in our `docker-compose.yml` file *do* have profiles defined, we need to specify which profiles (and * means everything) to use when bringing compose services up and down.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ SHELL := /bin/bash
 ### DOCKER ENVIRONMENTAL VARS #################################################
 export DOCKER_BUILDKIT:=1
 export COMPOSE_DOCKER_CLI_BUILD:=1
-export docker_compose:=docker-compose --env-file .env.ecr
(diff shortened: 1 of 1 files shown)


lvreynoso (approved):

mmmmmmaya:
Thank you!!! ⭐

PR #991, opened 2022-03-15
Migrate Sample create endpoint to FastAPI

jgadling:
### Summary:
- **What:** Migrate `samples/create` to fastapi
- **Ticket:** [sc183168](https://app.shortcut.com/genepi/story/183168)
- **Env:** https://samplecreate-frontend.dev.czgenepi.org

### Notes:
This is pretty much a plain transliteration of the old API to SA 2.0 query syntax w/async.

### Checklist:
- [x] I added labels to my PR


src/backend/aspen/api/schemas/samples.py
```diff
@@ -67,7 +67,7 @@ def get(self, key: Any, default: Any = None) -> Any:
         return default_response
 
 
-class SampleResponseSchema(BaseResponse):
+class SampleResponse(BaseResponse):
```
jgadling, line 72: Putting Schema in the name of classes imported from a `schema` module is starting to feel redundant.

src/backend/aspen/api/settings.py
```diff
@@ -120,6 +121,7 @@ class Settings(BaseSettings):
 
     # Env vars usually pulled from AWS SSM Parameters
     AWS_NEXTSTRAIN_SFN_PARAMETERS: Dict
+    AWS_PANGOLIN_SFN_PARAMETERS: Dict
```
jgadling, line 124: FastAPI didn't start pangolin jobs before, so it didn't need a config option for it.

src/backend/aspen/database/models/sample.py
```diff
@@ -253,6 +253,22 @@ class Sample(idbase, DictMixin):  # type: ignore
     sequencing_reads_collection: Optional[SequencingReadsCollection]
     uploaded_pathogen_genome: Optional[UploadedPathogenGenome]
 
+    def generate_public_identifier(self):
```
jgadling, line 256: This is one of the more "interesting" bits of code in this PR. If someone doesn't send us a public identifier, we generate that identifier via invoking SQL functions, like so:
```INSERT INTO SAMPLES (..., public_identifier, ...) VALUES (...,  CONCAT("hCov-19/USA/CADPH-", currval('aspen.samples_id_seq'), "/2022")```

We do this so we don't have to pre-generate DB identifiers outside a transaction, with nothing guarantees that they'll match the row ID's, like the previous code did.

File: src/backend/scripts/setup_dev_data.sh [hunk 4]
```diff
@@ -63,6 +63,23 @@ echo "Creating SSM Parameters"
...
+      "remote_dev_prefix":""}
+    },
+  "OutputPrefix":"s3://genepi-batch/pangolin-ondemand-sfn/results",
+  "RUN_WDL_URI":"s3://genepi-batch/pangolin-ondemand.wdl-v0.0.1.wdl",
+  "RunEC2Memory":64000,
+  "RunEC2Vcpu":10,
+  "RunSPOTMemory":64000,
+  "RunSPOTVcpu":10,
+  "StateMachineArn":'${LOCAL_SFN_ARN}'
+}'
```
jgadling, line 82: Add pangolin config info to local dev.

src/backend/aspen/api/schemas/samples.py
```diff
@@ -118,9 +118,52 @@ class UpdateSamplesRequest(BaseRequest):
...
+    public_identifier: Optional[str]
+    sample_collected_by: Optional[str]
+    sample_collector_contact_email: Optional[str]
+    sample_collector_contact_address: Optional[str]
+    authors: Optional[str]
+    organism: Optional[str]
+    host: Optional[str]
+    purpose_of_sampling: Optional[str]
+    specimen_processing: Optional[str]
+    czb_failed_genome_recovery: Optional[bool]
```
lvreynoso: I think it might be good to note which fields are currently collected on the frontend and which are not, though that might be a pain to keep in sync.

lvreynoso (approved):

PR #994, opened 2022-03-17
Get Pango Lineages

vincent-czi:
### Summary:
- **What:** Pipeline for pulling Pango lineages and stuffing them into the DB.
- **Ticket:** [sc<184612>](https://app.shortcut.com/genepi/story/<184612>) (part of work, still need to make a BE endpoint to use the data in DB now)
- **Env:** `vince-pango-lineages` -- customized rdev where I commented out the `pangolin_workflow` call from pangolin.wdl so it would only run the `pango_lineages_loading` call to be able to do a quick test and verify it was working. 

### Demos:

From `vince-pango-lineages` after running the new pipeline:
image

Above is as expected.

### Notes:

This is hitching a ride on the general `pangolin.wdl` job since it's kinda-sorta related and also a good thing to have run nightly. They really do different things and have virtually no relationship though, so I'd also understand not wanting to do this. The two big downsides I can see are: (1) Mixed up logging [I think? I don't really know enough here, but I think miniwdl will run both task calls simultaneously since there's no dependency, so the logs would get interspersed, right?] which might make sifting through cloudwatch logs when something goes wrong a pain; (2) A failure in one breaks the other, b/c miniwdl kills all tasks if any fails [I did some local testing to check the behavior and that's what I found], but the outcome of one task really shouldn't impact the other task.

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR
~I tested in multiple browsers~ (not browser based)


src/backend/aspen/workflows/import_pango_lineages/import.py
```diff
@@ -0,0 +1,198 @@
...
+    This function takes that file and parses out list of lineages.
+    """
+    # First line is header, no lineage data, but we use as a safety check.
+    first_line = next(lineage_notes_file)
+    safety_check_first_line(first_line)
+    # Now we parse remainder of lines from file and extract lineages
+    return [
+        lineage
+        for line in lineage_notes_file
+        if (lineage := extract_lineage_from_line(line)) is not None
```
vincent-czi: This is the first time I have ever used a walrus operator. I'll be honest, I was only vaguely aware of their existence before, but I learned about it because I was getting a mypy error from my original version I wrote that was like this:
```
[
extract_lineage_from_line(line)
for line in lineage_notes_file
if extract_lineage_from_line(line)
]
```
Basically the guarantees from the `if` conditional were clear to me the dev, but not really usable by mypy. But I actually like the walrus operator way of writing it! It feels remarkably readable to me while also solving the mypy problem I was having. (also yeah, I totally could have done this with `[].append` stuff, but list comprehensions are so pythonnnnnniiiiiccc)

If you were like me and had not previously interacted with them before, here's a good writeup of how I'm using it
https://realpython.com/python-walrus-operator/#list-comprehensions

src/backend/aspen/workflows/import_pango_lineages/import.py
```diff
@@ -0,0 +1,198 @@
...
+@click.command()
+@click.option(
+    "pango_lineages_file",
+    "--lineages-file",
+    type=click.File("r"),
+    required=True,
+    help="Pango lineages file to parse and import.",
+)
+@click.option(
+    "--parse-without-import",
```
vincent-czi: I know all the other workflow Python scripts have a `test` flag for doing smoke tests that goes like this:
```
@click.option("--test", type=bool, is_flag=True)

... later in main code block ...

    if test:
        print("Success!")
        return  # end script early

```

Should I be doing it as well? Or since there's another testing approach via `--parse-without-import`, it's fine without the smoke test version?

src/backend/aspen/workflows/import_pango_lineages/import.py
jgadling: Thanks for the writeup! I'm also only vaguely aware of weird python operators.

src/backend/aspen/workflows/import_pango_lineages/import.py
```diff
@@ -0,0 +1,198 @@
...
+        lineage
+        for line in lineage_notes_file
+        if (lineage := extract_lineage_from_line(line)) is not None
+    ]
+
+
+def create_temp_table(session, source_table):
+    """Creates a new table that is structured same as source.
+    Intended to be temporary, should drop by end of this script.
+    copy-pasta from import_gisaid/save.py"""
```
jgadling: &lt; insert nitpicky whining about not making a shared module />

src/backend/aspen/workflows/import_pango_lineages/import.py
jgadling: This is fine. Basically we just have those `--test` flags to do the barest-minimum spot checks of our scripts that don't have better tests (😢) in CI - do they compile? Do all their imports work? So it might be nice to make sure this is wired into the `make check-images` target in the root `Makefile`

src/backend/database_migrations/versions/20220315_205645_create_pango_lineages_table.py
```diff
@@ -0,0 +1,33 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "20220315_205645"
+down_revision = "20220215_194838"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
jgadling: Remove these auto-generated comments pls!

jgadling (approved):
Approved with nits:
- please remove the goofy comments in the migration file
- I'd really like to see the new script tacked into our `check-images` target (or write real unit tests!)
- Shared module for table imports is nice-to-have, but we can refactor it later if need be.

src/backend/aspen/workflows/import_pango_lineages/import.py
vincent-czi: You know, I actually thought the workflows dir was outside the scope of the installed aspen package, but I see I was wrong, so this is way easier than I had originally thought.

Also, I have now discovered that naming your python file `import.py` is a bad idea if you want to import from it! 😅 

Yeah, I'll take a bit of time to move stuff around and create a `shared_utils` folder in workflows, totally agree that's a better way!

jgadling (approved):

PR #999, opened 2022-03-29
Make sure we build/run x86 images everywhere

jgadling:
### Summary:
- **What:** Force `docker compose` and `happy` to build/run amd64 images

### Notes:
I recently switched to an ARM based mac, and the `happy` cli (and my local dev environment) were trying to build/run arm64 images by default. I looked into a few different ways to fix this and it turns out that I think all of our build/push scripts except for our GitHub actions eventually use `docker compose build` and recent versions of `docker compose` support a `platform` parameter. I think this should make our dev/prod architectures consistent.

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,6 +4,7 @@ services:
   database:
     image: "${DOCKER_REPO}genepi-devdb:latest"
     profiles: [ "backend", "web", "all" ]
+    platform: linux/amd64
     ports:
       - "5432:5432"
     environment:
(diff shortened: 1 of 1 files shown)


edulop91 (approved):
thank you!

lvreynoso (approved):
good catch

alexlokshin-czi (approved):

vincent-czi (approved):
I doubt it will change how anything is actually working, but I wonder if we should add a `--platform` arg to the `docker buildx build` command in GitHub actions, just to be super explicit. I don't really have a great grasp on how our build process works yet, but that's the moment (in `build-push-images`) that actually generates the images being deployed to Staging and Prod, right? I expect it will make no difference -- since I'm all but certain that's what the GitHub action is running in -- but if we need to have it explicit in one place, being explicit everywhere just seems like a reasonable policy.

edulop91:
> 

I can take this on in the centralized docker-build-push github actions we're writing

PR #1000, opened 2022-03-29
Sample edit tsv download

phoenixAja:
### Summary:
- **What:** Allow user to download sample edit tsv template, also moved some similar code with the upload tsv to a new place that was more central (src/components/DownloadMetadataTemplate)
- **Ticket:** [sc189667](https://app.shortcut.com/genepi/story/189667)
- **Env:** https://tsv-download-frontend.dev.czgenepi.org/data/samples

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/frontend/src/components/CollapsibleInstructions/index.tsx
```diff
@@ -28,6 +30,7 @@ interface Props {
 
 const CollapsibleInstructions = ({
   buttonSize = "xxxs",
+  downloadTSVOption,
```
mmmmmmaya: I feel weird about this. The collapsible instructions component is generic and shouldn't have knowledge of other components or app functionality. At the very least, this should be given a more generic name, maybe something like `additionalHeaderLink`.

src/frontend/src/components/CollapsibleInstructions/style.ts
```diff
@@ -117,3 +117,12 @@ export const InstructionsWrapper = styled("div", {
     `;
   }}
 `;
+
+export const Divider = styled.span`
+  ${(props) => {
+    const colors = getColors(props);
+    return `
+      color: ${colors?.gray[400]}
```
mmmmmmaya:
```suggestion
      color: ${colors?.gray[400]};
```

src/frontend/src/components/DownloadMetadataTemplate/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+  headers,
+  instructions,
+  className,
+}: SampleEditTsvProps): JSX.Element => {
+  return (
+    <CSVLink
+      className={className}
+      filename="metadata_template.tsv"
+      separator={TSV_SEPARATOR}
+      data={[...instructions, [...headers], ...rows]}
```
mmmmmmaya:
```suggestion
      data={[...instructions, headers, ...rows]}
```

I think the above suggestion is correct. If headers is an array, and then you put it inside an array only to spread it then you just end up with the original array again, right?

src/frontend/src/components/CollapsibleInstructions/index.tsx
phoenixAja: is it ok if i change it to `additionalHeaderLink` or would you rather i remove it altogether?

phoenixAja: i'll move it to it's own thing

src/frontend/src/components/DownloadMetadataTemplate/prepMetadataTemplate.ts
```diff
@@ -0,0 +1,189 @@
...
+    "USA/CA-CZB-0001/2021", // publicId
+    DATE_FORMAT, // collectionDate
+    "North America/USA/California/San Francisco County", // collectionLocation
+    DATE_FORMAT, // sequencingDate
+    "No", // keepPrivate
+  ],
+];
+
+const SAMPLE_EDIT_INSTRUCTIONS = [
+  ["# Only fill out columns or cells you want to update the content"],
```
mmmmmmaya: this isn't valid english but maybe it's fine

src/frontend/src/components/DownloadMetadataTemplate/prepMetadataTemplate.ts
```diff
@@ -0,0 +1,189 @@
...
+  ["# Empty cells will not change what is currently in CZ GEN EPI database"],
+  ["# Fill cells with 'Delete' if you want to remove the existing content"],
+  ["# Save in .tsv .csv or .txt format"],
```
mmmmmmaya, line 85: there should be commas here

src/frontend/src/components/CollapsibleInstructions/index.tsx
mmmmmmaya: I think renaming is fine for now

src/frontend/src/components/DownloadMetadataTemplate/prepMetadataTemplate.ts
phoenixAja: this was from the copy provided, but i can fix

src/frontend/src/components/CollapsibleInstructions/index.tsx
phoenixAja: thanks, after this i'll leave `CollapsibleInstructions` alone 😆

src/frontend/src/components/DownloadMetadataTemplate/prepMetadataTemplate.ts
phoenixAja: taken from the copy again, i think the goal was to not have commas in case users want to import the tsv into csv.

vincent-czi: Yeah, the need for lack of commas was something I flagged. If they use it as a CSV and there's commas anywhere in the line boom boom. Similarly can't have tabs in the text, but we're not lunatics, so that one was never going to be an issue.

mmmmmmaya (approved):

PR #1001, opened 2022-04-01
move hero image into its own file

mmmmmmaya:
### Summary
- **What:** Move home page hero svg into its own file.
- **Why:** Rebrand cleanup
- **Ticket:** [[sc-187987]](https://app.shortcut.com/genepi/story/187987)
- **Env:** https://heroimg-frontend.dev.czgenepi.org/

### Demos
No visual changes.

### Notes
The reason this change is so complicated is because we use `svgr` to bring svg into react, and that loader makes use of the default `svgo` [config](https://github.com/svg/svgo#built-in-plugins). The default has two configuration options that were getting in the way of what we were trying to do (just load the SVG directly instead of having it inline). 

First, it tries to be smart and minify/otherwise shrink down the size of your files for you. One way it does this is by [removing hidden elements](https://github.com/svg/svgo/blob/main/plugins/removeHiddenElems.js#L51) (defined in a variety of ways, including `opacity: 0;`). Since the main groups for the animation were all set to `opacity: 0;` by default, webpack decided to remove those groups altogether before bundling the image.

Second, since you can have multiple svg on a page, svgo tries to modify every #id on every svg to something more specific than what is actually defined in the file. It normally does this by [pre-pending some part of the path or file name](https://github.com/svg/svgo/blob/main/plugins/prefixIds.js) to the specified id. This got in the way of styles being applied appropriately since the styling relied on the ids, which webpack modified.

I did not want to disable these rules wholesale because they have obvious value in the majority of our use cases (read: all of them besides this one, so far). So I added an extra rule which only applies to this hero file to modify the way those plugins work.

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR


vincent-czi (approved):
Nice work! Great sleuthing, and great explanation of what's going on!

PR #1003, opened 2022-04-01
Endpoint for pulling Pango lineages

vincent-czi:
### Summary:
- **What:** Adds backend endpoint for pulling list of all Pango lineages
- **Ticket:** [sc<184612>](https://app.shortcut.com/genepi/story/<184612>)
- **Env:** no rdev currently, need to add lineages to DB "seed" for both rdev and local still

### Demos:
image

### Notes:
- I opted to just pull the lineage names and return that as a simple list of strings within the response JSON. Normally we would hold on to the `id` info, but because the id PK is just ephemeral and can change with every workflow run, I think it's better to only return the "real" data of the lineage names.
- Kinda silly, but I realized I'd accidentally named the DB model `PangoLineages` with a plural **s** previously. And while our tables use the plural generally, the models are singular, and the grammar of it bugged me. So I changed to `PangoLineage` throughout. I had a tiny worry that this might somehow screw up the alembic auto-generation, but I verified that the model name change had no impact on auto-generation.
- I think this route would be a good candidate for a publicly available API (in case anyone else in the world needs a way to pull current lineage list). I discussed this with Jess and we both think it might be nice to eventually offer some of our public data in such a manner, but we haven't really prepped for that yet, so it's behind auth (via `Depends(get_auth_user)`) like the rest of the routes for now.

### Checklist:
- [x] I added labels to my PR
    - ^^^ still need to make modifications so the auto-loaded DBs for local and rdev have the new lineage data present


src/backend/aspen/api/schemas/pango_lineages.py
```diff
@@ -0,0 +1,5 @@
+from aspen.api.schemas.base import BaseResponse
+
+
+class PangoLineagesResponse(BaseResponse):
+    all_pango_lineages: list[str]
```
jgadling: can we rename this to `lineages` or something? I'm thinking this endpoint might support filtering at some point, and then the `all_pango_lineages` key name starts to look strange.

src/backend/aspen/api/views/pango_lineages.py
```diff
@@ -0,0 +1,24 @@
...
+
+    Note that the returned result is very simple: pretty  much just a list of
+    strings. While we could return `id` information, the ids (at least with the
+    current data workflow) are ephemeral and have no connection to the "real"
+    lineage info. The only real info we have is the names of the lineages,
+    so that's all we pull and return.
+    """
+    all_lineages_query = sa.select(PangoLineage.lineage)  # type: ignore
+    result = await db.execute(all_lineages_query)
+    return {"all_pango_lineages": result.scalars().all()}
```
jgadling: This should return a response schema rather than a raw object -- see https://github.com/chanzuckerberg/czgenepi/blob/trunk/src/backend/aspen/api/views/locations.py#L30

src/backend/aspen/api/views/pango_lineages.py
```diff
@@ -0,0 +1,24 @@
...
+
+from aspen.api.deps import get_db
+from aspen.api.schemas.pango_lineages import PangoLineagesResponse
+from aspen.database.models import PangoLineage
+
+router = APIRouter()
+
+
+@router.get("/", response_model=PangoLineagesResponse)
+async def list_pango_lineages(db: AsyncSession = Depends(get_db)):
```
jgadling: We probably want to include the output type here.

jgadling (changes requested):

src/backend/aspen/api/views/pango_lineages.py
vincent-czi: I thought that was accomplished via the `response_model=PangoLineagesResponse` in the decorator? When I was working on it, it definitely was using the model and would throw an exception if I tried to return the wrong structure.

src/backend/aspen/api/schemas/pango_lineages.py
vincent-czi: Yeah, I see your point. I wonder if we should go even farther and change the routing to be more like `/lineages/pango/` because we likely want to support other lineages (like WHO) down the road? Or maybe just `/lineages/` and we would eventually differentiate on more params in the request?

jgadling: I like `/lineages/pango/` actually! I don't expect that it would make sense to support returning lineages from different databases all mixed together, I think it's more likely that they'd have their own endpoints

src/backend/aspen/api/views/pango_lineages.py
jgadling: huh, this is interesting -- We probably need to figure out what the best practice here is, and then work on standardizing our results!

src/backend/aspen/api/views/pango_lineages.py
vincent-czi: Yeah, honestly I skipped over that because of the interaction between `response_model` and the view's brief return before the automagic alley-oop to Pydantic seemed weird to document. That would definitely be a benefit of explicitly using the schema response in the code rather than relying on `response_model`.

src/backend/aspen/api/schemas/pango_lineages.py
vincent-czi: I'm not sure. I expect the most common use-case will be: "Please give me all Pangos, all WHOs, and a mapping that shows how to get WHO<--> Pango". One route that returns that object all at once seems nice, although minorly overloaded.

jgadling: yeah, that sounds like a later-us problem right now?

src/backend/aspen/api/views/pango_lineages.py
vincent-czi: I already talked about this with Jess, but for anybody else following these comments: the `response_model` does execute on what gets returned from the view function. The downside is that you get screwy output typing if you try to type annotate the view func because of the interaction with the decorator. The unwrapped view function here returns `dict[str, Any]`, but once the router decorator with `response_model` has worked its magic, it's returning a `PangoLineagesResponse`.

It would be possible to drop the `response_model` and instead just do the schema in the view func code
```return PangoLineagesResponse.parse_obj({"lineages": all_lineages})```
or something similar, but the downside to that is FastAPI won't use the response schema in its automatically generated docs.

Another option would be to do the above, but _also_ put in the `response_model`. But -- after a bit of experimentation -- it looks like that just means FastAPI winds up init-ing the output schema twice. Once when we directly use the schema in the view func code, then again when FastAPI is doing the automatic processing via `response_model`. I was hoping it would see that the type matched and just directly use the schema object that was returned, but it seems to run twice.

Finally, I know FastAPI does a lot of implicit stuff based on types (eg, a view func arg like this `view_func(incoming: MyModel)` causes FastAPI to parse the request data view MyModel, then expose that as the arg `incoming`, so I was hoping there might be something similar with doing a type annotation of the output from `list_pango_lineages` in connection to the model, avoiding the double init, but I didn't see anything. As far as I can tell, FastAPI ignores the view func's output type.

Since the automatic API docs are nice and we don't really consume the view func's returned value in Python code, it seems like it's slightly better to use `response_model` and have the view func's return be a dict that matches to what the model uses.

jgadling (approved):

PR #1005, opened 2022-04-04
Fix local db snapshot creation tools.

jgadling:
### Summary:
- **What:** Fix the `make` targets we use to generate local db snapshots

# Important! #
- Once this PR is merged, everyone should run `make local-init` to repopulate their local databases with sample data!
- Also, to use these targets, your repo checkout directory should be named `czgenepi`

### Notes:
@vincent-czi discovered that our local db snapshot targets weren't persisting the changes he made to his local database. After some investigation, we discovered that the `postgresql` docker image we're using declares a `volume` in its dockerfile, causing the default PostgreSQL data directory to always be created as a mounted volume, instead of peristing data normally inside the container. This isn't a problem in normal usage, but when we use `docker commit` to snapshot the state of a running container to a temporary docker image, any mounted volumes are omitted from the snapshot!

This change forces our database to write its db data to `/var/lib/czge/data` which isn't mounted as a volume, so our snapshot flows work properly.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ LOCAL_DB_RW_PASSWORD = password_rw
 LOCAL_DB_RO_USERNAME = user_ro
 LOCAL_DB_RO_PASSWORD = password_ro
 LOCALDEV_PROFILE ?= web
-
+LOCAL_DB_CONN_STRING = postgresql://$(LOCAL_DB_ADMIN_USERNAME):$(LOCAL_DB_ADMIN_PASSWORD)@$(LOCAL_DB_SERVER)/$(LOCAL_DB_NAME)
 
 ### HELPFUL #################################################
 help: ## display help for this makefile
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -115,33 +115,35 @@ local-init: oauth/pkcs12/certificate.pfx .env.ecr local-ecr-login local-hostconf
...
 	$(docker_compose) up -d database
 	# Wait for psql to be up
-	while [ -z "$$($(docker_compose) exec -T database psql "postgresql://$(LOCAL_DB_ADMIN_USERNAME):$(LOCAL_DB_ADMIN_PASSWORD)@$(LOCAL_DB_SERVER)/$(LOCAL_DB_NAME)" -c 'select 1')" ]; do echo "waiting for db to start..."; sleep 1; done;
+	while [ -z "$$($(docker_compose) exec -T database psql $(LOCAL_DB_CONN_STRING) -c 'select 1')" ]; do echo "waiting for db to start..."; sleep 1; done;
 	$(docker_compose) exec -T backend alembic upgrade head
 	@echo
 	@echo "Ok, local db is prepared and ready to go -- make any additional changes you need to and then run:"
 	@echo "make create_new_db_image"
 
 .PHONY: create_new_db_image
```
vincent-czi: I don't know makefiles from a hole in the ground, but will the `_` vs `-` in the .PHONY vs command impact anything?

jgadling: not much, but good catch - I'll fix!

Makefile
```diff
@@ -115,33 +115,35 @@ local-init: oauth/pkcs12/certificate.pfx .env.ecr local-ecr-login local-hostconf
...
 	$(docker_compose) exec -T backend $(BACKEND_APP_ROOT)/scripts/setup_dev_data.sh
 	$(docker_compose) exec -T backend alembic upgrade head
 	$(docker_compose) exec -T backend python scripts/setup_localdata.py
 	$(docker_compose) exec -T backend pip install ./aspen
 
+# Assumes you've already run `make local-init` to configure localstack resources!
 .PHONY: prepare-new-db-snapshot
-prepare-new-db-snapshot:
+prepare-new-db-snapshot: local-start
+	$(docker_compose) rm -f database # Wipe out the additional dev rows that `make local-init` adds
```
vincent-czi, line 128: Glad to see this added -- I was doing it manually when working earlier today

Makefile
```diff
@@ -115,33 +115,35 @@ local-init: oauth/pkcs12/certificate.pfx .env.ecr local-ecr-login local-hostconf
...
 	@echo "Ok, local db is prepared and ready to go -- make any additional changes you need to and then run:"
 	@echo "make create_new_db_image"
 
 .PHONY: create_new_db_image
-create_new_db_image:
-	docker exec aspen_database_1 psql postgresql://$(LOCAL_DB_ADMIN_USERNAME):$(LOCAL_DB_ADMIN_PASSWORD)@$(LOCAL_DB_SERVER)/$(LOCAL_DB_NAME) -c VACUUM FULL
-	docker commit aspen_database_1 temp_db_image
+create-new-db-image:
+	docker compose exec database psql $(LOCAL_DB_CONN_STRING) -c VACUUM FULL
+	docker commit czgenepi-database-1 temp_db_image
```
vincent-czi, line 141: We talked on Slack, but just a reminder to let other team members know they should make sure their local folder is `czgenepi`

Makefile
```diff
@@ -115,33 +115,35 @@ local-init: oauth/pkcs12/certificate.pfx .env.ecr local-ecr-login local-hostconf
...
 	export AWS_ACCOUNT_ID=$$(aws sts get-caller-identity --profile $(AWS_DEV_PROFILE) | jq -r .Account); \
 	export DOCKER_REPO=$${AWS_ACCOUNT_ID}.dkr.ecr.us-west-2.amazonaws.com; \
 	export TAGGED_IMAGE="$${DOCKER_REPO}/genepi-devdb:build-$$(date +%F)"; \
-	echo docker build -t $${TAGGED_IMAGE} -f ./docker/Dockerfile.devdb ./docker; \
-	docker build -t $${TAGGED_IMAGE} -f ./docker/Dockerfile.devdb ./docker; \
+	echo dockerx build --platform linux/amd64 -t $${TAGGED_IMAGE} -f ./docker/Dockerfile.devdb ./docker; \
+	docker buildx build --platform linux/amd64 -t $${TAGGED_IMAGE} -f ./docker/Dockerfile.devdb ./docker && \
```
vincent-czi, line 146: Just curious -- why switching over to `&&` vs `;` like the other lines here? I'm very weak on Bash, curious if there was a particular reason here I'm not seeing.

docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: "3.8"
 
 services:
   database:
-    image: "${DOCKER_REPO}genepi-devdb:latest"
+    image: "${DOCKER_REPO}genepi-devdb:czge-2022-04-04"
```
vincent-czi: This should go back to being `:latest`, right? We'll need to put in a new image for that tag that that has the data under the new PGDATA as part of this merge though.

docker-compose.yml
```diff
@@ -11,6 +11,7 @@ services:
       - POSTGRES_USER=postgres
       - POSTGRES_PASSWORD=password_postgres
       - POSTGRES_DB=aspen_db
+      - PGDATA=/var/lib/czge/data # Avoid using the pgsql image's default volume.
```
vincent-czi: Nit: I think a bit more explanation here might be nice. It's such a niche thing and kind of unclear why we need to avoid the default volume at a glance. Maybe just toss in a link to the StackOverflow you showed me in case anyone needs to figure out why this is here in two years.

docker/Dockerfile.devdb
```diff
@@ -7,4 +7,4 @@ FROM temp_db_image as dumper
 # from continually building new images on top of old ones
 FROM postgres:13.1-alpine
 
-COPY --from=dumper /var/lib/postgresql/data /var/lib/postgresql/data
+COPY --from=dumper /var/lib/czge/data /var/lib/czge/data
```
vincent-czi, line 12: Nit: might want to mention that this is a magic string that needs to match what's in the docker compose PGDATA var

vincent-czi (approved):
Looks great! Thanks for figuring this out so fast!

The only important comment is that we probably want to go back to the tag being `:latest` once we upload a new image at that tag, right? I figure we can basically do those at the same time -- push up newly tagged image and merge this within a few minutes of that.

Makefile
jgadling: There are some edge cases & caveats, but for the most part: `&&` means that we'll stop and throw an error when a command fails. `;` means we'll just keep on mowing through the rest of our commands even if a previous one failed. `&&` is much safer in most cases.

docker-compose.yml
jgadling: Yep!

Makefile
vincent-czi: Thanks!

PR #1009, opened 2022-04-06
Change key for saving pango version

lvreynoso:
### Summary:
- **What:** Pangolin had a major version upgrade (3 -> 4), and now internally defaults to using `usher` instead of `pangolearn`. To keep compatibility, we needed to change the key used to access version information in our `save.py` script. This should fix broken pangolin runs for now; the release notes indicate other major changes in the future. 
- **Env:** `<rdev link>`

### Demos:
Screen Shot 2022-04-06 at 14 14 50

### Checklist:
- [x] I added labels to my PR


danrlu:
Lightening speed! Thanks Lucia! Looks good to me. Only thing is do we need to update the testing (I saw it while searching for the old column header)? https://github.com/chanzuckerberg/czgenepi/search?q=%22pangoLEARN_version%22
The `version` column can have test values of `PUSHER-1.2.133`

jgadling (approved):
I'm somewhat blindly approving this -- is it ok to just use the usher version directly? Do we need to test that everything still works properly when switching to usher?

danrlu:
Like sending the change to staging and see whether the scheduled rerun of all samples will work? With the testing Lucia did, I think the only thing that will break this is if any samples are weird enough to choke Pango.

lvreynoso:
I updated the tests (and found something else, the assignment hash wording changed - fixed it).

danrlu (approved):
Thanks very much Lucia!

PR #1018, opened 2022-04-08
Make pangolin job run once a day

danrlu:
### Summary:
- **What:** Running Pangolin on all our samples take 6+ hours now. There is not much added benefit to do a 11a run and a 4pm run in the same day. 

### Checklist:
- [x] I added labels to my PR


jgadling:
I thought we were changing our string matching and this job was supposed to go back to running quickly? Either way, we run pango on upload now so 1x/day is fine.

jgadling (approved):

danrlu:
Good point~~ I put this in since it's unclear when that change will be in place.... except I forgot about this PR and it's prob less useful now LOL Thanks Jess!

PR #1022, opened 2022-04-13
sample edit import tsv

phoenixAja:
### Summary:
- **What:** allow user to import tsv file for sample edit, also show warnings/ errors / success alerts related  to user import. Also tell user that data will be overwritten if importing new metadata file
- **Tickets:** 
   - [sc180576](https://app.shortcut.com/genepi/story/180576)
   - [sc180928](https://app.shortcut.com/genepi/story/180928) - partially completed (see note)
   - [sc180927](https://app.shortcut.com/genepi/story/180927) - partially completed (see note)
   - [sc180636](https://app.shortcut.com/genepi/story/180636)
   - [sc180636](https://app.shortcut.com/genepi/story/180636)

- **Env:** coming soon

### Notes:
* some warnings/ errors (duplicate IDS, Unknown data not imported, Updated value) have not been included in this PR as those are relatively recent warnings/errors that have been merged in. 

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/index.tsx
@@ -1,0 +1,208 @@
+import { pick } from "lodash";
+import React, { useEffect, useMemo, useState } from "react";
+import { EMPTY_OBJECT } from "src/common/constants/empty";
+import { createStringToLocationFinder } from "src/common/utils/locationUtils";
+import FilePicker from "src/components/FilePicker";
+import ImportFileWarnings from "src/components/ImportFileWarnings";
+import { WebformTableTypeOptions as MetadataUploadTypeOption } from "src/components/WebformTable";
+import { SAMPLE_EDIT_WEBFORM_METADATA_KEYS_TO_HEADERS } from "src/components/WebformTable/common/constants";
+import {
+  SampleEditMetadataWebform,
+  SampleIdToEditMetadataWebform,
+  WARNING_CODE,
+} from "src/components/WebformTable/common/types";
+import {
+  NamedGisaidLocation,
+  Props as CommonProps,
+} from "src/views/Upload/components/common/types";
+import {
+  getAutocorrectCount,
+  getMissingFields,
+} from "src/views/Upload/components/Metadata/components/ImportFile";
+import { getMetadataEntryOrEmpty } from "../../utils";
+import {
+  parseFileEdit,
+  ParseResult,
+  SampleIdToWarningMessages,
+} from "./parseFile";
+
+interface Props {
+  metadata: SampleIdToEditMetadataWebform | null;
+  changedMetadata: SampleIdToEditMetadataWebform | null;
+  namedLocations: NamedGisaidLocation[];
+  setMetadata: CommonProps["setMetadata"];
+  hasImportedMetadataFile: boolean;
+  setHasImportedMetadataFile(x: boolean): void;
+  setAutocorrectWarnings(x: SampleIdToWarningMessages): void;
+  setChangedMetadata: CommonProps["setMetadata"];
+}
+
+export default function ImportFile({
+  metadata,
+  namedLocations,
+  setMetadata,
+  setChangedMetadata,
+  hasImportedMetadataFile,
+  setHasImportedMetadataFile,
+  setAutocorrectWarnings,
+  changedMetadata,
+}: Props): JSX.Element {
+  const [missingFields, setMissingFields] = useState<string[] | null>(null);
+  const [autocorrectCount, setAutocorrectCount] = useState<number>(0);
+  const [filename, setFilename] = useState("");
+  const [parseResult, setParseResult] = useState<ParseResult | null>(null);
+  const [extraneousSampleIds, setExtraneousSampleIds] = useState<string[]>([]);
+  const [missingData, setMissingData] =
+    useState<SampleIdToWarningMessages>(EMPTY_OBJECT);
+  const [badFormatData, setBadFormatData] =
+    useState<SampleIdToWarningMessages>(EMPTY_OBJECT);
+
+  useEffect(() => {
+    // If no file uploaded yet, do nothing.
+    if (!parseResult) return;
+    // If file was missing any col header fields, we parsed no data from it
+    // and only display that error to the user to force them to fix problems.
+    if (getMissingFields(parseResult)) return;
+
+    const { data } = parseResult;
+    const parseResultSampleIds = Object.keys(data);
+    const sampleIds = Object.keys(metadata || EMPTY_OBJECT);
+
+    const sampleIdsSet = new Set(sampleIds);
+    const extraneousSampleIds = parseResultSampleIds.filter((parseId) => {
+      return !sampleIdsSet.has(parseId);
+    });
+    setExtraneousSampleIds(extraneousSampleIds);
+  }, [parseResult, metadata]);
+
+  // Used by file upload parser to convert location strings to Locations
+  const stringToLocationFinder = useMemo(() => {
+    return createStringToLocationFinder(namedLocations);
+  }, [namedLocations]);
+
+  const handleFiles = async (files: FileList | null) => {
+    if (!files) return;
+
+    const result = await parseFileEdit(files[0], stringToLocationFinder);
+
+    const { warningMessages, filename } = result;
+    const missingFields = getMissingFields(result);
+    const autocorrectCount =
+      getAutocorrectCount(warningMessages.get(WARNING_CODE.AUTO_CORRECT)) || 0;
+    setHasImportedMetadataFile(true);
+    setMissingFields(missingFields);
+    setAutocorrectCount(autocorrectCount);
+    setFilename(filename);
+    setParseResult(result);
+    setMissingData(
+      warningMessages.get(WARNING_CODE.MISSING_DATA) || EMPTY_OBJECT
+    );
+    setBadFormatData(
+      warningMessages.get(WARNING_CODE.BAD_FORMAT_DATA) || EMPTY_OBJECT
+    );
+
+    handleMetadataFileUpload(result);
+  };
+
+  function inferDeleteEntries(
+    value: string | boolean | NamedGisaidLocation
+  ): string | boolean | NamedGisaidLocation | undefined {
+    if (value && value.toString().toLowerCase() == "delete") {
+      return "";
+    }
+    return value;
+  }
+
+  function handleMetadataFileUpload(result: ParseResult) {
+    // If they're on the page but somehow have no samples (eg, refreshing on
+    // Metadata page), short-circuit and do nothing to avoid any weirdness.
+    if (!metadata) return;
+
+    const { data: sampleIdToUploadedMetadata, warningMessages } = result;
+
+    // Filter out any metadata for samples they did not just upload
+    // Note: Might be cleaner to do this filtering inside of file parse call,
+    // but would require changing the way some of the warnings work currently.
+    const uploadedMetadata: SampleIdToEditMetadataWebform = {};
+    const changedMetadataUpdated: SampleIdToEditMetadataWebform = {};
+    for (const sampleId of Object.keys(metadata)) {
+      // get current metadata and changed metadata entry for a sampleId
+      const existingMetadataEntry = metadata[sampleId];
+      const exitingChangedMetadataEntry = getMetadataEntryOrEmpty(
+        changedMetadata,
+        sampleId
+      );
+
+      if (existingMetadataEntry) {
+        const uploadedMetadataEntry = sampleIdToUploadedMetadata[sampleId];
+
+        if (uploadedMetadataEntry) {
+          // get metadata entries from upload that are not empty (means user wants to import new data)
+          const uploadedFieldsWithData: string[] = [];
+          Object.keys(uploadedMetadataEntry).forEach(function (item) {
+            if (
+              uploadedMetadataEntry[item as keyof SampleEditMetadataWebform] !=
+              ""
+            )
+              uploadedFieldsWithData.push(item);
+          });
+
+          // check if any entries need to be deleted/ cleared
+          for (const [key, value] of Object.entries(uploadedMetadataEntry)) {
+            (uploadedMetadataEntry[key as keyof SampleEditMetadataWebform] as
+              | string
+              | boolean
+              | NamedGisaidLocation
+              | undefined) = inferDeleteEntries(value);
+          }
+          uploadedMetadata[sampleId] = {
+            ...existingMetadataEntry,
+            ...pick(uploadedMetadataEntry, uploadedFieldsWithData),
+          };
+          changedMetadataUpdated[sampleId] = {
+            ...exitingChangedMetadataEntry,
+            ...pick(uploadedMetadataEntry, uploadedFieldsWithData),
+          };
+        }
+      }
+    }
+
+    setMetadata(uploadedMetadata);
+    setChangedMetadata(changedMetadataUpdated);
+    setHasImportedMetadataFile(true);
+
+    setAutocorrectWarnings(
+      warningMessages.get(WARNING_CODE.AUTO_CORRECT) || EMPTY_OBJECT
+    );
+  }
+
+  return (
+    <>
+      <FilePicker
+        handleFiles={handleFiles}
+        text="Select Metadata File"
+        accept=".tsv,.csv"
+        shouldConfirm={hasImportedMetadataFile}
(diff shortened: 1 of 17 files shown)


phoenixAja:
i promise there will be no force pushes until review is complete!

src/frontend/src/views/Upload/components/Metadata/components/ImportFile/components/Alerts/warnings.tsx
```diff
@@ -82,6 +83,52 @@ export function WarningExtraneousEntry({
...
+    />
+  );
+}
+
+export function WarningExtraneousEntrySampleEdit({
+  extraneousSampleIds,
+}: PropsExtraneousEntry): JSX.Element {
+  const count = extraneousSampleIds.length;
+  // "X Samples in metadata file were not used."
+  const title = `${count} ${pluralize(
```
mmmmmmaya, line 112: I think this works out to `5 Samples in metadata files were couldn't be matched and weren't imported.`. Is that intentional?

It seems like maybe the pluralization of "was" should be moved to just before imported

src/frontend/src/views/Upload/components/Metadata/components/ImportFile/components/Alerts/warnings.tsx
```diff
@@ -82,6 +83,52 @@ export function WarningExtraneousEntry({
...
+/**
+ * WARNING_CODE.EXTRANEOUS_ENTRY (SAMPLE EDIT)
+ * (mostly similar to the above implementation, wording is different enough to have it's own component)
+ */
+function MessageExtraneousEntrySampleEdit({
+  extraneousSampleIds,
+}: PropsExtraneousEntry) {
+  const tablePreamble =
+    "The following IDs in your file’s “Current Private ID” column did not match any selected " +
+    "samples, and weren’t imported. Please double check and correct any errors. ";
```
mmmmmmaya, line 95: Is the extra space at the end of this string needed for spacing?

src/frontend/src/views/Upload/components/Metadata/components/ImportFile/components/Alerts/warnings.tsx
```diff
@@ -225,17 +277,23 @@ function MessageBadFormatData({ badFormatData }: PropsBadFormatData) {
...
       title={title}
-      collapseContent={<MessageBadFormatData badFormatData={badFormatData} />}
-      intent={WARNING_SEVERITY}
+      collapseContent={
+        <MessageBadFormatData
+          badFormatData={badFormatData}
+          IdColumnNameForWarnings={IdColumnNameForWarnings}
+        />
+      }
+      intent={"error"}
```
mmmmmmaya:
```suggestion
      intent="error"
```

src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/parseFile.ts
```diff
@@ -0,0 +1,249 @@
...
+      header: true, // Imported file starts with a header row
+      // We parse the column headers to their corresponding metadata keys
+      comments: "#",
+      transformHeader: convertHeaderToMetadataKey,
+      // Because file parsing is async, we need to use callback on `complete`
+      complete: ({
+        data: rows,
+        meta: papaParseMeta,
+      }: Papa.ParseResult<Record<string, string>>) => {
+        const uploadedHeaders = papaParseMeta.fields as string[]; // available b/c `header: true`
```
mmmmmmaya:
```suggestion
        const uploadedHeaders: string[] = papaParseMeta.fields; // available b/c `header: true`
```

src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/parseFile.ts
```diff
@@ -0,0 +1,249 @@
...
+              IGNORED_SAMPLE_IDS
+            );
+            // If false-y, there was no parse result, so we just skip those
+            if (rowMetadata) {
+              // now that we've parsed the tsv we need to transform the metadata row
+              // from SampleEditTsvMetadata to SampleEditMetadataWebform to do this we
+              // need the a type to transition us (MergedSampleEditTsvWebformMetadata)
+              const rowMetadataNew: MergedSampleEditTsvWebformMetadata =
+                rowMetadata;
+              const rowPrivateID = rowMetadataNew.currentPrivateID as string;
```
mmmmmmaya:
```suggestion
              const rowPrivateID: string = rowMetadataNew.currentPrivateID;
```

src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+    );
+    setBadFormatData(
+      warningMessages.get(WARNING_CODE.BAD_FORMAT_DATA) || EMPTY_OBJECT
+    );
+
+    handleMetadataFileUpload(result);
+  };
+
+  function inferDeleteEntries(
+    value: string | boolean | NamedGisaidLocation
```
mmmmmmaya, line 109: I think I saw a few string | boolean | NamedGisaidLocation in a previous PR (not sure if I'm remembering right). If so, maybe consider adding a named type.

src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+    // Filter out any metadata for samples they did not just upload
+    // Note: Might be cleaner to do this filtering inside of file parse call,
+    // but would require changing the way some of the warnings work currently.
+    const uploadedMetadata: SampleIdToEditMetadataWebform = {};
+    const changedMetadataUpdated: SampleIdToEditMetadataWebform = {};
+    for (const sampleId of Object.keys(metadata)) {
+      // get current metadata and changed metadata entry for a sampleId
+      const existingMetadataEntry = metadata[sampleId];
+      const exitingChangedMetadataEntry = getMetadataEntryOrEmpty(
```
mmmmmmaya:
```suggestion
      const existingChangedMetadataEntry = getMetadataEntryOrEmpty(
```

src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+              | boolean
+              | NamedGisaidLocation
+              | undefined) = inferDeleteEntries(value);
+          }
+          uploadedMetadata[sampleId] = {
+            ...existingMetadataEntry,
+            ...pick(uploadedMetadataEntry, uploadedFieldsWithData),
+          };
+          changedMetadataUpdated[sampleId] = {
+            ...exitingChangedMetadataEntry,
```
mmmmmmaya:
```suggestion
            ...existingChangedMetadataEntry,
```

mmmmmmaya (approved):
Overall, I think it looks good. Would love for @vincent-czi to take a look and ok the logic on upload stuff, since he's more familiar with it than I am.

Beyond resolving the one copy wiggle, nothing here is blocking

src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+    useState<SampleIdToWarningMessages>(EMPTY_OBJECT);
+  const [badFormatData, setBadFormatData] =
+    useState<SampleIdToWarningMessages>(EMPTY_OBJECT);
+
+  useEffect(() => {
+    // If no file uploaded yet, do nothing.
+    if (!parseResult) return;
+    // If file was missing any col header fields, we parsed no data from it
+    // and only display that error to the user to force them to fix problems.
+    if (getMissingFields(parseResult)) return;
```
vincent-czi: I don't understand why it's running `getMissingFields` rather than just using the state of `missingFields` on its own. Won't the parseResult only exist after `handleFiles` has run, and as part of that there's a call to setting the `missingFields` state?

src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+        sampleId
+      );
+
+      if (existingMetadataEntry) {
+        const uploadedMetadataEntry = sampleIdToUploadedMetadata[sampleId];
+
+        if (uploadedMetadataEntry) {
+          // get metadata entries from upload that are not empty (means user wants to import new data)
+          const uploadedFieldsWithData: string[] = [];
+          Object.keys(uploadedMetadataEntry).forEach(function (item) {
```
vincent-czi: Nit: I think the purpose of what this is doing would be clearer if written as a `.filter` rather than a `.forEach` that is pushing the keys to an array.

src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+      warningMessages.get(WARNING_CODE.BAD_FORMAT_DATA) || EMPTY_OBJECT
+    );
+
+    handleMetadataFileUpload(result);
+  };
+
+  function inferDeleteEntries(
+    value: string | boolean | NamedGisaidLocation
+  ): string | boolean | NamedGisaidLocation | undefined {
+    if (value && value.toString().toLowerCase() == "delete") {
```
vincent-czi: This is JS territory, we need that `===` knee-jerk devotion (even though here I don't think there's any way for `==` to go weird on us given what's comparing)

src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+      warningMessages.get(WARNING_CODE.MISSING_DATA) || EMPTY_OBJECT
+    );
+    setBadFormatData(
+      warningMessages.get(WARNING_CODE.BAD_FORMAT_DATA) || EMPTY_OBJECT
+    );
+
+    handleMetadataFileUpload(result);
+  };
+
+  function inferDeleteEntries(
```
vincent-czi: Nit: At first glance on this name, I thought it was going to return a bool of if the entry needed to be deleted. I did skim the function, but I guess I didn't pay attention, because I only realized much later than when it infers deletion it means that it returns an effectively deleted value of empty string. I think a better name would be `passOrDeleteEntry` since that what it does: it checks one entry value at a time and then either passes it on or it deletes it. Probably also good to explain with a comment above the helper func saying that the user signals they want to delete a value by the keyword DELETE

src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+          };
+          changedMetadataUpdated[sampleId] = {
+            ...exitingChangedMetadataEntry,
+            ...pick(uploadedMetadataEntry, uploadedFieldsWithData),
+          };
+        }
+      }
+    }
+
+    setMetadata(uploadedMetadata);
```
vincent-czi, line 172: I _think_ this will delete data from the overall metadata if the user does not upload an entry for every single entry in the edit table. It's hard to tell, but I'm pretty sure the prior `if`s will only put in data for the `uploadedMetadata` if there is **both** `existingMetadataEntry` and `uploadedMetadataEntry` for the `sampleId`.

I don't know if it's in the spec, but if a user has rows A, B, and C, then downloads the TSV but removes the row for C entirely before uploading the TSV, I think that should be a no-op on C's data. But I think this will clear the data for C right now?

src/frontend/src/components/WebformTable/common/types.ts
```diff
@@ -46,7 +46,7 @@ export interface Metadata {
   keepPrivate?: boolean;
   publicId?: string;
   sequencingDate?: string;
-  collectionLocation?: NamedGisaidLocation;
```
vincent-czi, line 49: Is the `string` case only for editing? When the `collectionLocation` is coming from an existing sample and it's in edit mode? In a perfect world, I think the BE should return the location object as opposed to just the string, but that's more effort than we want to spend right now. (plus it would probably require a very painful migration for older samples)

I think it would be good to have a comment here explaining that it's not exactly that `collectionLocation` can be either at any time, it's that depending on the situation -- brand new or edit mode -- it determines the type.

vincent-czi: I just saw the comment on `collectionLocationName` elsewhere
```
  const collectionLocationName = () => {
    // collection location will always be a NamedGisaidLocation at this stage,
    // the only time collectionLocation will be a string is during tsv upload
    // where collectionLocation can be "DELETE" (when a user wants to clear a value)
```
If `string` type is purely used for the DELETE sentinel value, maybe have the other type here not be a generic string, but the constant string of just "DELETE", I think that would be a lot more explicit and clear.

Now, that said, we would need to make sure where the parsing is happening it's only returning either the `NamedGisaidLocation` or the "DELETE" string, so it would require re-writing that part a bit. I think it would make some of the logic around deletion and collectionLocation types less confusing in other places, but I'm not sure it's really worth the effort given time constraints. I dunno.

vincent-czi (commented):
Overall looks good. I mostly focused on carefully reading the `ImportFile/index.tsx` and didn't read the warnings part nearly as carefully, so I can't speak to that part as much.

src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/index.tsx
phoenixAja: yeah that's a good idea, i'll address that on my next PR

src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/index.tsx
phoenixAja: will do!

src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/index.tsx
phoenixAja: it shouldn't unless the user has a row for C, but i will double check

src/frontend/src/components/WebformTable/common/types.ts
phoenixAja: yeah that's a good idea! I will try to change that over in my later PR (since i want this to get merged in soon so maya can start on the review page)

src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/index.tsx
phoenixAja: ok i can change it

src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/index.tsx
phoenixAja: true true, will change

src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/index.tsx
phoenixAja: i'll make a note to change it

src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/index.tsx
phoenixAja: updated

src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/index.tsx
phoenixAja: you're right that this was wrong, thank you for these eagle eyes! 🦅 I fixed it

PR #1027, opened 2022-04-15
Lineage filter for Sample Filtering

vincent-czi:
### Summary:
- **What:** Adds in lineage filter for Sample Filtering feature
- **Ticket:** [sc<184610>](https://app.shortcut.com/genepi/story/<184610>)
- **Env:** https://vince-lineage-filter-frontend.dev.czgenepi.org/

### Demos:
image

image

### Notes:
- There is a **major issue** with the dropdown as currently implemented. When you type in a search term to the Dropdown (say, "BA"), filtering works as expected. However, as soon as you click an option, the search term gets cleared, and the filter resets. This is a problem, because it would be a common case for our users to want to type in "BA", then select every `BA.x` lineage that appears. But because of the reset, they would have to type in "BA" over and over again to get back to that section of the lineage choices.
  - After doing a bit of research into it, this is the standard behavior in the czifui Dropdown, and I think it's also the standard behavior in the MUI components czifui is building off of. I couldn't find a way to disable this search term clearing after selection. Hopefully I missed something, but I'm worried this might be very hard to solve. I'm going to bring this up in the SDS Support channel. I'm not sure if this should be considered a blocker for merging or if we can merge it in now and hopefully get a solution from SDS in the next week. @Janeece, thoughts?
- The lineages filter is currently set to only display 100 options at a time instead of all the lineages at once (~1600). If you try to render all the lineages at once, the Dropdown takes multiple seconds to render and is very slow when interacting. This is basically the approach that the Locations dropdown in the metadata upload flow takes -- even if there would be more than 100 matches, it only displays 100 at most. I discussed this with @Janeece and she thought it was an acceptable compromise. It's a little bit weird compared to Locations though because Locations only begins to show options after at least 3 characters are typed, while Lineages shows options right from the start. The alternative would be figuring out virtualizing the rows though, and I wasn't sure about how much effort that would entail, so I think this was the better choice? That said, depending on how the above issue with search clearing on selection goes, it might be necessary to virtualize so the user could just scroll down to the "BA" zone as a stopgap.
- There are a number of visual specs I wasn't able to work into the dropdown. I've discussed them with Janeece and she was tentatively okay with them not being in there. I tried to work on them for awhile, but with the way it interacts with czifui, I'm not sure we can accomplish them on our end, or if they would necessitate changes in the czifui codebase. Here's the major ones:
  - Coloring of the input dropdown click target: specs have it getting a colored border when the Dropdown is open. I couldn't figure out a way to target this aspect since the `open` state is abstracted away within Dropdown, but we need the `InputDropdownComponent` within Dropdown to be get that as a prop so it can do styling based on that. Also, this might be part of the planned design for Dropdown in SDS and it's just not implemented yet? I'm going to bring this up in the SDS Support channel.
  - `max-height` for the open dropdown menu: specs have it at `290px`. I tried for awhile to mess with this, but was unsuccessful. I can change the menu container's max-height, but the items don't respect it and just blow right past the edge making for very weird visuals. It seems that whatever is rendering the menu items is doing some cleverness based on window height, so the actual menu height varies depending on the window, so it seems that was beating out my attempts to control it.
    - In retrospect, maybe I just missed trying a CSS `overflow` property. I can go try that.
    - @Janeece might also be okay with not having the `max-height` -- since the menu re-sizing is fairly intelligent, it does a good job of never letting the bottom of the window cut off any of the menu, so it's not as much of a visual problem.
- The dropdown menu does an automatic "flip" if you scroll the page in such a way that a part of it would no longer be visible. So the menu will flip to being above the dropdown opener if scrolling would cause it to go out the edge of the window. This is the default behavior for czifui and MUI. @Janeece wasn't sure if we want it or not, it hadn't come up before in design discussions, so she was going to discuss with rest of Design. Probably not a blocker on merging now, but if we need to disable it, you can do so with code along the below lines:
```
	    import { DropdownPopper } from "czifui";
		<StyledDropdown
		  ... other props like normal ...
		  PopperComponent={(props: any) => (
		    <DropdownPopper
		      placement="bottom-start"
		      modifiers={{
		        flip: {
		          enabled: false,
		        },
		      }}
		      {...props}
		    />
		  )}
		/>
```

### Checklist:
- [x] I added labels to my PR


src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/style.tsx
```diff
@@ -0,0 +1,102 @@
...
+export const StyledDropdown = styled(Dropdown)`
+  width: ${DROPDOWN_WIDTH};
+  .MuiButton-label > span {
+    color: black;
+  }
+` as typeof Dropdown; // assert b/c `styled` causes an interface hiccup;
+
+// DropdownPopper (for use with Dropdown's PopperComponent prop) needs to
+// have a `placement` prop to set where it anchors against Dropdown opener.
+const ConfiguredDropdownPopper = (props: any) => {
```
mmmmmmaya: I kind of feel like this doesn't belong in the style sheet. Why not include it in the index.tsx file?

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/RadioLabel/index.tsx
```diff
@@ -16,11 +16,22 @@ import {
   StyledListItemIcon,
```
mmmmmmaya, line 16: Not necessarily relevant for this PR, but if we do work on this again, probably this file should be split up. I feel like it's trying to do too much.

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/index.tsx
```diff
@@ -11,6 +17,119 @@ import {
...
+
+/**
+ * Creates `options` for lineages Dropdown to display as choices.
+ *
+ * This handles a few things. First, the Dropdown component expects the options
+ * to come in a certain format, so this converts the internal lineage arrays
+ * into something Dropdown can display. Second, it handles moving up selected
+ * lineages to the top of the list to be displayed above unchosen lineages.
+ * Third, it ensures that the "All" choice -- reset back to having no selected
+ * lineages -- is always available and ad the top of the list.
```
mmmmmmaya:
```suggestion
 * lineages -- is always available and at the top of the list.
```

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/index.tsx
```diff
@@ -21,6 +140,12 @@ const SAMPLE_FILTERING_TOOLTIP_TEXT = (
   </div>
 );
 
+// Styling for the InputDropdown child of general Dropdown component
+const InputDropdownProps = {
+  sdsStage: "userInput",
+  sdsStyle: "square",
+} as const;
```
mmmmmmaya, line 151: what is this type assertion?

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/index.tsx
```diff
@@ -31,13 +156,114 @@ const SAMPLE_FILTERING_TOOLTIP_TEXT = (
...
+   * is not really a choice, but rather a pseudo-choice that means no lineages
+   * are selected or that the lineage filter should be reset, we get into some
+   * weird places. The Dropdown component expects every option to just be an
+   * object that is either selected or not selected. But b/c of the above, the
+   * options now have side-effects: choosing a lineage removes "All" from being
+   * selected, or choosing "All" can reset and de-select all the other options.
+   * So in addition to needing to handle this side-effect logic, we also have
+   * to avoid infinite render loops due to the side-effect setting a new
+   * `selectedLineages` upstream, which then goes down into the Dropdown, which
+   * then kicks off the onChange (b/c its controlled), which can then trigger
```
mmmmmmaya:
```suggestion
   * then kicks off the onChange (b/c it's controlled), which can then trigger
```

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/index.tsx
```diff
@@ -31,13 +156,114 @@ const SAMPLE_FILTERING_TOOLTIP_TEXT = (
...
+   * another new side-effect handling and spiral into an infinite loop...
+   *
+   * If this winds up being refactored into something better, that would be
+   * great, but make sure to test your work pretty aggressively if you do that
```
mmmmmmaya, line 216: can we do some basic automated testing on this to make it less scary?

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/index.tsx
```diff
@@ -31,13 +156,114 @@ const SAMPLE_FILTERING_TOOLTIP_TEXT = (
...
+   * refactor -- the above interactions cause a lot of edge cases.
+   */
+  function handleLineageDropdownChange(
+    newSelectedOptions: DefaultMenuSelectOption[] | null
+  ): void {
+    // No selection at all means empty all lineage choices.
+    // (Vince) Poked around: I don't think Dropdown emits this in our case?
+    // But interface for the component says its there, so defensive code here.
```
mmmmmmaya:
```suggestion
    // But interface for the component says it's there, so defensive code here.
```

mmmmmmaya (commented):
overall mostly looks good, just a few questions

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/RadioLabel/index.tsx
vincent-czi: Yeah, I agree. Also, I think it's a bit of a stretch for the `label` to carry as much weight as it's doing in the modal radio group. There's some weird scrolling effects that are starting to happen because the extended "label" of a selection is pretty much as tall as the entire modal. The radio group is trying to helpfully move the item into view when clicked, but because the whole item is so tall now, it makes things worse.

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/style.tsx
vincent-czi: I waffled on this one. My thinking was that this configuration was ultimately visual. So since `style.tsx` is about visual stuff (in my mind), I figured it made the most sense in here rather than in `index.tsx` because it's not really about anything logic-process-wise. I also just liked passing a `StyledDropdownPopper` as the `PopperComponent` without needing to give `StyledDropdownPopper` any additional props.

If you think the `style.tsx` is more about CSS though rather than visual stuff generally, I can move it around.

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/index.tsx
vincent-czi: It's a ["const assertion"](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions) -- it's so hard to find good documentation on it, I have no idea how I first learned about it, but basically it says that the thing preceding the `as const` should be read-only. It's weird how it looks like a type assertion, but it's effectively the opposite. Instead of a normal type assertion saying, "Trust me on this one, compiler", a const assertion is saying, "Yo compiler, enforce immutability and yell at me if I get out of line".

The reason I needed it here was that TS was complaining that my `InputDropdownProps` were no good when I tried to use them as props later because they could have been modified since their declaration and couldn't be guaranteed to conform to the props interface. The `as const` lets TS trust me because it knows I can't have any code changing it before passing it.

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/index.tsx
vincent-czi: I certainly like that idea, but I've been thinking about it for awhile and I can't come up with a way to write a good test for it quickly. I'm not great at testing, so maybe you've got a good way in mind or it's easier than I'm imagining? I'll Slack you about it, but I'm pretty much afraid that writing a test that would meaningfully improve its safety -- beyond the 10 minutes of manual QA I did when wrapping up that work -- would take multiple days.

Janeece:
RE Dropdown selection
> I'm not sure if this should be considered a blocker for merging or if we can merge it in now and hopefully get a solution from SDS in the next week. @Janeece, thoughts?

I dont think this is a blocker, I think it would be ok to push it back to SDS and updating when there is a solution. Can we make a ticket to represent that update? Just so we don't forget. 

Thanks for the reminder on the flipping menu question — probably wont get to sync with the other designers on it until at least next week. It isn't a blocker.

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/index.tsx
vincent-czi: Thinking a bit more, another very different way to approach this would be to have a `preprocessSelection` prop on the Dropdown component in czifui. Instead of directly passing the selections made in the Dropdown to its `onChange` prop func, an optional `preprocessSelection` could first run on the selection -- not sure if it would be after every selection click or if it would be only on blur, both approaches seem reasonable -- and then the result of that would get handed to `onChange`.

A lot of the difficulty of this implementation came from needing to handle side-effects for how certain dropdown choices would impact other choices. A `preprocessSelection` prop func would at least bring that side-effect occurrence into the Dropdown and might make the infinite loop aspect impossible. That said, I think this is a real, real edge case use of how Dropdowns work, so not sure if it's worth that effort and muddying the Dropdown API in czifui.

vincent-czi:
@Janeece I created a ticket to track the search text clearing aspect, it's in the Sample Filtering epic.
https://app.shortcut.com/genepi/story/193953/lineage-menu-persist-search-term-during-selections

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/style.tsx
mmmmmmaya: I do think the style files are stylesheets, and not visual stuff. I don't think component composition belongs in the style files, and I feel pretty strongly about that.

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/RadioLabel/index.tsx
mmmmmmaya: https://app.shortcut.com/genepi/story/193962/split-up-refactor-create-nextstrain-tree-modal-radiolabel-into-multiple-files

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/index.tsx
mmmmmmaya: No problem, let's tackle it after we have more science test infra in 
https://app.shortcut.com/genepi/story/193963/write-a-test-to-make-sure-the-lineage-filter-dropdown-has-sane-and-functional-logic

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/style.tsx
vincent-czi: That's reasonable, I'll keep with the stylesheet intention in the future. I moved the configuration over to the index.tsx and went back to a style.ts that's just a bunch of `styled` calls.

mmmmmmaya (approved):

PR #1030, opened 2022-04-18
Add latitude and longitude to the `locations` table when importing gisaid data

jgadling:
### Summary:
- **What:** Add lat/long data to 
- **Ticket:** [sc167081](https://app.shortcut.com/genepi/story/167081)
- **Env:** `<rdev link>`

### TL;DR:
Please run `make local-init` after this PR is merged to fetch the latest version of the local dev database

### Notes:
This PR adds another step to our GISAID import pipeline that reads geo locations from the [ncov latlongs database](https://github.com/nextstrain/ncov/blob/master/defaults/lat_longs.tsv) and populates our `locations` table with those locations.

#### OpenStreetMaps Support
The existing ncov latlongs database only has coverage for about 50% of the locations in our database. I added some functionality to our import script to allow it to also fetch locations from OpenStreetMaps where we have missing data. This is very slow, so we probably don't want to run this every night. For now, I'm using the output from the openstreetmaps script to update the lat/longs db in ncov.

#### Database extensions
Postgresql has several different geographic location calculation tools, and we've decided to use an extension called [earthdistance](https://www.postgresql.org/docs/current/earthdistance.html) to calculate distances between locations. The extension is simple enough to enable in RDS (see the migration script), but our local dev database required admin access to enable it. I've updated our base Docker image to allow non-admin users to enable these extensions, so our migration scripts work locally.

### Checklist:
- [x] I added labels to my PR


src/backend/database_migrations/versions/20220413_225151_add_location_latlongs.py
```diff
@@ -0,0 +1,31 @@
+"""add location latlongs
+
+Create Date: 2022-04-13 22:51:59.577986
+
+"""
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects.postgresql import DOUBLE_PRECISION
```
lvreynoso, line 8: thanks for changing this!

docker/Dockerfile.devdb
```diff
@@ -10,3 +10,5 @@ FROM postgres:13.1-alpine
 # This path needs to be the same as the path in the PGDATA environment
 # variable in docker-compose.yml
 COPY --from=dumper /var/lib/czge/data /var/lib/czge/data
+# Make earthdistance a trusted psql extension
+RUN echo "trusted = true" >> /usr/local/share/postgresql/extension/earthdistance.control
```
lvreynoso, line 14: Do we need to do the same for `cube`?

lvreynoso (approved):
Looks like it needs a `make backend-run-style`; the git hook to automatically style commits is broken at the moment, unfortunately. Also, do we need to add `cube` as a trusted extension as well? Otherwise it looks good, I was able to import lat/longs locally.

docker/Dockerfile.devdb
jgadling: Nope, cube is trusted by default.

PR #1032, opened 2022-04-19
Fix gisaid samples in tsv download

jgadling:
### Summary:
- **What:** Make sure gisaid samples are marked as selected in the tsv download file
- **Ticket:** [sc193207](https://app.shortcut.com/genepi/story/193207)
- **Env:** `<rdev link>`

### Notes:
If GISAID samples were added to a phylo tree with an `hCoV-19/` prefix, our filtering code wasn't stripping the prefix before checking whether samples in a tree (which don't have the prefix) match the samples input by the user.

This PR will force samples to match whether or not they have a `hCoV-19/` prefix on either side of the comparison, and also makes sure all matching is case-insensitive.

### Checklist:
- [x] I added labels to my PR


danrlu:
Looks good to me!! Thanks very much!

danrlu (approved):
hCoV-19, a great banana peel.

PR #1033, opened 2022-04-19
Add metadata for Auspice/Nextstrain to color code nearest countries on a group's tree

lvreynoso:
### Summary:
- **What:** Modifies metadata in the tree json served by the backend, adding a scale to country colorings with the tree owner's location and the 15 nearest countries. Moves installed psql extensions to the `public` schema.
- **Ticket:** [sc167081](https://app.shortcut.com/genepi/story/167081)
- **Env:** [https://countryclr-frontend.dev.czgenepi.org/data/phylogenetic_trees](https://countryclr-frontend.dev.czgenepi.org/data/phylogenetic_trees)

### Demos:

Screen Shot 2022-04-20 at 17 10 45
Screen Shot 2022-04-20 at 16 58 16

### Checklist:
- [x] I added labels to my PR


Makefile
```diff
@@ -9,7 +9,7 @@ export AWS_PROD_PROFILE=genepi-prod
 export BACKEND_APP_ROOT=/usr/src/app
 
 ### DATABASE VARIABLES #################################################
-DB_SEARCH_PATH=aspen
```
jgadling, line 12: TIL!

danrlu:
I am SO excited for this piece of change. Every time I see the tree coloring, it's a suffering and I'm seeing the end of it!!! XD Thank you, thank you! 

**Great to have:**
One suggestion is, if I understand the code correctly, we pull 15 countries that's closest to the user's country and assign colors to them in the JSON. But these 15 countries may or may not have samples on the tree, so some colors may not get used at all. Would it be doable, with a reasonable amount of work, to pull all existing countries from the JSON, and calculate their distance from the user's country, and pick the 15 closest ones to assign colors? 

**Nice to have:**
This ticket was originally made to accommodate LMIC trees where country-level coloring was horribly broken. I'm wondering whether we can modify the current PR to instead do county-level (`location` in the database) coloring for these reasons:
- The trees in the dev revealed a long running bug (note and fix see #1034). As a result, all our trees in 2022 don't have legit country designation in their JSON. So even if we fix country coloring, it won't really help the past trees. And applying this change to past trees is such a great strength of this PR and I'm sad it won't work out this way....
- One challenge with pulling the closest county (aka `Location.location`): I learned from the lat long discussion with Jess that there can be multiple entries in our database for the same place, for example `Union`, `Union MS`, `Union county MS` all point to the same lat long. If we build a tree for `Location.location`=A, which is near `Union`, and we pull 15 counties closest to `Location.location`=A, 3 of them will be `Union`, `Union MS`, `Union county MS` with the same lat long. Here is where the first suggestion comes in. If we pull locations from the tree JSON (which will be unique locations), calculate their lat long and distance to the tree location and pick the 15 closest, having those duplicates in database will actually be a strength (so for any locations in the tree, we definitely find a match in our database to pull lat long) instead of an issue.
- I read through the code, it's so nicely set up and may be not a big lift to change the `Location.country` to `Location.location`. This change will benefit all our current users that are mostly `Location.location` aka county-level users. And we can push the change to add country-level color fix to later sprints when we're actually going to have LMIC users.

Let me know how I can best help out! Thanks very much for working on this!

lvreynoso:
Hey Dan! Good points. So something we could do is pull the countries out of the tree json and find the 15 closest, and if there aren't enough then fill in the remainder with countries pulled from the db. That way we would deal with countries that are actually on the tree, and also have a fallback in the case of the bug in #1034 . We could also do the same with locations. You are right that this code is easily adapted to locations, but that's why I want to finish this up - we can do locations in a follow-up PR.

danrlu:
Ahhhh I see :D For countries that are not in the tree JSON, they wouldn't have any samples on the tree and won't need a color to show, right? What did I miss?
Could it be that you're thinking about the ["meta"] section where the color hex code goes:
image

And I'm thinking about the tree nodes (["tree"]["children"] or ["tree"]["children"]["children"] etc) to pull countries in JSON:
image

lvreynoso:
Oh yes you're right, we can make sure countries have a color label but that won't make a difference if they're not in the actual tree (since they won't get colored). So this doesn't help with #1034 . But I'm not sure what we can do about that.

danrlu:
Those trees are county-level trees, and once we get the following PR in for county/location colors, they will look refreshed and amazing on the county level!! Let's ignore them for now for our sanity XD

src/backend/aspen/api/utils/phylo.py
```diff
@@ -60,6 +92,81 @@ def _sample_filter(sample: Sample, can_see_pi_group_ids: Set[int], system_admin:
...
+        sa.select(Location.country, Location.latitude, Location.longitude)  # type: ignore
+        .where(
+            and_(
+                Location.division == None,  # noqa: E711
+                Location.location == None,  # noqa: E711
+                Location.country == tree_location.country,
+            )
+        )
+        .subquery()
+        .lateral()
```
jgadling: This looks good for the most part -- Only comment is that I think the `lateral` isn't doing anything for us here, and it might (?) be clearer to write this as a query on the locations table for the origin, with a subquery for neighbors, instead of a join between two subqueries. But that's more of a style choice since I suspect our DB will compile down most of these variations to the same execution plan.

src/backend/aspen/api/views/tests/test_auspice.py
```diff
@@ -9,12 +10,37 @@
...
 from aspen.api.views.tests.test_update_phylo_run_and_tree import make_shared_test_data
+from aspen.test_infra.models.location import location_factory
 from aspen.test_infra.models.usergroup import group_factory, user_factory
 
 # All test coroutines will be treated as marked.
 pytestmark = pytest.mark.asyncio
 
 
+Point = namedtuple("Point", ["region", "country", "latitude", "longitude"])
+test_country_data = [
```
jgadling: ❤️

src/backend/aspen/api/conftest.py
```diff
@@ -55,6 +55,21 @@ async def async_db() -> AsyncGenerator[AsyncPostgresDatabase, None]:
         f"grant all privileges on database {database_name} to {USERNAME}"
     )
 
+    # Install extensions we use to the test database.
+    ext_query = await admin_session.execute("SELECT extname FROM pg_extension")
+    extensions = [row["extname"] for row in ext_query]
+    test_admin_uri = (
+        f"postgresql+asyncpg://postgres:password_postgres@database:5432/{database_name}"
+    )
```
jgadling, line 63: it would be better to generate the admin uri based on some config (or better yet, figure out how to grant more access to our test user!) but we can sort that out later if this is getting out of hand.

jgadling (changes requested):

src/backend/aspen/api/conftest.py
lvreynoso: We can give the test user superuser privileges, but that seemed like a bad idea to me 🥴

jgadling: Wait - I take this whole comment back - we're already doing stupid hardcoded db uri's in this method. We don't have to make this smarter right now.

src/backend/aspen/api/utils/phylo.py
```diff
@@ -60,6 +90,86 @@ def _sample_filter(sample: Sample, can_see_pi_group_ids: Set[int], system_admin:
...
+        .where(
+            and_(
+                origin_location.country == tree_location.country,
+                origin_location.division == None,  # noqa: E711
+                origin_location.location == None,  # noqa: E711
+            )
+        )
+        .order_by(asc("distance"))
+        .limit(15)
+    )
```
danrlu, line 152: this task is so beautifully done 🤩

src/backend/aspen/api/utils/phylo.py
```diff
@@ -60,6 +90,86 @@ def _sample_filter(sample: Sample, can_see_pi_group_ids: Set[int], system_admin:
     return sample.submitting_group_id in can_see_pi_group_ids
 
 
+def _collect_countries(node: dict) -> Set[str]:
+    countries = set()
+    node_country = node.get("node_attrs", {}).get("country", {}).get("value", None)
+    if node_country:
+        countries.add(node_country)
+    for child in node.get("children", []):
+        countries |= _collect_countries(child)
```
danrlu, line 99: is this a function calling itself 🤯

lvreynoso: Yep, it recursively traverses the nodes in the tree.

jgadling (approved):
Code LGTM, I'll leave the intent-checking to @danrlu

danrlu:
Should I got to a new dev link? The tree on there only have 1 color:
image
and the JSON has 1 color for ["meta"]["key":"country"]
image

I also tried incognito mode

lvreynoso:
Ah, let me update it.

lvreynoso:
I added enough data to the dev db to get that tree working right:

Screen Shot 2022-04-20 at 16 44 29

In staging and prod, we will have location data to properly sort things. But if we DON'T have the data we should be adding countries present in the tree to the color scale anyways - I'll push a fix for that.

lvreynoso:
This is the full color scale:

Screen Shot 2022-04-20 at 16 58 16

danrlu:
You read my mind. All looks good to me!! Thanks very much!

danrlu (approved):

lvreynoso:
With the fix:
Screen Shot 2022-04-20 at 17 10 45

PR #1034, opened 2022-04-20
set region to global for tree build

danrlu:
### Summary:
- **What:** Adding this change #836 to yaml builder since we're having the same coloring issues 😉 

### Demos:
image
And tree run looks good. 

### Notes:
Will have some impact on Lucia and Jess's coloring by lat long work #1033 b/c all trees in 2022 don't really have functional color by country 😭 Very sorry!!

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/workflows/nextstrain_run/builder_base.py
+++ b/src/backend/aspen/workflows/nextstrain_run/builder_base.py
@@ -37,9 +37,11 @@ class BaseNextstrainConfigBuilder:
             self.tree_build_level = "country"
         elif not location.location:
             self.tree_build_level = "division"
-        # Fill out region/country/division/location fields if the group has them,
+        # The region field for tree build needs to be 'global' to avoid country names being converted to region
+        build["region"] = "global"
+        # Fill out country/division/location fields if the group has them,
         # or remove those fields if they don't.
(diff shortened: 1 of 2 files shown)


lvreynoso (approved):

src/backend/aspen/workflows/nextstrain_run/builder_base.py
```diff
@@ -37,9 +37,11 @@ def update_build(self, config):
             self.tree_build_level = "country"
         elif not location.location:
             self.tree_build_level = "division"
-        # Fill out region/country/division/location fields if the group has them,
+        # The region field for tree build needs to be 'global' to avoid country names being converted to region
+        build["region"] = "global"
```
jgadling: Can we put this change in the YAML template instead if it's going to be the same for all tree runs?

danrlu: I left it here thinking there is a slim chance this will become a parameter at some point (but not in a foreseeable future). Should I put it into the template and move it back when we actually use is as a parameter?

danrlu: Yeah this makes more sense. Let me move it~~~

jgadling (approved):

PR #1040, opened 2022-04-21
Maya/184616 date range menu

mmmmmmaya:
### Summary
- **What:** Add a collection date range filter to the ns tree modal
- **Ticket:** [[sc-184616]](https://app.shortcut.com/genepi/story/184616)
- **Env:** https://daterange-frontend.dev.czgenepi.org/
- **Design:** https://www.figma.com/file/nGRxk2lhJoWoiPSnM0GY8g/Self-Serve-Tree-V4---Sample-Filtering?node-id=2664%3A179344

#### Before: 
Screen Shot 2022-04-21 at 10 46 43 AM

#### After: 
after

### Notes
- Further styling required
- Still need to have this filter impact the resulting tree

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/CollectionDateFilter/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/CollectionDateFilter/index.tsx
@@ -1,0 +1,62 @@
+import React, { useState } from "react";
+import { getDateRangeLabel } from "src/common/utils/dateUtils";
+import { MENU_OPTION_ALL_TIME } from "src/components/DateFilterMenu/constants";
+import { StyledDateFilterMenu, StyledInputDropdown } from "./style";
+
+interface Props {
+  fieldKeyEnd: string;
+  fieldKeyStart: string;
+  menuOptions: DateMenuOption[];
+  updateDateFilter: UpdateDateFilterType;
+}
+
+const CollectionDateFilter = ({
+  updateDateFilter,
+  ...props
+}: Props): JSX.Element => {
+  // `startDate` and `endDate` represent the active filter dates. Update on filter change.
+  const [startDate, setStartDate] = useState<FormattedDateType>();
+  const [endDate, setEndDate] = useState<FormattedDateType>();
+  const [anchorEl, setAnchorEl] = useState<HTMLElement | undefined>();
+  // What menu option is chosen. If none chosen, `null`.
+  const [selectedDateMenuOption, setSelectedDateMenuOption] =
+    useState<DateMenuOption | null>(MENU_OPTION_ALL_TIME);
+
+  const handleClick: React.MouseEventHandler<HTMLButtonElement> = (event) => {
+    setAnchorEl(event.currentTarget);
+  };
+
+  const handleClose = () => {
+    setAnchorEl(undefined);
+  };
+
+  const dateLabel = getDateRangeLabel({
+    currentLabel: selectedDateMenuOption?.name,
+    startDate,
+    endDate,
+  });
+
+  return (
+    <>
+      <StyledInputDropdown
+        sdsStyle="square"
+        sdsType="singleSelect"
+        label={dateLabel}
+        // @ts-expect-error remove line when inputdropdown types fixed in sds
+        onClick={handleClick}
+      />
+      <StyledDateFilterMenu
+        {...props}
+        anchorEl={anchorEl}
+        onClose={handleClose}
+        onStartDateChange={setStartDate}
+        onEndDateChange={setEndDate}
+        selectedDateMenuOption={selectedDateMenuOption}
+        setSelectedDateMenuOption={setSelectedDateMenuOption}
+        updateDateFilter={updateDateFilter}
+      />
+    </>
+  );
+};
+
+export { CollectionDateFilter };
--- a/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/index.tsx
@@ -1,6 +1,11 @@
 import React from "react";
 import { noop } from "src/common/constants/empty";
+import {
+  MENU_OPTIONS_COLLECTION_DATE,
+  MENU_OPTION_ALL_TIME,
+} from "src/components/DateFilterMenu/constants";
 import { StyledInfoOutlinedIcon, StyledTooltip } from "../../style";
+import { CollectionDateFilter } from "./components/CollectionDateFilter";
 import {
   StyledContainer,
   StyledExplainerTitle,
@@ -37,7 +42,7 @@ const SAMPLE_FILTERING_TOOLTIP_TEXT = (
  * our designs to what's current available. Those are just there to get
  * a version of this out the door, nothing longterm at all there.
  */
-export function SampleFiltering() {
+export function SampleFiltering(): JSX.Element {
   return (
     <StyledContainer>
       <StyledExplainerTitle>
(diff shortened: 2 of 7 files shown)


mmmmmmaya:
added @Janeece as a reviewer for the interaction portion of the dropdown (I know the visual styling for the `InputDropdown` only is off, probably needs to be fixed in sds, but for now you can ignore it)

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/CollectionDateFilter/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+}
+
+const CollectionDateFilter = ({
+  updateDateFilter,
+  ...props
+}: Props): JSX.Element => {
+  // `startDate` and `endDate` represent the active filter dates. Update on filter change.
+  const [startDate, setStartDate] = useState<FormattedDateType>();
+  const [endDate, setEndDate] = useState<FormattedDateType>();
+  const [anchorEl, setAnchorEl] = useState<HTMLElement | undefined>();
```
phoenixAja, line 20: i've seen you use anchorEl multiple times, but i'm still not 100% on what it's doing, do you mind explaining?

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/index.tsx
```diff
@@ -65,12 +70,14 @@ export function SampleFiltering() {
...
-          <StyledInputDropdown
-            disabled={false}
-            label="Dummy placeholder"
-            onClick={noop}
-            sdsStage="userInput"
-            sdsStyle="square"
+          <CollectionDateFilter
+            fieldKeyEnd="collectionDateEnd"
+            fieldKeyStart="collectionDateStart"
+            updateDateFilter={noop}
```
phoenixAja, line 76: why is this noop?

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/CollectionDateFilter/index.tsx
```diff
@@ -0,0 +1,62 @@
+import React, { useState } from "react";
+import { getDateRangeLabel } from "src/common/utils/dateUtils";
+import { MENU_OPTION_ALL_TIME } from "src/components/DateFilterMenu/constants";
+import { StyledDateFilterMenu, StyledInputDropdown } from "./style";
+
+interface Props {
+  fieldKeyEnd: string;
+  fieldKeyStart: string;
+  menuOptions: DateMenuOption[];
+  updateDateFilter: UpdateDateFilterType;
```
phoenixAja, line 10: what is the `updateDateFilter` prop doing exactly? It's a little hard to infer from the name/usage

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/index.tsx
mmmmmmaya: Because the filter does not actually do anything yet. As I mentioned in the PR notes, I still need to have the filter impact the resulting tree. That is in the next PR

src/frontend/src/components/DateFilterMenu/constants.ts
```diff
@@ -37,3 +37,7 @@ export const MENU_OPTIONS_COLLECTION_DATE: DateMenuOption[] = [
     numDaysStartOffset: 365,
   },
 ];
+
+export const MENU_OPTION_ALL_TIME: DateMenuOption = {
+  name: "All time",
```
phoenixAja: should this be `All Time`? (haven't looked at the design spec, jw)

mmmmmmaya: yep!
```suggestion
  name: "All Time",
```

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/index.tsx
phoenixAja: ok got it! thanks

phoenixAja (approved):

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/CollectionDateFilter/index.tsx
mmmmmmaya: InputDropdown is an element that is always inline in the page. The menu sometimes appears in the page, but sometimes doesn't. When it _does_ appear, it does not occur inline with the other elements, it floats on top. Because it's a floating element and can't position itself based on the other elements around it, it needs to know the exact location on the page where it is supposed to float. The `anchorEl` (InputDropdown) is the element that the menu uses to anchor itself. That is, it's the element that the menu uses to position itself (by relatively offsetting its own position based on the position of the anchor).

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/CollectionDateFilter/index.tsx
mmmmmmaya: In the original implementation, the `updateDateFilter` prop was used to register updates to the state of a single filter (eg, collection date range) with the filter state manager. For example, the filter panel for sample table has one logical filtering manager that figures out how to combine all the active filters into a set of filtering rules that yield only those samples that match all the required restrictions. 

However, this case is simpler and doesn't require a filter manager. I will rename this to something more generic when  I do the next PR.

Janeece:
> added @Janeece as a reviewer for the interaction portion of the dropdown (I know the visual styling for the `InputDropdown` only is off, probably needs to be fixed in sds, but for now you can ignore it)

@mmmmmmaya behavior looks perfect, thank you!

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/CollectionDateFilter/index.tsx
phoenixAja: i wonder why i've never had to use it. what do you mean by `The menu sometimes appears in the page, but sometimes doesn't` you mean visually when a user opens the modal?

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/CollectionDateFilter/index.tsx
phoenixAja: ah ok got it, thanks for explaining

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/CollectionDateFilter/index.tsx
mmmmmmaya: > you mean visually when a user opens the modal?

Yep!

If you use and SDS complexfilter or dropdown, you don't need to use it because it's an implementation detail that is handled by the library. In this case, we're implementing the two pieces separately so we control the relationship between the two.

vincent-czi: @phoenixAja If you open the developer console and try to follow the elements, you'll see that the thing opening the menu and the menu itself are on totally different chains of DOM nodes. The menu isn't exactly owned in the traditional `<div><span> I am a child of span and a grandchild of div!! </span></div>` kind of way. The menu just lives outside of that chain of nodes entirely, waiting to be called into existence, then gets anchored to the page at the appropriate spot by that anchorEl.

PR #1041, opened 2022-04-21
Update custom colors

danrlu:
### Summary:
- **What:** Make the colors all very distinct from each other. See what you think~ 

### Demos:
image

image

### Checklist:
- [ ] I added labels to my PR


Janeece:
@danrlu Nice, these look very distinct. The only potentially troublesome color that stands out to me is the very light yellow one — this color will become invisible on a lot of monitors. 

Edit for solution: just needs to be a darker color (yellow or other)

danrlu:
Very good point. What about the brown swap (above)? If not good I'll try black and white before pink lol @Janeece

lvreynoso (approved):

Janeece:
@danrlu that looks great!

PR #1044, opened 2022-04-26
duplicate ids error

phoenixAja:
### Summary:
- **What:** throw error when user uploads a file with duplicate public identifiers or new private identifiers
- **Ticket:** [ part of sc180927](https://app.shortcut.com/genepi/story/180927)
- **Env:** `<rdev link>`

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/parseFile.ts
```diff
@@ -196,8 +212,19 @@ export function parseFileEdit(
           uploadedHeaders,
           HEADERS_TO_SAMPLE_EDIT_METADATA_KEYS
         );
-        if (missingHeaderFields) {
+        const duplicatePublicIds = getDuplicateIds(rows, "publicId");
+        const duplicatePrivateIds = getDuplicateIds(rows, "newPrivateID");
+
+        if (missingHeaderFields || duplicatePublicIds || duplicatePrivateIds) {
```
mmmmmmaya: I find it a little bit weird that there is one if statement for all the errors. I would expect one conditional per message.

src/frontend/src/views/Upload/components/Samples/components/AlertTable/index.tsx
```diff
@@ -19,6 +19,8 @@ const ERROR_CODE_MESSAGES: Record<ERROR_CODE, string> = {
   [ERROR_CODE.MISSING_FIELD]: "placeholder",
   [ERROR_CODE.OVER_MAX_SAMPLES]:
     "This file contains more than 500 samples, which exceeds the maximum for each upload process. Please limit the samples to 500 or less",
+  [ERROR_CODE.DUPLICATE_PUBLIC_IDS]: "placeholder",
```
mmmmmmaya: Do you want me to fill these parts in?

src/frontend/src/common/components/library/data_subview/components/EditSamplesConfirmationModal/components/ImportFile/parseFile.ts
phoenixAja: i rewrote this to be less wierd

src/frontend/src/views/Upload/components/Samples/components/AlertTable/index.tsx
phoenixAja: i'll write a comment for this, we actually don't need this to be filled in for `AlertTable` as this was used in the upload flow when uploading samples from a fasta. I only added these here to get around the type errors that i was getting from not providing these values. Same thing is happening in `src/frontend/src/views/Upload/components/Metadata/components/ImportFile/components/Alerts/Error.tsx` where i added the two new errors with placeholder even though the Error component does not handle the duplicateIds case. (the duplicate IDs error is different enough that it doesn't make sense to share the default `Error` components)

phoenixAja: i was trying not to get into type hell by changing the ERRORCODE type, but if you think i should go that route then i can, otherwise i can add some comments to explain what's happening

phoenixAja: i'll change the type, update coming soon

phoenixAja: ok type updated, was not as bad as i initially thought

PR #1053, opened 2022-04-27
Don't use the old usergroup api endpoint.

jgadling:
### Summary:
- **What:** don't rely on the old `/api/userdata` API endpoint anymore, switch to the new fastapi equivalent.
- **Ticket:** [sc194360](https://app.shortcut.com/genepi/story/194360)
- **Env:** https://newuserdata-frontend.dev.czgenepi.org/data/phylogenetic_trees

### Notes:
This is also an initial experiment in handling response deserialization without our old & increasingly difficult-to-maintain apiResponse converter.

### Checklist:
- [x] I added labels to my PR


src/frontend/src/common/queries/auth.ts
```diff
@@ -29,32 +24,28 @@ export const USE_USERDATA = {
   id: "userData",
 };
 
-export interface UserResponse extends APIResponse {
-  group: Group;
-  user: User;
-}
+const mapUserData = (obj: any): User => {
+  return <User>{
```
vincent-czi: I didn't actually know this was a way to write TS until now? Stating the type of the object with angle brackets up against the opening curly brace is new to me. I'm not sure it's necessary though since the function's return type already tells you that the returned object is a `User`. If it's complaining when you take out the `<User>` here, there might be something else going on, like maybe it's actually another way to do a type assertion?

src/frontend/src/common/queries/auth.ts
```diff
@@ -29,32 +24,28 @@ export const USE_USERDATA = {
...
-const USER_MAP = new Map<string, keyof User>([
-  ["auth0_user_id", "auth0UserId"],
-  ["group_admin", "groupAdmin"],
-  ["system_admin", "systemAdmin"],
-  ["group_id", "groupId"],
-  ["agreed_to_tos", "agreedToTos"],
-  ["acknowledged_policy_version", "acknowledgedPolicyVersion"],
-]);
+function mapGroupData(obj: any): Group {
+  return <Group>{
```
vincent-czi: Same thing with angle bracket type identification

src/frontend/src/common/queries/auth.ts
```diff
@@ -29,32 +24,28 @@ export const USE_USERDATA = {
...
+    id: obj.id,
+    name: obj.name,
+  };
+}
 
-const fetchUserData = (): Promise<V2User> => {
+export const fetchUserInfo = (): Promise<User> => {
```
vincent-czi: I believe this type claim is actually wrong, but won't be discovered by TS compiler. The result from `fetchUserInfo` will be an object of the raw JSON returned from the BE route, but that's not a `User` per se. For example, the thing returned here will have `acknowledged_policy_version` rather than `acknowledgedPolicyVersion` since it hasn't been run through your mapping func yet.

Bigger picture, maybe we want to have an intermediary interface for the "raw" object, _before_ we run the mapping/deserializer/parser func. Something like `UserRawResponse`? In a perfect world, we would have a convenient way to connect what's in our BE `Foobar` response schema to the equivalent `FoobarRawResponse` interface over here in the FE, but for now we can just duplicate manually. I know I'm kind of waffling now on what I said earlier about not really caring what the actual fetch calls return since it's outside the scope of what TS can "see" b/c it comes from external source, but I don't really know if we can get around it unless we're willing to just go ahead and say `fetchUserInfo = (): Promise<any>`.

src/frontend/src/common/queries/auth.ts
```diff
@@ -121,7 +107,7 @@ export function useProtectedRoute(): UseQueryResult<UserResponse, unknown> {
   useEffect(() => {
     // Wait for the `useUserInfo` call to complete
     if (!isLoading) {
-      const agreedToTOS = data?.user?.agreedToTos;
+      const agreedToTOS = data?.agreedToTos;
```
vincent-czi: Nit: I'd prefer to rename the `data` var during destructuring of the `result` above to something that makes the data it has clearer.
```
const { isLoading, data } = result;
\/\/\/\/\/
const { isLoading, data: userInfoData } = result;
```

src/frontend/src/components/NavBar/components/RightNav/index.tsx
```diff
@@ -6,9 +6,7 @@ import UserMenu from "./components/UserMenu";
 import { UploadButton } from "./style";
 
 export default function RightNav(): JSX.Element {
-  const { data } = useUserInfo();
-
-  const user = data?.user;
+  const { user } = useUserInfo();
```
vincent-czi: I could be wrong, but I don't think this works? I'm pretty sure the `useUserInfo` react-query thing will still have `data` as its top level "the result from the API (after optional post-process via `select`)", it's just that now the `data` will be a `User`. So it's now `user = data`, rather than `user = data?.user`.

Although -- assuming I'm right -- I would have the destructuring step also do the rename like in that other comment:
```
const { data: userInfoData } = useUserInfo();
```

Like I said, I could be wrong, but that appears to be how it works with a similar thing I did for named locations in src/frontend/src/views/Upload/index.tsx:
```
const { data: namedLocationsData } = useNamedLocations();
```

src/frontend/src/common/types/user.d.ts
```diff
@@ -1,36 +1,17 @@
 interface Group {
   [index: string]: JSONPrimitive;
-  type: "Group";
-  address: string;
-  email: string;
+  type: "V2Group";
```
vincent-czi: This is the part I'm most unsure what's happening. I _think_ it's a claim that a `Group` interface has a `type` property that is always the constant string of "V2Group" (and similar for `type: "User"` below). However, I don't see how that would actually get into the runtime object of any of the Groups (or Users, for that matter). I think the old `apiResponse` smushed it in somehow at some point... maybe? But I don't think that's happening for the new approach you're using with the `mapUserData`.

Can you try console.log-ing one of the `data` results from your new work and seeing if there's a `type` in there? I don't think there will be, but maybe I'm missing something (or if you're sure it's there, do you know what I'm missing?). Also kinda curious if the old `apiResponse` actually has the `type` key in there. In any case, I don't think we actually use the `type` value anywhere and I'm not sure what the reasoning behind having them in the interfaces was (maybe it was there a possible way to later set up [user-defined type guards](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)? I dunno).

src/frontend/src/common/queries/auth.ts
vincent-czi: Okay, I spent some time looking into this a bit more, and I don't think this is doing what's intended. Messing around with it locally, I'm seeing some weird stuff.

Here's what I have in my IDE
```
interface MyInterface {
  x: string;
  y: string;
}

const a = <MyInterface>{ // No error, TS is happy as a clam, even though it fails to satisfy MyInterface
  x: "blah"
};
const b: MyInterface = { // Error as expected ("Property 'y' is missing...", what you'd expect)
  x: "blah",
};
const c = <MyInterface>{ // Error, but uh, what? (I'll paste the error below, it's long)
  x: "blah",
  z: "bleh",
};
```

The error for `c` above:
```
Conversion of type '{ x: string; z: string; }' to type 'MyInterface' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Property 'y' is missing in type '{ x: string; z: string; }' but required in type 'MyInterface'.
```

So I don't think the `<Blah>{object}` syntax is a type assertion, but `a` should fail and it didn't, so something weird is definitely going on?

Also, this syntax will blow up if you're writing `.tsx` files because it will try to interpret it as JSX and the `<Blah>` isn't actually a React component tag (or at least it blew up on me at first until I moved my back of the envelope check to a `.ts` file instead of the`.tsx` I had open in my editor when I started messing around)

vincent-czi (commented):
Overall, I'm a fan of this approach. There's some TS-syntax tweaks I think it needs and a couple of minor nits, but big picture, I think this works very well.

src/frontend/src/common/queries/auth.ts
jgadling: Welp, running the linter/formatter on this code refactored it one way or another 😢

src/frontend/src/common/types/user.d.ts
jgadling: I think these extra fields were just present for compatibility with the auto-interface-converter that we use everywhere else. I've removed them.

src/frontend/src/common/queries/auth.ts
jgadling: I think all of this stuff (type de/serialization, error handling, etc etc) needs to be encapsulated in an API client library in the future, but for now, I'll add the intermediate interface just for clarity.

vincent-czi: I don't disagree about encapsulating it, but won't we still need to define that intermediate interface somewhere? (unless we can have a shared interface against the BE response schemas) We need to define somewhere what that initial fetched JSON looks like, right? ... Although, writing that out, maybe that deserializer step would handle it because it's going to parse the JSON and it either raises an error when it doesn't matched expected structure or it will match the expected structure, so the "raw" JSON interface aspect (what I called `FoobarRawResponse`) is implicit in the parser code?

src/frontend/src/common/queries/auth.ts
```diff
@@ -36,17 +31,17 @@ const mapUserData = (obj: any): User => {
...
 function mapGroupData(obj: any): Group {
-  return <Group>{
+  return {
     id: obj.id,
     name: obj.name,
   };
 }
 
-export const fetchUserInfo = (): Promise<User> => {
+export const fetchUserInfo = (): Promise<RawUserRequest> => {
```
vincent-czi, line 54: I see now that this type gets used both for the response from a GET and also the payload structure for a PUT. Maybe a better name would just be `RawUser` or something since it's used both in request and response

src/frontend/src/common/types/user.d.ts
```diff
@@ -1,17 +1,27 @@
...
   id: number;
   name: string;
-  group: V2Group;
+  group: Group;
   agreedToTos: boolean;
   acknowledgedPolicyVersion: string | null; // Date or null in DB. ISO 8601: "YYYY-MM-DD"
   splitId: string;
 }
+
+interface RawGroupRequest {
```
vincent-czi: Nit: I would put the `RawBlah` interfaces in the same file as where the API calling functions are since I expect those are the only functions that need to know the interface. I think of the `*.d.ts` files as being more about defining common interfaces that are used throughout the codebase. If the interface/type is restricted to only being used in a couple files, it feels like we normally just define them in one and export/import. It's a bit fuzzy, and I'd defer to Maya's opinion, I've never really thought this much about it since I almost never add new types to the `*.d.ts` files

jgadling: I moved them for now!

src/frontend/src/views/LandingPageV2/index.tsx
```diff
@@ -12,7 +12,7 @@ import UseCases from "./components/UseCases";
 import { Container, LoadingText } from "./style";
 
 export default function Homepage(): JSX.Element {
-  const { data: userInfo, isLoading } = useUserInfo();
+  const { data: userData, isLoading } = useUserInfo();
```
mmmmmmaya: I see sometimes you renamed to `user`, sometimes to `userData` and other times to `userInfo`. What's the pattern you used?

mmmmmmaya (approved):

src/frontend/src/views/LandingPageV2/index.tsx
jgadling: There is no pattern 😅 -- I prefer the plain `user` but not sure if you have a different opinion? I'll update them all to be the same.

mmmmmmaya: I don't have a preference

PR #1055, opened 2022-04-28
Migrate policy update endpoint to fastapi

jgadling:
### Summary:
- **What:** Create a new fastapi endpoint for updating user tos / version acceptance info
- **Ticket:** [sc183176](https://app.shortcut.com/genepi/story/183176)
- **Env:** `<rdev link>`

### Checklist:
- [x] I added labels to my PR


src/backend/aspen/api/views/users.py
```diff
@@ -14,3 +14,15 @@ async def get_current_user(
...
+
+
+@router.put("/me", response_model=UserMeResponse)
+async def update_user_info(
+    user_update_request: UserUpdateRequest,
+    db: AsyncSession = Depends(get_db),
+    user=Depends(get_auth_user),
+) -> UserMeResponse:
+    user.agreed_to_tos = user_update_request.agreed_to_tos
+    user.acknowledged_policy_version = user_update_request.acknowledged_policy_version
```
vincent-czi: I see you updated the FE `AcknowledgePolicyChanges` to send both, but I think I prefer the approach of having the various update fields be optional since since pretty soon we're going to allow the user to update other (new) attributes about themself (like that Bio submitting number for GISAID ingest, etc). If you feel strongly that it should be both pieces required and together, maybe make this a sub-route, like `/me/policies`, so that we can update other parts of the user later without any concern for policy agreement stuff.

vincent-czi (approved):
Overall, looks good! I've got one minor quibble about making the UserUpdateRequest have all fields be required, but I wouldn't really be bothered if you just merged it right now.

lvreynoso (approved):

src/backend/aspen/api/views/users.py
jgadling: Fair enough. Updated to make all fields optional

vincent-czi: Great, thanks!

PR #1058, opened 2022-04-28
Maya/refactor 2

mmmmmmaya:
### Summary
- **What:**
  - Restrict state management by child components
  - Inline some small subcomponents to improve readability
  - Rename modal-closing related event handlers so the logic is easier to follow and match used apis
- **Why:** In general, the parent component should specify an interface for a callback, rather than directly exposing every state management function. This way, if the parent wants to refactor its internals, the children don't need to worry about it. Also, it limits the number of props you need to pass down, and makes children components easier to reason about.

### Demos
No visual changes.

### Notes
Read this PR by viewing each commit on its own. The changes are atomic per commit.

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR


phoenixAja (approved):
this is nice

PR #1063, opened 2022-04-30
upgrade react csv

phoenixAja:
### Summary:
- **What:** [old version of react-csv was not properly escaping quote characters](https://github.com/react-csv/react-csv/issues/241), which was messing up our tsv template downloads 
- **Ticket:** [sc195535](https://app.shortcut.com/genepi/story/195535)
- **Env:** no rdev

### Notes:
checked that all tsv downloads were working as expected

### Checklist:
- [x] I added labels to my PR


src/frontend/src/components/DownloadMetadataTemplate/index.tsx
```diff
@@ -45,6 +45,7 @@ const SampleEditTsvTemplateDownload = ({
     <CSVLink
       className={className}
       filename="metadata_template.tsv"
+      enclosingCharacter={""}
```
vincent-czi, line 49: So if I'm reading this correctly, we're saying that the enclosingCharacter -- for when we need to wrap something that would otherwise have a control character, like `,` in CSV -- is the empty string? So we don't ever wrap anything? I think this works in our current case because we're downloading TSVs and we're not lunatics with tabs in our data, but I am a little worried about the possibility of copy-pasting this code somewhere else and things breaking: say if someone needs to support downloading a CSV and some of the fields have commas. I think this would be good to have a comment nearby with a warning saying that enclosing characters have been completely disabled because we don't need them in this case and we need no interference with comments.

vincent-czi, line 49: So if I'm reading this correctly, we're saying that the enclosingCharacter -- for when we need to wrap something that would otherwise have a control character, like `,` in CSV -- is the empty string? So we don't ever wrap anything? I think this works in our current case because we're downloading TSVs and we're not lunatics with tabs in our data, but I am a little worried about the possibility of copy-pasting this code somewhere else and things breaking: say if someone needs to support downloading a CSV and some of the fields have commas. I think this would be good to have a comment nearby with a warning saying that enclosing characters have been completely disabled because we don't need them in this case and we need no interference with comments.

vincent-czi (approved):

src/frontend/src/components/DownloadMetadataTemplate/index.tsx
phoenixAja: i think you're confusing the `enclosingCharacter` prop with the `separator` prop? enclosingCharacter means that we want to be downloading raw strings without being wrapped in `"` characters.

vincent-czi: I don't think I'm wrong: the `"` enclosing characters in CSV or TSV are necessary to handle raw strings that contain the separator (like `,` in CSV). Possible I'm wrong about what it means in this library, but we can't have raw strings in all cases since the raw string can contain a control character. I _think_ this basically says to always dump raw strings only, but that's dangerous in some situations (not ours now, but could break on others).

phoenixAja: oh ok i get it now.. i can add a comment to be aware when csv or tsv contains control characters

phoenixAja: i'm pretty sure the library handle those cases, but i'll add the comment since i haven't tested

PR #1066, opened 2022-05-03
Automate scheduled group runs

lvreynoso:
### Summary:
- **What:** Adds a scheduled (daily) batch job that launches an overview tree build for each group in the database, depending on the schedule expression in the db. Adds a JSONB column to the groups table, `tree_parameters`, to allow flexibility around scheduled trees that are built for a group.
- **Ticket:** [sc159631](https://app.shortcut.com/genepi/story/159631)
- **Env:** [https://autorun-frontend.dev.czgenepi.org/](https://autorun-frontend.dev.czgenepi.org/)

### Demos:
The actual tree builds failed as expected due to the rdev db lacking needed information, but otherwise the coordinated launch worked:
Screen Shot 2022-05-03 at 13 13 35

### Notes:
Groups without a defined `schedule_expression` key in their `tree_parameters` json blob will have a tree built for them every time the master autorun job runs. To make sure a group does not have any trees automatically built for them, the key's value must be an empty list, e.g. `tree_parameters` would look like `{ schedule_expression: [] }`. To schedule a group's runs on certain days of the week, the list should contain ints for the desired days, where Monday = 0 and Sunday = 6. E.g. for Tuesday & Thursday builds `tree_parameters` would look like `{ schedule_expression: [1, 3] }`.

### Checklist:
- [x] I added labels to my PR


src/backend/aspen/workflows/nextstrain_run/autorun_scheduled.py
```diff
@@ -0,0 +1,92 @@
...
+def launch_scheduled_run(aws_client, sfn_params: Dict, group: Group):
+    start_datetime = datetime.datetime.now()
+
+    sfn_params["Input"]["Run"] |= {
+        "group_name": group.name,
+        "s3_filestem": f"{group.prefix}/scheduled",
+        "template_args": json.dumps(TEMPLATE_ARGS),
+        "tree_type": SCHEDULED_TREE_TYPE,
+    }
+    sfn_params[
```
jgadling, line 40: I think a comment about why it's important to make output_prefix unique here would be good!

src/backend/aspen/api/settings.py
```diff
@@ -122,6 +122,7 @@ class Settings(BaseSettings):
     # Env vars usually pulled from AWS SSM Parameters
     AWS_NEXTSTRAIN_SFN_PARAMETERS: Dict
     AWS_PANGOLIN_SFN_PARAMETERS: Dict
+    AWS_NEXTSTRAIN_SCHEDULED_SFN_PARAMETERS: Dict
```
jgadling: I don't think we need any of the changes to this file?

src/backend/aspen/workflows/nextstrain_run/run_nextstrain_autorun.sh
```diff
@@ -0,0 +1,29 @@
...
+
+start_time=$(date +%s)
+
+aws configure set region $AWS_REGION
+
+if [ -n "${BOTO_ENDPOINT_URL-}" ]; then
+  export aws="aws --endpoint-url ${BOTO_ENDPOINT_URL}"
+else
+  export aws="aws"
+fi
```
jgadling, line 23: Not relevant to this PR specifically, but I think we need to look for a way to reduce all the copy/paste boilerplate in our wdl and bash wrappers.

jgadling (changes requested):
let's clean up the unused changes to the settings module

jgadling (approved):

PR #1076, opened 2022-05-09
Don't json-quote pango_lineages; add debug option to dump yaml config

lvreynoso:
### Summary:
- **What:** Fixes a bug in nextstrain builds where augur did not accept our pango lineage filters as formatted in the builds yaml
- **Env:** [https://pangoarray-frontend.dev.czgenepi.org/](https://pangoarray-frontend.dev.czgenepi.org/)

### Notes:
Added a debug option to nextstrain run's `export.py` that only writes out a `builds.yaml`.

YAML output now looks like this, an array:
Screen Shot 2022-05-10 at 13 34 33

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/workflows/nextstrain_run/build_config.py
+++ b/src/backend/aspen/workflows/nextstrain_run/build_config.py
@@ -1,4 +1,3 @@
-import json
 import re
 from math import ceil
 
--- a/src/backend/aspen/workflows/nextstrain_run/build_config.py
+++ b/src/backend/aspen/workflows/nextstrain_run/build_config.py
@@ -165,8 +165,9 @@ def apply_filters(config, subsampling, template_args):
 
     pango_lineages = template_args.get("filter_pango_lineages")
     if pango_lineages:
-        # Techically pango_lineages should be a *python* encoded list, but we're
-        # cheating since json is interoperable as long as we remove bad characters
+        # Nextstrain is rather particular about the acceptable syntax for
+        # values in the pango_lineages key. Before modifying please see
+        # https://discussion.nextstrain.org/t/failure-when-specifying-multiple-pango-lineages-in-a-build/670
         clean_values = [re.sub(r"[^0-9a-zA-Z.]", "", item) for item in pango_lineages]
         config["builds"]["aspen"]["pango_lineage"] = clean_values
         # Remove the last " from our old query so we can inject more filters
--- a/src/backend/aspen/workflows/nextstrain_run/build_config.py
+++ b/src/backend/aspen/workflows/nextstrain_run/build_config.py
@@ -169,7 +169,7 @@ def apply_filters(config, subsampling, template_args):
         # Techically pango_lineages should be a *python* encoded list, but we're
         # cheating since json is interoperable as long as we remove bad characters
         clean_values = [re.sub(r"[^0-9a-zA-Z.]", "", item) for item in pango_lineages]
-        config["builds"]["aspen"]["pango_lineage"] = json.dumps(clean_values)
+        config["builds"]["aspen"]["pango_lineage"] = clean_values
         # Remove the last " from our old query so we can inject more filters
         old_query = subsampling["group"]["query"][:-1]
         pango_query = " & (pango_lineage in {pango_lineage})"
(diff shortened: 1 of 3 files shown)


src/backend/aspen/workflows/nextstrain_run/build_config.py
```diff
@@ -166,10 +165,11 @@ def apply_filters(config, subsampling, template_args):
 
     pango_lineages = template_args.get("filter_pango_lineages")
     if pango_lineages:
-        # Techically pango_lineages should be a *python* encoded list, but we're
-        # cheating since json is interoperable as long as we remove bad characters
+        # Nextstrain is rather particular about the acceptable syntax for
```
vincent-czi, line 168: This is such a good comment to have here

src/backend/aspen/workflows/nextstrain_run/export.py
```diff
@@ -74,15 +74,21 @@
     help="Should the status of this workflow be set to 'STARTED'?",
 )
 @click.option("--test", type=bool, is_flag=True)
+@click.option("--dump-config", type=bool, is_flag=True)
```
vincent-czi: So to be clear, this flag will _only_ print the config, right? It does not go on to also do the run? If that's the case, could you tweak the option to being `--dump-config-only` so it's explicit?

src/backend/aspen/workflows/nextstrain_run/export.py
```diff
@@ -97,6 +103,32 @@ def cli(
...
+        context = {
+            "num_sequences": num_sequences,
+            "num_included_samples": num_included_samples,
+        }
+        builder: BaseNextstrainConfigBuilder = builder_factory(
+            phylo_run.tree_type, group, phylo_run.template_args, **context
+        )
+        builder.write_file(builds_file_fh)
+
+        return f"YAML Build Config dumped to {builds_file_fh.name}"
```
vincent-czi: Nit: Since this is intended only ever to be printed/logged, I'd just `print` it here and have no return on the function. When I first read the cli block
```
    if dump_config:
        msg = dump_yaml_template(phylo_run_id, builds_file_fh)
        print(msg)
        return
```
I had assumed `msg` was actually the resulting YAML string, then getting printed up there. Since I can't see any other reason this returned value would be used than to print it, I think it makes more sense to just print it here and have the caller do nothing other than call the function.

vincent-czi (approved):
Code wise, looks great! Some minor suggestions, but not terribly important.

Were you able to successfully run and generate a tree with the code change, or does that still need to be tested?

src/backend/aspen/workflows/nextstrain_run/export.py
```diff
@@ -97,6 +103,32 @@ def cli(
     print(json.dumps(aligned_gisaid_dump))
 
 
+def dump_yaml_template(
```
danrlu, line 106: (sorry wrong question earlier). When do I use this one: https://github.com/chanzuckerberg/czgenepi/blob/48132693794ca0047b2a4a2b129d5b4854a939b8/src/backend/aspen/workflows/nextstrain_run/run_nextstrain_ondemand.sh#L51-L52
and when do I use this `dump_yaml_template` function?

lvreynoso: I'm not sure what your confusion is, they do very different things. `dump_yaml_template()` generates a `builds.yaml` file. It is only useful for local debugging, like in my case I didn't have any sequence or metadata files, but I still wanted to output a `builds.yaml` since we were debugging that. This line you highlighted is an s3 command that copies a `builds.yaml` file to s3.

src/backend/aspen/workflows/nextstrain_run/export.py
lvreynoso: Sure

src/backend/aspen/workflows/nextstrain_run/export.py
danrlu: Got it! So the `dump_yaml_template` is just for debugging. The other one is part of a regular run. I'll put it into my wiki. Thanks Lucia <3

PR #1085, opened 2022-05-13
allow no smaple tress for overview and noncontextualized trees

mmmmmmaya:
### Summary
- **What:** Allow user to create non-contextulaized and overview trees without providing any sample IDs
- **Ticket:** [[sc-196733]](https://app.shortcut.com/genepi/story/196733)
- **Env:** https://nosampletree-frontend.dev.czgenepi.org/

### Demos
after

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
@@ -74,7 +74,6 @@ export const CreateNSTreeModal = ({
   const [validatedInputSamples, setValidatedInputSamples] = useState<string[]>(
     []
   );
-  const [isValidTreeType, setIsValidTreeType] = useState<boolean>(false);
 
   // Certain tree types can filter based on lineages and date ranges
   const { data: lineagesData } = useLineages();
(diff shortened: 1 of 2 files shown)


danrlu (approved):
The demo is so cooooooooool!!! Thanks Maya!

danrlu:
Oh forgot to mention, I cannot read the code, but the behavior is exactly what we want

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
```diff
@@ -87,19 +86,10 @@ export const CreateNSTreeModal = ({
     if (shouldReset) setShouldReset(false);
   }, [shouldReset]);
 
-  useEffect(() => {
-    if (treeType !== undefined && Object.values(TreeTypes).includes(treeType)) {
```
vincent-czi, line 91: I just want to double-check: we're removing the check that the provided `treeType` is actually one of the existing TreeTypes we have defined, right? I don't see this moved over to a new runtime check, although the check is still mostly present because of the downstream Props type check with `treeType: TreeType | undefined`.

I am totally fine with this -- I think it's good and reasonable! -- but wanted to make sure I am reading things correctly.

vincent-czi (approved):
Looks good. Not a blocker, just curious: is the downstream BE kicking off a tree build when there are no samples also working as expected?

danrlu:
by code yes, and i always test tree changes in staging after PRs are in so we shall see ;)

Janeece:
Agreed with Dan, the demo is so great! I'm late to the review party but it looks good!

PR #1088, opened 2022-05-16
edit cells highlighting

phoenixAja:
### Summary:
- **What:** cells are highlighted when in edit state
- **Ticket:** [sc180630](https://app.shortcut.com/genepi/story/180630)
- **Env:** `<rdev link>`

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/frontend/src/components/DateField/index.tsx
```diff
@@ -14,19 +14,41 @@ interface Props {
...
+    setInitialValue(initialValues[fieldKey]);
+  }, [fieldKey, initialValues, values]);
+
+  useEffect(() => {
+    if (initialValue !== changedValue && shouldShowEditedCellsAsMarked) {
+      setBackgroundColorShown(true);
+    } else {
+      setBackgroundColorShown(false);
+    }
+  }, [initialValue, changedValue, shouldShowEditedCellsAsMarked]);
```
phoenixAja: these hooks are reused between the `DateField`, `FreeTextField`, and `LocationField`, is there a more dry way to do this? (@mmmmmmaya @vincent-czi )

mmmmmmaya: you can turn it into a custom hook. https://reactjs.org/docs/hooks-custom.html

vincent-czi: Does the `initialValue` change over time? I would think that would be constant for the entirety of time the Edit modal is open, because it's what the metadata started as before editing process begins. So you could get rid of that state, possibly move it up to props, but if you need the overall `initialValues` object at this level, just have it pull as a constant on every render or `useMemo` it.

Similarly -- and I don't know enough to be sure, haven't read PR or have a ton of context -- it seems like this doesn't need to be done with a useEffect or have a state for the background color. Just do a comparison like that if-conditional, and pass the boolean as a prop to the `StyledBlah` component.

As for underlying question though, I don't really know how doable it is to extract cleanly and generalize to all the various Fields. This comparison will probably have to happen inside each of these Fields since that's where the comparison can have access to the `currentValue`, right? I guess you could use the formik at the parent level and compare up there? Not sure that's really so much better. But if you shrink the usage of state and effect hooks, the boilerplate aspect will probably be less cumbersome.

phoenixAja: @vincent-czi yeah you're right the initialValue does not change over time, so i could get rid of that state. the UseEffect i think is necessary though for when the user changes the value (changedValue) that would trigger the color change

src/frontend/src/components/DateField/style.ts
```diff
@@ -1,14 +1,26 @@
...
+})`
+  ${(props: FormFieldProps) => {
+    const spaces = getSpaces(props);
+    const colors = getColors(props);
+    const { isBackgroundColorShown } = props;
     return `
       width: ${(spaces?.l || 0) + 120}px;
       margin: 0;
+      input {
+        background-color: ${isBackgroundColorShown && colors?.primary[200]}
```
mmmmmmaya:
```suggestion
        background-color: ${isBackgroundColorShown && colors?.primary[200]};
```

src/frontend/src/components/DateField/index.tsx
vincent-czi: I'm still not sure why the useEffect has to happen though? That seems like it could just be determined on every render which would happen with `changedValue` changing. If `changedValue` changes, it will kick off a re-render, and you can just have a comparison in "normal" code without a useEffect hook.

src/frontend/src/components/DateField/index.tsx
```diff
@@ -14,19 +14,41 @@ interface Props {
...
 }
 
+export type valueType = string | boolean | NamedGisaidLocation | undefined;
+
 export default function DateField({
   fieldKey,
   formik,
   helperText,
   onChange,
+  shouldShowEditedCellsAsMarked,
```
mmmmmmaya: This prop name implies that the date field component knows about the concept of a "cell", which it shouldn't because it is an input field which could totally exist outside of a table. I'm wondering if there's a more generic name that you can think of for this prop?

src/frontend/src/components/DateField/index.tsx
```diff
@@ -14,19 +14,41 @@ interface Props {
...
 }: Props): JSX.Element {
-  const { handleChange, handleBlur, values, touched, errors } = formik;
+  const { handleChange, handleBlur, values, touched, errors, initialValues } =
+    formik;
+  const [isBackgroundColorShown, setBackgroundColorShown] =
```
vincent-czi: Nit: I want us to have a convention for React state values and their setters that is to blindly follow the `fooBar` and `setFooBar` structure. I know it's cumbersomely long, but I think the consistency of `setIsBackgroundColorShown` is worthwhile

mmmmmmaya: +1

src/frontend/src/components/WebformTable/components/Row/components/FreeTextField/index.tsx
```diff
@@ -1,20 +1,41 @@
...
+import React, { useEffect, useState } from "react";
+import { valueType } from "src/components/DateField";
 import { Metadata } from "src/components/WebformTable/common/types";
 import { StyledTextField } from "./style";
 
 interface Props {
   fieldKey: keyof Metadata;
   formik: FormikContextType<Metadata>;
   isShown?: boolean;
+  shouldShowEditedCellsAsMarked?: boolean; // used to mark edited cells as purple for crud
```
mmmmmmaya: same prop name comment

src/frontend/src/components/WebformTable/components/Row/components/FreeTextField/style.ts
```diff
@@ -1,15 +1,26 @@
...
+    const { isBackgroundColorShown } = props;
     const spaces = getSpaces(props);
+    const colors = getColors(props);
 
     return `
       width: ${(spaces?.l || 0) + 200}px;
       padding-right: ${spaces?.l}px;
       margin: 0;
+      input {
+        background-color: ${isBackgroundColorShown && colors?.primary[200]}
```
mmmmmmaya:
```suggestion
        background-color: ${isBackgroundColorShown && colors?.primary[200]};
```

src/frontend/src/components/WebformTable/components/Row/components/LocationField/index.tsx
```diff
@@ -14,6 +15,7 @@ interface Props {
   applyToAllColumn: (fieldKey: keyof Metadata, value: unknown) => void;
   isFirstRow: boolean;
   locations: NamedGisaidLocation[];
+  shouldShowEditedCellsAsMarked?: boolean; // used to mark edited cells as purple for crud
```
mmmmmmaya: same

mmmmmmaya:
My main thing is: When they click continue to the review page, and then click back, the cells lose purple highlighting. But I'm pretty sure I remember Janeece saying that they should stay highlighted until the changes are persisted

src/frontend/src/components/DateField/index.tsx
phoenixAja: yeah i guess this is a little overkill

mmmmmmaya:
Also comment on the approach. I'm not 100% whether this is the best approach or not because I don't have enough context on design direction. Are we going to highlight these cells in other cases/under differing conditions? Or is it possible that in the future we will have many, eg, DateFields that only highlight when their value changes from some initial value? If it's the former (and I would guess that's so), I think the way we need to handle this is by having a parent component determine whether or not the date field needs to be highlighted, and then only passing down a boolean of whether or not highlight is on for the input.

why?

if other inputs will be highlighted under different conditions, then having all this code inside the input to highlight when the value changes doesn't really make sense, and I can't imagine adding a ton of conditionals for something that's mainly supposed to be a pure input.

if the parent wants the child to highlight itself, the parent can decide when to do so by passing down a `isHighlighted` prop, or similar when conditions are right. 

@Janeece can you shed some light on how input highlighting will show up in our app in the future (will there be other circumstances where we highlight cells)?

phoenixAja:
yeah i think it's still helpful though for the users to see the edited cells when they first get onto the page? Maybe we should leave this out then

mmmmmmaya:
I would defer to @Janeece on whether a partial implementation is valuable

phoenixAja:
https://crudv1-frontend.dev.czgenepi.org/data/samples

Janeece:
@mmmmmmaya @phoenixAja we dont have any plans now but I would guess we'd want to highlight changes anywhere we have editing. Only thing coming up that I can think of is editing group or user account details (onboarding V1, V2). If it is recommended, it would be a nice to have on those. 

Since all of this is nice to have, I'll take what I can get. The ideal is to keep the highlights until user saves the changes (even coming back from review step), but only highlighting when they are initially editing would still be beneficial IMO.

src/frontend/src/components/DateField/index.tsx
phoenixAja: i'll change `shouldShowEditedInputAsMarked`

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -152,7 +152,7 @@ module backend_service {
   api_url               = local.backend_url
   frontend_url          = local.frontend_url
   remote_dev_prefix     = local.remote_dev_prefix
-  health_check_path     = "/health"
+  health_check_path     = "/v2/health/"  # the trailing slash here is vital.
```
mmmmmmaya: it looks like something went wrong with your merge?

mmmmmmaya:
Is it intentional trunk is not the base branch on this PR?

phoenixAja:
> Is it intentional trunk is not the base branch on this PR?

yes, i wanted the changes here to be just for the highlighted cells, i branched off of the warnings for uneditable samples branch. I was going to merge in warnings for unedited samples first

.happy/terraform/modules/ecs-stack/main.tf
phoenixAja: i think this is because i merged in my warnings branch and trunk, i would just ignore this as this file won't get overwritten on trunk as the commits exist on this branch and trunk and the files are identical.

phoenixAja:
@mmmmmmaya is this good to merge now? 😄

mmmmmmaya:
yes, as long as my commits look good to you

mmmmmmaya (approved):

phoenixAja:
hooray 🎉 🎉 🎉 i do think it would be good to revisit this logic in the future and send down isHighlighted from a parent component, but we can make a ticket for that and fix later.

PR #1095, opened 2022-05-17
Use group name instead of S3 Prefix for tree names

lvreynoso:
### Summary:
- **What:** Scheduled tree builds are being named according to the s3 prefix for the tree, instead of using the group's name, which is creating some confusing names on the user's end.
- **Env:** `<rdev link>`

### Notes:
Current:
Screen Shot 2022-05-17 at 12 03 05
After PR (and previous behavior):
Screen Shot 2022-05-17 at 12 03 28

### Checklist:
- [ ] I added labels to my PR


danrlu (approved):
Beautiful. Thanks very much Lucia!!

PR #1100, opened 2022-05-18
Auth0 User/Groups sync script

jgadling:
### Summary:
- **What:** An outline for a script to sync users/groups/memberships between our DB & Auth0

### Notes:
This CLI currently supports some basic functionality for syncing Auth0 state to our db, or our db state to Auth0. It has the following capabilities:
- Create/Delete Auth0 users
- Create/Delete Auth0 groups
- Adding Auth0 users to Auth0 Orgs
- Create DB groups
- Adding DB users to DB Groups

And it's **missing** these capabilities (with explanations). Some of these capabilities will likely need to be fleshed out as our group membership and authorization tools evolve, but I *think* the current version does enough to satisfy this current ticket.
- **Delete DB users** - DB users can have any number of samples associated with their user account. Do we want to just null out the "who created this sample" field?
- **Create DB users** - DB users currently *must* have an associated group in order to be created, and figuring out which group to associate a new user with via the Auth0 API is slightly complicated. This should be easy enough to add in if/when we need it.
- **Delete DB groups** -  DB groups have users, samples, trees, etc etc associated with them. We need to be able to delete groups when our users ask us to, but including that functionality in a bulk-operation script feels like a footgun and I don't think it's important right now.
- **Removing DB users from groups** - Much the same issue as not supporting creating new users: `group_id` is a required field on the `users` table, so if we disassociate a user with a group, we need to associate them with another. This just requires a bit more code and we don't need it right now.
- **Handling Auth0 member/admin roles** - This ticket doesn't require any special authorization differences between group members and admins, so we can update it to handle multiple roles when we get to the next ticket.

The current UI looks like this:
```
# ./sync_auth0.py --help
Usage: sync_auth0.py [OPTIONS]

Options:
  --source-of-truth [db|auth0]    Create Auth0 users/groups for any DB
                                  users/groups without Auth0 accounts

  --dry-run                       Whether to only print what actions this
                                  script would take instead of actually
                                  applying changes

  --delete-ok                     Whether to delete objects from the
                                  destination data store if there aren't
                                  matching objects in the source of truth

  --sync-groups / --no-sync-groups
                                  Sync groups
  --sync-users / --no-sync-users  Sync users
  --sync-memberships / --no-sync-memberships
                                  Sync membership
  --help                          Show this message and exit.
```

### Checklist:
- [ ] I added labels to my PR


src/backend/aspen/cli/sync_auth0.py
```diff
@@ -0,0 +1,420 @@
...
+        auth_req = auth0_authentication.GetToken(domain)
+        token = auth_req.client_credentials(
+            client_id, client_secret, f"https://{domain}/api/v2/"
+        )
+
+        self.client = Auth0(domain=domain, token=token["access_token"])
+
+    def get_all_results(self, endpoint, key):
+        # Auth0 paginates results. We don't have a crazy amount of data in auth0 so we can
+        # afford to just paginate through all the results and hold everything in memory.
```
mmmmmmaya, line 64: at what scale would this stop being true?

src/backend/aspen/cli/sync_auth0.py
```diff
@@ -0,0 +1,420 @@
...
+            "password": generate_password(),
+            "connection": "Username-Password-Authentication",
+        }
+        self.client.users.create(body)
+
+    def delete_user(self, auth0_user_id):
+        self.client.users.delete(auth0_user_id)
+
+
+def compare_stuff(auth0_objects, db_objects, auth0_match_field, db_match_field):
```
mmmmmmaya: very technical name 😆

mmmmmmaya (approved):

src/backend/aspen/cli/sync_auth0.py
lvreynoso: yeah it's cute, but if it's comparing user data I'd like the name be more reflective of it

src/backend/aspen/cli/sync_auth0.py
jgadling: As far as memory concerns go, we can handle well into the tens, or even low hundreds of thousands of objects at a time, but I think it's likely that we'll get grouchy earlier than that about all of the up-front latency for making so many API calls to Auth0 before we even start processing. I think we'll might want to revisit the structure of this script in the ~5k-10k users/groups range.

src/backend/aspen/cli/sync_auth0.py
```diff
@@ -0,0 +1,420 @@
...
+            # that "good enough" for the purposes of this sync script. We need to *ALSO*
+            # validate that they have the correct role, but that will be more important
+            # in the next version, so we're skipping that right now.
+            auth0_only, db_only = compare_stuff(
+                auth0_memberships, db_group_users, "user_id", "auth0_user_id"
+            )
+            group_manager = UserGroupManager(
+                self.auth0_client, self.db, self.dry_run, org, db_group
+            )
+            self.update_objects(auth0_only, db_only, group_manager)
```
lvreynoso: Be aware that when we first created the auth0 tenant, the sign-up flow on the user-facing auth0 login page was enabled (presumably by default). We had at least several dozen people sign up before we shut that off, so there are in fact many auth0 accounts present in our tenant that *are not actual users*. Right now I would consider the prod database to be the source of truth for that.

lvreynoso (commented):
This looks good, I just have a few more general questions. This is just the first step towards an event-driven, user-initiated process for user management, right? Honestly it's unclear to me why this task is part of the view members ticket, and I'm unsure how this is going to help with the invitation flow.

src/backend/aspen/cli/sync_auth0.py
jgadling: > This looks good, I just have a few more general questions. This is just the first step towards an event-driven, user-initiated process for user management, right? Honestly it's unclear to me why this task is part of the view members ticket, and I'm unsure how this is going to help with the invitation flow.

[For reasons that I don't remember right now] we included "listing group invitations" in this ticket. We're planning to have Auth0 handle as much as possible for the user invitations flow, and in order to do that we need to have Auth0 groups for all of the groups in our DB. Some of the code in this script will be copied in to API endpoints for managing users/groups/etc as we build out this functionality. However, until we have our API managing all of our user/group objects, we're probably going to need to re-sync our db to Auth0 - for example, when we manually create a new db group. Even after we have full API coverage in place, having two sources of truth for the medium term makes it possible (or likely) that someone will manually update one system or another, and we'll need to re-sync these stores in either direction to make them consistent.

jgadling: > Be aware that when we first created the auth0 tenant, the sign-up flow on the user-facing auth0 login page was enabled (presumably by default). We had at least several dozen people sign up before we shut that off, so there are in fact many auth0 accounts present in our tenant that _are not actual users_. Right now I would consider the prod database to be the source of truth for that.

Yep, we'll probably run this script in a mode that lets it delete all Auth0 users that don't have matching DB rows in that case.

src/backend/aspen/cli/sync_auth0.py
jgadling: Yep, updated!

PR #1105, opened 2022-05-24
Add list invitations endpoint

lvreynoso:
### Summary:
- **What:** Adds an API endpoint to list auth0 org invitations for a given group.
- **Env:** `<rdev link>`

### Demos:
Screen Shot 2022-05-23 at 17 28 10

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/conftest.py
+++ b/src/backend/aspen/api/conftest.py
@@ -13,6 +13,7 @@ from aspen.api.auth import get_auth_user, setup_userinfo
 from aspen.api.deps import get_db
 from aspen.api.error import http_exceptions as ex
 from aspen.api.main import get_app
+from aspen.auth.auth0_mock import MockAuth0Client
 from aspen.database import connection as aspen_connection
 from aspen.database import schema
 from aspen.database.connection import init_async_db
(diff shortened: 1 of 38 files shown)


src/backend/aspen/cli/sync_auth0.py
```diff
@@ -99,6 +116,10 @@ def get_org_members(self, org: Auth0Org) -> List[Auth0User]:
             "members",
         )
 
+    def get_org_invitations(self, org: Auth0Org) -> List[Auth0Invitation]:
+        # Not a paginated endpoint
```
jgadling: are we sure we're sure? I think this is paginated -- https://auth0.com/docs/api/management/v2/#!/Organizations/get_invitations

lvreynoso: Okay, that's what I assumed at first but `get_all_results` kept giving me strange errors and when I examined the API response there was no pagination info, but I don't doubt I could have made a mistake here (after all why would the docs have page number fields?). I will take a look at it again

lvreynoso: Okay, I figured out what is going on. For whatever reason this endpoint does not return the usual object structured like `{ 'start': 0, 'limit': 25, 'length': 25, 'keyword': [...] }`. Instead the callable just returns a bare array of invitations 🥴

jgadling: yeah, there's a flag like 'include_totals=true' or something that needs to be passed to the API to tell it to return the extra structure!

lvreynoso: The library doesn't support it on that call for some reason: https://github.com/auth0/auth0-python/blob/0be88df898dcd1ed0a870d996fefd08b4a06ed12/auth0/v3/management/organizations.py#L341
🥴

jgadling: Welp. That sucks!

Can we use the internal `Organizations.client` directly now to compose the call we want, and ideally submit a PR to the auth0-python organizations module, so it conforms to the interface that most of the rest of their endpoints do, where it accepts an `include_totals` parameter for most/all methods?

lvreynoso: Yep, sounds good

src/backend/aspen/api/conftest.py
```diff
@@ -124,13 +127,25 @@ async def override_get_auth_user(
     return found_auth_user
 
 
+async def override_get_auth0_client(
```
jgadling, line 130: ❤️

jgadling (approved):

PR #1106, opened 2022-05-24
Group invitations

jgadling:
### Summary:
- **Env:** `<rdev link>`

### Checklist:
- [ ] I added labels to my PR


src/backend/aspen/api/auth.py
```diff
@@ -87,3 +88,13 @@ async def get_auth_user(
...
+
+
+async def get_auth0_client(
+    request: Request, settings: Settings = Depends(get_settings)
+):
+    client_id: str = settings.AUTH0_MANAGEMENT_CLIENT_ID
+    client_secret: str = settings.AUTH0_MANAGEMENT_CLIENT_SECRET
+    domain: str = settings.AUTH0_MANAGEMENT_DOMAIN
+    auth0_client = Auth0Client(client_id, client_secret, domain)
+    return auth0_client
```
lvreynoso, line 100: 🥳

src/backend/aspen/api/views/groups.py
```diff
@@ -27,3 +35,35 @@ async def get_group_members(
...
+    client_id = settings.AUTH0_CLIENT_ID
+    responses = []
+    for email in group_invitation_request.emails:
+        success = True
+        try:
+            auth0_client.invite_member(
+                organization["id"], client_id, user.name, email, "member"
+            )
+        except Auth0Error as err:
+            # TODO - we need to learn more about possible exceptions here.
```
lvreynoso, line 64: Raising an `Auth0Error` looks like this (taken from the library)
````python
raise Auth0Error(
                status_code=self._status_code,
                error_code=self._error_code(),
                message=self._error_message(),
            )
````
which should be enough to make sense of in Sentry. Unless we want to implement more advanced error handling (probably not right now) this should be good enough. The one case I'm concerned about is sending an invite to an email that already has a pending invite, does auth0 handle that gracefully? Will that create another invite or raise an error here?

File: src/backend/aspen/auth/auth0_management.py [hunk 3]
```diff
@@ -0,0 +1,189 @@
...
+    )
+    return "".join(random.choice(possible_characters) for _ in range(length))
+
+
+class Auth0Client:
+    def __init__(self, client_id, client_secret, domain) -> None:
+        # TODO these will need to be read from settings instead of env.
+        # client_id: str = settings.AUTH0_MANAGEMENT_CLIENT_ID
+        # client_secret: str = settings.AUTH0_MANAGEMENT_CLIENT_SECRET
+        # domain: str = settings.AUTH0_MANAGEMENT_DOMAIN
```
lvreynoso: We can delete these lines now, right?

lvreynoso (commented):
Looks good, I just have one major question about an edge case for invites.

lvreynoso (approved):

src/backend/aspen/api/views/groups.py
jgadling: You're right that there are a *lot* of potential error cases we're not handling here. Based on some of the reading I've been doing about auth0's recommended group invitation flows though, I think this endpoint is going to change significantly before this feature is released. I'm planning to try getting these flows working properly with success cases first and *then* handling whatever failures are possible in those cases.

PR #1108, opened 2022-05-24
Set up OneTrust, put Segment behind it

vincent-czi:
### Summary:
- **What:** Adds OneTrust framework, uses it to block Segment (analytics) unless user has opted-in via OneTrust
- **Ticket:** [sc<192931>](https://app.shortcut.com/genepi/story/<192931>)
- **Env:** https://vince-onetrust-frontend.dev.czgenepi.org/

### Demos:
image

### Notes:
There is still a fair amount of visual stuff that needs to get handled for OneTrust:
- Branding as CZ GE (our version from OneTrust was forked off the CZ ID version, and it still has their branding in place when you open the cookie settings menu)
- General visual styling
- Copywriting and any specific language we need depending on conversations with Trust team
- Adding in a "cookie settings" button or similar somewhere in app so user can change mind after initial banner presentation (probably in User dropdown menu in nav, need to finalize)

I've created a [new ticket to track that here](https://app.shortcut.com/genepi/story/199423).

That said, all the framework and interaction with blocking Segment should be in place with this PR.

### Checklist:
- [x] I merged latest `trunk` [minor whoops here, should actually be based off `vince/analytics`, but it's not a big deal]
- [x] I added labels to my PR

### DIFF
--- a/src/backend/Makefile
+++ b/src/backend/Makefile
@@ -55,6 +55,7 @@ remote-db-migrations:
 		aspen-cli db --remote setup; \
 	fi
 	DB=remote alembic upgrade head
+	sleep 60
 
 .PHONY: remote-db-drop
 remote-db-drop:
--- a/src/backend/Makefile
+++ b/src/backend/Makefile
@@ -13,7 +13,7 @@ black:
 	black --check $(STYLE_CHECK_PYTHON_CODE_DIRECTORIES) --exclude $(STYLE_CHECK_PYTHON_CODE_SKIPPED_DIRECTORIES)
 
 isort:
-	isort --profile black --check $(STYLE_CHECK_PYTHON_CODE_DIRECTORIES) --skip $(STYLE_CHECK_PYTHON_CODE_SKIPPED_DIRECTORIES) 
+	isort --profile black --check $(STYLE_CHECK_PYTHON_CODE_DIRECTORIES) --skip $(STYLE_CHECK_PYTHON_CODE_SKIPPED_DIRECTORIES)
 
 mypy:
 	mypy --ignore-missing-imports aspen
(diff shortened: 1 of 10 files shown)


src/backend/database_migrations/versions/20220524_174911_add_auth0_org_id_column_to_groups_table.py
```diff
@@ -0,0 +1,35 @@
+"""Add auth0 org id column to groups table
```
vincent-czi, line 1: This file being in this PR is an oopsie, but incredibly minor and can be ignored.

Since the analytics stuff is a long-lived feature branch, everything should really base off `vince/analytics` branch, with that branch directly getting merges from `trunk` now and then. But muscle memory and I accidentally merged in latest trunk as I was getting ready for PR. But I was going to merge this same stuff directly in `vince/analytics` anyway, so... I'm just going to be lazy and leave it as part of this PR if no one yells at me.

mmmmmmaya:
I can't figure out how to get the modal to appear ...

vincent-czi:
@mmmmmmaya 
> I can't figure out how to get the modal to appear ...

... I forgot to pass the env var with the OneTrust key to the FE. I have the env var in ENV.js, and in the AWS secrets, but forgot to change the terraform... thank you!

mmmmmmaya (approved):

src/backend/Makefile
```diff
@@ -55,6 +55,7 @@ remote-db-migrations:
 		aspen-cli db --remote setup; \
 	fi
 	DB=remote alembic upgrade head
+	sleep 60
```
mmmmmmaya: @jgadling what do you think about this?

vincent-czi: I plan to remove that commit before I merge it in, but would be interested in hearing Jess' thoughts generally on the issue and why this fixes it. There was a Slack thread about it here:
https://chanzuckerbergteam.slack.com/archives/C01C310V70X/p1653587525484089

PR #1109, opened 2022-05-24
fix update endpoint for deleted fields

phoenixAja:
### Summary:
- **What:** update endpoint was not properly clearing values that were None
- **Ticket:** [sc199051](https://app.shortcut.com/genepi/story/199051)
- **Env:** `<rdev link>`

### Checklist:
- [x] I merged latest `trunk`
- [ ] I added labels to my PR
- [x] I updated relevant unit tests


src/backend/aspen/database/models/sample.py
```diff
@@ -29,44 +29,6 @@
     from .sequences import UploadedPathogenGenome
 
 
-def create_public_ids(
```
phoenixAja, line 32: old code that's not referenced anywhere else

src/backend/aspen/api/views/samples.py
```diff
@@ -194,20 +194,20 @@ async def update_samples(
...
             if not loc:
                 raise ex.BadRequestException("location is invalid")
             sample.collection_location = loc
+
         # Sequencing date is handled specially
-        if update_data.sequencing_date:
-            sample.uploaded_pathogen_genome.sequencing_date = (
-                update_data.sequencing_date
-            )
+        sample.uploaded_pathogen_genome.sequencing_date = update_data.sequencing_date  # type: ignore
```
lvreynoso, line 206: What happens in the case that a user does not want to make a change to the sequencing date? Will the old sequencing date be in the request? Sequencing date is nullable, so I am worried that any existing date could be accidentally overwritten.

src/backend/aspen/api/views/samples.py
```diff
@@ -194,20 +194,20 @@ async def update_samples(
...
+        # workaround for our response serializer
         sample.show_private_identifier = True
+
+        sample.generate_public_identifier(already_exists=True)
```
lvreynoso, line 210: Could it be possible that samples with a public identifier don't conform to the format in this function? I find it odd that we have a metadata field for this but then we would always be overwriting it on update here.

lvreynoso (commented):
A couple comments!

src/backend/aspen/api/views/samples.py
phoenixAja: so how this endpoint is being used currently is that all field data is being sent in, so it will send either the  old sequencing date (if the user does not want to make changes), the new sequencing date, or null if they wish to delete the value. so to answer your question yes the old sequencing_date will be sent in the request.

src/backend/aspen/api/views/samples.py
phoenixAja: we would not be overwriting it if the sample has a public identifier, it only autogenerates if the sample has no public_identifier. it's a good callout though because it does look like it overwrites!

phoenixAja:
@lvreynoso do you mind taking over my PR and merging it in when you approve? Just so it gets merged in soon since i'm leaving on vacation. You would be doing me a huge favor!

lvreynoso:
sure thing!

lvreynoso (approved):

PR #1112, opened 2022-05-25
Auto create new users on first login with an org.

jgadling:
### Summary:
- **What:** This is just a draft for now, but I'm posting it to get early[ish] feedback
- **Ticket:** [sc187621](https://app.shortcut.com/genepi/story/187621)
- **Env:** `<rdev link>`

#### Leaning on Auth0 built-in functionality
For now we're using the Auth0 [Organization Invitation flow](https://auth0.com/docs/manage-users/organizations/configure-organizations/invite-members) to invite users to organizations. This flow is convenient because it's 100% managed by auth0, but it comes with the following tradeoffs, which we're going to accept *temporarily*:
- Auth0 doesn't allow us to add First/Last name fields to the new user creation flow
- We need to enable self-signup to our auth0 user db, which will *not* grant any access to users who create accounts without an organization invitation. This may be confusing for users who wander on to our login screen and create accounts, but hopefully we'll be adding a nicer flow soon.

#### How we need to wire it up
To make this flow work, even with its compromises, we need to do the following (and this PR is an example implementation):
- Update our login callback to handle invitation id's and organization id's 
- Provide some way to create a DB user when users accept organization invitations and create new user accounts.

#### What this PR is doing
The invitation/organization query string parameters is pretty unobjectionable imop.

The new-user-creation code here is a little more interesting. Basically any time someone logs into our application and meets the following conditions, we create a new DB user for them:
- They don't already have a user ID in our db
- Their access token includes an auth0 organization id
  - This will be set for users completing the invitation flow
  - Any user visiting our login form can also add a `?organization_id=xxx` query string parameter to the login URL and (if they're a member of that org in Auth0) log in to an organization directly.
- The organization id in their access token matches an auth0_org_id for a group in our database

#### Alternatives?
Some possible other paths forward here:
- When we "invite" a user to an organization, we can actually create an Auth0 user at that point, so we'll have the auth0 user ID available, and we can create a matching DB row. We can let the organization invitation flow just handle granting access to an organization instead of creating new user accounts.
  - This might be closer to a best-practices final flow that doesn't require us to enable arbitrary signups, but in the meantime we'd need to figure out how to repurpose the "change password" flow to look more like an "accept invitation" flow
   - We may well wind up with a bunch of extra auth0 & db users that never get used / invitations never get accepted.
- We can use Auth0's callbacks to create new users when they're created in auth0, like so: https://auth0.com/docs/customize/hooks/extensibility-points/post-user-registration
  - This should also be doable, but it looks more complicated
### Checklist:
- [ ] I added labels to my PR


src/backend/aspen/api/views/auth.py
```diff
@@ -20,17 +25,66 @@
...
+        kwargs["organization"] = organization
+    if organization_name:
+        kwargs["organization_name"] = organization_name
+    return await auth0.authorize_redirect(
+        request, settings.AUTH0_CALLBACK_URL, **kwargs
+    )
+
+
+async def create_user_if_not_exists(db, userinfo):
+    if "org_id" not in userinfo:
```
vincent-czi, line 47: Is `org_id` just a string by itself, or can it also be a list? My mental model might be wrong -- I had thought Auth0 would return a list of all the orgs that a user was a member of. So if I was a member of groupA and groupB, I thought Auth0 would send something like [groupA, groupB] back as a list _somewhere_ in the identity payload. Am I wrong?

src/backend/aspen/api/views/auth.py
```diff
@@ -20,17 +25,66 @@
...
+        group = groupquery.scalars().one()  # type: ignore
+    except NoResultFound:
+        return  # We didn't find a matching group, we can't create this user.
+    user_fields = {
+        "name": userinfo["email"],
+        "email": userinfo["email"],
+        "auth0_user_id": auth0_user_id,
+        "group_admin": False,
+        "system_admin": False,
+        "group": group,
```
vincent-czi, line 75: What does `group` mean in this brave new world? If a user can be the member of multiple groups, is the group a user entry is associated with their original group on first creation?

src/backend/aspen/api/views/auth.py
```diff
@@ -20,17 +25,66 @@
...
+    if organization_name:
+        kwargs["organization_name"] = organization_name
+    return await auth0.authorize_redirect(
+        request, settings.AUTH0_CALLBACK_URL, **kwargs
+    )
+
+
+async def create_user_if_not_exists(db, userinfo):
+    if "org_id" not in userinfo:
+        return  # We're currently only creating new users if they're confirming an org invitation
```
vincent-czi, line 48: Feels like it should be an exception getting raised here

src/backend/aspen/api/views/auth.py
```diff
@@ -20,17 +25,66 @@
...
+    auth0_user_id = userinfo.get("sub")
+    if not auth0_user_id:
+        return  # User ID really needs to be present
```
vincent-czi, line 51: Feels like it should be an exception getting raised here

src/backend/aspen/api/views/auth.py
jgadling: This will always be a string, because users will *specifically* log in to a particular org. Afaik auth0 has no way to send us a list of all the org ID's a user is part of, as part of an auth token.

src/backend/aspen/api/views/auth.py
jgadling: Yeah, but it'll get raised later anyway.

src/backend/aspen/api/views/auth.py
jgadling: Our main Auth0 application supports two login modes: 
1. Logging into our app without an organization. This is what we do right now, and will still be the default login mode for CZGE
2. Logging into our app *with* an organization. This is the flow that gets used when accepting an org invitation, and this flow can be forced by adding a query string parameter to our login URL.

We want to continue supporting both flows (at least for now) so not all auth tokens will contain an org ID.

src/backend/aspen/api/views/auth.py
jgadling: We're not tackling users in multiple groups right now. This is the user's only group.

src/backend/aspen/api/views/auth.py
vincent-czi: ahhh, got it, thank you. I thought (2) was totally supplanting (1), that makes sense

mmmmmmaya:
generally looks ok to me, but I'm no python wizard.

src/backend/aspen/api/views/auth.py
```diff
@@ -20,17 +25,66 @@
 @router.get("/login")
 async def login(
     request: Request,
+    organization: Optional[str] = None,
+    invitation: Optional[str] = None,
+    organization_name: Optional[str] = None,
```
lvreynoso, line 30: These are coming from the query string?

lvreynoso (approved):
Looks good to me, and pretty straightforward. I'm not sure how hard it would be to fake a callback in the test (probably annoyingly hard).

src/backend/aspen/api/views/auth.py
jgadling: yep!

jgadling:
> Looks good to me, and pretty straightforward. I'm not sure how hard it would be to fake a callback in the test (probably annoyingly hard).

Actually probably not that hard... I'll take a stab at it and update the PR if it's not too crazy.

PR #1114, opened 2022-05-27
Allow user and group deactivation

lvreynoso:
### Summary:
- **What:** Allows us to mark users and groups as "deleted" by setting a value of `FALSE` in a column `active`. This makes it possible to keep samples in a group even if it was uploaded by a now-deleted user.
- **Ticket:** [sc190605](https://app.shortcut.com/genepi/story/190605)
- **Env:** `<rdev link>`

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/database/models/usergroup.py
+++ b/src/backend/aspen/database/models/usergroup.py
@@ -23,6 +23,7 @@ class Group(idbase, DictMixin):  # type: ignore
     __tablename__ = "groups"
 
     name = Column(String, unique=True, nullable=False)
+    active = Column(Boolean, nullable=False, default=True)
     address = Column(String, nullable=True)
     prefix = Column(
         String,
@@ -33,7 +34,7 @@ class Group(idbase, DictMixin):  # type: ignore
     division = Column(String, nullable=True)
     location = Column(String, nullable=True)
 
-    auth0_org_id = Column(String, unique=True, nullable=False)
+    auth0_org_id = Column(String, unique=True, nullable=True)
 
     # Default location context (int'l or division or location level)
     default_tree_location_id = Column(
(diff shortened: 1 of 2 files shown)


jgadling:
... stepping up on soapbox... 

Firstly, bools in a database are almost never the best representation of an idea. Maybe today we add a bool called "active" ... but down the road we might want to add a bool called "accepted_invitation" or something else. We wind up having a sprawl of specialty booleans when we might be better off modeling a "user_status" table with a list of viable statuses.

Secondly, keeping data in a db, in its usual home with some sort of "disabled" status is a nightmare. It means basically every single query we make to those tables needs appropriate filters added, and 99.99% of the time we're just never going to need that data again and we're doing all that work for no reason. If product says we must have this functionality (auditing purposes, the ability to "undelete" stuff", whatever) then we do what we have to do (and even then, there are other ways to model this), but in the meantime I vote for actually deleting deleted users.

lvreynoso:
> ... stepping up on soapbox...
> 
> Firstly, bools in a database are almost never the best representation of an idea. Maybe today we add a bool called "active" ... but down the road we might want to add a bool called "accepted_invitation" or something else. We wind up having a sprawl of specialty booleans when we might be better off modeling a "user_status" table with a list of viable statuses.
> 
> Secondly, keeping data in a db, in its usual home with some sort of "disabled" status is a nightmare. It means basically every single query we make to those tables needs appropriate filters added, and 99.99% of the time we're just never going to need that data again and we're doing all that work for no reason. If product says we must have this functionality (auditing purposes, the ability to "undelete" stuff", whatever) then we do what we have to do (and even then, there are other ways to model this), but in the meantime I vote for actually deleting deleted users.

I think those are great points. What should we do about previously uploaded samples then? `uploaded_by_id` is non-nullable, but I don't think we should remove a group's access to a sample just because we delete a user.

PR #1115, opened 2022-05-27
Maya/187629 members page

mmmmmmaya:
### Summary
- **What:** 
  - Add new page to see information about group members
  - Tab navigation at two levels
  - Group details page styles at small and large screen sizes
  - Reduce horizontal padding on small screen sizes
- **Env:** https://maya-members-frontend.dev.czgenepi.org/members
- **Design:** https://www.figma.com/file/SALQZGfIJntBes0HvIzVeX/Onboarding-New-Users-V0?node-id=280%3A48892

### Demos
- Here's a before and after of the uploads page, as far as padding changes go
Screen Shot 2022-05-26 at 6 15 54 PM
Screen Shot 2022-05-26 at 6 15 41 PM

- And the new content
Screen Shot 2022-05-26 at 6 15 29 PM
Screen Shot 2022-05-26 at 6 15 34 PM

### Notes
Tagging @Janeece for a review of:
- Group Details page, large and small screen layouts
- The outer level of tab navigation (I had to do some css hacks to get the tab border to extend all the way across the page. Mainly checking in whether it looks right)

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/styles/mixins/global.ts
+++ b/src/frontend/src/common/styles/mixins/global.ts
@@ -1,6 +1,15 @@
 import styled from "@emotion/styled";
+import { CommonThemeProps, getSpaces } from "czifui";
 import { NAV_BAR_HEIGHT_PX } from "src/components/NavBar";
 
 export const PageContent = styled.div`
   height: calc(100% - ${NAV_BAR_HEIGHT_PX}px);
 `;
+
+export const ContentStyles = (props: CommonThemeProps) => {
+  const spaces = getSpaces(props);
+
+  return `
+    padding: ${spaces?.xl}px 125px ${spaces?.l}px 125px;
+  `;
+};
--- a/src/frontend/src/common/styles/mixins/global.ts
+++ b/src/frontend/src/common/styles/mixins/global.ts
@@ -2,6 +2,8 @@ import styled from "@emotion/styled";
 import { CommonThemeProps, getSpaces } from "czifui";
 import { NAV_BAR_HEIGHT_PX } from "src/components/NavBar";
 
+export const PAGE_PADDING = 125;
+
 export const PageContent = styled.div`
   height: calc(100% - ${NAV_BAR_HEIGHT_PX}px);
 `;
@@ -11,7 +13,7 @@ export const ContentStyles = (props: CommonThemeProps) => {
 
   return `
     @media only screen and (min-width: 768px) {
-      padding: ${spaces?.xl}px 125px;
+      padding: ${spaces?.xl}px ${PAGE_PADDING}px;
     }
 
     @media only screen and (max-width: 768px) {
--- a/src/frontend/src/common/styles/mixins/global.ts
+++ b/src/frontend/src/common/styles/mixins/global.ts
@@ -10,6 +10,12 @@ export const ContentStyles = (props: CommonThemeProps) => {
   const spaces = getSpaces(props);
 
   return `
-    padding: ${spaces?.xl}px 125px ${spaces?.l}px 125px;
+    @media only screen and (min-width: 768px) {
+      padding: ${spaces?.xl}px 125px;
+    }
+
+    @media only screen and (max-width: 768px) {
+      padding: ${spaces?.xl}px;
+    }
   `;
 };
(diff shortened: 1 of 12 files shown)


Janeece (commented):
@mmmmmmaya group details (large and small sizes) + top level tabs look great to me!

I just checked with Christine and she said we won't have the destinations for the "Learn More" links ready before the feature, so we can remove those for now.

src/frontend/src/common/styles/mixins/global.ts
```diff
@@ -1,6 +1,23 @@
...
+
 export const PageContent = styled.div`
   height: calc(100% - ${NAV_BAR_HEIGHT_PX}px);
 `;
+
+export const ContentStyles = (props: CommonThemeProps) => {
+  const spaces = getSpaces(props);
+
+  return `
+    @media only screen and (min-width: 768px) {
```
vincent-czi, line 15: So I don't have a problem with this specific code per se, but I think needing to write media queries like this in a bunch of places is going to feel cumbersome the more and more we aim to have responsive design. I don't have a solution really, but I'd like to think about if we could come up with a nice dev experience to indicate "this big CSS, this small CSS" without needing to have this kind of boilerplate code in a bunch of style files.

At previous job, we used [styled-system](https://styled-system.com/responsive-styles) and I thought that was a really smooth experience. Now, that said, previous job used a _totally different_ approach where we passed style props down to some core, heavily prop-styled component as opposed to what we do at CZ GE where we style the component more directly and much less prop styling. But just as a point of inspiration, I like how _clean_ styled-system felt for an API as a way of presenting "big CSS, small CSS" side-by-side without much need for boilerplate.

In no way do I think this blocks this PR or immediate work, but I would like to just kinda spitball some ideas for awhile amongst FE folks. Is there some clean, easy way we could abstract this? I'm just vaguely gesturing at the idea of mushing together the idea of a HOC and a `styled` call, but I think if we could figure out something it would be a nice dev experience down the road.

src/frontend/src/common/styles/mixins/global.ts
```diff
@@ -1,6 +1,23 @@
...
+      padding: ${spaces?.xl}px ${PAGE_PADDING}px;
+    }
+
+    @media only screen and (max-width: 768px) {
```
vincent-czi, line 19: We should probably abstract out `768px` to a constant and import it everywhere we do media query stuff since we -- I assume! -- want all the pages to use that same value for the breakpoint

src/frontend/src/views/GroupMembersPage/components/MembersTab/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+type TabType = "active" | "invitations";
+
+interface Props {
+  invites: any[];
+  members: any[];
+}
+
+const MembersTab = ({ invites, members }: Props): JSX.Element => {
+  const [tabValue, setTabValue] = useState<TabType>("active");
```
vincent-czi, line 15: nit: Feels like TabType should be an enum

src/frontend/src/views/GroupMembersPage/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { GroupDetailsTab } from "./components/GroupDetailsTab";
+import { MembersTab } from "./components/MembersTab";
+import {
+  StyledHeader,
+  StyledName,
+  StyledPageContent,
+  StyledTabs,
+} from "./style";
+
+type PrimaryTabType = "members" | "details";
```
vincent-czi, line 12: Nit: feels like an enum to me

vincent-czi (approved):
Looks good! No critical comments, could be merged in as-is and I wouldn't be bothered at all

src/frontend/src/views/GroupMembersPage/components/MembersTab/index.tsx
mmmmmmaya: It's in the follow up pr

src/frontend/src/views/GroupMembersPage/index.tsx
mmmmmmaya: follow up

src/frontend/src/common/styles/mixins/global.ts
mmmmmmaya: agree. I'll work this into a follow up

src/frontend/src/common/styles/mixins/global.ts
mmmmmmaya: I think this should be an eng excellence ticket

PR #1117, opened 2022-05-27
Expanded group information in usergroup API response

lvreynoso:
### Summary:
- **What:** Expand group information response in `users/me` endpoint. Includes fields necessary for group management frontend work.
- **Env:** [https://groupinfo-backend.dev.czgenepi.org/v2/users/me](https://groupinfo-backend.dev.czgenepi.org/v2/users/me)

### Demos:
Screen Shot 2022-05-27 at 11 47 56

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/schemas/usergroup.py
+++ b/src/backend/aspen/api/schemas/usergroup.py
@@ -7,9 +7,9 @@ from aspen.api.schemas.base import BaseRequest, BaseResponse
 class GroupResponse(BaseResponse):
     id: int
     name: str
-    address: str
+    address: Optional[str]
     prefix: str
-    location: str
+    location: Optional[str]
 
 
 class UserBaseResponse(BaseResponse):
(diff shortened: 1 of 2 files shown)


mmmmmmaya:
seems good to me, but would love for BE person to approve!

src/backend/aspen/api/schemas/usergroup.py
```diff
@@ -7,6 +7,9 @@
 class GroupResponse(BaseResponse):
     id: int
     name: str
+    address: Optional[str]
+    prefix: str
+    location: Optional[str]
```
jgadling, line 12: Should this be the group's default tree location? I'm not sure if we use the old location field at all anymore

jgadling (approved):
I'm not sure what we're doing with the `location` field at the moment, but we can fix that now or later.

src/backend/aspen/api/schemas/usergroup.py
lvreynoso: Yeah, that makes sense.

PR #1123, opened 2022-06-01
Maya/187629 members page

mmmmmmaya:
### Summary
- **What:** Routing for group details page
- **Why:** Keeps the users browser address bar in line with the tab they are on
- **Env:** https://maya-routing-frontend.dev.czgenepi.org/

### Demos
No visual changes.

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/.github/workflows/push-rdev.yml
+++ b/.github/workflows/push-rdev.yml
@@ -86,4 +86,4 @@ jobs:
         with:
           tfe-token: ${{ secrets.TFE_TOKEN }}
           stack-name: ${{ steps.stack-name.outputs.stack-name }}
-          happy_version: "0.19.0"
+          happy_version: "0.20.0"
--- a/src/frontend/src/views/GroupMembersPage/index.tsx
+++ b/src/frontend/src/views/GroupMembersPage/index.tsx
@@ -1,6 +1,7 @@
 import { Tab } from "czifui";
 import { useRouter } from "next/router";
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
+import { ROUTES } from "src/common/routes";
 import { GroupDetailsTab } from "./components/GroupDetailsTab";
 import { MembersTab } from "./components/MembersTab";
 import {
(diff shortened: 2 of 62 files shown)


.github/workflows/push-rdev.yml
```diff
@@ -86,4 +86,4 @@ jobs:
         with:
           tfe-token: ${{ secrets.TFE_TOKEN }}
           stack-name: ${{ steps.stack-name.outputs.stack-name }}
-          happy_version: "0.19.0"
+          happy_version: "0.20.0"
```
vincent-czi, line 89: I think this is fine to just version bump, but I'd want to be sure that @jgadling agrees before merging it.

src/frontend/src/views/GroupMembersPage/index.tsx
```diff
@@ -9,15 +11,39 @@ import {
...
+
+const isValidPrimaryTab = (token: string) => {
+  return token === PrimaryTabType.MEMBERS || token === PrimaryTabType.DETAILS;
+};
+
+const GroupMembersPage = ({ pathTokens }: Props): JSX.Element => {
+  const [primaryQueryParam = "members", secondaryQueryParam = "details"] =
+    pathTokens ?? [];
+  const initialPrimaryTab = (
+    isValidPrimaryTab(primaryQueryParam)
```
vincent-czi, line 35: Nit: Because there's this tertiary that defaults to "members" when invalid, it seems like the default assignment above `[primaryQueryParam = "members", secondaryQueryParam = "details"]` is unnecessary? As it is, there's two places where it defaults to making the primaryQueryParam members -- first in the array when it's missing from pathTokens, and then here when the value is invalid. But since the second default will catch the first case (I'm pretty confident, maybe I'm wrong), it seems a bit confusing to have two defaulting approaches when one would do.

vincent-czi, line 35: Nit: Because there's this tertiary that defaults to "members" when invalid, it seems like the default assignment above `[primaryQueryParam = "members", secondaryQueryParam = "details"]` is unnecessary? As it is, there's two places where it defaults to making the primaryQueryParam members -- first in the array when it's missing from pathTokens, and then here when the value is invalid. But since the second default will catch the first case (I'm pretty confident, maybe I'm wrong), it seems a bit confusing to have two defaulting approaches when one would do.

vincent-czi (approved):

src/frontend/src/views/GroupMembersPage/index.tsx
vincent-czi: ... dunno why GitHub added this comment twice.

.github/workflows/push-rdev.yml
jgadling: lgtm

PR #1124, opened 2022-06-01
refactor breakpoint code

mmmmmmaya:
### Summary
- **What:** Breakpoint refactor
- **Why:** Keep our breakpoint definition DRY
- **Env:** https://maya-breakpoint-frontend.dev.czgenepi.org/

### Testing
Spot testing.

### Demos
No visual changes.

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/views/GroupMembersPage/style.ts
+++ b/src/frontend/src/views/GroupMembersPage/style.ts
@@ -4,6 +4,7 @@ import {
   ContentStyles,
   PageContent,
   PAGE_PADDING,
+  SmallerThanBreakpoint,
 } from "src/common/styles/mixins/global";
 
 export const StyledName = styled.div`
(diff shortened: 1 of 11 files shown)


src/frontend/src/views/GroupMembersPage/style.ts
```diff
@@ -35,9 +36,9 @@ export const StyledHeader = styled.div`
       border-bottom: 2px solid ${colors?.gray[200]};
       padding: 0 ${PAGE_PADDING}px;
 
-      @media only screen and (max-width: 768px) {
```
vincent-czi, line 38: I don't know enough about CSS off hand: should we care about this difference

```
@media only screen and (max-width: 768px)
VS
@media (max-width: 768px)
```

Is it fine for them to all just use the same thing, or are there cases when the `only screen` aspect would be an important differentiator?

vincent-czi (approved):
Nice! I am satisfied with this approach, thanks!

I do have one comment in there that I don't think affects the approval status of the PR, but should probably be looked at before merging just in case it matters.

src/frontend/src/views/GroupMembersPage/style.ts
mmmmmmaya: screen only implies the changes don't apply for non-screen applications (for example, if you were going to print the page out)

PR #1125, opened 2022-06-01
Maya/187629 group dropdown api

mmmmmmaya:
### Summary
- **What:**
  - Add api functions to grab group members and info
  - Also, add open/close functionality for invite modal (mainly because I added this early on by accident and it needs to go in anyway)
- **Why:** No hard coded data!
- **Env:** https://maya-api-frontend.dev.czgenepi.org/

### Demos
No visual changes.

### Notes
Visually the same, but now we grab data from the API instead of from hard-coded values.

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/frontend/src/common/utils/userUtils.ts
```diff
@@ -0,0 +1,9 @@
+export const getGroupIdFromUser = (user?: User): number | undefined => {
+  if (!user) return;
+
+  const { group } = user;
+  if (!group) return;
```
phoenixAja, line 5: when would the user not have a group?

mmmmmmaya: The users returned with the group members api call don't have a group attached to them.

src/frontend/src/components/NavBar/components/GroupDetailsDropdown/index.tsx
```diff
@@ -14,46 +19,43 @@ import {
...
-  const { memberCount, name, location } = groupInfo;
+  // right now users can only have one group, but will be able to have more in the future.
+  // ui already knows how to render for multiple groups, so we still want to give an array.
+  const usersGroups: Group[] = [groupInfo];
 
-  const userInfo = {
-    isOwner: true,
-  };
-  const { isOwner } = userInfo;
+  const { name, location } = groupInfo ?? {};
```
phoenixAja, line 45: will this unpack correctly if `{}`?

mmmmmmaya: yes, it will unpack to `undefined`, which is what I would expect.

src/frontend/src/components/NavBar/components/GroupDetailsDropdown/index.tsx
```diff
@@ -14,46 +19,43 @@ import {
...
+  const displayLocation = stringifyGisaidLocation(location);
+
+  // is the current user a group owner
```
phoenixAja, line 48: this comment doesn't really make sense, we're just finding the members of the group, not if the current user is an owner right?

phoenixAja (approved):

src/frontend/src/components/NavBar/components/GroupDetailsDropdown/index.tsx
mmmmmmaya: sorry, this describes the next 3 lines.

PR #1128, opened 2022-06-03
Maya/187620 invite api call

mmmmmmaya:
### Summary
- **What:** 
  - Add functions that call api to list and create group invites
  - Send invites from invite modals
  - Add default render of `--` for cells in the tables
  - read group members page row data from db instead of hard code
- **Ticket:** [[sc-187620]](https://app.shortcut.com/genepi/story/187620)
- **Env:** https://maya-invite-api-frontend.dev.czgenepi.org/
- **Design:** https://www.figma.com/file/SALQZGfIJntBes0HvIzVeX/Onboarding-New-Users-V0?node-id=280%3A48892

### Notes
- I noticed there is some wonkiness on the routing if you navigate to a sub-page directly using the address bar rather than by clicking in the UI
- The endpoint to read invitations doesn't seem to work. I pinged Jess to check it out, and I will add code to use the new function in a follow up PR

### TODO
- Read invites from DB to list in table
- write types on props and vars for all new components that use `any` or other vague types
- clean up all `// TODO (mlila)` for this feature
- fix and remove all `@ts-expect-error` that were added for this feature
- fix and remove all `eslint-disable-line` that were added for this feature 
- go back to previous PRs and address nits
- design tweaks

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/frontend/src/views/GroupMembersPage/components/MembersTab/index.tsx
```diff
@@ -54,9 +56,32 @@ const MembersTab = ({
...
+      dateSent: "2022-05-24",
+      role: "Member",
+    },
+    {
+      email: "frank@fake.com",
+      status: "expired",
+      dateSent: "2022-03-24",
+      role: "Member",
+    },
+  ];
```
phoenixAja, line 73: do we still need this stub?

mmmmmmaya: yep! I added the basic function to read invites, but it doesn't actually work yet (auth0 migration has not happened), so the last piece for me will be to double check that the function works as expected after migration and to take out these test values.

phoenixAja (approved):
looks good

PR #1129, opened 2022-06-03
Update the sync script to write auth0 group ID's to the db

jgadling:
### Summary:
This PR contains a few different changes:
- Now that we have a separate auth0 tenant for localdev, we're using @vincent-czi's new flow to pull in auth0 management credentials to local dev, so we should actually be able to generate invitations from local dev once we get the DB's sync'd properly.
- Update the auth0 sync script to write auth0 group ID's to our DB when we sync data.
- @mmmmmmaya was noticing that since we've changed the `setup_dev_data.sh` script to run outside of a container, all the AWS commands are paginating if they run in a small terminal (like in an IDE pane). We're disabling pagination in that script now

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/scripts/setup_dev_data.sh
+++ b/scripts/setup_dev_data.sh
@@ -41,8 +41,11 @@ done
 
 
 ONETRUST_FRONTEND_KEY=$(jq -c .ONETRUST_FRONTEND_KEY <<< "${EXTRA_SECRETS}")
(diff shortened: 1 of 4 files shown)


lvreynoso (approved):
LGTM, have you tried running it? Does it work with our new localdev tenant?

jgadling:
> LGTM, have you tried running it? Does it work with our new localdev tenant?

Yep, it works with the new localdev tenant. We still need to update the localdev db so that the auth0 id's are consistent across our local devs, but this is one step closer.

PR #1140, opened 2022-06-08
don't show invite button if user isn't admin

phoenixAja:
### Summary:
- **What:** only admins should be allowed to invite new users to group
- **Ticket:** [sc200919](https://app.shortcut.com/genepi/story/200919)
- **Env:** `<rdev link>`

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/views/GroupMembersPage/components/MembersTab/index.tsx
+++ b/src/frontend/src/views/GroupMembersPage/components/MembersTab/index.tsx
@@ -1,7 +1,9 @@
 import { Button, Tab } from "czifui";
+import { find } from "lodash";
 import { useRouter } from "next/router";
 import React, { useState } from "react";
 import { HeadAppTitle } from "src/common/components";
+import { useUserInfo } from "src/common/queries/auth";
 import { useGroupInvitations } from "src/common/queries/groups";
 import { ROUTES } from "src/common/routes";
 import { TabEventHandler } from "../../index";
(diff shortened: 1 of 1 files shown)


src/frontend/src/views/GroupMembersPage/components/MembersTab/index.tsx
```diff
@@ -27,13 +29,16 @@ const MembersTab = ({
...
+}: Props): JSX.Element | null => {
   const [tabValue, setTabValue] =
     useState<SecondaryTabType>(initialSecondaryTab);
   const [isInviteModalOpen, setIsInviteModalOpen] = useState<boolean>(false);
   const router = useRouter();
   const { data: invitations = [] } = useGroupInvitations(groupId);
+  const { data: userInfo } = useUserInfo();
+  const currentUser = find(members, (m) => m.id === userInfo?.id);
 
+  const isOwner = currentUser?.isGroupAdmin === true;
```
vincent-czi, line 41: Nothing inherently wrong with this implementation, but it strikes me as weird that `isGroupAdmin` isn't just part of the `useUserInfo` data. I think it would be more reasonable to add this to the `/me` endpoint than needing to look in the general members info for the group to get that info.

In the case where we have other things we might want to display for group_admins but not general group members, we'd have to get this member list and compare in those situations as well, which seems weird when it's really just a question about the user, not all the members of the group.

vincent-czi (approved):
I made a comment with a pretty major change suggestion, but since we're crunched on time, I think it would be fine to just turn it into a TODO (assuming you agree with my comment).

Additionally, this isn't the responsibility of the FE, but have you verified that the BE invite endpoint will block/deny requests from non-group_admin users? I'm sure it's the case, but if you haven't double-checked, this is probably a good time just to make sure.

src/frontend/src/views/GroupMembersPage/components/MembersTab/index.tsx
phoenixAja: yeah totally agree with you! I was just doing the easy solution due to time crunch, i'll update the /me endpoint if time permits.

PR #1141, opened 2022-06-09
Add support for division and location colorings in trees

lvreynoso:
### Summary:
- **What:** Like #1033 , but for divisions and locations. Adds support for giving tree json colorings metadata for categories `division` and `location`.
- **Ticket:** [sc195034](https://app.shortcut.com/genepi/story/195034)
- **Env:** [https://colorings-frontend.dev.czgenepi.org](https://colorings-frontend.dev.czgenepi.org)

### Checklist:
- [ ] I added labels to my PR


src/backend/aspen/api/utils/phylo.py
```diff
@@ -90,40 +97,65 @@ def _sample_filter(sample: Sample, can_see_pi_group_ids: Set[int], system_admin:
...
+    extracted_location_list = [
+        node.get("node_attrs", {}).get(key, {}).get("value", None)
+        for key in LOCATION_KEYS
+    ]
+    locations.add(ExtractedLocation(*extracted_location_list))
     for child in node.get("children", []):
-        countries |= _collect_countries(child)
-    return countries
+        locations |= _collect_locations(child)
+    return locations
```
danrlu, line 113: Niceeeee <3

src/backend/aspen/api/utils/phylo.py
```diff
@@ -90,40 +97,65 @@ def _sample_filter(sample: Sample, can_see_pi_group_ids: Set[int], system_admin:
...
-    defined_countries -= set(countries)
+    category_defines_index = None
+    for index, defines in enumerate(tree_json["meta"]["colorings"]):
+        if defines["key"] == category:
+            category_defines_index = index
 
+    locs_in_tree = set(extracted_locations)
     origin_location = aliased(Location)
 
+    # SQLAlchemy black magic going on here
```
danrlu: 🤣

src/backend/aspen/api/utils/phylo.py
```diff
@@ -134,43 +172,55 @@ async def _set_countries(
...
             and_(
                 origin_location.country == tree_location.country,
-                origin_location.division == None,  # noqa: E711
-                origin_location.location == None,  # noqa: E711
+                origin_location.division == tree_location.division,  # noqa: E711
+                origin_location.location == tree_location.location,  # noqa: E711
             )
         )
         .order_by(asc("distance"))
         .limit(15)
```
danrlu, line 185: (you prob figured it out already. just adding notes in case helpful) i think the query could benefit from a deduplication based on the `category` column (for example, `Location.division`), and in best-case scenario keep the one entry that has the smallest `distance` among duplicates. So when `category=division`, we don't end up w 15 entries w identical `division` which then become 1 color. 
This way we would always end up with 15 entries from the sql query and no need to fill extra ones below. On the tree many of the colored locations had no lontitude/latitude data (from the map under the tree) so they are likely not pulled from the database from the sql query and instead filled with the locations on the tree, indicating the query is not pulling enough locations.

src/backend/aspen/api/utils/phylo.py
```diff
@@ -134,47 +186,74 @@ async def _set_countries(
...
                 origin_location.country == tree_location.country,
-                origin_location.division == None,  # noqa: E711
-                origin_location.location == None,  # noqa: E711
+                origin_location.division == tree_location.division,
+                origin_location.location == tree_location.location,
             )
         )
         .order_by(asc("distance"))
-        .limit(15)
+        .limit(16)
```
danrlu, line 199: we're pulling 16 closest directly from the database which will include the `tree_location` itself, so we don't need to add into the `tree_location` later.

src/backend/aspen/api/utils/phylo.py
```diff
@@ -208,6 +287,7 @@ async def process_phylo_tree(
     json_data = json.loads(data)
 
     if id_style == "public":
+        json_data = await _set_colors(db, json_data, phylo_run)
```
danrlu, line 282: modify color when we download trees with public IDs as well. thanks for the tip Jess!!

danrlu:
I populated rdev database lat and long w random numbers and made tree location the center of universe:
```
update locations set latitude=floor(random() * 80 + 0);
update locations set longitude=floor(random() * 80 + 0);
update locations set latitude=0,longitude=0 where location='San Mateo County';
```
So the code is pulling out tree location itself on top and the closest locations:

County:
image
image

State:
image

image

Country (rdev database don't have entries that division is `null`, so w Lucia's set up, countries were filled with whatever is in the tree)
image

src/backend/aspen/api/views/tests/test_download_phylo_tree.py
```diff
@@ -85,7 +85,7 @@ async def test_phylo_tree_id_style_public(
         headers=auth_headers,
     )
     returned_tree = result.json()
-    assert returned_tree == TEST_TREE
+    assert returned_tree["tree"] == TEST_TREE["tree"]
```
danrlu, line 88: We update the `meta` portion of the tree json with colors so here we only check the IDs (which are in `tree` portion) stay the same on public ID tree download

danrlu:
I'll merge to staging after prod deploy tomorrow and check again all looks good. Thanks very much!!

danrlu:
To help with PR review: 
What changed since Lucia's last commit: https://github.com/chanzuckerberg/czgenepi/compare/10a34ed...22ab85c
What changed since Jess's last commit: https://github.com/chanzuckerberg/czgenepi/compare/3001195...22ab85c

jgadling (approved):

danrlu:
On staging 🤩 (see how the colored dots surround the teal dot which is the tree location) 

image

image

image

PR #1142, opened 2022-06-09
email conditional logic if no organization name

phoenixAja:
### Summary:
- **What:** if we trigger a verification email to be sent through the auth0 management console no organization context is provided, if that is the case use more generic language when inviting user to a group. I also added the prod verification template since the image paths are different
- **Ticket:** no ticket
- **Env:** no rdev

### Demos:
Screen Shot 2022-06-09 at 9 52 40 AM
Screen Shot 2022-06-09 at 9 52 16 AM

### Checklist:
- [ ] I added labels to my PR


vincent-czi (approved):
Looks good! Hard to tell about all the CSS stuff for sure, but I definitely don't see anything wacky.

Have you gotten the chance to verify the email looks right on mobile? If you haven't, I'd make sure to do that before merging and calling it done.

Just curious, is there any difference between the Prod and Staging email templates? Or is it just that we need two different files to be targeted because of a tenant/env thing?

phoenixAja:
> Just curious, is there any difference between the Prod and Staging email templates? Or is it just that we need two different files to be targeted because of a tenant/env thing?

They're only different in that they reference the asset png files differently (ex: https://czgenepi.org/CZGenEpiLogoWhite.png vs. https://staging.czgenepi.org/CZGenEpiLogoWhite.png)

PR #1143, opened 2022-06-10
Handle multi-delimited emails; handle spaces

vincent-czi:
### Summary:
- **What:** Fixes to email address parsing in the "invite" new members dialog
- **Ticket:** Part of [sc<200919>](https://app.shortcut.com/genepi/story/<200919>)
- **Env:** none

### Demos:
image

Would send out these three invites

image

---

Still capturing bad emails

image

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/constants/inputDelimiters.ts
+++ b/src/frontend/src/common/constants/inputDelimiters.ts
@@ -1,1 +1,2 @@
-export const INPUT_DELIMITERS = /[\n\t,]/g;
+export const INPUT_DELIMITERS = /[\n\t, ]/g;
+export const GREEDY_SPACES = / +/g;
--- a/src/frontend/src/views/GroupMembersPage/components/MembersTab/components/InviteModal/index.tsx
+++ b/src/frontend/src/views/GroupMembersPage/components/MembersTab/components/InviteModal/index.tsx
@@ -63,10 +63,6 @@ const InviteModal = ({ groupName, onClose, open }: Props): JSX.Element => {
    *     user1@example.com,   user2@example.com, ,, ,,  , user3@example.com,,
    *   user2@example.com, user4@example.com,,
    * Would return as four emails in total, as you'd expect to read them.
-   *
-   * Arguable if it's best UX -- should double commas bring up an error? etc --
-   * but it's easiest to code and reasonably sane to just ignore that sort
-   * of stuff. Approach was approved by Design.
    */
   const getAddressArrayFromInputValue = (): string[] => {
     // First pass may have multiple spaces between chunks or pre/suffix spaces
--- a/src/frontend/src/views/GroupMembersPage/components/MembersTab/components/InviteModal/index.tsx
+++ b/src/frontend/src/views/GroupMembersPage/components/MembersTab/components/InviteModal/index.tsx
@@ -2,7 +2,10 @@ import { Button, Dialog, DialogActions, DialogTitle } from "czifui";
 import { compact, uniq } from "lodash";
 import React, { ChangeEvent, useState } from "react";
 import { noop } from "src/common/constants/empty";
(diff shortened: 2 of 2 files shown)


src/frontend/src/views/GroupMembersPage/components/MembersTab/components/InviteModal/index.tsx
```diff
@@ -48,12 +51,35 @@ const InviteModal = ({ groupName, onClose, open }: Props): JSX.Element => {
...
+   * in the input, only one copy of the chunk is returned in output array.
+   *
+   * Example:
+   *     user1@example.com,   user2@example.com, ,, ,,  , user3@example.com,,
+   *   user2@example.com, user4@example.com,,
+   * Would return as four emails in total, as you'd expect to read them.
+   *
+   * Arguable if it's best UX -- should double commas bring up an error? etc --
+   * but it's easiest to code and reasonably sane to just ignore that sort
+   * of stuff. Approach was approved by Design.
```
phoenixAja: i don't think this part of the comment is needed, the rest of the description is good though 👍

src/frontend/src/views/GroupMembersPage/components/MembersTab/components/InviteModal/index.tsx
```diff
@@ -48,12 +51,35 @@ const InviteModal = ({ groupName, onClose, open }: Props): JSX.Element => {
...
+   */
+  const getAddressArrayFromInputValue = (): string[] => {
+    // First pass may have multiple spaces between chunks or pre/suffix spaces
+    const delimitBySpaces = inputValue.replace(INPUT_DELIMITERS, " ");
+    const delimitBySingleSpaceTrimmed = delimitBySpaces
```
phoenixAja, line 70: why re-assign to `delimitBySpaces` to `delimitBySingleSpaceTrimmed `?

phoenixAja (approved):
looks good, i would just take out that section of the comment if that sounds ok to you

src/frontend/src/views/GroupMembersPage/components/MembersTab/components/InviteModal/index.tsx
vincent-czi: That's fair. I guess I was thinking that the behavior is kind of surprising --`x@a.com,,,,,,  ,,   y@a.com` being valid is weird, frankly -- so I wanted to call out that the weirdness was acknowledged and accepted, not just inadvertent.

src/frontend/src/views/GroupMembersPage/components/MembersTab/components/InviteModal/index.tsx
vincent-czi: Sorry, I don't follow? It appears to be a weird auto-styling thing, but the methods that get called show up on the next line, so while it looks like a simple aliasing at first glance, it's actually running all those methods on original `delimitBySpaces` before setting result as `delimitBySingleSpaceTrimmed`. Does that clear it up, or did you mean something else, like why not just do it all in a single giant run of method calls? (which could be done, I just figured having an intermediate var made the process a bit easier to follow)

src/frontend/src/common/constants/inputDelimiters.ts
```diff
@@ -1 +1,2 @@
-export const INPUT_DELIMITERS = /[\n\t,]/g;
+export const INPUT_DELIMITERS = /[\n\t, ]/g;
```
mmmmmmaya, line 1: @vincent-czi You might want to check other sites where this is used to ensure the functionality of the inputs aren't affected.

vincent-czi: Good catch! Yeah, I realized over the weekend that might be a problem, sure enough it is. I need to make a PR and split out two versions of the delimiters.

PR #1145, opened 2022-06-13
Automatically strip leading/trailing whitespace in input strings.

jgadling:
### Summary:
- **What:** Enable pydantic automatic leading/trailing whitespace stripping feature

### Notes:
Pydantic docs are here: https://pydantic-docs.helpmanual.io/usage/model_config/#options
We've already encountered some issues with samples with leading/trailing tabs and spaces in their identifier names, and keeping extra whitespace is never the right behavior for this app.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/schemas/base.py
+++ b/src/backend/aspen/api/schemas/base.py
@@ -4,7 +4,12 @@ from pydantic import BaseModel
 
 
 class BaseRequest(BaseModel):
-    pass
+    class Config:
+        """Extra configuration options"""
+
+        anystr_strip_whitespace = (
+            True  # remove leading/trailing whitespace from strings
+        )
 
 
 def convert_datetime_to_iso_8601(dt: datetime) -> str:
--- a/src/backend/aspen/api/views/tests/test_create_samples.py
+++ b/src/backend/aspen/api/views/tests/test_create_samples.py
@@ -7,7 +7,7 @@ from httpx import AsyncClient
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm import joinedload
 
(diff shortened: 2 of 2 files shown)


src/backend/aspen/api/schemas/base.py
```diff
@@ -4,7 +4,12 @@
 
 
 class BaseRequest(BaseModel):
-    pass
+    class Config:
+        """Extra configuration options"""
+
+        anystr_strip_whitespace = (
+            True  # remove leading/trailing whitespace from strings
```
vincent-czi, line 11: Why is this in parens? Is it just an auto-formatter thing? It wanted to stay under 80chars or something? My first thought was there was some kind of tuple thing going on for the config, but I realize that's not the case now.

vincent-czi (approved):
Seems reasonable! Looks good, thanks for the test!

src/backend/aspen/api/schemas/base.py
lvreynoso: In this case the parens are just decorative and show an expression is written over multiple lines. Parentheses in Python can indicate a tuple if there's a comma separating a sequence, but in a single item tuple you would need a comma (e.g.`my_tuple = (item1,)`). They can also indicate a generator expression if it uses the comprehension syntax (`x for x in y if some_cond(x)`).

lvreynoso (approved):

src/backend/aspen/api/schemas/base.py
jgadling: the formatting tool didn't like my comment making the line super long so it wrapped the value onto another line with parens.

PR #1147, opened 2022-06-14
Add create group command

lvreynoso:
### Summary:
- **What:** Adds a command to aspencli to create a group and its corresponding auth0 org.
- **Ticket:** [sc200306](https://app.shortcut.com/genepi/story/200306)
- **Env:** [https://groupadd-backend.dev.czgenepi.org](https://groupadd-backend.dev.czgenepi.org)

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/schemas/usergroup.py
+++ b/src/backend/aspen/api/schemas/usergroup.py
@@ -1,16 +1,18 @@
 import datetime
 from typing import List, Optional
 
+from pydantic import constr
+
 from aspen.api.schemas.base import BaseRequest, BaseResponse
 from aspen.api.schemas.locations import LocationResponse
 
 
 class GroupCreationRequest(BaseRequest):
-    name: str
-    prefix: str
-    address: Optional[str]
-    division: Optional[str]
-    location: Optional[str]
+    name: constr(min_length=3, max_length=128, strict=True)  # type: ignore
+    prefix: constr(min_length=2, max_length=20, strict=True)  # type: ignore
+    address: Optional[constr(min_length=1, max_length=128, strict=True)]  # type: ignore
+    division: Optional[constr(min_length=1, max_length=128, strict=True)]  # type: ignore
+    location: Optional[constr(min_length=1, max_length=128, strict=True)]  # type: ignore
     default_tree_location_id: int
 
 
--- a/src/backend/aspen/api/schemas/usergroup.py
+++ b/src/backend/aspen/api/schemas/usergroup.py
@@ -5,6 +5,15 @@ from aspen.api.schemas.base import BaseRequest, BaseResponse
 from aspen.api.schemas.locations import LocationResponse
 
 
+class GroupCreationRequest(BaseRequest):
+    name: str
+    prefix: str
+    address: Optional[str]
+    division: Optional[str]
+    location: Optional[str]
+    default_tree_location_id: int
+
+
 class GroupResponse(BaseResponse):
     id: int
     name: str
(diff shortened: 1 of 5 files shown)


src/backend/aspen/api/views/groups.py
```diff
@@ -23,6 +24,32 @@
 router = APIRouter()
 
 
+@router.post("/", response_model=GroupInfoResponse)
+async def create_group(
+    group_creation_request: GroupCreationRequest,
+    db: AsyncSession = Depends(get_db),
+    auth0_client: Auth0Client = Depends(get_auth0_apiclient),
+    settings: Settings = Depends(get_settings),
+    user: User = Depends(get_auth_user),
```
jgadling: I think that if this endpoint is admin-only we can use get_admin_user here instead of get_auth_user

src/cli/aspencli.py
```diff
@@ -191,6 +193,12 @@ class CliConfig:
             "verify": True,
             "oauth_api_config": default_oauth_api,
         },
+        "rdev": {
```
jgadling: The CLI won't work with rdev unfortunately -- there's two layers of authorization in front of our rdev web services that mean device auth flow doesn't work properly here. Maybe we should look into a workaround at some point :'(

src/backend/aspen/api/views/groups.py
```diff
@@ -23,6 +24,32 @@
...
+) -> GroupInfoResponse:
+    if not user.system_admin:
+        raise ex.UnauthorizedException("Not authorized")
+    group = Group(**dict(group_creation_request))
+    organization = auth0_client.add_org(group.id, group.name)
```
jgadling: I don't think we have a group ID yet here. We have something of a catch-22 where db & auth0 groups each need to refer to each other. Maybe we should use `group.prefix.lower()` here instead?

src/backend/aspen/api/schemas/usergroup.py
```diff
@@ -5,6 +5,15 @@
 from aspen.api.schemas.locations import LocationResponse
 
 
+class GroupCreationRequest(BaseRequest):
+    name: str
+    prefix: str
+    address: Optional[str]
```
jgadling: can/should we use constr() on these input strings to restrict length or anything?

jgadling (approved):

src/backend/aspen/api/schemas/usergroup.py
lvreynoso: I think only `prefix` needs to be restricted, but I'm not sure what a good max character length would be.

lvreynoso: I'll add some reasonable restrictions.

vincent-czi: I'm a bit late to the conversation, but I wonder if we should add some sort of suffix requirement to `prefix` (heh). Most of our group prefixes end in `-` (eg, `CA-SLOC-`), but it's not a hard and fast rule right now, probably about 35% of our real user groups don't follow the convention. Seems like we should probably enforce a convention here, one way or the other?

... huh... Looks like the code changed at some point. We used to handle this via `create_public_ids` func which handled a dynamic dash, but it appears that function doesn't get called anywhere and the prefix is now directly used in the `generate_public_identifier` now. So if you look at the External Demo Group's account on Prod for an old sample (looking at 2022-03-17 right now), you see `hCoV-19/USA/demo-29624/2022`, but if you look at something more recent (2022-05-03) you get `hCoV-19/USA/demo--34277/2022` because the External Demo Group's group's prefix is `demo-` (not to be confused with DEMO ACCOUNT, which has a prefix of plain `demo`).

While we're discussing it, I wonder if we should also restrict spaces in `prefix`. I think there might be some stuff downstream that assumes public IDs never have a space included. I doubt we'd be crazy enough to enter a space when making a new group's `prefix`, but could be nice to enforce in code since this will be our point of intake from here on out.

PR #1149, opened 2022-06-14
chore: Replace happy deploy with a deployment trigger

alexlokshin-czi:


jgadling:
I think we also need to nix the "environment: prod" from this job

jgadling (approved):
added a commit to nix the "environment" parameter from this job

PR #1153, opened 2022-06-15
Additional backend docs.

jgadling:
### Summary:
- **What:** Update some developer docs

## Notes
This is just a start at making it easier to understand how to work with our backend application as a developer, somewhat more in line with the documentation we have available for the frontend.

### DIFF
--- a/docs/backend/README.md
+++ b/docs/backend/README.md
@@ -88,6 +88,11 @@ async def list_items(
     request: NewRequest, # GET requests often don't need input model validation, but this is here for completeness.
     db: AsyncSession = Depends(get_db),
     settings: Settings = Depends(get_settings),
+    # If this endpoint requires authentication, we need to make sure to depend on get_auth_user
+    # or get_admin_user here to validate the user's credentials and return 401/403 responses
+    # if their credentials are invalid. For most endpoints, get_auth_user is added at the root
+    # router in aspen/api/main.py, and it only needs to be included again here if we want to
+    # *use* the user object in our endpoint.
     user: User = Depends(get_auth_user),
 ) -> NewResponseList:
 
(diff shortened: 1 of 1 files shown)


docs/backend/README.md
```diff
@@ -1,3 +1,104 @@
...
+
+
+# Specify the HTTP method, sub-path (appending to the path added to main.py) and response model type.
+# The response model type here populates our API documentation.
+@router.get("/", response_model=NewResponseList) # All endpoints that return a list must end with a trailing slash
+async def list_items(
+    request: NewRequest, # GET requests often don't need input model validation, but this is here for completeness.
+    db: AsyncSession = Depends(get_db),
+    settings: Settings = Depends(get_settings),
+    user: User = Depends(get_auth_user),
```
vincent-czi, line 96: I think this line deserves a special call out. While `user` does not actually get used below, the call to `get_auth_user` is actually the hinge of our auth right now. It being called is what prevents access by unauthenticated users. I don't think that's particularly obvious, but since it's a core part of our auth process it's important to make it explicit.

Unlike the `settings` above, which is just serving as an example of how you could pull the app settings if you needed them in your view function, the `Depends(get_auth_user)` is doing more than just pulling the user's data: it's also kicking a 401 if the request is from a non-authenticated user and blocking the view code from firing.

vincent-czi (approved):
This is great! Very clear! I think the one comment I have is pretty deserving of a tweak, but that's my only note.

docs/backend/README.md
jgadling: fair enough, I've added a comment block with more detail.

vincent-czi: Perfect, that's extremely clear. Thanks!

PR #1159, opened 2022-06-15
Add location search functionality in API and CLI

lvreynoso:
### Summary:
- **What:** Adds an api endpoint to search for locations in the database and return them as a list of json dicts. Adds a cli command to query this endpoint that prints results in a pretty table. Enables the `fuzzystrmatch` Postgres extension to make this all work.
- **Ticket:** [sc200306](https://app.shortcut.com/genepi/story/200306)
- **Env:** [https://locationsearch-backend.dev.czgenepi.org](https://locationsearch-backend.dev.czgenepi.org)

### Demos:
You will need to navigate to the `src/cli` directory and do a `pip install -r requirements.txt` to add a package needed for the cli (`tabulate`).

The cli command is `aspencli.py locations search`. The command has four available options: `--region`, `--country`, `--division`, `--location`. You must enter a value for at least one option. 

Querying with only a `--country` or `--division` without a `--location` will show you all the divisions in that country, and all the locations in that division respectively.
Screen Shot 2022-06-16 at 10 21 16
Screen Shot 2022-06-15 at 15 49 38
Screen Shot 2022-06-15 at 15 50 11

Querying using a `--location` will return the first 30 results.
Querying just a `--location` will attempt to do a closest text match, but it's rather simple and only uses levenshtein distance. 
Screen Shot 2022-06-15 at 15 50 59
Screen Shot 2022-06-15 at 15 52 04

Querying a `--location` with a `--country` or a `--division` will restrict the text search for that location to that country or division, increasing your odds of finding what you want.
Screen Shot 2022-06-15 at 15 52 47
Screen Shot 2022-06-15 at 15 53 04

### Notes:
I'm sure the API endpoint code could be made more readable somehow, but it works as-is and I can't easily see a better way at the moment.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/views/locations.py
+++ b/src/backend/aspen/api/views/locations.py
@@ -1,20 +1,23 @@
+import functools
+
 import sqlalchemy as sa
 from fastapi import APIRouter, Depends
+from sqlalchemy import asc
 from sqlalchemy.ext.asyncio import AsyncSession
 from starlette.requests import Request
 
 from aspen.api.auth import get_auth_user
 from aspen.api.deps import get_db, get_settings
-from aspen.api.schemas.locations import LocationListResponse, LocationResponse, LocationSearchRequest
+from aspen.api.schemas.locations import (
+    LocationListResponse,
+    LocationResponse,
+    LocationSearchRequest,
+)
 from aspen.api.settings import Settings
 from aspen.database.models import Location, User
 
 router = APIRouter()
 
-import logging
-
-logger = logging.getLogger(__name__)
-logger.setLevel(logging.DEBUG)
 
 @router.get("/", response_model=LocationListResponse)
 async def list_locations(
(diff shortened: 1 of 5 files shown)


src/backend/aspen/api/views/locations.py
```diff
@@ -28,3 +40,126 @@ async def list_locations(
         response.append(LocationResponse.from_orm(row))
 
     return LocationListResponse.parse_obj({"locations": response})
+
+
+# The idea in this route is that,
+# given a country, show all divisions in that country,
```
jgadling, line 46: We also need to list the country without any divisions, in case someone wants to build country-level trees.

lvreynoso: added!

jgadling (approved):

PR #1160, opened 2022-06-15
Prevent duplicate users / invitations.

jgadling:
### Summary:
- **What:** Don't send invitations for users that already exist, and add in MagicMock
- **Ticket:** [sc200169](https://app.shortcut.com/genepi/story/200169)

### Notes:
There is still an edge case here - I'm only checking against our list of *accepted* users since it's pretty expensive to check all existing groups to see if any of them have a valid invitation for the same email address. I think in practice this is expected to be rare, and we're going to drop the uniqueness requirement soon, so ... any objections?

Also this replaces our MockAuth0Client with an instance of [Mock](https://docs.python.org/3/library/unittest.mock.html#the-mock-class) that's shared between our test code and our test API, so we can mock out our auth0 interface dynamically.

### Checklist:
- [x] I added labels to my PR


lvreynoso (approved):
lgtm

PR #1168, opened 2022-06-17
Sync admin flags between auth0 and czge

jgadling:
### Summary:
- **What:** Makes sure users invited with the admin role have that role reflected in our app.

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/schemas/usergroup.py
+++ b/src/backend/aspen/api/schemas/usergroup.py
@@ -43,6 +43,7 @@ class UserUpdateRequest(BaseRequest):
 class UserMeResponse(UserBaseResponse):
     split_id: str
     group: GroupResponse
+    group_admin: bool
 
 
 class GroupInvitationsRequest(BaseRequest):
--- a/src/backend/aspen/api/views/auth.py
+++ b/src/backend/aspen/api/views/auth.py
@@ -13,8 +13,10 @@ from starlette.requests import Request
 from starlette.responses import Response
 
 import aspen.api.error.http_exceptions as ex
+from aspen.api.auth import get_auth0_apiclient
 from aspen.api.deps import get_auth0_client, get_db, get_settings
 from aspen.api.settings import Settings
+from aspen.auth.auth0_management import Auth0Client
 from aspen.database.models import Group, User
 
 # From the example here:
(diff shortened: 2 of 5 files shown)


src/backend/aspen/api/schemas/usergroup.py
```diff
@@ -43,6 +43,7 @@ class UserUpdateRequest(BaseRequest):
 class UserMeResponse(UserBaseResponse):
     split_id: str
     group: GroupResponse
+    group_admin: bool
```
jgadling, line 46: Adds group admin flag to the /users/me response @mmmmmmaya

src/backend/aspen/api/views/auth.py
```diff
@@ -66,11 +68,14 @@ async def create_user_if_not_exists(db, userinfo):
         group = groupquery.scalars().one()  # type: ignore
     except NoResultFound:
         return  # We didn't find a matching group, we can't create this user.
+    # Get the user's roles for this organization and tag them as group admins if necessary.
+    roles = auth0_mgmt.get_org_user_roles(userinfo["org_id"], auth0_user_id)
```
jgadling, line 72: Get a list of roles associated with the auth0 user and group

src/backend/aspen/api/schemas/usergroup.py
mmmmmmaya: Will this need updates when we get to multigroup, or will this stay the same structure and just update the value depending on which group is shown?

src/backend/aspen/api/views/groups.py
```diff
@@ -134,7 +134,11 @@ async def invite_group_members(
         try:
             if success:
                 auth0_client.invite_member(
-                    organization["id"], client_id, user.name, email, "member"
+                    organization["id"],
+                    client_id,
+                    user.name,
+                    email,
+                    group_invitation_request.role,
```
jgadling, line 141: Send the role name that we get from the `group invite` cli to Auth0 so new users are created with a role.

lvreynoso (approved):
lgtm

src/backend/aspen/api/schemas/usergroup.py
jgadling: hehe, I think that's a much bigger conversation -- let's talk next week!

mmmmmmaya (approved):

PR #1184, opened 2022-06-23
Install oso & some initial fastapi hooks as a no-op

jgadling:
### Summary:
- **What:** This sets up oso so that we can start integrating it with our backend. **It doesn't actually put oso in any functionality codepaths yet though**.
- **Ticket:** [sc187623](https://app.shortcut.com/genepi/story/187623)
- **Env:** https://osov0-frontend.dev.czgenepi.org/

### Demos:
I've already linked out some oso design docs to the team, this is basically the integrated version of those docs, please reach out if

### Notes:
A few key points in this PR:
- This PR creates new `roles`, `user_roles` and `group_roles` tables *and migrates existing user and group roles to these tables.* Any API endpoints that get migrated to use Oso will use these roles tables, and any endpoints that haven't been migrated will still use our old `user.group_admin`/`user.group_id` columns and the `can_see table` for authz. It's *very important* that we keep these data sets in sync until the oso migration is complete.
- This updates the "user creation on first login" code to sync new user roles to *both* the old and the new access control columns/tables, so that should help everything stay in sync for now.
- This installs the Oso authorization framework [from our fork](https://github.com/chanzuckerberg/oso/tree/czge-oso25) that includes SQLAlchemy 2.0 query syntax, async support, and postgres `distinct on` functionality. The installation method is hacky right now, but I want to make sure it will work for us before investing in better packaging.
- There are a few starter policies and model mappings included in this PR that should be pretty close to what we're going to need to migrate the first few endpoints, but these will probably be adjusted as our understanding evolves. 

### Checklist:
- [x] I merged latest `<base branch>`
- [x] I manually verified the change
- [x] I added labels to my PR
- [ ] I tested in multiple browsers
- [x] I added relevant unit tests
- [ ] I have notified others of changes they need to make locally (migrations, jobs, package updates, etc)

PR #1185, opened 2022-06-24
Return group roles as part of /users/me response.

jgadling:
### Summary:
- **What:** Add groups & group roles to `/v2/users/me` response.
- **Ticket:** [sc192742](https://app.shortcut.com/genepi/story/192742)
- **Env:** `<rdev link>`

### Notes:
This builds on #1184

### Demos:
The response format looks like this for now, we can tweak it as necessary if it helps with anything @mmmmmmaya 
Screen Shot 2022-06-24 at 11 11 17 AM

- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/test_infra/models/usergroup.py
+++ b/src/backend/aspen/test_infra/models/usergroup.py
@@ -1,6 +1,9 @@
 from typing import Optional
 from uuid import uuid1
 
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from aspen.auth.role_manager import RoleManager
 from aspen.database.models import Group, Location, User
 
 
@@ -47,7 +50,7 @@ def user_factory(
     system_admin=False,
     agreed_to_tos=True,
 ) -> User:
-    return User(
+    user = User(
         name=name,
         auth0_user_id=auth0_user_id,
         email=email,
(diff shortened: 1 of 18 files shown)


mmmmmmaya:
I assume this is right but confirming top level `group` = currently authed group and the `groups` is just all group user belongs to?

this is 😻

jgadling:
> I assume this is right but confirming top level `group` = currently authed group and the `groups` is just all group user belongs to?
> 
> this is 😻

We can do this two ways.

First way (my preference): The top level group and group_admin flag are going away once we finish this migration. A user will always have access to all their groups, and it's up to the front end to send the current group context to the backend when asking for samples/trees/etc. 

Second way, also will work fine but seems a little weird to me?: The frontend is welcome to send a group ID to the `/users/me` endpoint (it'll be like `/v2/orgs/1234/users/me` and I can populate the top level group and group admin fields with the group referenced in the URL.

mmmmmmaya:
The first way you proposed sounds good to me!

phoenixAja (approved):
looks good to me!

src/backend/aspen/test_infra/models/usergroup.py
```diff
@@ -47,7 +50,7 @@ def user_factory(
     system_admin=False,
     agreed_to_tos=True,
 ) -> User:
-    return User(
+    user = User(
```
lvreynoso, line 53: Will this function still be used after this PR, or is not every instance replaced with `userrole`? Could it be deleted?

lvreynoso (approved):
lgtm

src/backend/aspen/test_infra/models/usergroup.py
jgadling: Ohhh good point! our workflow tests still use this method but I think we can work on removing it as part of our oso migration.

PR #1196, opened 2022-06-30
Maya/192742 redux

mmmmmmaya:
### Summary
- **What:** Implement from end state management
  - include examples of how to write actions, selectors, and how to dispatch actions to update the store
- **Why:** Allow us to start storing certain state on the front end, such as current /last viewed group/pathogen 
- **Ticket:** [[sc-192742]](https://app.shortcut.com/genepi/story/192742)
- **Env:** https://maya-redux-frontend.dev.czgenepi.org/

### Testing
0. install [redux dev tools](https://chrome.google.com/webstore/detail/redux-devtools/lmhkpmbekcpmknklioeibfkpmmfibljd?hl=en) and restart chrome
1. load the app and make sure it doesn't crash
2. check out chrome dev tools `Application` tab and see that localstorage is empty
1. check out the Redux dev tools tab and see the state looks like this
>   ```
>     current: {
>       group: null,
>       pathogen: null,
>     },
>  ```

### Demos
after

### Next steps
- stop depending on /me to know what group is current
- if no group in localstorage, choosing to display the oldest group to the user by default
- updating api calls to inculde group id in path

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/frontend/src/common/types/redux.d.ts
```diff
@@ -0,0 +1,8 @@
+type ActionType<T> = (payload: T | null) => {
+  type: string;
+  payload: T | null;
```
vincent-czi: This feels to me like it should be `payload?:` for those actions that don't need any kind of payload at all. But I don't know a way to do it with generics. Needing to explicitly pass `payload: null` when creating actions that have no payload aspect seems overly verbose, but I can't see a way to do it beyond `ActionType<T>` and `ActionNoPayloadType`. If you can think of an elegant way to do it, I'd think of that as a nice-to-have.

src/frontend/src/common/redux/middleware/index.ts
```diff
@@ -0,0 +1,18 @@
...
+ * to learn more about how these functions work.
+ */
+
+import { AnyAction, Middleware } from "redux";
+import { setLocalStorage } from "src/common/utils/localStorage";
+import { groupPersistedName, setGroupActionType } from "../actions";
+
+export const setGroupMiddleware: Middleware =
+  () => (next) => (action: AnyAction) => {
+    if (action.type === setGroupActionType) {
```
vincent-czi: Nit: If we have the `getLocalStorage(pathogenPersistedName)` happening elsewhere to read the persisted pathogen, it feels like we might as well have setting of the persisted pathogen also happen here. I'd also be fine with removing reading it, but it feels like they should be implemented (or not) as a pair.

src/frontend/src/common/redux/actions/index.ts
```diff
@@ -0,0 +1,15 @@
+// export action type for use in reducers and middleware
+// persisted names are for use in localstorage
+export const groupPersistedName = "currentGroup";
+export const setGroupActionType = "group/setGroup";
```
vincent-czi: I have a strong belief that we should have a dedicated `actionTypes` enum/var -- I'd lean to it being its own dedicated file, but I'm not against it living with the actions (especially while we have so few types right now) -- rather than independent actionType variables.

If I interrogate my reasoning, a good chunk of it is that it's just what I'm used to from prior projects, but I do think it's useful to have a canonical location for them. Also it helps prevent the temptation to just directly use a string in your action/reducer. Also also, if we have an enum of them, we can then require the ActionType TS type to have its `type` property be one of those values, which seems like a good safety feature.

vincent-czi (approved):
Very nice! I have one comment -- the one about a dedicated `actionTypes` collection -- I would really like to see integrated. The others are more minor.

src/frontend/src/common/types/redux.d.ts
mmmmmmaya: This is a good callout

src/frontend/src/common/types/redux.d.ts
```diff
@@ -1,8 +1,20 @@
...
+  type: CZGEReduxActions;
+  payload?: T;
 };
 
 enum Pathogen {
   COVID = "covid",
 }
+
+// export action type for use in reducers and middleware
+enum CZGEReduxActions {
```
vincent-czi: This comment goes for both `CZGEReduxActions` and `ReduxPersistenceTokens` -- 
I personally always find the `.d.ts` files a little... invisible? I dunno, they never come first to mind as places to go look, and because you aren't explicitly importing them elsewhere, I have a tendency to forget about them. Maybe that's on me. For types, I think it's reasonable since types sort of make up the soup that all the code swims in, so it being a bit invisible isn't a big deal. But it just feels like the enums of actions and persistence tokens are more in the "concrete" world of stuff I would _expect_ to import?

I dunno, if you think this is the right place for these, that's fine. But my gut says I would expect `ActionType` to be pulled in automatically from `.d.ts` somewhere, but I would need to explicitly import `CZGEReduxActions` from some `/common/redux` or whatever if I was going to be dispatching or reducing.

vincent-czi (approved):
Really like the changes, thanks! I've got one comment about where the actions and tokens live that I think is reasonable, but if you disagree, you can ignore it.

src/frontend/src/common/types/redux.d.ts
mmmmmmaya: It's not possible to pull the enums outs without also pulling out ActionType, because the declaration files aren't modules so you can't import anything into them.

However, I put it into a types.ts file which will require importing for consuming modules, as you requested (although, I personally like the magic of not having to import common type, and prefer to just search for them when needed).

vincent-czi: I think this comes down to me not seeing the redux action type as a _type_, per se. To me, it's a not a TypeScript type indicating the underlying data structure, but rather a constant that just serves as a signal. Now, considering that TS has a whole thing about constant valued types, maybe that's just silly of me. It could simply be that all my experience with redux was in a JS context, so seeing an action type as _also_ being a TS type feels weird to me just because it's new. So yeah, if you prefer it the other way around, you can revert this change in a future PR and I'll just get used to it.

PR #1199, opened 2022-07-05
Add authorizedsession object for using oso in Depends calls

lvreynoso:
### Summary:
- **What:** Adds a function, `require_access(permission, resource)` that can be used as part of a `Depends()` call in the function signature of an API endpoint. Returns an oso authorized query.
- **Env:** `<rdev link>`

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/authz.py
+++ b/src/backend/aspen/api/authz.py
@@ -166,16 +166,6 @@ async def get_authz_session(
     return AuthZSession(session, auth_context)
 
 
-async def get_oso(session: AsyncSession = Depends(get_db)) -> AsyncOso:
-    oso = AsyncOso()
-    oso.set_data_filtering_adapter(AsyncSqlAlchemyAdapter(session))
-    register_classes(oso)
-    await oso.load_files(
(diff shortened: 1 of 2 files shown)


src/backend/aspen/api/authz.py
```diff
@@ -163,3 +164,41 @@ async def get_authz_session(
...
+        self.model = model
+
+    async def __call__(
+        self,
+        auth_context: AuthContext = Depends(get_auth_context),
+        oso: AsyncOso = Depends(get_oso),
+    ):
+        self.auth_context = auth_context
+        self.oso = oso
+        return await self.authorized_query()
```
jgadling: This should be returning an authorized query from AuthZSession.authorized_query() instead of an oso query directly:
https://github.com/chanzuckerberg/czgenepi/blob/trunk/src/backend/aspen/api/authz.py#L138

jgadling (changes requested):

src/backend/aspen/api/authz.py
lvreynoso: yeah, my bad

jgadling (approved):

PR #1201, opened 2022-07-07
Vince/segment methods

vincent-czi:
### Summary:
- **What:** Establish process to call all the analytics framework methods we care about
- **Ticket:** [sc<192935>](https://app.shortcut.com/genepi/story/<192935>) -- This ticket is really not a good indication of the work done here. I basically started at proof of concept and figured I might as well implement all the structure for actually using the framework in one go, but never updated the ticket to match.
- **Env:** no rdev, going to launch an rdev with next PR once all the various events are in place and verify everything is landing as expected at once (I was verifying locally though as I worked)

### Notes:
Big goal of this work was twofold:
1. Set up the analytics stuff to be very easy to use in the future (just import function and fire away, no need to really understand any of how analytics is set up)
2. Set it up in such a way so that it would be hard to use wrong (try to ensure the functions all handle opting-out, already deal with edge case / race condition and will just quietly work as expected, decent guard rails in place to avoid leaking PII)

### Checklist:
- [x] I merged latest `vince/analytics` (long lived analytics branch)
- [x] I added labels to my PR
~[ ] I have notified others of changes they need to make locally (migrations, jobs, package updates, etc)~

### DIFF
--- a/src/frontend/src/common/analytics/methods.ts
+++ b/src/frontend/src/common/analytics/methods.ts
@@ -1,0 +1,184 @@
+/**
+ * Funcs that handle calling the various Segment methods we use for analytics.
+ *
+ * Broadly speaking, our data capture for analytics splits into three types:
+ * - Events: A discrete "thing" happening in our app, generally initiated by
+ *   a user, but could also could be an event to track an internal happening
+ *   (e.g., noting an upload failing because of a BE error).
+ *     Handled by `analyticsTrackEvent`
+ * - User info: Capturing info on the user. We make sure to de-identify the
+ *   user and only send up non-PII data to analytics. This occurs when the
+ *   app loads, and also if they later change something about how they are
+ *   seen as a user (e.g., which group they are viewing).
+ *     Handled by `analyticsSendUserInfo`
+ * - Page/Route: Noting when the user's page changes and recording what page
+ *   they are now on. This happens for every route change.
+ *     Handled by `analyticsRecordRouteChange`
+ *
+ * For all of the functions here, if a user has opted-out of analytics, the
+ * functions will do nothing. That way these functions can be safely called
+ * in the rest of our code and all of the opt-in/out logic is taken care of.
+ */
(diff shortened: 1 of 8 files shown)


src/frontend/src/common/analytics/methods.ts
```diff
@@ -0,0 +1,184 @@
...
+ * expose any sensitive data from FE to analytics.
+ *
+ * TODO -- This function will need to grow in complexity and probably alter
+ * its signature once we have multi-group membership in place.
+ */
+export function extractAnalyticsUserInfo(userInfo: User): AnalyticsUserInfo {
+  // TODO [Vincent, near future] Below is a temporary anonymous ID for dev work
+  // and will be replaced soon by a proper anonymized user ID once that aspect
+  // is in place in the backend and comes with rest of userInfo data.
+  const PLACEHOLDER_ANON_USER_ID = "warrrbargl";
```
mmmmmmaya, line 102: ❤️

src/frontend/src/common/analytics/methods.ts
```diff
@@ -0,0 +1,184 @@
...
+ * `page` calls until we can fix whatever is causing the weird router stuff.
+ */
+let previouslyRecordedRoute = "";
+export function analyticsRecordRouteChange(pageUrl: string): void {
+  if (
+    window.analytics &&
+    window.isCzGenEpiAnalyticsEnabled &&
+    previouslyRecordedRoute !== pageUrl
+  ) {
+    window.analytics.page();
```
mmmmmmaya, line 181: do you need to pass `pageUrl` to `window.analytics.page`?

mmmmmmaya (approved):
just one q, thought maybe we needed to tell analytics.page _which_ page the user went to, but not sure. 

otherwise lgtm

src/frontend/src/common/analytics/methods.ts
vincent-czi: Nah, it just pulls it automatically

PR #1202, opened 2022-07-07
Update groups view to support oso

lvreynoso:
### Summary:
- **What:** Updates endpoints in the groups view to support oso, except for group creation.
- **Env:** [https://osogroup-frontend.dev.czgenepi.org/](https://osogroup-frontend.dev.czgenepi.org/)

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/views/groups.py
+++ b/src/backend/aspen/api/views/groups.py
@@ -1,13 +1,12 @@
 import re
-from typing import Union
 
 import sentry_sdk
 import sqlalchemy as sa
 from auth0.v3.exceptions import Auth0Error
 from fastapi import APIRouter, Depends
-from sqlalchemy.sql.expression import Select
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm import joinedload
+from sqlalchemy.sql.expression import Select
 from starlette.requests import Request
 
 from aspen.api.authn import get_admin_user, get_auth0_apiclient, get_auth_user
@@ -28,10 +27,6 @@ from aspen.database.models import Group, User
 
 router = APIRouter()
 
-import logging
-
-logger = logging.getLogger(__name__)
-logger.setLevel(logging.DEBUG)
 
 @router.post("/", response_model=GroupInfoResponse)
 async def create_group(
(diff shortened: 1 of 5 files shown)


src/backend/aspen/api/views/groups.py
```diff
@@ -62,17 +64,15 @@ async def get_group_info(
     group_id: int,
     request: Request,
     db: AsyncSession = Depends(get_db),
-    user: User = Depends(get_auth_user),
+    authorized_query: Select = Depends(require_access("read", Group)),
```
jgadling, line 67: ooooh.... doesn't have to happen right now, but maybe we have a dependency that finds the group(/[object]) by the id referenced in the endpoint path automatically!!

src/backend/aspen/api/views/groups.py
```diff
@@ -62,17 +64,15 @@ async def get_group_info(
...
-    if user.group.id != group_id and not user.system_admin:
-        raise ex.UnauthorizedException("Not authorized")
-    group_query = (
-        sa.select(Group)  # type: ignore
-        .options(joinedload(Group.default_tree_location))
-        .where(Group.id == group_id)
-    )
+    group_query = authorized_query.options(  # type: ignore
+        joinedload(Group.default_tree_location)
+    ).where(Group.id == group_id)
```
jgadling, line 71: hrm... then again we join more stuff to these queries so that might not be as awesome as it sounds

src/backend/aspen/database/models/usergroup.py
```diff
@@ -55,6 +55,8 @@ class Group(idbase, DictMixin):  # type: ignore
         server_default=text("'{}'::jsonb"),
     )
 
+    members = relationship("User", back_populates="group")  # type: ignore
```
jgadling, line 58: Not totally sure why these mappings are updated -- just a heads-up that user.group_id is eventually going to go away.

src/backend/aspen/api/views/groups.py
```diff
@@ -81,32 +81,31 @@ async def get_group_members(
...
-    if user.group.id != group_id and not user.system_admin:
-        raise ex.UnauthorizedException("Not authorized")
-    group_members_query = (
-        sa.select(User).where(User.group_id == group_id).order_by(User.name.asc())  # type: ignore
-    )
-    group_members_result = await db.execute(group_members_query)
-    group_members = group_members_result.scalars().all()
-    return GroupMembersResponse.parse_obj({"members": group_members})
+    group_with_members_query = authorized_query.options(  # type: ignore
+        joinedload(Group.members)
```
jgadling: Ahhh - here we want to update the query to grab data from `user_roles` and include the role info!

jgadling (changes requested):
Looks good! Only one thing - we're ready to update the members endpoints to use the `user_roles` table now instead of `user.group_id` and we should probably also update the response to return roles for all the group members

src/backend/aspen/api/views/groups.py
lvreynoso: Yeah. I think SQLAlchemy has methods to populate model attributes like this but I've never played with them, so I don't know how tricky it would be to use them.

src/backend/aspen/database/models/usergroup.py
lvreynoso: 😵‍💫

jgadling (approved):

PR #1204, opened 2022-07-08
Maya/192742 redux

mmmmmmaya:
### Summary
- **What:**
  - Replace BE calls for useUserInfo to calculate group id with redux store hydration
  - If no defaults set in localstorage, set them both in redux and localstorage
- **Why:** because user.group is gonna go 💀 
- **Ticket:** [[sc-192742]](https://app.shortcut.com/genepi/story/192742)
- **Env:** https://maya-redux-frontend.dev.czgenepi.org/

### Testing
1. Clear localstorage (you can check on Application tab of chrome dev tools)
1. Load app
2. Shouldn't crash
3. Check localstorage again
4. Should now have your groupid and pathogen=covid set in localstorage
5. App ui shouldn't be janked

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/frontend/src/common/redux/index.ts
```diff
@@ -1,12 +1,33 @@
...
+ * comes online.
+ * So we have a compromise to use defaults for a fast initial hydration, and then immediately
+ * make reqeusts for "true" default values from the server if needed.
+ */
+
+// first, load state from localstorage if any exists and use it to initialize redux
 const getInitialState = () => {
   const storedGroupStr = getLocalStorage(ReduxPersistenceTokens.GROUP);
-  const storedGroup = storedGroupStr ? parseInt(storedGroupStr) : null;
+  const storedGroup = storedGroupStr ? parseInt(storedGroupStr) : -1;
```
vincent-czi: Nit: I generally prefer `null` as a way of informing that something does not exist at all rather than using `-1` or `""`. I know for interfaces it means that there's a bit of extra boilerplate since you need to union on a `| null` possibility, but it feels -- to me at least -- more obvious what the interface is representing. With just a `key: number` I assume the property is always present and known, while with `key: number | null` it feels more clear that it can also be in a "weird" state (`key?: number` also accomplishing the same, albeit with different undertones).

src/frontend/src/views/GroupMembersPage/components/MembersTab/components/InviteModal/index.tsx
```diff
@@ -47,8 +51,26 @@ const InviteModal = ({ groupName, onClose, open }: Props): JSX.Element => {
...
+      if (successCount > 0) setIsSuccessNotificationOpen(true);
+
+      handleClose();
+    },
+    componentOnError: noop,
+  });
+
+  // can't send invites if we don't know what group they are in
+  const groupId = useSelector(selectCurrentGroup);
+  if (!groupId) return null;
```
vincent-czi: Maybe I'm mistaken, but won't `storedGroup` potentially load as `-1` while we wait for `setDefaults` in redux-land to complete, so this would evaluate (briefly) as truth-y?

vincent-czi (approved):
Overall, looks good, but I think there might be an edge case issue with `groupId` loading as `-1` when missing, then evaluating as truth-y elsewhere when it's assumed missing means false-y. I could be wrong though.

src/frontend/src/views/GroupMembersPage/components/MembersTab/components/InviteModal/index.tsx
mmmmmmaya: Yes, which is what it already currently does. Jess said she's cool with the failed requests while we wait to figure out what group the user is in

src/frontend/src/common/redux/index.ts
mmmmmmaya: I don't want it to be possible to ever set the group to `null`, though. I prefer to load it in as a placeholder value for a second while the app loads and always have it be a number from that point on.

src/frontend/src/views/GroupMembersPage/components/MembersTab/components/InviteModal/index.tsx
mmmmmmaya: that's true. Just fixed!

PR #1205, opened 2022-07-08
Add a permission for creating samples/phyloruns within groups

jgadling:
### Summary:
- **What:** adding `create` permissions
- **Env:** `<rdev link>`

### Notes:
I'm honestly not sure at what level the authz checking for "can create a new thing in this group" should be. It feels like it should be on samples/phyloruns directly, but without a group we can't check that permission directly. I think this will get the job done for now, but open to alternatives here.

### Checklist:
- [ ] I added labels to my PR


lvreynoso (approved):
lgtm - I think adding those permissions onto a Group is reasonable.

PR #1211, opened 2022-07-13
fix: When happy errors out, workflow assumed stack needs to be created

alexlokshin-czi:


jgadling (changes requested):
I think we need to fix the problem in the upstream (deploy-happy-stack) action

jgadling (approved):

PR #1212, opened 2022-07-13
update nextstrain docker

danrlu:
### Summary:
- **What:** Brought in the latest upstream from `nextstrain/ncov` which included the Nextclade V2 update https://github.com/chanzuckerberg/ncov/pull/8. This PR releases the pin on our nextstrain docker so we can use the latest one which has Nextclade V2. 

### Demos:
Tree build w the new docker and new ncov pin worked together. Will test again in staging.

### Checklist:
- [x] I added labels to my PR


vincent-czi (approved):
Seems reasonable! Only thing I'm wondering is if we should pin to whatever the latest/current sha is as a defensive move. If they tend to change things and cause breakage on our end, pinning now could be worth it, but the downside is we won't be automatically getting updates. But I wouldn't blink at merging this as is.

danrlu:
Let me investigate how often they change it and see. Thank you!!

PR #1215, opened 2022-07-13
Allow easier selection of multiple lineages

lvreynoso:
### Summary:
- **What:** Adds wildcard entries to the lineage list for building phylo trees.
- **Ticket:** [sc199195](https://app.shortcut.com/genepi/story/199195)
- **Env:** [https://wildcards-frontend.dev.czgenepi.org/](https://wildcards-frontend.dev.czgenepi.org/)

### Demos:
Option available in dropdown:
Screen Shot 2022-07-13 at 16 48 54
Screen Shot 2022-07-18 at 15 50 43

Wildcard expansion:
Screen Shot 2022-07-13 at 16 50 15
Screen Shot 2022-07-18 at 15 51 03

### Notes:
TODO: add tests

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/utils/lineage.py
+++ b/src/backend/aspen/api/utils/lineage.py
@@ -6,31 +6,38 @@ from sqlalchemy.ext.asyncio import AsyncSession
 
 from aspen.database.models import PangoLineage
 
-"""
-Takes in a list of lineages and wildcard lineage expressions (as strings) and
-returns a list of specific lineages. Wildcards are expanded to include both
-the base expression and all sublineages.
-
-Example:
-We encounter the wildcard expression 'BA.1.*'.
-We add:
-'BA.1'
-'BA.1.1'
-'BA.1.1.7',
-etc.
-
-but not 'BA.11'.
-"""
+NEXTSTRAIN_LINEAGE_MAP = {
+    "BA.1*": "21K",
+    "BA.2*": "21L",
+    "BA.4*": "22A",
+    "BA.5*": "22B",
+}
 
 
 async def expand_lineage_wildcards(db: AsyncSession, lineage_list: List[str]):
+    """
+    Takes in a list of lineages and wildcard lineage expressions (as strings) and
+    returns a list of specific lineages. Wildcards are expanded to include both
+    the base expression and all sublineages.
+
+    Example:
+    We encounter the wildcard expression 'BA.1*'.
+    We add:
+    'BA.1'
+    'BA.1.1'
+    'BA.1.1.7',
+    etc., but not 'BA.11'.
+    """
     all_lineages_query = sa.select(PangoLineage.lineage)  # type: ignore
     result = await db.execute(all_lineages_query)
     all_lineages = set(result.scalars().all())
 
     expanded_lineage_list = set()
     for entry in lineage_list:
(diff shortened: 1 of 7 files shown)


src/backend/aspen/api/utils/lineage.py
```diff
@@ -0,0 +1,42 @@
...
+
+from aspen.database.models import PangoLineage
+
+"""
+Takes in a list of lineages and wildcard lineage expressions (as strings) and
+returns a list of specific lineages. Wildcards are expanded to include both
+the base expression and all sublineages.
+
+Example:
+We encounter the wildcard expression 'BA.1.*'.
```
danrlu: Very minor: I was a bit on the fence whether ppl would interpret `BA.1.*` to include `BA.1` or `BA.1*` to include `BA.1`, and noticed here: https://github.com/cov-lineages/pango-designation/blob/master/pango_designation/alias_key.json they use `BA.1*`. I think non-coding ppl may be less likely to expect `BA.1*` to include `BA.11`?

lvreynoso: That makes sense, I'll update it to keep it consistent with what's in the `cov-lineages` repo. My coder brain definitely interprets `BA.1*` to include `BA.11`

mmmmmmaya:
were you also going to add *fancy* options for delta/omicron?

mmmmmmaya:
The wildcards look great
Screen Shot 2022-07-14 at 12 58 04 PM

but would still need options that expand out for, eg, Delta and Omicron groupings
Screen Shot 2022-07-14 at 12 59 39 PM

danrlu:
My understanding is we're going for the fancy version to include:
BA.1* / 21K
BA.2* / 21L
BA.4* / 22A
BA.5* / 22B
while Delta and Omicron would be in the extra fancy version for later?? Since it's more than expanding the wildcard
image
Lucia did I get it right about the plan?

danrlu:
For Maya's ref, I was trying to give diff options and eng can decide which one to go with depending on bandwidth and interest. Now I see this can be confusing 😆 Next time we'll agree upon which option to implement earlier on.

mmmmmmaya:
Sorry, I think the confusion came in because Jess and I had kinda agreed we were going to do the extra fancy option, and then Jess ended up not picking up the ticket. I'm interested in going for extra fancy now so we can close this ticket and never think about it again.

Is that possible?

danrlu:
I'll leave it to @lvreynoso. And I'm sorry that Lucia asked for clarification for the "fancy versions" and I interpreted that as the "fancy version". I'll leave the rest to you to coordinate since I'm not helping positively 🙇‍♀️ 

I'm happy to do any tests if needed. Thank you all!!

jgadling:
It's pretty easy to support Delta and Omicron now, so we may as well do it?

lvreynoso:
Yeah might as well add Delta and Omicron. Just those two?

danrlu:
Just those two is good for now. If any other lineages come up in the future or by user requests I will be able to add them in. Thanks very much Lucia!

lvreynoso:
Added Delta and Omicron. Should be everything

jgadling (approved):
LGTM!

PR #1226, opened 2022-07-15
Don't check for duplicate users before sending invitations

jgadling:
### Summary:
- **What:** This fixes two issues: it allows inviting users with existing CZGE accounts to work and it also allows admins to invite users to any group.
- **Ticket:** [sc187619](https://app.shortcut.com/genepi/story/187619)
- **Env:** https://invite-frontend.dev.czgenepi.org/

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/authn.py
+++ b/src/backend/aspen/api/authn.py
@@ -196,7 +196,7 @@ class AuthContext:
     def __init__(
         self,
         user: User,
-        group: Group,
+        group: Optional[Group],
         user_roles: List[str],
         group_roles: List[ACGroupRole],
     ):
@@ -219,7 +219,7 @@ async def get_group_context(
     return group
 
 
-async def require_group_membership(
+async def get_user_roles(
     group_id: Optional[int] = Depends(get_group_context),
     user: User = Depends(get_auth_user),
     session: AsyncSession = Depends(get_db),
@@ -239,12 +239,11 @@ async def require_group_membership(
     user_roles = rolewait.unique().scalars().all()
     return user_roles
 
-
 async def get_auth_context(
     group_id: Optional[int] = Depends(get_group_context),
     user: User = Depends(get_auth_user),
     session: AsyncSession = Depends(get_db),
-    user_roles: MutableSequence[UserRole] = Depends(require_group_membership),
+    user_roles: MutableSequence[UserRole] = Depends(get_user_roles),
 ) -> AuthContext:
     # TODO TODO TODO
     # we need to be able to generate an authcontext without any
@@ -252,24 +251,32 @@ async def get_auth_context(
     # roles, change my username, etc) so we need to handle that case
     group = None
     roles: List[str] = []
+    groles: List[ACGroupRole] = []
     for row in user_roles:
         roles.append(row.role.name)
         group = row.group
-    # If you don't have any roles in this group, go away
-    if not group:
-        raise ex.UnauthorizedException("not authorized")
-    query = (
-        sa.select(GroupRole)  # type: ignore
-        .options(  # type: ignore
-            joinedload(GroupRole.role, innerjoin=True),  # type: ignore
-            joinedload(GroupRole.grantor_group, innerjoin=True),  # type: ignore
+    if group:
+        query = (
+            sa.select(GroupRole)  # type: ignore
+            .options(  # type: ignore
+                joinedload(GroupRole.role, innerjoin=True),  # type: ignore
+                joinedload(GroupRole.grantor_group, innerjoin=True),  # type: ignore
+            )
+            .filter(GroupRole.grantee_group_id == group_id)  # type: ignore
         )
-        .filter(GroupRole.grantee_group_id == group_id)  # type: ignore
-    )
-    rolewait = await session.execute(query)
-    group_roles = rolewait.unique().scalars().all()
-    groles: List[ACGroupRole] = []
-    for row in group_roles:
-        groles.append({"group_id": row.grantor_group.id, "role": row.role.name})
+        rolewait = await session.execute(query)
+        group_roles = rolewait.unique().scalars().all()
+        for row in group_roles:
+            groles.append({"group_id": row.grantor_group.id, "role": row.role.name})
     ac = AuthContext(user, group, roles, groles)
     return ac
+
+
+async def require_group_membership(
+    ac: AuthContext = Depends(get_auth_context),
+):
+    # If you don't have any roles in this group, go away
+    if not ac.group:
+        print(f"HAHAHAHA NOPE")
+        raise ex.UnauthorizedException("Not authorized")
+    print(f"FOUND GROUP {ac.group} ACCESS ?!?!?!?!")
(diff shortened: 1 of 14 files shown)


src/backend/aspen/api/authn.py
```diff
@@ -196,7 +196,7 @@ class AuthContext:
     def __init__(
         self,
         user: User,
-        group: Group,
+        group: Optional[Group],
```
jgadling, line 199: Not all AuthContexts will have groups!

src/backend/aspen/api/authn.py
```diff
@@ -219,7 +219,7 @@ async def get_group_context(
     return group
 
 
-async def require_group_membership(
+async def get_user_roles(
```
jgadling, line 222: The name of this method was misleading

src/backend/aspen/api/authn.py
```diff
@@ -244,32 +244,44 @@ async def get_auth_context(
...
-    user_roles: MutableSequence[UserRole] = Depends(require_group_membership),
+    user_roles: MutableSequence[UserRole] = Depends(get_user_roles),
 ) -> AuthContext:
-    # TODO TODO TODO
-    # we need to be able to generate an authcontext without any
-    # org info *intentionally* for user self-management endpoints (get all my
-    # roles, change my username, etc) so we need to handle that case
     group = None
     roles: List[str] = []
+    groles: List[ACGroupRole] = []
```
jgadling, line 251: The changes to this method specifically allow creating authcontexts without a group context

src/backend/aspen/api/authn.py
```diff
@@ -244,32 +244,44 @@ async def get_auth_context(
...
+
+
+# We *specifically* want to allow authcontexts to be generated without
+# group context in several situations, however some endpoints *require*
+# a group role, so this is here to enforce that requirement.
+# This is a *stopgap* -- it's better to integrate Oso to do endpoint-level
+# policy enforcement, which will let us check membership and access more flexibly.
+async def require_group_membership(
+    ac: AuthContext = Depends(get_auth_context),
+):
```
jgadling, line 282: A dependency to *require* a group context where applicable

src/backend/aspen/api/authz.py
```diff
@@ -214,3 +216,37 @@ def require_group_privilege(
...
+            )
+            return res
+        except NoResultFound:
+            raise ex.UnauthorizedException(f"unauthorized")
+
+
+# NOTE - any endpoint that uses `fetch_authorized_row` must use "row_id" as the path
+#        parameter name to identify the row to be fetched.
+@lru_cache
+def fetch_authorized_row(
```
jgadling, line 248: A shortcut for endpoints that try to read/write/etc a particular item to just get the item they care about, or return an Unauthorized exception if it doesn't exist or they don't have the required permissions.

src/backend/aspen/api/main.py
```diff
@@ -107,32 +107,11 @@ def get_app() -> FastAPI:
...
-    _app.include_router(
-        phylo_trees.router,
-        prefix="/v2/phylo_trees",
-        dependencies=[Depends(get_auth_user)],
-    )
-    _app.include_router(
-        samples.router,
-        prefix="/v2/samples",
-        dependencies=[Depends(get_auth_user)],
-    )
```
jgadling, line 124: I refactored these into the loop below -- we're removing these routes once the FE is fully converted.

src/backend/aspen/api/main.py
```diff
@@ -142,15 +121,22 @@ def get_app() -> FastAPI:
...
         "sequences": sequences.router,
         "phylo_trees": phylo_trees.router,
         "phylo_runs": phylo_runs.router,
         "samples": samples.router,
     }
     for suffix, router in org_routers.items():
+        _app.include_router(
+            router,
+            prefix="/v2/" + suffix,
+            dependencies=[Depends(require_group_membership)],
```
jgadling, line 138: Make sure any access to these endpoints is limited to members of the group in question.

src/backend/aspen/api/policy.polar
```diff
@@ -76,6 +76,9 @@ resource PhyloTree {
   "write" if "member";
 }
 
+has_permission(ac: AuthContext, "invite_members", group: Group) if
+    has_role(ac, "admin", group) or ac.user.system_admin;
```
jgadling, line 80: Support letting system admins send invitations (we basically need this to invite the first user to a brand new group)

src/backend/aspen/auth/auth0_management.py
```diff
@@ -84,21 +84,9 @@ def get_org_user_roles(self, org_id: str, user_id: str) -> List[str]:
         res = self.client.organizations.all_organization_member_roles(org_id, user_id)
         return [item["name"] for item in res]
 
-    @cache
-    def get_org_by_name(self, org_name: str) -> Auth0Org:
```
jgadling, line 88: This method seemed kinda dangerous since we don't have much of a unique constraint on group names, and we're already storing auth0 group ID's in the db. I refactored all the callers to use `get_org_by_id` instead.

src/backend/aspen/auth/auth0_management.py
```diff
@@ -84,21 +84,9 @@ def get_org_user_roles(self, org_id: str, user_id: str) -> List[str]:
...
-        raise Exception("Organization not found")
-
     @cache
     def get_org_by_id(self, org_id: str) -> Auth0Org:
-        orgs = self.get_orgs()
-        for org in orgs:
-            if org["id"] == org_id:
-                return org
-        raise Exception("Organization not found")
+        return self.client.organizations.get_organization(org_id)
```
jgadling, line 89: heh, oops. there was a simpler way to fetch this info 😳

src/cli/aspencli.py
```diff
@@ -294,7 +294,7 @@ def usher():
 def get_link(ctx, sample_ids):
     api_client = ctx.obj["api_client"]
     payload = {"samples": sample_ids, "downstream_consumer": "USHER"}
-    resp = api_client.post("/v2/sequences/getfastaurl", json=payload)
+    resp = api_client.post_with_org("/v2/sequences/getfastaurl", json=payload)
```
jgadling, line 297: I updated all the relevant CLI commands to be able to fetch resources with org ID's in the url.

src/backend/aspen/api/authz.py
lvreynoso: What is a `row_id`? Is it literally the `id` column in the `groups` table?

lvreynoso (approved):
lgtm, just one question.

src/backend/aspen/api/authz.py
jgadling: if the model object passed in is a group, yes. It refers to the `id` column on whatever `model` class gets passed in. Thoughts on whether I can make this clearer?

jgadling:
> lgtm, just one question.

Yeah, it can be a little confusing:
`row_id` needs to be the path parameter for a given endpoint, for example:
```
@router.get("/{row_id}/", response_model=GroupInfoResponse)
```

Then that row_id gets picked up to AuthorizedRow and mapped to a DB query for an object with `modelclass.id = row_id`:
https://github.com/chanzuckerberg/czgenepi/blob/cdbd322440cb90411a03192f74865f36f4b4e4b3/src/backend/aspen/api/authz.py#L236

And since the model classes passed into `fetch_authorized_row` is `Group` that row_id corresponds to a group id:
https://github.com/chanzuckerberg/czgenepi/blob/cdbd322440cb90411a03192f74865f36f4b4e4b3/src/backend/aspen/api/views/groups.py#L80

PR #1227, opened 2022-07-15
sc189557 Use SDS icons

ehoops-cz:
### Summary:
- **What:** `Swapping svg icons for SDS icons`
- **Ticket:** [sc189557](https://app.shortcut.com/genepi/story/189557)
- **Env:** `<rdev link>` (I'm not sure what to put here)

Remove IconArrowDownSmall.svg
### Demos:
Before:
Screen Shot 2022-07-15 at 2 49 40 PM
Screen Shot 2022-07-15 at 2 49 36 PM
After:
Screen Shot 2022-07-15 at 2 53 04 PM
Screen Shot 2022-07-15 at 2 52 58 PM

### Checklist:
- [x] I added labels to my PR


mmmmmmaya (approved):

src/frontend/src/components/AlertAccordion/style.ts
```diff
@@ -1,34 +1,19 @@
...
   display: flex;
   height: 100%;
 
   .MuiCollapse-root {
     overflow-y: auto;
   }
+
+  svg {
+    fill: black;
+  }
```
ehoops-cz, line 16: I'm not sure why putting the fill into the `parent component > svg` works while putting the fill elsewhere doesn't seem to work.  I also tried putting it into `smallIcon` above.

Is this an acceptable way to do this?  It seems weird, but I'm not seeing another way.

mmmmmmaya:
You could also remove IconArrowUp svg, if it isn't being used anywhere else.

src/frontend/src/common/styles/iconStyle.ts
```diff
@@ -1,33 +1,117 @@
...
 `;
 
-export const StyledEditIcon = styled(EditIcon)`
-  fill: black;
+export const StyledTrashIconWrapper = styled.div`
+  ${rightMarginM}
+  ${iconFillError}
+`;
+
+export const StyledTreeIconWrapper = styled.div`
```
mmmmmmaya: It looks like this one is only used in one place. Is there a benefit to putting it here rather than in the component specific style file?

ehoops-cz: I liked the idea of keeping icon styles together since they use some of the same mixins.  I don't feel strongly either way.  It is a bit more modular to have icons styled in their respective components.  It's also more likely to have repetition.  Thoughts?

ehoops-cz: alternatively this file could include the mixins and then icon styling could be in whichever component.  There might be some duplication, but it wouldn't be too much.

src/frontend/src/components/AlertAccordion/style.ts
mmmmmmaya: This is the only way to do this atm. There is a small bug in sds.

src/frontend/src/common/styles/iconStyle.ts
mmmmmmaya: I like the idea of the mixins living in the shared area and then you can import them into whatever styling files need them!

src/frontend/src/components/AcknowledgePolicyChanges/index.tsx
```diff
@@ -53,7 +53,7 @@ const determineIfAcknowledgmentNeeded = (user: User | undefined): boolean => {
   return true;
 };
 
-const AcknowledgePolicyChanges = () => {
+const AcknowledgePolicyChanges = (): JSX.Element | null => {
```
mmmmmmaya, line 56: 🙌

mmmmmmaya (approved):

PR #1230, opened 2022-07-19
Fix group endpoint authz.

jgadling:
### Notes:
I broke the authz for the group membership endpoints when I switched to `row_id` -- this fixes the breakage and adds a test for multi-group-membership.

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/authz.py
+++ b/src/backend/aspen/api/authz.py
@@ -1,11 +1,13 @@
 from functools import lru_cache
 from pathlib import Path
+from typing import Optional
 
 from fastapi import Depends
 from oso import AsyncOso, Relation
 from polar.data.adapter.async_sqlalchemy2_adapter import AsyncSqlAlchemyAdapter
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm.exc import NoResultFound
+from starlette.requests import Request
 
 from aspen.api.authn import AuthContext, get_auth_context
 from aspen.api.deps import get_db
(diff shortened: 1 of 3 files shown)


src/backend/aspen/api/authz.py
```diff
@@ -219,21 +221,25 @@ def require_group_privilege(
...
     async def __call__(
         self,
-        row_id: int,
+        request: Request,
         auth_context: AuthContext = Depends(get_auth_context),
         session: AsyncSession = Depends(get_db),
     ) -> idbase:
         authz_session = AuthZSession(session, auth_context)
         query = await authz_session.authorized_query(self.privilege, self.model)
+        id_value = int(request.path_params[self.id_field])
```
lvreynoso, line 173: I should have caught this when reviewing, my bad.

src/backend/aspen/api/views/tests/test_groups.py
```diff
@@ -75,6 +77,70 @@ async def test_list_members(
         assert resp_data[key] == expected[key]
 
 
+async def test_list_members_secondary_group(
```
lvreynoso, line 80: 👍

lvreynoso (approved):
It looks like I broke a phylo run test that just happened to work when my PR's CI ran, I'll fix that.

src/backend/aspen/api/authz.py
jgadling: nah, I was intentionally trying to standardize everything to `row_id` at the time but just looks like I took it toooo far.

PR #1232, opened 2022-07-20
Make sure localdev users can still log in.

jgadling:
### Notes:
Now that the callback endpoint syncs auth0 user roles on login, any localdev users that are tied to our local OIDC service can't log in locally. This puts in an exception for users without 'auth0|' prefixes to not require a matching user in auth0.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/conftest.py
+++ b/src/backend/aspen/api/conftest.py
@@ -6,12 +6,13 @@ from typing import AsyncGenerator
 from unittest.mock import create_autospec, MagicMock
 
 import pytest
+from authlib.integrations.starlette_client import StarletteOAuth2App
 from fastapi import Depends, FastAPI, Request
 from httpx import AsyncClient
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from aspen.api.authn import get_auth0_apiclient, get_cookie_userid
-from aspen.api.deps import get_db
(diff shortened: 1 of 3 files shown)


src/backend/aspen/api/conftest.py
```diff
@@ -134,15 +135,22 @@ async def auth0_apiclient() -> MagicMock:
     return create_autospec(Auth0Client)
 
 
+@pytest.fixture()
+async def auth0_oauth() -> MagicMock:
```
jgadling, line 139: This lets us mock the auth0 oauth client in our test suite

src/backend/aspen/api/views/tests/test_auth.py
```diff
@@ -273,3 +274,58 @@ async def test_sync_complicated_roles(
             [(group.auth0_org_id, role) for group, role in final_roles]
         )
         await check_roles(async_session, user.auth0_user_id, expected_roles)
+
+
+async def test_callback_syncs_auth0_user_roles(
```
jgadling, line 279: Make sure we do try to sync roles if userid starts with auth0| and we don't if it doesn't.

phoenixAja (approved):
looks good!! thank you for addressing this so quickly!

PR #1235, opened 2022-07-21
Maya/group switch flake b gone

mmmmmmaya:
### Summary
- **What:**
  - be smarter about where groupId is really required for cache subscription (ie: remove the need to pass groupId in to custom hooks which calls apis that return different data depending which group you are viewing)
  - if user's current group ends up in a weird state, don't prevent the group dropdown from opening
  - pull user's current role and available groups from /me
  -  When user changes group, expire all user-related caches
- **Why:** 
  - easier to subscribe to fe cache if you dont have to give a group id, and the id was only used to generate the url anyway
  - Allows user to get out of wonky group state themselves
  - shape of api response is changing
  - update samples table, group info, etc for new group
- **Env:** https://maya-multigroup-frontend.dev.czgenepi.org/

### Demos
No visual changes, besides dropdown in navbar should work again.

### Notes
I did a bad thing and made a very large PR. Please review each of the commits individually with `?w=1` for easier review.

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/api/index.tsx
+++ b/src/frontend/src/common/api/index.tsx
@@ -11,6 +11,7 @@ export enum API {
   LOCATIONS = "/v2/locations/",
   PANGO_LINEAGES = "/v2/lineages/pango",
   GROUPS = "/v2/groups/",
+  ORGS = "/v2/orgs/",
 }
 
 export enum ORG_API {
@@ -213,7 +214,7 @@ export const fetchSamples = (): Promise<SampleResponse> =>
   apiResponse<SampleResponse>(
     ["samples"],
     [SAMPLE_MAP],
-    generateGroupSpecificUrl(ORG_API.SAMPLES)
+    generateOrgSpecificUrl(ORG_API.SAMPLES)
   );
 
 export interface PhyloRunResponse extends APIResponse {
@@ -231,5 +232,5 @@ export const fetchPhyloRuns = (): Promise<PhyloRunResponse> =>
   apiResponse<PhyloRunResponse>(
     ["phylo_runs"],
     [PHYLO_RUN_MAP],
-    generateGroupSpecificUrl(ORG_API.PHYLO_RUNS)
+    generateOrgSpecificUrl(ORG_API.PHYLO_RUNS)
   );
--- a/src/frontend/src/common/api/index.tsx
+++ b/src/frontend/src/common/api/index.tsx
@@ -1,5 +1,7 @@
 import ENV from "src/common/constants/ENV";
 import { jsonToType } from "src/common/utils";
+import { store } from "../redux";
+import { selectCurrentGroup } from "../redux/selectors";
 
 export enum API {
   USERDATA = "/v2/users/me",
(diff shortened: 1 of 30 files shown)


src/frontend/src/common/queries/groups.ts
```diff
@@ -250,3 +229,24 @@ export function useSendGroupInvitations({
...
+    USE_PHYLO_RUN_INFO,
+    USE_SAMPLE_INFO,
+    USE_GROUP_INFO,
+    USE_GROUP_INVITATION_INFO,
+    USE_GROUP_MEMBER_INFO,
+  ];
+
+  forEach(queriesToRefetch, async (q) => {
+    await queryClient.fetchQuery([q]);
+    await queryClient.invalidateQueries([q]);
```
phoenixAja: wouldn't you want to invalidate first before refetching?

src/frontend/src/common/api/index.tsx
```diff
@@ -20,11 +23,14 @@ export enum ORG_API {
   GET_FASTA_URL = "sequences/getfastaurl",
```
phoenixAja, line 23: should we add a trailing slash here as well to follow the pattern of the other paths?

src/frontend/src/common/queries/groups.ts
phoenixAja: also won't it automatically refetch if you invalidate the queries?

src/frontend/src/common/utils/userInfo.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Takes the stored id for the current group and returns the available details about that group
+ * from /me response
+ */
+export const getCurrentGroupFromUserInfo = (
+  userInfo?: User
+): UserGroup | undefined => {
+  const state = store.getState();
+  const currentGroupId = selectCurrentGroup(state);
+  return find(userInfo?.groups, (g) => g.id === currentGroupId);
+};
```
phoenixAja, line 15: do we need a fallback for if userInfo is undefined? Or do we not need to worry about that anymore? I remember that causing some strange side effects before

phoenixAja:
i think there might be something wrong, i tried logging into your rdev with contra costa and fresno and both showed having 0 samples and trees?

src/frontend/src/common/api/index.tsx
jgadling: Nope - this endpoint doesn't have a trailing slash because it's not a "list" endpoint

jgadling:
Clicking around the rdev, this looks awesome!! 😍

jgadling:
> i think there might be something wrong, i tried logging into your rdev with contra costa and fresno and both showed having 0 samples and trees?

Yep, you totally uncovered a problem. In this case it's a problem with the backend though. 😳 -- shouldn't impact this PR.

src/frontend/src/common/queries/groups.ts
mmmmmmaya: It does not always automatically refetch when you invalidate the queries. Invalidate refetches "smartly". React query has some internal logic to determine whether/when to refetch a query that has been invalidated. When you use refetch directly, the refetch is guaranteed. In this case, whatever smart thing react query was trying to do was preventing the refetch from happening, even though we wanted it to happen

mmmmmmaya: I do agree fetch should happen before invalidate. fixing...

src/frontend/src/common/utils/userInfo.ts
mmmmmmaya: It caused strange side effects in the past because we were using that -1 group id to call the api, which we aren't doing anymore. we shouldn't need a fallback here.

phoenixAja (approved):
looks good!! 🤩 thank you maya!

PR #1238, opened 2022-07-21
optionally accept nextstrain headers for upload [sc190569]

phoenixAja:
### Summary:
- **What:** accept nextstrain default metadata headers on upload
- **Ticket:** [sc190569](https://app.shortcut.com/genepi/story/190569)
- **Env:** no rdev link

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/views/Upload/components/Metadata/components/ImportFile/components/Instructions/index.tsx
+++ b/src/frontend/src/views/Upload/components/Metadata/components/ImportFile/components/Instructions/index.tsx
@@ -1,5 +1,6 @@
 import { List, ListItem } from "czifui";
 import React from "react";
+import { NewTabLink } from "src/common/components/library/NewTabLink";
 import {
   ReImportDataItem,
   StyleDownloadTemplate,
@@ -23,14 +24,21 @@ export default function Instructions({ headers, rows }: Props): JSX.Element {
           included in your import file.
         </ListItem>
         <ListItem ordered fontSize="xs">
-          We recommend that you copy your metadata into our{" "}
+          We recommend that you copy your metadata into our
           <StyleDownloadTemplate headers={headers} rows={rows}>
             TSV template
           </StyleDownloadTemplate>
           , but you can use your own TSV or CSV file as well.
         </ListItem>
         <ListItem ordered fontSize="xs">
-          Make sure your column headers match our naming convention.
+          Make sure your column headers match our naming convention or the&nbsp;
+          <NewTabLink
+            href={
+              "https://docs.nextstrain.org/projects/ncov/en/latest/reference/metadata-fields.html"
+            }
+          >
+            Nextstrain defaults.
+          </NewTabLink>
         </ListItem>
         <ListItem ordered fontSize="xs">
           Make sure your metadata values are in the correct format.
--- a/src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
+++ b/src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
@@ -17,6 +17,7 @@ import {
   FORBIDDEN_NAME_CHARACTERS_REGEX,
   HEADERS_TO_METADATA_KEYS,
   MAX_NAME_LENGTH,
+  NEXTSTRAIN_FORMAT_HEADERS_TO_METADATA_KEYS,
 } from "../../../common/constants";
 /**
  * (Vince) Regarding interfaces for Warnings/Errors:
@@ -63,7 +64,13 @@ interface ParsedRow {
 // Helper -- Takes column header from file, converts to internal metadata key
 // If header is unrecognized, leaves it alone (useful for warnings, etc).
 function convertHeaderToMetadataKey(headerName: string): string {
-  return HEADERS_TO_METADATA_KEYS[headerName] || headerName;
+  if (headerName in HEADERS_TO_METADATA_KEYS) {
+    return HEADERS_TO_METADATA_KEYS[headerName];
+  } else if (headerName in NEXTSTRAIN_FORMAT_HEADERS_TO_METADATA_KEYS) {
+    return NEXTSTRAIN_FORMAT_HEADERS_TO_METADATA_KEYS[headerName];
+  } else {
+    return headerName;
+  }
 }
 
 // Helper -- Returns any missing col header field names based on what header
--- a/src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
+++ b/src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
@@ -63,6 +63,7 @@ interface ParsedRow {
 
 // Helper -- Takes column header from file, converts to internal metadata key
 // If header is unrecognized, leaves it alone (useful for warnings, etc).
+// User can also use Nextstrain header defaults as an alias
 function convertHeaderToMetadataKey(headerName: string): string {
   if (headerName in HEADERS_TO_METADATA_KEYS) {
     return HEADERS_TO_METADATA_KEYS[headerName];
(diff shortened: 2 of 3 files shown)


src/frontend/src/views/Upload/components/common/constants.ts
```diff
@@ -10,6 +10,16 @@ export const HEADERS_TO_METADATA_KEYS = invert(
...
 ) as Dictionary<keyof SampleUploadTsvMetadata>;
 
+export const NEXTSTRAIN_FORMAT_HEADERS_TO_METADATA_KEYS: Record<
+  string,
+  string
+> = {
+  date: "collectionDate",
+  location: "collectionLocation",
+  gisaid_epi_isl: "publicId",
+  strain: "sampleId",
```
mmmmmmaya, line 20: I'm surprised this isn't lineage or something. What's going on here?

src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
```diff
@@ -63,7 +64,13 @@ interface ParsedRow {
 // Helper -- Takes column header from file, converts to internal metadata key
 // If header is unrecognized, leaves it alone (useful for warnings, etc).
 function convertHeaderToMetadataKey(headerName: string): string {
-  return HEADERS_TO_METADATA_KEYS[headerName] || headerName;
+  if (headerName in HEADERS_TO_METADATA_KEYS) {
+    return HEADERS_TO_METADATA_KEYS[headerName];
+  } else if (headerName in NEXTSTRAIN_FORMAT_HEADERS_TO_METADATA_KEYS) {
+    return NEXTSTRAIN_FORMAT_HEADERS_TO_METADATA_KEYS[headerName];
```
mmmmmmaya, line 71: What is the use for this second dictionary? What are the instances where the headers differ?

phoenixAja: sorry i didn't do a good job at explaining, so this was to handle the case for if the users wanted to upload metadata with different headers (that are default for nextstrain), we're handling this for the users so that they don't need to convert the headers to match to our system, and can just upload directly.

src/frontend/src/views/Upload/components/common/constants.ts
phoenixAja: I was just following the ticket description, wondering if @sidneymbell can double check this?

src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
phoenixAja: does that answer your question? lol

src/frontend/src/views/Upload/components/common/constants.ts
ehoops-cz: It sounds like this column is called `sampleId` in our template, but it is called `strain` in the Nextstrain metadata, so we're mapping both to `"sampleId"`?

phoenixAja: yep!

src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
mmmmmmaya: How will this process generalize in multi pathogen? For example, will there also be nextstrain specific templates for tb, etc?

src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
```diff
@@ -63,7 +64,13 @@ interface ParsedRow {
 // Helper -- Takes column header from file, converts to internal metadata key
 // If header is unrecognized, leaves it alone (useful for warnings, etc).
```
vincent-czi, line 65: I think it would be good to update these docs/comments to explain that users can provide headers as we give them in template from CZ GE, or they can also use the ones from Nextstrain as an alias.

src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
sidneymbell: @mmmmmmaya -- Nextstrain template should be pretty universal across pathogens, at least for the fields we care about. Exceptions:   
* `gisaid_epi_isl` is only applicable to flu and sars-cov-2; might want to consider mapping [`genbank accession`](https://docs.nextstrain.org/projects/ncov/en/latest/reference/metadata-fields.html#column-4-genbank-accession) to `public_id` if `gisaid_epi_isl` is either missing altogether or is empty. I think it's OK to file this as a "de-covidifying" ticket for later.
* Some other fields may change, but not the ones we care about -- those should be consistent across pathogens (i.e., `location`, `date`, `strain`, `gisaid_epi_isl`, `genbank_accession`).

sidneymbell:
@phoenixAja -- if the user has uploaded a file that contains all the required fields, plus a bunch of other ones we are going to discard, do we gracefully ignore all the other fields?

phoenixAja:
@sidneymbell yes we gracefully ignore all other headers provided (as in no errors appear and the form gets populated as expected)

src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
phoenixAja: ok i will update the comment and am also updating the helper instructions text

src/frontend/src/views/Upload/components/Metadata/components/ImportFile/components/Instructions/index.tsx
```diff
@@ -23,14 +24,21 @@ export default function Instructions({ headers, rows }: Props): JSX.Element {
           included in your import file.
         </ListItem>
         <ListItem ordered fontSize="xs">
-          We recommend that you copy your metadata into our{" "}
+          We recommend that you copy your metadata into our
```
phoenixAja, line 27: `{" "}` wasn't being respected so i took it out, `StyledDownloadTemplate` Margin on the left was creating a sort of faux space

src/frontend/src/views/Upload/components/Metadata/components/ImportFile/components/Instructions/index.tsx
```diff
@@ -23,14 +24,21 @@ export default function Instructions({ headers, rows }: Props): JSX.Element {
...
           <StyleDownloadTemplate headers={headers} rows={rows}>
             TSV template
           </StyleDownloadTemplate>
           , but you can use your own TSV or CSV file as well.
         </ListItem>
         <ListItem ordered fontSize="xs">
-          Make sure your column headers match our naming convention.
+          Make sure your column headers match our naming convention or the&nbsp;
```
phoenixAja, line 34: `{" "}` wasn't being respected so i replaced with `&nbsp;` wondering if either @mmmmmmaya or @ehoops-cz knows why and can enlighten me on the reasoning 😆

ehoops-cz: Apparently it's part of the css spec to trim trailing white space when `white-space` is set to normal (the default).  I honestly had no idea there was so much to read about this 😂 .  I usually use `&nbsp` as well.  I think another option may be wrapping the space in a span (iirc) like 
```
<ListItem ordered fontSize="xs">
          Make sure your column headers match our naming convention or the
          <span>{' '}</span>
          <NewTabLink ...
</ListItem>
```

phoenixAja: interesting! thanks Erin!

mmmmmmaya (approved):

PR #1240, opened 2022-07-21
Support syncing roles to/from auth0

jgadling:
### Summary:
This updates the syncer script to be able to sync organization roles from the db to auth0, or from auth0 to the db.

We'll need to run it in rdev and staging and prod to sync the db roles (currently the source of truth) to auth0 before enabling the "sync on login" feature flag.

All of this needs to happen before we can properly support users joining multiple groups.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/auth/auth0_management.py
+++ b/src/backend/aspen/auth/auth0_management.py
@@ -143,6 +143,18 @@ class Auth0Client:
             org["id"], user_id, {"roles": [member_role["id"]]}
         )
 
+    def add_org_roles(self, org: Auth0Org, user_id: str, roles: List[str]) -> None:
+        roleids = [self.get_auth0_role(role)["id"] for role in roles]
+        self.client.organizations.create_organization_member_roles(
+            org["id"], user_id, {"roles": roleids}
+        )
+
+    def remove_org_roles(self, org: Auth0Org, user_id: str, roles: List[str]) -> None:
+        roleids = [self.get_auth0_role(role)["id"] for role in roles]
+        self.client.organizations.delete_organization_member_roles(
+            org["id"], user_id, {"roles": roleids}
+        )
+
     def remove_org_member(self, org: Auth0Org, user_id: str) -> None:
         self.client.organizations.delete_organization_members(
             org["id"], {"members": [user_id]}
--- a/src/backend/aspen/auth/auth0_management.py
+++ b/src/backend/aspen/auth/auth0_management.py
@@ -144,15 +144,15 @@ class Auth0Client:
         )
 
     def add_org_roles(self, org: Auth0Org, user_id: str, roles: List[str]) -> None:
-        roleids = [self.get_auth0_role(role)["id"] for role in roles]
+        role_ids = [self.get_auth0_role(role)["id"] for role in roles]
         self.client.organizations.create_organization_member_roles(
-            org["id"], user_id, {"roles": roleids}
+            org["id"], user_id, {"roles": role_ids}
         )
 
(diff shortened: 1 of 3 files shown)


src/backend/aspen/database/models/usergroup.py
```diff
@@ -103,7 +103,7 @@ class UserRole(base):  # type: ignore
     """User role grants"""
 
     __tablename__ = "user_roles"
-    role_id = Column(Integer, ForeignKey(Role.id), nullable=False)
+    role_id = Column(Integer, ForeignKey(Role.id), nullable=False, primary_key=True)
```
lvreynoso, line 106: Primary key for this table is a `btree (user_id, group_id)`, is there a missing migration to this PR?

lvreynoso (changes requested):
just a small issue.

src/backend/aspen/auth/auth0_management.py
```diff
@@ -143,6 +143,18 @@ def add_org_member(self, org: Auth0Org, user_id: str) -> None:
             org["id"], user_id, {"roles": [member_role["id"]]}
         )
 
+    def add_org_roles(self, org: Auth0Org, user_id: str, roles: List[str]) -> None:
+        roleids = [self.get_auth0_role(role)["id"] for role in roles]
```
phoenixAja: this is the nittiest nit, but can this variable be named `role_ids`? for some reason `roleids` is upsetting to me

src/backend/aspen/database/models/usergroup.py
jgadling: I checked every environment, and this is what I see:
```  
"user_roles_pkey" PRIMARY KEY, btree (role_id, user_id, group_id)
```

And also the migration file that creates the user_roles table is what established this primary key.

src/backend/aspen/cli/sync_auth0.py
```diff
@@ -17,7 +18,98 @@
...
+        self.dry_run = dry_run
+        self.groupmap: Dict[str, Group] = {}
+        self.rolemap: Dict[str, Role] = {}
+
+    def load_context(self):
+        roles = self.db.execute(sa.select(Role)).scalars().all()
+        self.rolemap = {role.name: role for role in roles}
+
+        groups = self.db.execute(sa.select(Group)).scalars().all()
+        self.groupmap = {group.auth0_org_id: group for group in groups}
```
phoenixAja, line 38: why not just call `load_context` in `__init__`?

jgadling: Doing work in constructors (especially interacting with external systems, like a db in this case) can make testing more difficult, so I usually try to avoid it.

src/backend/aspen/database/models/usergroup.py
phoenixAja: i'm confused, how can the primary key be `btree (role_id, user_id, group_id)`, but then you set the primary_key as just `role_id` above? Is this some sqlalchemy wierdness?

jgadling: There are 3 properties/columns in this model/table with the primary key attribute set to true now, where there were only two before:
https://github.com/chanzuckerberg/czgenepi/blob/867ee81e49da4f15695c6a0e4c7ec9b041900d1b/src/backend/aspen/database/models/usergroup.py#L102-L111

lvreynoso: I see, my local machine is off for some reason 🥴

lvreynoso (approved):

PR #1244, opened 2022-07-26
Redirect to nice error pages on invitation failure.

jgadling:
### Summary:
- **What:** redirect to pretty error pages on invitation code expiration

### Notes:
Unfortunately auth0 sends the same error message to our backend if an invitation has already been used, never existed, or has expired 😭 We may want to update the text on the "already used" error page to reflect the fact that we have no way to differentiate what happened. I think we can basically redirect to a "something went wrong with your invitation" page and a general "something went wrong and we don't know what" page.

### Checklist:
- [ ] I added labels to my PR


mmmmmmaya (approved):

Janeece:
Design and Prod should look at writing a generic message if we can't specify why it failed. The suggestions above work, but I wonder if we can provide some small troubleshooting to the user.

PR #1246, opened 2022-07-26
Send slack updates on prod success/failure.

jgadling:
### Notes:
1. Re-enable Tues morning deploys
2. Our slack notifications for prod deploy success/failure were being sent from the wrong action, so we were getting a success message like 1min after the "starting" message, even if the deploy ultimately failed. I put the slack notifications in the action that *actually* does the deploy so we should get a success notification at the end of a deploy
3. We probably want to get failure notifications for staging deploys in our ops channel as well, so I tried to update the notification to message us on any failures instead of just prod.

I'm also not a GHA syntax expert 😬 

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/.github/workflows/deploy-happy-stack.yml
+++ b/.github/workflows/deploy-happy-stack.yml
@@ -73,7 +73,7 @@ jobs:
         with:
           status: ${{ job.status }}
           steps: ${{ toJson(steps) }}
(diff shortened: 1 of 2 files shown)


lvreynoso (approved):
lgtm

PR #1249, opened 2022-07-28
Update pangolin version comparison

lvreynoso:
### Summary:
- **What:** Sources current pangolin version from updated program running on the container, not from github. Compares pangolin version with version strings stored in database while keeping PangoLEARN / PUSHER information.
- **Ticket:** [sc191868](https://app.shortcut.com/genepi/story/191868)
- **Env:** [https://pangov4-frontend.dev.czgenepi.org](https://pangov4-frontend.dev.czgenepi.org)

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/workflows/pangolin/find_samples.py
+++ b/src/backend/aspen/workflows/pangolin/find_samples.py
@@ -1,6 +1,7 @@
+import re
 import subprocess
-import urllib.request
-from typing import Iterable
+from re import Match
+from typing import Iterable, List, Optional
 
 import click
 from sqlalchemy.orm import joinedload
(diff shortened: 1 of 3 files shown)


src/backend/aspen/workflows/pangolin/find_samples.py
```diff
@@ -1,6 +1,7 @@
+import re
```
jgadling, line 1: Mind renaming this file to something that better reflects what it does?

src/backend/aspen/workflows/pangolin/run_mass_update.sh
```diff
@@ -0,0 +1,15 @@
...
+# TODO: fix pipefail flags to be informative
+#set -Eex pipefail
+
+# activate miniconda
+eval "$($HOME/miniconda/bin/conda shell.bash hook)"
+conda init
+conda activate pangolin
+
+# record pangolin version:
+PANGOLIN_VERSION=$(pangolin -pv)
```
jgadling, line 12: Any idea what's up with the scorpio versions? I think useher/pangolearn versions match pangolin versions, but this is odd:

```
SCORPIO_v0.1.10
```

src/backend/aspen/workflows/pangolin/find_samples.py
```diff
@@ -14,20 +15,23 @@
 )
 from aspen.database.models import Sample
 
+VERSION_REGEX = re.compile(r"[0-9.]+$")
 
-def check_latest_pangolin_version() -> str:
-    contents = urllib.request.urlopen(
-        "https://github.com/cov-lineages/pangoLEARN/releases/latest"
+
+def same_pangolin_version(sample_version_string: str, pangolin_version_string: str):
```
jgadling, line 21: tests?

src/backend/aspen/workflows/pangolin/run_mass_update.sh
lvreynoso: No, I was not aware this was a possible value until now. That doesn't match the pangolin version so to support this we would have to extract a version string somewhere :/ @danrlu do you know what this is?

lvreynoso:
Back to the drawing board.

lvreynoso:
I think going forward the best way to deal with this would be to separately store the pangolin software version from the model version string spit out in the csv file of lineage calls. Doing this right now would mean a db migration, but since we want to get rid of pangolin-specific columns anyways with our decovidify work I think it would be best to come back to this after that work is done, since we can just add another key in the data stored in the future jsonb column that will be replacing pangolin-specific db columns.

danrlu:
If I understand correctly, the PR as it stands will skip all samples that has `PANGO-v.1.12` and `PUSHER-v.1.12` and rerun those with `SCORPIO_v0.1.10`, right? This still solves 99.9% of our problem and is totally sufficient for now. Can we just merge this in and mark it as done until future problem arises?

PR #1256, opened 2022-08-01
User needs longer max lengths

vincent-czi:
### Summary:
- **What:** Increase max length limits for certain fields pertaining to groups
- **Ticket:** [sc<209498>](https://app.shortcut.com/genepi/story/<209498>)
- **Env:** None

### Notes:
One of our new user groups needed longer field lengths than we currently allow for some. Since these were mostly arbitrary, bumping them up to 1000 each so we don't encounter this again.

The important exception is `prefix`, since that string gets put into SFN, which has a max length of 80 overall. We can't really go much larger than we currently have it -- 32 would be the absolute max before stuff would definitely begin breaking -- so bumping it to 25 in the meantime until we come up with some sort of shortening/truncation strategy. The referenced ticket above,  [sc<209498>](https://app.shortcut.com/genepi/story/<209498>), will be used for tracking that.

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/schemas/usergroup.py
+++ b/src/backend/aspen/api/schemas/usergroup.py
@@ -8,11 +8,14 @@ from aspen.api.schemas.locations import LocationResponse
 
 
 class GroupCreationRequest(BaseRequest):
-    name: constr(min_length=3, max_length=128, strict=True)  # type: ignore
-    prefix: constr(min_length=2, max_length=20, strict=True)  # type: ignore
-    address: Optional[constr(min_length=1, max_length=128, strict=True)]  # type: ignore
-    division: Optional[constr(min_length=1, max_length=128, strict=True)]  # type: ignore
-    location: Optional[constr(min_length=1, max_length=128, strict=True)]  # type: ignore
+    name: constr(min_length=3, max_length=1000, strict=True)  # type: ignore
+    # group prefix currently is used in the SFN name, which has max char limit
+    # `prefix` cannot be arbitrarily increased until this ticket is resolved:
+    # https://app.shortcut.com/genepi/story/209498
+    prefix: constr(min_length=2, max_length=25, strict=True)  # type: ignore
+    address: Optional[constr(min_length=1, max_length=1000, strict=True)]  # type: ignore
+    division: Optional[constr(min_length=1, max_length=1000, strict=True)]  # type: ignore
+    location: Optional[constr(min_length=1, max_length=1000, strict=True)]  # type: ignore
     default_tree_location_id: int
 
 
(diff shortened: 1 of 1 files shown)


src/backend/aspen/api/schemas/usergroup.py
```diff
@@ -8,11 +8,14 @@
 
 
 class GroupCreationRequest(BaseRequest):
-    name: constr(min_length=3, max_length=128, strict=True)  # type: ignore
-    prefix: constr(min_length=2, max_length=20, strict=True)  # type: ignore
-    address: Optional[constr(min_length=1, max_length=128, strict=True)]  # type: ignore
-    division: Optional[constr(min_length=1, max_length=128, strict=True)]  # type: ignore
-    location: Optional[constr(min_length=1, max_length=128, strict=True)]  # type: ignore
+    name: constr(min_length=3, max_length=1000, strict=True)  # type: ignore
```
jgadling: I get address/location/etc being long, but are we sure we need to make group *name* so long? This name is displayed in the navbar for example and I don't think we've had any cases where groups pushed the 128 char limit, let alone a 200 char limit.

lvreynoso: This has to be smaller, since we use `group.name` as part of an s3 object key for sequence downloads. s3 object key names have a limit of 1024-bytes UTF-8 encoded, which is roughly the same number of characters since we mostly use ASCII chars. So an actual 1000 char name will break that endpoint. I think 512 or even 256 is much more reasonable here.

lvreynoso (changes requested):
The name constraint has to be smaller. ~~Also you can get some more characters for a group prefix by changing how we name SFN executions, if you're up to it - I think we can leave the datetime-stamp out of it since the executions log that.~~ Never mind, I think it gets used in too many places for a quick fix to deal with.

src/backend/aspen/api/schemas/usergroup.py
lvreynoso: See https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html

vincent-czi: Ah, thanks for pointing that out. I didn't realize it got used in S3 keys.

vincent-czi: UI starts to get funky even before hitting 128, but I see your point. Considering Lucia's note about how it's also used in S3, I think we should just leave the `name` alone entirely and if we get a group with a crazy long name in the future we can address it then.

image

lvreynoso (approved):

PR #1260, opened 2022-08-02
SC-194965 Add callout if download will use multiple files.

ehoops-cz:
### Summary:
- **What:** `If the user selects more than 999 samples for gisaid or genbank download, then let them know that the download will be chopped into multiple files.`
- **Ticket:** [sc194965](https://app.shortcut.com/genepi/story/194965/as-a-user-if-i-select-more-than-999-samples-i-want-to-be-notified-that-my-template-files-will-be-chopped-up)
- **Env:** `none`

### Demos:
Screen Shot 2022-08-02 at 2 49 18 PM

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
@@ -29,6 +29,7 @@ import {
   DownloadTypeInfo,
   Header,
   StyledButton,
+  StyledCallout,
   StyledCheckbox,
   StyledFileTypeItem,
   Title,
@@ -288,6 +289,14 @@ const DownloadModal = ({
                     still be included in your Sample Metadata download.
                   </DownloadTypeInfo>
                 </Alert>
+              )}
+            {checkedSampleIds.length - failedSampleIds.length > 999 &&
+              (isGisaidSelected || isGenbankSelected) && (
+                <StyledCallout intent="info" autoDismiss={false}>
+                  Your submission template download will be generated in
+                  multiple files of up to 999 samples in order to comply with
+                  GISAID/GenBank upload limits.
+                </StyledCallout>
               )}
             {getDownloadButton()}
           </Content>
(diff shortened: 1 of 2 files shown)


src/frontend/src/common/components/library/data_subview/components/DownloadModal/style.ts
```diff
@@ -111,6 +118,7 @@ export const DownloadType = styled.div`
 `;
 
 export const StyledButton = styled(Button)`
+  max-width: fit-content;
```
ehoops-cz, line 121: Changing to flex made the button full-width, so this puts it back to the correct size.

src/frontend/src/common/components/library/data_subview/components/DownloadModal/style.ts
```diff
@@ -133,3 +141,16 @@ export const StyledCheckbox = styled(Checkbox)`
     }
   }
 `;
+
+export const StyledCallout = styled(Callout)`
+  ${fontBodyXxs}
+  width: 100%;
+  max-width: 434px;
```
ehoops-cz, line 148: Callout has a width set to 360 px, so I added the width: 100% to counteract that.  Without this max-width, width: 100% was making the whole modal bigger.  Open to alternatives here.

src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
```diff
@@ -289,6 +290,14 @@ const DownloadModal = ({
                   </DownloadTypeInfo>
                 </Alert>
               )}
+            {checkedSampleIds.length - failedSampleIds.length > 999 &&
```
ehoops-cz, line 293: I removed failed samples from this count since we wouldn't be putting them into the file anyway.  Is this correct?

vincent-czi: That seems reasonable to me? If you verified that failed samples don't get put into the download -- I think you're right, just haven't looked myself -- than this seems good to me.

vincent-czi (approved):
Looks good!

src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
ehoops-cz: @phoenixAja I'm going to go ahead and merge this, but lmk if it needs alteration.

PR #1261, opened 2022-08-03
Add submitting_lab column to Group

lvreynoso:
### Summary:
- **What:** Adds a column to the groups table, `submitting_lab`, which is a nullable string. Populates this field with the value of the `name` column in the migration. In the group creation API, defaults to using the `name` value for this field. Returns this field as part of the group info.
- **Ticket:** [sc194989](https://app.shortcut.com/genepi/story/194989), [sc194991](https://app.shortcut.com/genepi/story/194991)
- **Env:** [https://submit-lab-frontend.dev.czgenepi.org](https://submit-lab-frontend.dev.czgenepi.org)

### Demos:
Screen Shot 2022-08-03 at 15 51 02

Screen Shot 2022-08-03 at 16 17 44

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/views/groups.py
+++ b/src/backend/aspen/api/views/groups.py
@@ -43,6 +43,8 @@ async def create_group(
     )
     organization = auth0_client.add_org(auth0_safe_prefix, group_creation_request.name)
     group_values = dict(group_creation_request) | {"auth0_org_id": organization["id"]}
+    if group_values.get("submitting_lab") is None:
+        group_values["submitting_lab"] = group_values["name"]
     group = Group(**group_values)
     db.add(group)
     await db.commit()
--- a/src/backend/aspen/api/views/groups.py
+++ b/src/backend/aspen/api/views/groups.py
@@ -43,8 +43,6 @@ async def create_group(
     )
     organization = auth0_client.add_org(auth0_safe_prefix, group_creation_request.name)
     group_values = dict(group_creation_request) | {"auth0_org_id": organization["id"]}
-    if group_values.get("submitting_lab") is None:
-        group_values["submitting_lab"] = group_values["name"]
     group = Group(**group_values)
     db.add(group)
     await db.commit()
--- a/src/backend/database_migrations/versions/20220803_224221_add_submitting_lab_to_groups.py
+++ b/src/backend/database_migrations/versions/20220803_224221_add_submitting_lab_to_groups.py
@@ -1,0 +1,32 @@
+"""add submitting lab to groups
+
+Create Date: 2022-08-03 22:42:27.655366
+
+"""
+import enumtables  # noqa: F401
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "20220803_224221"
+down_revision = "20220621_232148"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        "groups",
+        sa.Column("submitting_lab", sa.String(), nullable=True),
+        schema="aspen",
+    )
+
+    conn = op.get_bind()
+    populate_submitting_lab_sql = sa.sql.text(
+        "UPDATE aspen.groups SET submitting_lab = name"
+    )
+    conn.execute(populate_submitting_lab_sql)
+
+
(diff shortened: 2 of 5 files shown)


src/backend/aspen/api/views/groups.py
```diff
@@ -43,6 +43,8 @@ async def create_group(
     )
     organization = auth0_client.add_org(auth0_safe_prefix, group_creation_request.name)
     group_values = dict(group_creation_request) | {"auth0_org_id": organization["id"]}
+    if group_values.get("submitting_lab") is None:
+        group_values["submitting_lab"] = group_values["name"]
```
phoenixAja: i was thinking that we would have the group name be used as a default in the model field directly, something like [this from the sqlalchemy docs](https://docs.sqlalchemy.org/en/14/core/defaults.html#context-sensitive-default-functions):

```
def mydefault(context):
    return context.get_current_parameters()['counter'] + 12

t = Table('mytable', metadata_obj,
    Column('counter', Integer),
    Column('counter_plus_twelve', Integer, default=mydefault, onupdate=mydefault)
)
```

src/backend/database_migrations/versions/20220803_224221_add_submitting_lab_to_groups.py
```diff
@@ -0,0 +1,32 @@
+"""add submitting lab to groups
+
+Create Date: 2022-08-03 22:42:27.655366
+
+"""
+import enumtables  # noqa: F401
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
```
phoenixAja, line 10:
```suggestion
```

src/backend/database_migrations/versions/20220803_224221_add_submitting_lab_to_groups.py
```diff
@@ -0,0 +1,32 @@
...
+
+    conn = op.get_bind()
+    populate_submitting_lab_sql = sa.sql.text(
+        "UPDATE aspen.groups SET submitting_lab = name"
+    )
+    conn.execute(populate_submitting_lab_sql)
+
+
+def downgrade():
+    pass
```
phoenixAja: can we do this instead of passing? Otherwise it's not clear that the back migration didn't run, also i think it's fairly easy to add the downgrade migration? You just drop the `submitting_lab` column? 
`op.drop_column("groups", "submitting_lab", schema="aspen")`
 
```suggestion
raise NotImplementedError("Reversing this migration is not supported")
```

src/backend/aspen/api/views/groups.py
lvreynoso: nice find!

src/backend/database_migrations/versions/20220803_224221_add_submitting_lab_to_groups.py
lvreynoso: What's the issue here? This is in all our migrations.

src/backend/database_migrations/versions/20220803_224221_add_submitting_lab_to_groups.py
lvreynoso: It is easy to add a downgrade, but the consensus we have been working with is to never run downgrades 🤷 @jgadling perhaps you can speak more to this

src/backend/database_migrations/versions/20220803_224221_add_submitting_lab_to_groups.py
phoenixAja: i just thought it was redundant/ not useful. If you like it leave it in though!

src/backend/database_migrations/versions/20220803_224221_add_submitting_lab_to_groups.py
phoenixAja: i mean that makes sense.. can we still raise the NotImplementedError though? 😄

lvreynoso: Yes, I added it

phoenixAja (approved):
looks good!

PR #1262, opened 2022-08-04
SC-194969 account page

ehoops-cz:
### Summary:
- **What:** `Create page where users can view and update their GIDSAID ID`
- **Ticket:** [sc194969](https://app.shortcut.com/genepi/story/194969/as-a-user-i-want-to-be-able-to-access-my-account-details-and-edit-my-gisaid-user-id)
- **Env:** [account-page](https://account-page-frontend.dev.czgenepi.org/account)

### Demos:
Screen Shot 2022-08-04 at 9 29 11 AM

### Notes:
TODO before merge: 
* Update href for Learn More link

### Checklist:
- [x] I added labels to my PR


src/frontend/src/views/Account/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+  StyledSection,
+  SubText,
+  WhiteIconWrapper,
+} from "./style";
+
+export default function Account(): JSX.Element {
+  const router = useRouter();
+  // TODO:194969 - get user's current gisaid id for the default value
+  const [gisaidId, setGisaidId] = useState("");
+  const [hasUnsavedChanges, setUnsavedChanges] = useState<boolean>(false);
```
mmmmmmaya:
```suggestion
  const [hasUnsavedChanges, setHasUnsavedChanges] = useState<boolean>(false);
```

we like to use a convention where the set is essentially identically named as get.

src/frontend/src/views/Account/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  const [isModalOpen, setModalOpen] = useState<boolean>(false);
```
mmmmmmaya:
```suggestion
  const [isModalOpen, setIsModalOpen] = useState<boolean>(false);
```

same here

src/frontend/src/views/Account/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+    };
+    const handleBrowseAway = () => {
+      if (!hasUnsavedChanges) return;
+      openModal();
+      router.events.emit("routeChangeError");
+      throw "routeChange aborted.";
+    };
+    window.addEventListener("beforeunload", handleWindowClose);
+    router.events.on("routeChangeStart", handleBrowseAway);
+    return () => {
```
mmmmmmaya, line 51: 👏  love this

src/frontend/src/views/Account/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+          }
+        >
+          {hasUnsavedChanges ? "Save" : "Saved"}
+        </Button>
+      </StyledHeaderRow>
+      <StyledDivider />
+      <StyledSection>
+        <H2>Details</H2>
+        <StyledRow>
+          <StyledH3>GISAID Submitter ID &nbsp;</StyledH3>
```
mmmmmmaya: whoa, is the nbsp really needed here?

mmmmmmaya: also, is nbsp better than {" "} in some way? that's the convention we've been using so far.

src/frontend/src/views/Account/index.tsx
ehoops-cz: ah! I do too, just saw some that weren't so I was trying to match.  Happy to change it.

src/frontend/src/views/Account/index.tsx
ehoops-cz: Unfortunately, I haven't been able to figure out how to get the modal to do the action once the user decides to leave.  For route changes, I could pass in the route and have the modal redirect them.  However, for closing the page, this doesn't seem to work. `window.close()` apparently only works if we opened the window in the first place. Do you know if it's possible to close the page for them?

src/frontend/src/views/Account/index.tsx
ehoops-cz: For some reason {" "} isn't actually adding the space.  I added a right-margin to the H3 and removed the `&nbsp;`

src/frontend/src/views/Account/index.tsx
mmmmmmaya: Yea, there are some legacy ones that dont match and we never updated them yet. feel free to change them whenever you see them!

src/frontend/src/views/Account/index.tsx
mmmmmmaya: Do you mean in handleWindowClose? If you remove the e.preventDefault, I think the default event handler also runs after your custom code _if_ no other handlers stop propagation, prevent default action, or otherwise cancel it. Does removing that line generate the desired effect?

ehoops-cz: It got weird with mounting a new component.  I ended up not worrying about it per [this comment](https://czi-sci.slack.com/archives/C01HYJYA50W/p1659719347193429?thread_ts=1659715202.644959&cid=C01HYJYA50W) from Janeece.

ehoops-cz:
The page is not behind the feature flag, but it is currently unaccessible from the rest of the application.  Someone would need to know the URL.  Is this a problem?

mmmmmmaya (approved):

mmmmmmaya:
I think it's ok to not put it behind the feature flag 🚀

PR #1263, opened 2022-08-05
Make sample_collector_contact_address nullable

lvreynoso:
### Summary:
- **What:** Fixes a bug where sample uploads fail for groups without an address. The field `sample_collector_contact_address` in the `samples` table was non-nullable, but the inserted value was sourced from `groups.address`, a nullable field. 
- **Env:** `<rdev link>`

### Checklist:
- [x] I added labels to my PR


phoenixAja (approved):
looks good

sidneymbell:
I believe this was initially non-nullable because we needed this information to submit to GISAID on their behalf. I guess now that we're just giving them the template it's probably fine for it to be nullable, although it'll save them the headache of filling it in themselves when they go to submit if they provide it to us. Thanks Lucia.

PR #1264, opened 2022-08-08
Analytics: Various events

vincent-czi:
### Summary:
- **What:** Adds in majority of analytics events for initial version of analytics
- **Ticket:** [sc<199706>](https://app.shortcut.com/genepi/story/<199706>)
- **Env:** None. Have tested locally. Planning to do an rdev once analytics is basically wrapped and ready for trunk merge.

### Demos:
All the events added have been checked with Segment debugger, payloads are landing as expected for each. This is just a screenshot from checking one of the event types, not all of them.

image

### Notes:
When reviewing, I recommend reading through the additions to `eventTypes.ts` first, as the documentation for the various new events will give you context for all the code changes to support and send those events.

There is a lot of documentation here -- probably too much -- and some annoyingly fancy TS stuff to try to bake in a certain pattern for handling analytics events in code. This was the result of a lot of thinking about making events that are easy to understand for data analysts and hard to break with code changes elsewhere. The documentation on `analyticsTrackEvent` goes into detail.

### Checklist:
- [x] I merged latest `vince/analytics`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/analytics/eventTypes.ts
+++ b/src/frontend/src/common/analytics/eventTypes.ts
@@ -18,6 +18,9 @@
 export enum EVENT_TYPES {
   // User has been sent to Nextstrain to view a phylo tree
   TREE_VIEW_NEXTSTRAIN = "TREE_VIEW_NEXTSTRAIN",
+
+  // User has successfully uploaded new samples
+  SAMPLES_UPLOAD_SUCCESS = "SAMPLES_UPLOAD_SUCCESS",
 }
 
 /**
@@ -43,7 +46,7 @@ export type EventData = Record<string, EventValue | undefined>;
 
 // While we only send values of EventValue, sometimes we send a JSON string.
 // For ease of readability below, we alias string for those cases.
-// type JsonString = string; // TODO VOODOO uncomment on first use
+type JsonString = string;
 
 /**
  * Structure of additionalEventData for each EVENT_TYPES type that sends it.
@@ -87,4 +90,12 @@ export type EventData = Record<string, EventValue | undefined>;
 export type AnalyticsTreeViewNextstrain = {
   // Tree that user is being sent to view
   tree_id: number;
+};
+
+/** EVENT_TYPES.SAMPLES_UPLOAD_SUCCESS*/
+export type AnalyticsSamplesUploadSuccess = {
+  // How many samples the user just uploaded
+  sample_count: number;
+  // JSON array of all the IDs for newly created samples for this upload
+  sample_ids: JsonString;
 };
--- a/src/frontend/src/common/analytics/eventTypes.ts
+++ b/src/frontend/src/common/analytics/eventTypes.ts
@@ -21,6 +21,11 @@ export enum EVENT_TYPES {
 
   // User has successfully uploaded new samples
   SAMPLES_UPLOAD_SUCCESS = "SAMPLES_UPLOAD_SUCCESS",
+
+  // User downloading data about samples to a file(s)
+  // Does not currently address success/failure, but download failures are very
+  // rare, so generally safe to not be concerned about that aspect for now.
+  SAMPLES_DOWNLOAD_FILE = "SAMPLES_DOWNLOAD_FILE",
 }
 
 /**
@@ -98,4 +103,18 @@ export type AnalyticsSamplesUploadSuccess = {
   sample_count: number;
   // JSON array of all the IDs for newly created samples for this upload
   sample_ids: JsonString;
+};
+
+/** EVENT_TYPES.SAMPLES_DOWNLOAD_FILE*/
+export type AnalyticsSamplesDownloadFile = {
+  // How many samples the user is downloading info on
+  sample_count: number;
+  // JSON array of the public identifiers for these samples
+  // Not the same as a sample PK, but it plus group_id is unique in DB
+  // (based on context of how download happens, PKs are hard to obtain)
+  sample_public_ids: JsonString;
+  // User downloaded info on consensus genome (FASTA) for these samples
+  includes_consensus_genome: boolean;
+  // User downloaded info on metadata for these samples
+  includes_sample_metadata: boolean;
 };
--- a/src/frontend/src/common/analytics/eventTypes.ts
+++ b/src/frontend/src/common/analytics/eventTypes.ts
@@ -4,11 +4,87 @@
  * This exists to ensure consistency/clarity across types and to provide
  * a reference for data analysts working on CZ GE.
  *
- * - Each event type should have a plain-English description of what it means.
+ * --- REQUIREMENTS for adding new event types ---
+ * - Each event type should have a comment with a plain-English description
+ *   of what the event means.
  * - Event types should be written in SCREAMING_SNAKE_CASE
  * - Each entry in the enum should be a duplicate key/value string enum
+ * - If an event type entails sending additional data about what happened
+ *   (most events do), there should be a matching event TypeScript type below
+ *   in this file that shows the structure of the additional event data.
+ *   => The TS type should be named according to the event type as follows:
+ *      Event of `SOME_EVENT_FOO` <--> TS type of `AnalyticsSomeEventFoo`
  */
 export enum EVENT_TYPES {
   // User has been sent to Nextstrain to view a phylo tree
   TREE_VIEW_NEXTSTRAIN = "TREE_VIEW_NEXTSTRAIN",
 }
+
+/**
+ * Values we consider reasonable to send to Segment as properties of event.
+ *
+ * Note that `undefined` is also acceptable within the event properties object
+ * as a value, but it is instead interpreted as "delete this key" when Segment
+ * receives. We allow `undefined` to be passed as part of EventData, but for
+ * clarity it's not considered an EventValue since it's not a value, per se.
+ *
+ * This is a convention we are choosing to make. Segment allows us to send
+ * pretty much anything, but we want to ensure a flat structure of key-value
+ * pairs for downstream analysis since the structure of the event properties
+ * implicitly determines the schema used for the event table when Segment
+ * syncs to the the data warehouse.
+ *
+ * We enforce this structure for EventData in the `analyticsTrackEvent` method
+ * that kicks off the underlying Segment `track` by ensuring that all events
+ * extend from that base structure.
+ */
+type EventValue = string | number | boolean | null;
+export type EventData = Record<string, EventValue | undefined>;
+
+// While we only send values of EventValue, sometimes we send a JSON string.
+// For ease of readability below, we alias string for those cases.
+// type JsonString = string; // TODO VOODOO uncomment on first use
+
+/**
+ * Structure of additionalEventData for each EVENT_TYPES type that sends it.
+ *
+ * Most of the event types we track include some kind of additionalEventData
+ * beyond just which EVENT_TYPES type occurred. For those event types, we have
+ * a TypeScript type that details the structure of that data. These TS types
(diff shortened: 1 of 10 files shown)


src/frontend/src/views/Data/components/TreeActionMenu/components/TreeTableDownloadMenu/index.tsx
```diff
@@ -1,13 +1,19 @@
...
+import { analyticsTrackEvent } from "src/common/analytics/methods";
 import { NewTabLink } from "src/common/components/library/NewTabLink";
 import { TREE_STATUS } from "src/common/constants/types";
 import DownloadIcon from "src/common/icons/IconDownloadSmall.svg";
 import { stringGuard } from "src/common/utils";
 import { StyledIcon, StyledIconWrapper } from "../../style";
 
 interface Props {
-  item: TableItem;
+  item: PhyloRun;
```
vincent-czi, line 16: The type change here is safe. This component only gets used in one place, and that place always passes a `PhyloRun`. The more generic `TableItem` left a lot of wiggle room that could possibly break my newly written `analyticsGetData` (and the component itself, it relies on PhyloRun structure implicitly), so this change is to increase type safety if we accidentally change the upstream and forget that this component isn't very generic.

src/frontend/src/common/analytics/eventTypes.ts
```diff
@@ -4,11 +4,168 @@
...
+};
+
+/** EVENT_TYPES.TREE_DOWNLOAD_SELECTED_SAMPLES_TEMPLATE */
+export type AnalyticsTreeDownloadSelectedSamplesTemplate = {
+  // Tree user is downloading template in regards to.
+  // Can download template before tree done or tree failed. Null indicates such
+  tree_id: number | null;
+  // PK of the workflow that kicked off the creation of this tree.
+  // Should never be null, but TS for underlying item does not guarantee it, so
+  // the null possibility is mostly to keep TS happy. If null, app has a bug.
```
mmmmmmaya, line 147: I'm not sure all the extra `| null` is the best solution for "not technically guaranteed" types here. Is there a reason we can't, for example, check for null-ish values and only track if it exists? or add some logic guards to make TS happy while keeping the safety features we want in place? 

If something here could indicate the app potential has a type-related bug, I would be interested in having the code changed to prevent that possibility from a TS perspective. The point of TS is to catch and prevent these types of mistakes, so adding `| null` seems counter to having TS in the app, from an actual type safety perspective.

src/frontend/src/common/analytics/eventTypes.ts
```diff
@@ -4,11 +4,168 @@
  * This exists to ensure consistency/clarity across types and to provide
  * a reference for data analysts working on CZ GE.
  *
- * - Each event type should have a plain-English description of what it means.
+ * --- REQUIREMENTS for adding new event types ---
```
mmmmmmaya, line 7: I like the idea of having the requirements outlined in english so new peeps can follow along. Do you feel these requirements are also codified, such that a new person who skips the comments can't make the mistake (ie: programmatic enforcement)?

src/frontend/src/common/analytics/methods.ts
```diff
@@ -73,31 +84,64 @@ type EventValue = string | number | boolean | null;
...
- *     - All the keys in additionalEventData should be in **snake_case** to
- *     match expectation when working with analytics database.
- *     - additionalEventData should be a flat object of simple key-value pairs.
- *     You can send `undefined` for a value, but it will cause the key to be
- *     considered deleted for that event when it lands in Segment.
+ * When passing additionalEventData, it's best to directly construct it as an
+ * object in part of the function call to trigger "Excess Property Checking"
+ * and help ensure that the passed event data matches the event TS type being
+ * specified for the event. If instead you do it like this,
+ *     NOT GREAT  ---> analyticsTrackEvent<AnalyticsFooEvent>(
```
mmmmmmaya, line 91: can we avoid this with fancier typing?

src/frontend/src/common/analytics/methods.ts
```diff
@@ -73,31 +84,64 @@ type EventValue = string | number | boolean | null;
...
+ *
+ * Part (A) is pretty self-explanatory, but (B) is less so and it's why we need
+ * all the complicated TS stuff below. Basically, whatever event data we send
+ * up to Segment via a `track` call will **implicitly** define the DB schema
+ * used to store that event downstream. But because this creation is implicit,
+ * if we change the structure of that payload later it can cause problems for
+ * downstream analysis. It's possible to do without issue, but accidentally
+ * changing it is generally bad. So all the TS fanciness is to ensure that,
+ * when additionalEventData is being sent, we have a clear structure for the
+ * additional data, we adhere to that, and we only change it on purpose.
```
mmmmmmaya, line 115: ❤️

src/frontend/src/common/analytics/eventTypes.ts
vincent-czi: In general, I agree about the type-related bug, but I didn't really want to go back and figure out why `id?:` is in PhyloRun right now. I think it doesn't make sense for a PhyloRun to be missing an id, but being sure about that was outside what I wanted to figure out right now.

For the TSV download event, I actually disagree: it's intentional for the `tree_id` to be null in the case when the tree is not completed yet or failed, but the user can still download the associated TSV. In that case, the tree_id really should be null. The other event of downloading a tree really should never have a null `tree_id`, so I agree on that count, but fixing it felt like it would take a decent amount of effort since the underlying PhyloRun has no requirement that a tree exists yet. It really comes down to `disabled` being used as form of type guard that I can't easily expand to the analytics call. (Although that's not even a "real" type guard, because we're assuming a contract of how the BE returns `TREE_STATUS.Completed` in terms of the tree, but that feels reasonable to me.)

src/frontend/src/common/analytics/eventTypes.ts
vincent-czi: Some are, some aren't. Flat event structures are enforced, but things like naming and commenting what an event is/contains are not. I don't really have any ideas for I could enforce the naming/documentation aspect beyond some sort of extra tool that is running on the raw text of the file, and that feels bad for other reasons. I'm mostly pinning this on hope that A) devs do the same as they see already done in a file, B) PR reviews, and C) devs read the docs.

src/frontend/src/common/analytics/methods.ts
vincent-czi: I do not believe so. I'm fairly confident that "Excess Property Checking" only happens in the context of an object literal. I spent a fair amount of time looking into the idea of an `Exact` type and as far as I could find, it does not exist. (There are possible workarounds, but when I tested them out I never was able to find a satisfactory approach.) You can see this GH issue on TS repo for what felt, to me at least, like the official "we're not going to do exact types": https://github.com/microsoft/TypeScript/issues/12936#issuecomment-284590083  -- it is a bit old from 2017, but I don't think the needle moved.

I spent way too much time poking at this, trying different approaches over a few days and searching the internet. This is the approach I was happiest with, although I'm still not in love with it.

src/frontend/src/common/analytics/eventTypes.ts
phoenixAja: It would be nice to have this documented somewhere in a readme page for frontend how-tos (and link to that in the code), I worry about the integrity of big comment blocks over time

src/frontend/src/common/analytics/eventTypes.ts
```diff
@@ -4,11 +4,168 @@
...
+  // User has kicked off the creation of a Nextstrain phylo tree
+  TREE_CREATION_NEXTSTRAIN = "TREE_CREATION_NEXTSTRAIN",
+
+  // User has been sent over to UShER site. UShER will now create tree and
+  // display it on their site. External, so can't tell if tree succeeds/fails.
+  // **No `additionalEventData`**. Just tracking event occurrence right now.
+  TREE_CREATION_VIEW_USHER = "TREE_CREATION_VIEW_USHER",
+
+  // User has downloaded a file of the actual phylo tree
+  TREE_DOWNLOAD_TREE_FILE = "TREE_DOWNLOAD_TREE_FILE",
```
phoenixAja, line 31: nit: I would say a better name for this would be `TREE_DOWNLOAD_JSON_FILE`

src/frontend/src/common/analytics/eventTypes.ts
```diff
@@ -4,11 +4,168 @@
...
+
+  // User downloaded a template with sample identifiers and metadata showing
+  // which selected. User could then add more metadata to overlay on tree view.
```
phoenixAja:
```suggestion
  // which samples were selected. User could then add more metadata to overlay on tree view.
```

src/frontend/src/common/analytics/eventTypes.ts
```diff
@@ -4,11 +4,168 @@
...
+  TREE_DOWNLOAD_SELECTED_SAMPLES_TEMPLATE = "TREE_DOWNLOAD_SELECTED_SAMPLES_TEMPLATE",
+
+  // User has successfully uploaded new samples
+  SAMPLES_UPLOAD_SUCCESS = "SAMPLES_UPLOAD_SUCCESS",
+
+  // User downloading data about samples to a file(s)
+  // Does not currently address success/failure, but download failures are very
+  // rare, so generally safe to not be concerned about that aspect for now.
+  SAMPLES_DOWNLOAD_FILE = "SAMPLES_DOWNLOAD_FILE",
```
phoenixAja, line 43: for samples download we have TSV and fasta downloads, does this event capture both of those events or just one?

phoenixAja: oh i see later in it captures both events

src/frontend/src/common/analytics/eventTypes.ts
vincent-czi: I'm happy to point it out or duplicate it in the readme, but I actually have a differing opinion. I agree that comment blocks getting out of sync with reality is a common issue, but I personally think documentation located in separate places drifts even faster. I think a giant comment block right next to the code in question has a (slightly) better chance of getting updated than a readme, simply because the dev doing the change might not be aware the readme exists or documents that stuff.

src/frontend/src/common/analytics/eventTypes.ts
vincent-czi: So here's my thinking. I considered naming it that when I was working, but I wound up thinking it was better to avoid naming the file type in case of later changes/addition. I know that right now all trees come as JSON, but it's conceivable another file type (YAML or maybe something totally new) could come on the scene  eventually. It felt like something where you can name the implementation or name the concept. If we start offering different filetypes down the road, we could then add a `tree_file_type` key to the event that names which one, but right now it's just JSON, so we don't need that yet.

That said, it sure is awkward naming! We're downloading from the category of tree stuff, and we're downloading a file of the tree. I think the best name would be `TREE_DOWNLOAD_FILE`, but it's confusing, because we also have other file downloads around trees (eg, `TREE_DOWNLOAD_SELECTED_SAMPLES_TEMPLATE`), so the nice name is too generic, because it gets confused with the other flavors of download involved in tree stuff.

I'm basically looking at it like this `TREE_DOWNLOAD` means we're downloading something from the category of tree stuff, we're on the tree table, etc. Then the next part if what we're downloading from there, and we are downloading the file representation of the tree, so `TREE_FILE`.

mmmmmmaya (approved):
overall lgtm, please make follow up tickets for docs and typing stuff in the eng excellence epic

phoenixAja (approved):

PR #1266, opened 2022-08-09
populate optional fields on upload (Nextstrain header support)

phoenixAja:
### Summary:
- **What:** autopopulate sampleId and PrivateId with `strain` when users use the nextstrain template, remove keepPrivate from required fields
- **Ticket:** [sc208436](https://app.shortcut.com/genepi/story/208436)
- **Env:** `<rdev link>`

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
+++ b/src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
@@ -66,6 +66,8 @@ interface ParsedRow {
 // If header is unrecognized, leaves it alone (useful for warnings, etc).
 // User can also use Nextstrain header defaults as an alias
 function convertHeaderToMetadataKey(headerName: string): string {
+  // console.log("headerName in convert headers", headerName); // REMOVE
+  // console.log("headerName in convertHeader", headerName); // REMOVE
   if (headerName in HEADERS_TO_METADATA_KEYS) {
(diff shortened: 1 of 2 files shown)


phoenixAja:
this feels hacky, but i couldn't find a way to transform the data another way through papaparse that was cleaner. if anyone knows a better way lmk

src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
```diff
@@ -89,9 +89,17 @@ function getMissingHeaderFields(
   )) {
     if (
       !uploadedHeaders.includes(metadataKey) &&
-      !headerField.includes("Optional")
+      !headerField.includes("Optional") &&
+      metadataKey !== "keepPrivate" // TODO: rename field to have -Optional flag
```
ehoops-cz, line 93: What is stopping us from making the "keepPrivate" key optional?

ehoops-cz (approved):
Looks good to me.  Left a question.

src/frontend/src/views/Upload/components/Metadata/components/ImportFile/parseFile.ts
phoenixAja: it's a larger product/design decision. Mainly we don't want users to have to use a different template if it's not really a breaking change

PR #1268, opened 2022-08-09
Make it easier for existing users to accept invitations instead of creating a new account

jgadling:
### Summary:
- **Ticket:** [sc208570](https://app.shortcut.com/genepi/story/208570)
- **Env:** https://acceptinvitation-frontend.dev.czgenepi.org

### Notes:
Alright, this PR does a few things:
- Adds a setup script to convert local dev to relying 100% on our localdev auth0 tenant for auth
- Updates local dev to enable the "sync roles" feature flag by default
- Adds a new endpoint for processing a user's invitation (basically trying to duplicate what auth0 does when processing an invitation acceptance 😢 )
- Updates the login endpoint to check whether an invitation is for a user that already exists in the db. If it does, it tries to encourage the user to *log in* instead of using the default auth0 flow which prompts users to *create a new account* by default.
- Updates the callback endpoint to read session parameters that might be set by the login endpoint, to redirect the user to the "process invitation" endpoint if necessary.

The flow diagram is here: https://lucid.app/lucidchart/77a5ba77-7e05-4987-847c-2250ded9ddfa/edit?invitationId=inv_7ac1d716-e2a0-4936-9928-8af7cd55a8b0&page=0_0#

I'm happy to go over this in person if anyone's interested!

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/scripts/auth0_login.sh
+++ b/scripts/auth0_login.sh
@@ -54,4 +54,4 @@ ${local_aws} secretsmanager update-secret --secret-id genepi-config --secret-str
   "AUTH0_MANAGEMENT_DOMAIN": '"${AUTH0_MANAGEMENT_DOMAIN}"'
 }' || true
 
-docker compose restart backend
+docker compose exec backend supervisorctl restart fastapi
--- a/scripts/auth0_login.sh
+++ b/scripts/auth0_login.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+# Fetch certain secrets that make local dev work better from *real* AWS
+# so we can feed them to localstack (fake aws)
+
+# CI doesn't support profiles right now, so work around it.
+PROFILE="--profile genepi-dev"
+# GitHub actions can't handle our remapped DNS or AWS profiles :'(
+if [ -n "${CI}" ]; then
+	PROFILE=""
+fi
+# Fetch some additional data from real-aws to populate in our fake-aws secret.
+EXTRA_SECRETS=$(aws ${PROFILE} secretsmanager get-secret-value --secret-id localdev/genepi-config-secrets --query SecretString --output text)
+
+export AWS_REGION=us-west-2
+export AWS_DEFAULT_REGION=us-west-2
+export AWS_ACCESS_KEY_ID=nonce
+export AWS_SECRET_ACCESS_KEY=nonce
+export AWS_PAGER=""
+
+export FRONTEND_URL=http://frontend.genepinet.localdev:8000
+export BACKEND_URL=http://backend.genepinet.localdev:3000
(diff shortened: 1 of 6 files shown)


src/backend/aspen/api/views/auth.py
```diff
@@ -26,23 +27,121 @@
...
+) -> Optional[Response]:
+    # If this invitation is for an email address that already exists in our db,
+    # we'll use our custom invitation acceptance flow to associate their existing
+    # account with the invitation. If we haven't seen that email address before,
+    # we'll send them to the standard auth0 "create a new account" flow.
+
+    # Load more information about the invitation from auth0
+    ticket_info = get_invitation_ticket(a0, organization, invitation)
+    if not ticket_info:
+        return None
```
phoenixAja, line 70: when would there be no ticket_info?

jgadling: If the invitation they tried to use doesn't exist or is expired.

phoenixAja (approved):
looks good!

scripts/auth0_login.sh
```diff
@@ -0,0 +1,57 @@
...
+  "S3_external_auspice_bucket": "genepi-external-auspice-data",
+  "S3_db_bucket": "genepi-db-data",
+  "ONETRUST_FRONTEND_KEY": '"${ONETRUST_FRONTEND_KEY}"',
+  "PLAUSIBLE_FRONTEND_KEY": '"${PLAUSIBLE_FRONTEND_KEY}"',
+  "AUTH0_MANAGEMENT_CLIENT_ID": '"${AUTH0_MANAGEMENT_CLIENT_ID}"',
+  "AUTH0_MANAGEMENT_CLIENT_SECRET": '"${AUTH0_MANAGEMENT_CLIENT_SECRET}"',
+  "AUTH0_MANAGEMENT_DOMAIN": '"${AUTH0_MANAGEMENT_DOMAIN}"'
+}' || true
+
+docker compose exec backend supervisorctl restart fastapi
```
lvreynoso, line 57: Q - what is this script for, exactly? Couldn't this just be a command in the Makefile?

jgadling: It tells local dev to use our auth0 localdev tenant for authentication (the login link goes to auth0 instead of the local OIDC container) and imop, this is a pretty long block that would be hard to translate to a make target. We could add a make target that calls it though!

lvreynoso: Oh, I didn't catch that! Don't worry about it then; I was just wondering if it could be consolidated because it seems to duplicate a lot of commands in the setup dev data script, but I just wasn't reading closely enough.

PR #1269, opened 2022-08-10
Add gisaid_submitter_id column to users table and api support

lvreynoso:
### Summary:
- **What:** Adds `gisaid_submitter_id` column to users table, as well as api support.
- **Ticket:** [sc209738](https://app.shortcut.com/genepi/story/209738), [sc210080](https://app.shortcut.com/genepi/story/210080
- **Env:** `<rdev link>`

### Demos:
Screen Shot 2022-08-09 at 17 12 27

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/schemas/usergroup.py
+++ b/src/backend/aspen/api/schemas/usergroup.py
@@ -44,6 +44,7 @@ class UserUpdateRequest(BaseRequest):
     agreed_to_tos: Optional[bool] = None
     acknowledged_policy_version: Optional[datetime.date] = None
     name: Optional[str] = None
+    gisaid_submitter_id: Optional[str] = None
 
 
 class GroupRoleResponse(BaseResponse):
@@ -56,6 +57,7 @@ class GroupRoleResponse(BaseResponse):
 class UserMeResponse(UserBaseResponse):
     split_id: str
     analytics_id: str
+    gisaid_submitter_id: Optional[str]
     group: GroupResponse
     groups: List[GroupRoleResponse]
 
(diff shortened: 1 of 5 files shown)


lvreynoso:
Tagging @danrlu and @ehoops-zz for visibility

jgadling:
PR looks good! Only comment I have is that I think we should start running migrations as their own deployment, and then do a separate deployment that depends on the migration. Or we can start running migrations before updating the app. Any thoughts? also cc @phoenixAja and @vincent-czi

vincent-czi:
@jgadling I think my preference would be to have migrations run before updating the app. I think in most cases, for simple migrations, that change would be enough. For more complicated migrations, having a way to pre-run the migration would be good, but I'm not sure what that would look like. If we merge to `trunk`, we need to have it hold off on a deploy until migration was run. I guess we could have a separate branch called `migration` that would only run migrations but not deploy? And if that succeeds, we can go back and merge to `trunk`? And then there's the issue of our scheduled Prod push, I don't know how the migration would work with what I just sketched out. I think this is mostly me just not being well-versed in the options we have: would love to discuss your ideas on different ways to approach. But for simple stuff, as a default, just having migrations run before deployment would probably fix most of our problems. We add more frequently than we subtract, so updating DB before model would probably have fewer issues.

phoenixAja (approved):

phoenixAja:
why did you add this field to `UserMeResponse` over the `UserBaseResponse` ? (UserBaseResponse feels like a better place for it to me)

phoenixAja:
thanks for tagging the team for visibility 🙌 !!

phoenixAja (approved):

lvreynoso:
> why did you add this field to `UserMeResponse` over the `UserBaseResponse` ? (UserBaseResponse feels like a better place for it to me)

I wasn't sure if we wanted to expose the gisaid submitter id to other group members. Should we?

PR #1270, opened 2022-08-10
SC-210372 MUIv5 and Czifuiv7 upgrades

ehoops-cz:
### Summary:
- **What:** `Upgrading MUI and czifui to get the latest updates`
- **Ticket:** [sc210372](https://app.shortcut.com/genepi/story/210372/upgrade-to-muiv5-and-czifui-v7-0-0)
- **Env:** https://czifui-upgrade-frontend.dev.czgenepi.org/data/samples

### Checklist:
- [x] I added labels to my PR


src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
```diff
@@ -155,12 +155,7 @@ const DownloadModal = ({
         </B>{" "}
         <ContactUsLink />
       </Notification>
-      <Dialog
-        disableEscapeKeyDown
-        disableBackdropClick
-        open={open}
-        onClose={handleCloseModal}
-      >
+      <Dialog disableEscapeKeyDown open={open} onClose={handleCloseModal}>
```
mmmmmmaya, line 163: why was `disableBackdropClick` removed?

src/frontend/src/components/NavBar/components/StaticPageNavBar/style.ts
```diff
@@ -1,4 +1,5 @@
 import styled from "@emotion/styled";
+import { fontBodyS } from "czifui";
```
mmmmmmaya: looks like this isn't being used anywhere

mmmmmmaya:
Looks like your push to rdev didn't work. Generally this looks good, but have a few q's and would love to see it working on an env

src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
ehoops-cz: The prop was removed as part of the migration ([docs](https://mui.com/material-ui/migration/v5-component-changes/#%E2%9C%85-remove-disablebackdropclick-prop)).  They recommend handling this in onClose if we want to.  Do you know why we're overriding the default here?

ehoops-cz: I will either remove the prop from our BaseDialog component or add it to onClose there

mmmmmmaya: I have no idea why.

mmmmmmaya:
also, if any of the github actions fail, your rdev doesn't deploy properly (any lint failures or ts errors cause the front end to not deploy so you can't access the app _even if_ it works on your local machine)

ehoops-cz:
Thanks @Janeece for looking through the rdev and noting a few issues [here](https://docs.google.com/document/d/17eg1FKDEU_1KAS88XQXHYAzAU7DLZwQCWjuhm4tnG5Q/edit).  Fixes incoming

ehoops-cz:
Polish fixes
Screen Shot 2022-08-19 at 2 59 11 PM
Screen Shot 2022-08-19 at 2 58 55 PM
Screen Shot 2022-08-19 at 3 00 28 PM
Screen Shot 2022-08-19 at 3 00 18 PM
Screen Shot 2022-08-19 at 2 59 43 PM

src/frontend/pages/welcome/[groupId].tsx
```diff
@@ -1,4 +1,3 @@
-import React from "react";
```
Comment (mmmmmmaya) [inline src/frontend/pages/welcome/[groupId].tsx:1]: what's up with this?

src/frontend/src/common/components/library/data_table/index.tsx
```diff
@@ -54,7 +54,7 @@ export function defaultSampleCellRenderer({
 
   return (
     <RowContent header={header}>
-      <Cell data-test-id={`row-${header.key}`}>{displayData}</Cell>
+      <Cell data-test-id={`row-${header.key}`}>{`${displayData}`}</Cell>
```
mmmmmmaya: why was this change necessary?

src/frontend/src/common/components/library/data_table/index.tsx
```diff
@@ -67,7 +67,7 @@ export function defaultTreeCellRenderer({
 
   return (
     <TreeRowContent>
-      <Cell data-test-id={`row-${header.key}`}>{displayData}</Cell>
+      <Cell data-test-id={`row-${header.key}`}>{`${displayData}`}</Cell>
```
mmmmmmaya: same here

src/frontend/src/components/WebformTable/components/Row/style.ts
```diff
@@ -19,7 +19,7 @@ export const Id = styled.div`
     const spaces = getSpaces(props);
 
     return `
-      padding: ${spaces?.m}px 0 ${spaces?.m}px ${spaces?.s}px
+    padding: ${spaces?.m}px ${spaces?.m}px ${spaces?.m}px ${spaces?.s}px;
```
mmmmmmaya:
```suggestion
      padding: ${spaces?.m}px ${spaces?.m}px ${spaces?.m}px ${spaces?.s}px;
```

src/frontend/src/views/Data/cellRenderers.tsx
```diff
@@ -179,7 +178,7 @@ const TREE_CUSTOM_RENDERERS: Record<string | number, CellRenderer> = {
     <TreeTypeTooltip value={value as string}>
       <TreeRowContent>
         <UnderlinedCell data-test-id={`row-${header.key}`}>
-          {value}
+          {`${value}`}
```
mmmmmmaya: same here

src/frontend/src/views/Data/index.tsx
```diff
@@ -154,7 +154,7 @@ const Data: FunctionComponent = () => {
   // create JSX elements from categories
   dataCategories.forEach((category) => {
     dataJSX.menuItems.push(
-      <StyledMenuItem>
+      <StyledMenuItem key={`${category.text}`}>
```
mmmmmmaya:
```suggestion
      <StyledMenuItem key={category.text}>
```

src/frontend/src/common/components/library/data_table/index.tsx
mmmmmmaya:
```suggestion
      <Cell data-test-id={`row-${header.key}`}>{displayData}</Cell>
```

src/frontend/src/common/components/library/data_table/index.tsx
mmmmmmaya:
```suggestion
      <Cell data-test-id={`row-${header.key}`}>{displayData}</Cell>
```

src/frontend/src/views/Data/cellRenderers.tsx
mmmmmmaya:
```suggestion
          {value}
```

mmmmmmaya (commented):
This mostly looks fine besides (what I assume is) some code mod weirdness. Though I'd really like to understand what all the react imports were removed.

mmmmmmaya:
Screen Shot 2022-08-22 at 8 33 20 AM

mmmmmmaya:
also downloading files from the tree table seems to be broken

mmmmmmaya (changes requested):
some broken functionality

ehoops-cz:
> This mostly looks fine besides (what I assume is) some code mod weirdness. Though I'd really like to understand what all the react imports were removed.

As of React 17, React no longer needs to be in scope to use JSX. ([doc](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html)).  The linter was insisting I take them out - I'm not sure why it cared now, but not before.  I tried setting the tsconfig compilerOption for jsx to `    "jsx": "react",` which stopped the lint issues, but then I found that Nextjs changes the compilerOption back to "preserve" when the dev server starts.  Not having write permission to change this was why my rdev couldn't deploy. 
 Apparently there's a way to stop the overwrite, but since the React imports are no longer necessary, I just took them out.  If there's a reason to keep them in, I can look into the tsconfig to see about appeasing nextjs and the linter.

src/frontend/src/common/components/library/data_table/index.tsx
ehoops-cz: Without this, there is a typescript error.  I think this is because displayData isn't a string.
Screen Shot 2022-08-22 at 9 51 59 AM

src/frontend/src/views/Data/cellRenderers.tsx
ehoops-cz: All of these are related to typescript.  I'm definitely open to suggestions if there's a more elegant way to address the type errors.

Screen Shot 2022-08-22 at 9 53 56 AM

Janeece:
> Screen Shot 2022-08-22 at 8 33 20 AM

Good catch on these progress dialogs -- in-progress, success, and probably all have the same issue with top padding missing. 

Also, idk if this is expected or not, but I'm not able to send any samples to make an UShER placement. Wanted to call it out in case that should be possible in this environment.

ehoops-cz:
> also downloading files from the tree table seems to be broken

@mmmmmmaya what are you seeing that is broken?  I tested and I see files getting downloaded, but I'm not sure if they might be the wrong ones or something.

mmmmmmaya:
the id in the link (shown in the lower left of my browser when I hover the link), is `undefined`, so it just loads the samples page again in a new tab instead of downloading anything
after

ehoops-cz:
> the id in the link (shown in the lower left of my browser when I hover the link), is `undefined`, so it just loads the samples page again in a new tab instead of downloading anything after [ after ](https://user-images.githubusercontent.com/7562933/186015402-3e5ef8b1-9c57-4a13-bc6c-273cf0263773.gif) [ ](https://user-images.githubusercontent.com/7562933/186015402-3e5ef8b1-9c57-4a13-bc6c-273cf0263773.gif)

That's how it looks on staging for failed trees.  Potentially also broken there or maybe expected?
Screen Shot 2022-08-22 at 1 50 31 PM

This is how it looks on the branch for completed trees:
Screen Shot 2022-08-22 at 1 52 00 PM

ehoops-cz:
Fixed dialog spacing:
Screen Shot 2022-08-22 at 1 47 20 PM

src/frontend/pages/_document.tsx
```diff
@@ -86,3 +55,70 @@ export default class MyDocument extends Document {
...
+    });
+
+  const initialProps = await Document.getInitialProps(ctx);
+  // This is important. It prevents Emotion to render invalid HTML.
+  // See https://github.com/mui/material-ui/issues/26561#issuecomment-855286153
+  const emotionStyles = extractCriticalToChunks(initialProps.html);
+
+  const emotionStyleTags = emotionStyles.styles.map(
+    (style: { key: string; ids: Array<string>; css: string }) => (
+      <style
```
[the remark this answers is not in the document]
ehoops-cz: /helpful

src/frontend/src/views/Data/cellRenderers.tsx
vincent-czi: I only read through the various comments on the PR, not the actual files changed, so I might be missing additional context, but here's my thoughts on this:

- If the only reason it's here is to do a type cast in to a string, I think it's clearer to the reader to do `String(value)` than the string interpolation, even if the result is the same.
- I think the root cause here is how generic the parent components and data being moved around is trying to be. From my experience messing with table stuff in the past, a lot of this comes down to everything being written in a very generic way at a high level, but then the actual low level display of data not being very generic. But instead of an explicit narrowing moment somewhere in the middle, it's more kinda hand-waved.
  - I'm curious how this was passing without error before? Was there a type assertion or something that got removed?
- If I'm right and the root cause is that types are just rather amorphous in regards to our table implementation, I don't think there's any value in figuring out a good solution right now. When Maya refactors the tables, a lot of this will have to change anyway, and we can either fix it then or the refactoring itself will do the fixing.

ehoops-cz: @vincent-czi There were some type upgrades in this PR, which is likely why typescript wasn't complaining before but is now.  I agree about not fighting too much with the current table.  I will swap to using String(value) instead for clarity 👍

mmmmmmaya (approved):

PR #1279, opened 2022-08-13
Patch up localstack persistence.

jgadling:
### Notes:
Reverts to an older version of localstack that supports persistence and enables it, so we won't loose localstack data (secrets, s3 buckets, etc) every time the localstack container restarts.

### Checklist:
- [ ] I added labels to my PR


vincent-czi (approved):
modern problems require modern solutions

PR #1280, opened 2022-08-15
Wrapping up analytics

vincent-czi:
### Summary:
- **What:** Wrap up of analytics: OneTrust styling, Privacy Policy tweaks, other small finishing touches
- **Ticket:** [sc<199423>](https://app.shortcut.com/genepi/story/<199423>), [sc<210424>](https://app.shortcut.com/genepi/story/<210424>)
- **Env:** http://vince-analytics-wrap-frontend.dev.czgenepi.org/

### Demos:
Styled "first time on site with analytics now merged" OneTrust banner

image

Styled Cookie Settings (opt-in/out of analytics) OneTrust modal

image

Cookie Settings now available in the top right user menu, opens same modal as above

image

### Notes:
- I wound up creating a circular dependency in my previous PR which was causing some very weird errors on app init. Extracted some sections of code to their own files and/or moved around where functions were located to fix the problems.
- Most of the work here is in styling the OneTrust banner we use for letting users opt-in/out of analytics. Shout out to Damian, who wrote the majority of the stylesheet: I mostly just integrated it along with adding images and a few tiny tweaks.
- Damian wrote the stylesheet in Sass (.scss) and I think we want to keep it in Sass. However, we also just went through a bunch of effort to remove the last of our Sass stylesheets and almost completely convert over to styling in React and not through stylesheets. Because OneTrust stuff is outside the flow of React, we have to style it with a stylesheet. But I think we're almost never going to touch this stylesheet again, so I'm of the -- possibly controversial opinion -- that it's not worth it to bring in any kind of Sass dep to our app. We can just manually run a Sass -> CSS tool if we need to change it again in a year. And if we ever change it more than once a year, then I was wrong and silly, and we should bring in a dependency. I don't feel super strongly about this opinion, so if someone feels the opposite I'm willing to change, but I don't think the dependency is worth the other trade offs from where I'm standing right now.
- Our FAQ has moved off site and in to the Help Center. I need to do a separate PR against trunk at some point in the next day or two to remove our in-app FAQ and ensure we only ever point at the new, correct Help Center version of the FAQ and never the old, stale in-app FAQ.

### Checklist:
- [x] I merged latest `vince/analytics`
- [x] I added labels to my PR


ehoops-cz:
I agree with the sass vs css opinion above.  I like the idea of not over-complicating this until we actually need it.

src/frontend/src/common/styles/oneTrust.scss
```diff
@@ -0,0 +1,366 @@
...
+//   npx sass oneTrust.scss ./oneTrust.css
+// to compile it to CSS for React to actually use in the app.
+//
+// If we have to do the above more than once a year, I (Vince) was wrong and
+// silly and we should totally install a Sass dev dep and do this the right
+// way. But from where I'm standing right now, I think we'll probably
+// never update this unless we go through a huge re-styling of the entire app.
+// Also, resulting CSS is currently stylelint ignored b/c it has some
+// minor, but not-worth-solving issues. If we have to actively develop this
+// stylesheet, we should remove that and probably fix those.
```
ehoops-cz, line 24: 👍

src/frontend/src/components/NavBar/components/RightNav/components/UserMenu/index.tsx
```diff
@@ -20,6 +20,14 @@ const UserMenu = ({ user }: UserMenuProps): JSX.Element => {
     setAnchorEl(null);
   };
 
+  const handleCookieSettings = () => {
+    // OneTrust _should_ always be loaded by time user doing real interaction
```
ehoops-cz, line 24: Where does OneTrust load from?  I assume it's not blocked by ad blockers?

ehoops-cz (approved):

src/frontend/src/components/NavBar/components/RightNav/components/UserMenu/index.tsx
vincent-czi: OneTrust loads from a script hosted by OneTrust's CDN. Honestly, it probably is blocked by certain ad blockers, but I think we would consider that an alternative from of opt-out. I don't think OT would normally get blocked, but very aggressive ones might.

mmmmmmaya (approved):

src/frontend/src/components/NavBar/components/RightNav/components/UserMenu/index.tsx
vincent-czi: I realize now that if it has been blocked, that "Cookie Settings" menu option will permanently do nothing for that user. I had just thought of it as a wacky edge case where if the OneTrust script hadn't loaded yet or failed to load due to a network error, it would be broken temporarily, but probably fixed after waiting a couple more seconds or on a refresh. I should bring this up to Olivia and/or Christine that if OneTrust is fully blocked, the Cookie Settings button will do nothing. Maybe we should have an `alert` message pop up to cover this case? I expect it's very rare, but thanks for making me think about it.

PR #1285, opened 2022-08-17
Test/sc 210124/switch out jest playwright

charles-testco:
### Summary:
- **What:** `Switching out Jest-Playwright. E2E tests to run using Playwright and unit tests continue using Jest until we decide otherwise. No test to use Jest-Playwright any more`
- **Ticket:** [sc-210124](https://app.shortcut.com/genepi/story/sc-210124)
- **Env:** ``

### Checklist:
- [ ] I added labels to my PR


src/frontend/.eslintrc.js
```diff
@@ -11,7 +11,6 @@ module.exports = {
     "plugin:react/recommended",
     "plugin:@typescript-eslint/recommended",
     "plugin:sonarjs/recommended",
-    "plugin:jest-playwright/recommended",
```
mmmmmmaya, line 14: why did you remove this linter plugin?

src/frontend/package.json
```diff
@@ -10,20 +10,22 @@
...
   "dependencies": {
+    "@babel/eslint-parser": "^7.18.9",
     "@emotion/css": "^11.1.3",
     "@emotion/react": "^11.1.5",
     "@emotion/styled": "^11.3.0",
     "@material-ui/core": "^4.11.3",
     "@material-ui/icons": "^4.11.3",
     "@material-ui/lab": "^4.0.0-alpha.57",
     "@material-ui/styles": "^4.11.3",
+    "@redux-devtools/extension": "^3.2.3",
```
mmmmmmaya, line 28: I think this can be a dev dependency right? Don't think we need to deploy it with our app.

mmmmmmaya: can this be a devdependency?

charles-testco: I only updated redux-devtools-extensions to @redux-devtools/extension because it has been migrated and the former is no longer maintained but I can move to devDependency as well

src/frontend/.eslintrc.js
charles-testco: It has a dependency on jest-playwright runner which we are trying to switch-off

mmmmmmaya (approved):
would love to see work on an rdev so we can check for app errors with changesets

charles-testco:
@phoenixAja Luis says this PR does not build locally. Would be nice to get this out of the way so the session management PR can be progressed. Do you think you can have look. Sorry for dumping this on you but I am still struggling with setting up local env.

phoenixAja:
Yeah sure i can check it out

PR #1287, opened 2022-08-17
add gisaid prefixes to samples on trees

jgadling:
### Summary:
- **What:** Add gisaid prefix to trees
- **Ticket:** [sc209810](https://app.shortcut.com/genepi/story/209810)
- **Env:** https://gisaidids-frontend.dev.czgenepi.org

### Notes:
This PR updates our tree & metadata download endpoints so that:
- regardless of whether or not the tree json in s3 has 'hCoV-19/' as a prefix for any given sample, *all* public identifiers in the tree will be updated to include the prefix before it gets sent back to the user.
- We try to match samples on a tree to samples in our db both *with* and *without* the 'hCoV-19/' prefix.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/utils/phylo.py
+++ b/src/backend/aspen/api/utils/phylo.py
@@ -54,13 +54,30 @@ def _rename_nodes_on_tree(
     """Given a tree, a mapping of identifiers to their replacements, rename the nodes on
     the tree.  If `save_key` is provided, then the original identifier is saved using
     that as the key."""
-    name = node["name"]
-    renamed_value = name_map.get(name, None)
+    gisaid_prefix = "hCoV-19/"
+
+    # The mixed situations we're dealing with here:
+    #  - The public identifiers in our database *do not* have gisaid prefixes on them
+    #  - The samples on a tree *sometimes* do and sometimes don't have gisaid prefixes on them.
+    #  - At the end of this method, we want gisaid prefixes added to *all* tree samples.
+    # So this means we have to:
+    #  - Strip gisaid prefixes from tree nodes before trying to match them to db samples
+    #  - Add gisaid prefixes to all public tree identifiers if they aren't already prefixed
+
+    # Strip the gisaid prefix from our tree identifier if it's present
+    tree_identifier = node["name"]
+    if tree_identifier.lower().startswith(gisaid_prefix.lower()):
+        tree_identifier = tree_identifier[len(gisaid_prefix) :]
+
+    renamed_value = name_map.get(tree_identifier, None)
+
+    # Make sure we have the gisaid prefix on this node when we output it.
+    node["name"] = f"{gisaid_prefix}{tree_identifier}"
     if renamed_value is not None:
         # we found the replacement value! first, save the old value if the caller
         # requested.
         if save_key is not None:
-            node[save_key] = name
+            node[save_key] = node["name"]
         node["name"] = renamed_value
     for child in node.get("children", []):
         _rename_nodes_on_tree(child, name_map, save_key)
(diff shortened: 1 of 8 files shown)


lvreynoso (approved):

danrlu:
Thanks very much Lucia for checking the code! I combined it with my GISAID PR and poked around and noticed that tree json has "samples" with name starting with `NODE_` that are not real samples (they were added to the tree to represent the most recent common ancestors MRCA, we didn't really sequence them but we infer them to exist). 
image

B/c they are not real samples, (1) we don't want to rename them and (2) for some reason, this PR puts them into the sample ID download, which didn't use to be the case and is not desired. 

Side note I will modify my GISAID PR to let CZGE add the prefix instead of having Nextstrain to do it. I like the current set up here!

phoenixAja:
@danrlu do we need to address the NODE_ naming before merging in? or can we merge this and do a follow up PR?

danrlu:
Hehe I love EST working hour ;) That is a blocker, we need to fix it first

src/backend/aspen/api/utils/phylo.py
```diff
@@ -54,13 +54,31 @@ def _rename_nodes_on_tree(
...
+    # So this means we have to:
+    #  - Strip gisaid prefixes from tree nodes before trying to match them to db samples
+    #  - Add gisaid prefixes to all public tree identifiers if they aren't already prefixed
+
+    # Strip the gisaid prefix from our tree identifier if it's present
+    tree_identifier = node["name"]
+    if tree_identifier.lower().startswith(gisaid_prefix.lower()):
+        tree_identifier = tree_identifier[len(gisaid_prefix) :]
+
+    renamed_value = name_map.get(tree_identifier, None)
```
phoenixAja, line 73: when does this return None? Just curious

src/backend/aspen/api/utils/phylo.py
```diff
@@ -54,13 +54,31 @@ def _rename_nodes_on_tree(
...
+
+    # Make sure we have the gisaid prefix on this node when we output it.
+    node["name"] = f"{gisaid_prefix}{tree_identifier}"
```
phoenixAja: can we just add a check here to fix the NODE renaming issue? 

ex:  `node["name"] =f"{gisaid_prefix}{tree_identifier}" if not tree_identifier.startswith("NODE_") else tree_identifier`

src/backend/aspen/api/views/phylo_trees.py
```diff
@@ -66,9 +66,12 @@ async def _get_selected_samples(db: AsyncSession, phylo_tree_id: int):
     # identifier check here since the process_phylo_tree method already does that
     # filtering, and this data is only used to match any identifiers that are
     # *already* on the tree.
+    gisaid_prefix = "hCoV-19/"
     for uploaded_pathogen_genome in phylo_run.inputs:
         sample = uploaded_pathogen_genome.sample
-        selected_samples.add(prefix_regex.sub("", sample.public_identifier))
+        stripped_identifier = prefix_regex.sub("", sample.public_identifier)
+        selected_samples.add(stripped_identifier)
+        selected_samples.add(f"{gisaid_prefix}{stripped_identifier}")
```
phoenixAja, line 74: so we're adding the idenitifers twice to the set of selected samples? Once with the gisaid_prefix stripped and then once more with the prefix added?

danrlu: it looks to me that `selected_samples` here are what samples the users selected, that will be later used to mark them accordingly in the sample id download, so here we include their public ID with or without gisaid_prefix and private ID (see line 75) to be comprehensive so regardless which ID we will use later for matching, it will not be missed. I think...

src/backend/aspen/api/utils/phylo.py
danrlu: cool this is what i'm looking for. let me plug this in and test out 😉

PR #1291, opened 2022-08-23
list pathogens

phoenixAja:
### Summary:
- **What:** list pathogens endpoints
- **Ticket:** [sc211438](https://app.shortcut.com/genepi/story/211438)
- **Env:** `<rdev link>`

### Demos:
response looks like: 

`{"pathogens":[{"id":1,"slug":"SC2","name":"SARS-CoV-2"}]}`
### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/views/pathogens.py
+++ b/src/backend/aspen/api/views/pathogens.py
@@ -17,7 +17,5 @@ async def list_pathogens(
 ) -> PathogensResponse:
     query = sa.select(Pathogen)
     results = await db.execute(query)
-    t = []
-    for row in results.scalars():
-        t.append(PathogenResponse.from_orm(row))
-    return PathogensResponse.parse_obj({"pathogens": t})
+    pathogens = [PathogenResponse.from_orm(row) for row in results.scalars()]
+    return PathogensResponse.parse_obj({"pathogens": pathogens})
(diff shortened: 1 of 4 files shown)


vincent-czi (approved):
Looks good!

src/backend/aspen/api/views/pathogens.py
```diff
@@ -0,0 +1,21 @@
...
+
+@router.get("/", response_model=PathogensResponse)
+async def list_pathogens(
+    db: AsyncSession = Depends(get_db),
+    user: User = Depends(get_auth_user),
+) -> PathogensResponse:
+    query = sa.select(Pathogen)
+    results = await db.execute(query)
+    pathogens = [PathogenResponse.from_orm(row) for row in results.scalars()]
+    return PathogensResponse.parse_obj({"pathogens": pathogens})
```
vincent-czi: This isn't even a nit, but I just wish we didn't need the `from_orm` and `parse_obj` boilerplate like this for so many of our responses. We have it in a bunch of places so I think it's necessary, but it feels like the serializer/schema ought to be able to handle this kind of interpretation without needing it explicitly called. Since we specify the `response_model`, it feels to me like the response model should be able to destructure whatever it gets handed based on the various attributes in the model and either send off the serialized response or raise an exception appropriately. In my perfect world, the last two lines of code would just look like
```
pathogens = [row for row in results.scalars()]
return pathogens
```

Anyway, I'm not actually requesting a change, just carping.

phoenixAja: no makes sense, i can try it

lvreynoso (approved):

PR #1292, opened 2022-08-24
Maya/194962 194961 gisaid genbank api

mmmmmmaya:
### Summary
- **What:** Genericize file downloads from the samples table
- **Why:** Support for gisaid and genbank templates
- **Ticket:**
  - [[sc-194961]](https://app.shortcut.com/genepi/story/194961)
  - [[sc-194962]](https://app.shortcut.com/genepi/story/194962)
- **Env:** https://maya-api-frontend.dev.czgenepi.org/

### Testing
1. Try downloading fasta and metadata files to make sure they still work
1. Gisaid and genbank downloads don't actually return anything yet

### Demos
No visual changes.

### Notes
This will need a small amount of work once the BE endpoints are in place to handle the case where multiple files need to be downloaded for GISAID/Genbank templates. I don't know how to handle that without having endpoint access. I think this is ok to merge, since the fasta and metadata downloads both still work as expected, and the other two options are behind a feature flag.

### Checklist
- [X] I merged latest `trunk`
- [x] I added labels to my PR


src/frontend/src/common/analytics/eventTypes.ts
```diff
@@ -169,6 +169,10 @@ export type AnalyticsSamplesDownloadFile = {
   sample_public_ids: JsonString;
   // User downloaded info on consensus genome (FASTA) for these samples
   includes_consensus_genome: boolean;
+  // User downloaded genbank template for these samples
```
mmmmmmaya, line 172: @vincent-czi would like to make sure this is an ok change/what I should do

vincent-czi: Perfect: this change and the other one to pass te data looks great!

src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
```diff
@@ -244,8 +255,11 @@ const DownloadModal = ({
                     <DownloadTypeInfo>
                       Download concatenated consensus genomes and metadata files
                       formatted to prepare samples for submission to GISAID.{" "}
-                      {/* TODO: (194961) - update href */}
-                      <Link href="" target="_blank" rel="noreferrer">
+                      <Link
+                        href="https://help.czgenepi.org/hc/en-us/articles/8179880474260"
```
ehoops-cz, line 259: 🙇‍♀️

src/frontend/src/common/queries/samples.ts
```diff
@@ -24,42 +24,43 @@ import { MutationCallbacks } from "./types";
 /**
  * Download fasta file for samples
  */
-interface SampleFastaDownloadPayload {
+
+type FileDownloadEndpointType =
```
ehoops-cz, line 28: nit: do these names need 'type' in them?  Seems redundant.

ehoops-cz (approved):
Looks good!

src/frontend/src/common/queries/samples.ts
mmmmmmaya: I prefer for types and interfaces to end with Type so it's clear which vars are types and which aren't. It's a pattern we have throughout the app

PR #1299, opened 2022-08-26
avoid console warning about props forwarded to div

mmmmmmaya:
Remove this warning from console.

Screen Shot 2022-08-25 at 8 32 25 PM


ehoops-cz (approved):
🎉

PR #1300, opened 2022-08-26
Maya/206011 group pathogen in url

mmmmmmaya:
### Summary
- **What:** Add groupId and pathogen to FE routing and api requests
- **Why:** ✨ decovidify ✨ 
- **Ticket:**
  - [[sc-206011]](https://app.shortcut.com/genepi/story/206011)
  - [[sc-211258]](https://app.shortcut.com/genepi/story/211258)
- **Env:** https://maya-routing-frontend.dev.czgenepi.org/
- **Depends on:** https://github.com/chanzuckerberg/czgenepi/pull/1290 (api requests using the new structure will fail until that PR is merged in)

### Expected Behavior
- Visit a "workspace" page (uploads, samples table, trees table) without additional params --> groupId and pathogen should be added for you
- Try to visit a url that links to a group you don't have access to --> redirects you to the same page in a group you _do_ have access to
- Try to visit a url that links to a pathogen that doesn't exist --> redirects you to the same page for covid
- The additional params must come after all the path token (for example, on the samples page, the path need to start with `/data/samples`), but besides that they can come in any order
- Visiting a page that doesn't require group info should remove the addition info from the url
  - Main thing to note is that some pages (privacy, for example) don't accept params and will 404 if you try to add them. I think this is sensible, since you shouldn't be able to get to those parts of the app with the params in place. We could also, optionally, just add `[[...params]]` to every page, so they will still render without a 404, even if you _try_ to visit them with those params, but it feels like an unnecessary optimization at the moment.

### Testing
Literally just do everything you can to break this. 
Visit parts of the app by clicking buttons and links.
Visit parts of the app by adding groupId and pathogen in the url where they aren't expected.
Randomly add these params in any order.

### Notes
Sorry this PR is kinda all over the place.

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/frontend/src/common/redux/utils/pathogenUtils.ts
```diff
@@ -0,0 +1,43 @@
...
+  getLocalStorage,
+  isWindowDefined,
+} from "src/common/utils/localStorage";
+import { store } from "..";
+import { setPathogen } from "../actions";
+import { selectCurrentPathogen } from "../selectors";
+import { Pathogen, ReduxPersistenceTokens } from "../types";
+
+export const isValidPathogen = (pathogen: string): boolean => {
+  // make TS happy enough to search Pthogen enum for arbitrary strings
```
phoenixAja, line 11:
```suggestion
  // make TS happy enough to search Pathogen enum for arbitrary strings
```

src/frontend/src/common/api/index.tsx
```diff
@@ -26,8 +27,12 @@ export enum ORG_API {
 }
 
 export const generateOrgSpecificUrl = (path: ORG_API): string => {
```
phoenixAja, line 29: should we rename to `generateOrgPathogenSpecificUrl`? or may `generateContextSpecificUrl`?

phoenixAja: also should maybe consider renaming `ORG_API` to `ORG_PATHOGEN_API`, but that may be a little out of scope for this

phoenixAja (approved):

PR #1304, opened 2022-08-30
[SC-211260] update strings and variable names in the FE

ehoops-cz:
### Summary:
- **What:** `Typescript type for strings.  Example strings.ts file`
- **Ticket:** [sc211260](https://app.shortcut.com/genepi/story/211260/update-strings-and-variable-names-in-the-frontend-to-support-multiple-pathogens)
- **Env:** `none`

### Demos:
Screen Shot 2022-08-30 at 8 24 54 AM
Screen Shot 2022-08-30 at 8 25 36 AM

Screen Shot 2022-08-30 at 8 55 25 AM

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/types/strings.ts
+++ b/src/frontend/src/common/types/strings.ts
@@ -1,0 +1,3 @@
+import { Pathogen } from "src/common/redux/types";
+
+export type StringsType<T> = Record<Required<Pathogen>, T>;
--- a/src/frontend/src/views/Upload/components/Samples/index.tsx
+++ b/src/frontend/src/views/Upload/components/Samples/index.tsx
@@ -16,6 +16,7 @@ import {
 import { ParseErrors, Props } from "../common/types";
 import AlertTable from "./components/AlertTable";
 import Table from "./components/Table";
+import { strings } from "./strings";
 import {
   ContentWrapper,
   SemiBold,
@@ -88,7 +89,7 @@ export default function Samples({ samples, setSamples }: Props): JSX.Element {
       <Content>
         <CollapsibleInstructions
           buttonSize="xxs"
-          header="Select SARS-CoV-2 Consensus Genome Files"
+          header={strings.covid.header}
           headerSize="xl"
           instructionListTitle="File instructions"
           listPadding="xl"
--- a/src/frontend/src/views/Upload/components/Samples/index.tsx
+++ b/src/frontend/src/views/Upload/components/Samples/index.tsx
@@ -101,10 +101,7 @@ export default function Samples({ samples, setSamples }: Props): JSX.Element {
               </SemiBold>{" "}
               Your sample name should be the sample&apos;s Private ID.
             </span>,
-            <span key="2">
-              Accepted file formats: fasta (.fa or .fasta), fasta.gz (.fa.gz),
-              fasta.zip, plain text (.txt)
-            </span>,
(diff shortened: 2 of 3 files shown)


src/frontend/src/common/types/strings.ts
```diff
@@ -0,0 +1,3 @@
+import { Pathogen } from "src/common/redux/types";
+
+export type StringsType<T> = Record<Required<Pathogen>, T>;
```
vincent-czi, line 3: I don't love `StringsType` as the name: it feels too general, like it could encompass too much. I'm struggling to come up with something better though that isn't a mouthful: `PerPathogenStrings` seems a little awkward. This type is specifically about having a structure for doing lookups of user-facing strings for each kind of pathogen we support. I would like a name that gets that across a bit more.

src/frontend/src/views/Upload/components/Samples/index.tsx
```diff
@@ -16,6 +16,7 @@ import {
 import { ParseErrors, Props } from "../common/types";
 import AlertTable from "./components/AlertTable";
 import Table from "./components/Table";
+import { strings } from "./strings";
```
vincent-czi: Similar to my other comment about the name `StringsType`: if our convention is to just call our per-pathogen set of strings the name `strings`, I think that is a bit vague. We might have other constant strings, possibly also being looked up on a different basis (eg, we look up different chunks of copywriting depending on if a user is a group admin or just a member). I think `strings.ts` is fine for the file: having a consistent bucket of strings adjacent to each component seems good. It's just the specific name of `strings` as the var we export for per-pathogen wording.

src/frontend/src/views/Upload/components/Samples/strings.ts
```diff
@@ -0,0 +1,14 @@
+import { StringsType } from "src/common/types/strings";
+
+interface SamplesStrings {
```
vincent-czi: Nit: I think this would be better as `UploadSamplesStrings`. To me, these strings are as much about the upload aspect as the samples aspect.

vincent-czi (approved):
Engineering-wise, I'm perfectly happy with this approach, looks good. I do have some thoughts about the naming though. I think some of the stuff would be nice to be named a bit more explicitly/specifically.

That said, this might be just a convention thing where we'll all get used to it once there's a `strings` next to every component and we'll all know what it means. If you think strongly it's the right way to go, I'm fine with it as is.

src/frontend/src/views/Upload/components/Samples/index.tsx
ehoops-cz: Modified to name the variable `pathogenStrings`

src/frontend/src/views/Upload/components/Samples/strings.ts
ehoops-cz: I used the component name, but I agree - especially since we have a nearly identical UI but different codepath for the edit flow.

ehoops-cz:
Addressed feedback here and will merge this.  Follow up PRs to come for renaming related to lineage caller and public data repository updates.

src/frontend/src/common/types/strings.ts
ehoops-cz: missed this comment previously.  Will address in one of the follow up PRs.

PR #1305, opened 2022-08-30
pathogen field support for PhyloRun, Sample, PathogenGenome tables

phoenixAja:
### Summary:
- **What:** add pathogen foreign keys to Sample, PthogenGenome, and PhyloRun tables
- **Ticket:** [sc211473](https://app.shortcut.com/genepi/story/211473)
- **Env:** `<rdev link>`

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/backend/database_migrations/versions/20220830_171547_add_pathogen_field.py
+++ b/src/backend/database_migrations/versions/20220830_171547_add_pathogen_field.py
@@ -64,7 +64,7 @@ def upgrade():
     # backpopulate pathogen fields to sars-cov-2
     op.execute(
         """
-        UPDATE aspen.samples 
+        UPDATE aspen.samples
         SET pathogen_id=(
             SELECT id from aspen.pathogens WHERE name='SARS-CoV-2'
         )
@@ -72,7 +72,7 @@ def upgrade():
     )
     op.execute(
         """
-        UPDATE aspen.phylo_runs 
+        UPDATE aspen.phylo_runs
         SET pathogen_id=(
             SELECT id from aspen.pathogens WHERE name='SARS-CoV-2'
         )
@@ -80,7 +80,7 @@ def upgrade():
     )
     op.execute(
         """
-        UPDATE aspen.pathogen_genomes 
+        UPDATE aspen.pathogen_genomes
         SET pathogen_id=(
             SELECT id from aspen.pathogens WHERE name='SARS-CoV-2'
         )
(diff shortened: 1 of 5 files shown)


src/backend/database_migrations/versions/20220830_171547_add_pathogen_field.py
```diff
@@ -0,0 +1,92 @@
...
+revision = "20220830_171547"
+down_revision = "20220816_000535"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+
+    op.add_column(
+        "pathogen_genomes",
```
jgadling, line 20: I think our data model had pathogen_id added to samples, I don't think we need it added to pathogen_genomes as well.

src/backend/database_migrations/versions/20220830_171547_add_pathogen_field.py
```diff
@@ -0,0 +1,92 @@
...
+        "pathogens",
+        ["pathogen_id"],
+        ["id"],
+        source_schema="aspen",
+        referent_schema="aspen",
+    )
+
+    op.add_column(
+        "phylo_runs",
```
jgadling, line 35: We also need to add a pathogen_id to phylo_trees

jgadling (changes requested):
We should merge & deploy migrations before updating our SA models to use the new columns, or else we'll cause errors while upgrading our app.

vincent-czi:
> We should merge & deploy migrations before updating our SA models to use the new columns, or else we'll cause errors while upgrading our app.

I agree with this, but how do we handle the Prod aspect? Would we make a separate PR of just the migrations, merge that to Staging, verify things didn't break in Staging and the migration succeeded, then do a Prod push before coming back with the PR to handle the SA models and rest of the changes? Or are you thinking something different?

jgadling:
> I agree with this, but how do we handle the Prod aspect? Would we make a separate PR of just the migrations, merge that to Staging, verify things didn't break in Staging and the migration succeeded, then do a Prod push before coming back with the PR to handle the SA models and rest of the changes? Or are you thinking something different?

Yep, we can deploy to prod once we've verified that staging looks good, and then open up a PR for other code changes that use new fields.

phoenixAja:
ok i'll close this and reopen the migration/ sqlalchemy support separately

PR #1306, opened 2022-08-30
[SC-211260] NextStrain Modal strings

ehoops-cz:
### Summary:
- **Ticket:** [sc211260](https://app.shortcut.com/genepi/story/211260/update-strings-and-variable-names-in-the-frontend-to-support-multiple-pathogens)
- **Env:** `none`

### Demos:
Screen Shot 2022-08-30 at 1 53 04 PM
Screen Shot 2022-08-30 at 1 52 57 PM
Screen Shot 2022-08-30 at 1 52 48 PM
Screen Shot 2022-08-30 at 1 52 20 PM

### Notes:
I had to copy/paste the strings, so it would be helpful to have a check that I didn't copy/pasta.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/RadioLabel/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/RadioLabel/index.tsx
@@ -1,7 +1,10 @@
 // eslint-disable @typescript-eslint/explicit-member-accessibility
 import ListItemText from "@mui/material/ListItemText";
 import { Icon, List } from "czifui";
+import { useSelector } from "react-redux";
+import { selectCurrentPathogen } from "src/common/redux/selectors";
 import { SampleFiltering } from "../SampleFiltering";
+import { pathogenStrings } from "./strings";
 import {
   Label,
   LabelMain,
@@ -36,15 +39,20 @@ export const RadioLabelOverview = ({
   setStartDate,
   setEndDate,
 }: TreeChoiceWithFilteringProps): JSX.Element => {
+  const pathogen = useSelector(selectCurrentPathogen);
+  const {
+    overviewDescription = "",
+    overviewBestFor = "",
+    overviewGoodFor1 = "",
+    overviewGoodFor2 = "",
+  } = pathogen ? pathogenStrings[pathogen] : {};
+
   return (
     <div>
       <Label>
         <LabelMain>Overview </LabelMain>
       </Label>
-      <SmallText>
-        Includes samples from both within and outside of your jurisdiction, at a
-        ratio of roughly 2:1.
-      </SmallText>
+      <SmallText>{overviewDescription}</SmallText>
       {selected && (
         <>
           <StyledList>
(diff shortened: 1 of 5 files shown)


src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/RadioLabel/index.tsx
```diff
@@ -36,15 +39,20 @@ export const RadioLabelOverview = ({
   setStartDate,
   setEndDate,
 }: TreeChoiceWithFilteringProps): JSX.Element => {
+  const pathogen = useSelector(selectCurrentPathogen);
+  const {
+    overviewDescription = "",
+    overviewBestFor = "",
+    overviewGoodFor1 = "",
+    overviewGoodFor2 = "",
+  } = pathogen ? pathogenStrings[pathogen] : {};
```
ehoops-cz, line 48: If pathogen is null, use an empty object to avoid errors related to trying to look up `null[someString]`

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/RadioLabel/strings.ts
```diff
@@ -0,0 +1,41 @@
...
+
+interface RadioLabelStrings {
+  nonContextualizedBestFor: string;
+  nonContextualizedDescription: string;
+  nonContextualizedGoodFor: string;
+  nonContextualizedNotRecommended: string;
+  overviewBestFor: string;
+  overviewDescription: string;
+  overviewGoodFor1: string;
+  overviewGoodFor2: string;
```
vincent-czi, line 12: Overall I'm happy with this approach, but I wonder how we'll handle things like if a certain pathogen has 3 bullet points or just 1 bullet point. If the various options start to have significantly divergent meanings for different pathogens, I don't know how standardized the interface will really be. I'm not trying to offer a solution or actually think this is important to address now, just something to consider in case it does come up in the hypothetical future.

vincent-czi (approved):

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/RadioLabel/strings.ts
ehoops-cz: I agree with the concern.  

Hopefully by keeping the strings directly with the component it will help.  Maybe we could do something like use a subset of the pathogens for the type rather than all of them so it would only require the pathogens that have that feature.

Another option would be to use a nested object rather than just a string and then each pathogen could define what is in the object.  So every pathogen would be required to have an object, but we wouldn't enforce the exact shape. 
 ie:
```
overviewGoodFor: {
   one: "some thing",
   two: "some other thing",
}
```

Janeece (approved):
LGTM for the screenshots

ehoops-cz:
[] fix style

ehoops-cz:
Going ahead with this merge.  We should re-evaluate the structure of the RadioLabelStrings interface as we go.  Likely this will be more clear when we update the nextstrain tree creation modal to support a second pathogen.

PR #1310, opened 2022-08-31
model code changes for pathogen support

phoenixAja:
### Summary:
- **What:** add sql alchemy model support for pathogen_id
- **Ticket:** [sc211473](https://app.shortcut.com/genepi/story/211473)
- **Env:** `<rdev link>`

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/backend/aspen/database/models/pathogens.py
```diff
@@ -27,6 +27,10 @@ class Pathogen(idbase):  # type: ignore
         comment=("full pathogen abbreviated name, ex: SARS-CoV-2"),
     )
 
+    samples = relationship("Sample", back_populates="pathogen")
+    phylo_runs = relationship("PhyloRun", back_populates="pathogen")
```
jgadling: I don't think we need to set up these relationships in both directions? Just setting up the relationships in the Samples/Trees/Runs tables should make these backrefs work.

src/backend/aspen/database/models/phylo_tree.py
```diff
@@ -61,6 +62,11 @@ class PhyloTree(Entity):
     s3_key = Column(String, nullable=False)
     name = Column(String, nullable=True)
 
+    pathogen_id = Column(
+        Integer, ForeignKey(Pathogen.id)
+    )  # TODO: change to nullable=False once we update workflows
+    pathogen = relationship(Pathogen, back_populates="phylo_trees")
```
jgadling: most of the relationships we set up in SA rely on a `foreign_keys` parameter. Do we not need those here?

src/backend/aspen/database/models/pathogens.py
phoenixAja: i thought it might be nice to setup this so that we could pull all the related samples, phylotrees, phyloruns for a given pathogen easily? For example I feel like Pathogen.samples might be nice and make that query easy?

src/backend/aspen/database/models/phylo_tree.py
phoenixAja: i thought we would only need that if we had multiple foreign keys to the Pathogen table

src/backend/aspen/database/models/pathogens.py
jgadling: Yeah, but that's handled by the `back_populates` parameter of the relationship in the other models!

jgadling: errr  --- maybe it's backref:
https://docs.sqlalchemy.org/en/14/orm/relationship_api.html#sqlalchemy.orm.relationship.params.backref

jgadling: OK, TIL! 

We've [been using `backref`](https://docs.sqlalchemy.org/en/14/orm/relationship_api.html#sqlalchemy.orm.relationship.params.backref) in most of our relationships, which populates relationships in both directions. It's super convenient, but the docs say that current best practice is to use `back_populates` in both directions since it plays nicer with type checking. 

This looks good then, I think we'll have to start migrating our old `backref` parameters at some point.

jgadling (approved):

src/backend/aspen/database/models/pathogens.py
phoenixAja: it's almost like i read the docs or something 😆 (i did read the docs, i had to refresh my memory on a lot of this stuff!)

jgadling: lol yeah your docs reading was way more up to date than mine was!! 🥇 😍

PR #1311, opened 2022-08-31
[SC-211770] bug - TOS requires multiple clicks to close

ehoops-cz:
### Summary:
- **What:** `Use multiple invalidateQueries calls. Add icon loading to button.`
- **Ticket:** [sc211770](https://app.shortcut.com/genepi/story/211770/agree-to-terms-of-service-requires-double-agree-click)
- **Env:** `<rdev link>`

### Demos:
TOS-one-click

### Notes:
This bug looked like an issue with the Terms of Service page.  However, the actual issue was that TOS redirected to the samples page and since the user data was not properly invalidated, the user was then redirected back to the TOS page.  This redirect loop continued until the new user data was fetched.  Invalidating (and refetching) the data fixed the issue.

The click looks a little slower now because it is actually invalidating and refetching data.  I added a spinner to the button so the user knows we're working on it.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/queries/auth.ts
+++ b/src/frontend/src/common/queries/auth.ts
@@ -16,11 +16,7 @@ import { selectCurrentGroup } from "../redux/selectors";
 import { ensureValidGroup } from "../redux/utils/groupUtils";
 import { ROUTES } from "../routes";
 import { ENTITIES } from "./entities";
-import {
-  USE_GROUP_INFO,
-  USE_GROUP_INVITATION_INFO,
-  USE_GROUP_MEMBER_INFO,
-} from "./groups";
+import { GROUP_QUERY_PREFIX } from "./groups";
 
 const { API_URL } = ENV;
 
@@ -75,12 +71,10 @@ export function useUpdateUserInfo(): UseMutationResult<
 
   return useMutation(updateUserInfo, {
     onSuccess: async () => {
-      await queryClient.invalidateQueries([
-        USE_USER_INFO,
-        USE_GROUP_INFO,
-        USE_GROUP_INVITATION_INFO,
-        USE_GROUP_MEMBER_INFO,
-      ]);
+      // invalidateQueries only works for multiple queries if they start with the same key
+      // https://github.com/TanStack/query/discussions/2057#discussioncomment-549333
+      await queryClient.invalidateQueries([USE_USER_INFO]);
+      await queryClient.invalidateQueries([GROUP_QUERY_PREFIX]);
     },
   });
 }
--- a/src/frontend/src/common/queries/auth.ts
+++ b/src/frontend/src/common/queries/auth.ts
@@ -16,7 +16,7 @@ import { selectCurrentGroup } from "../redux/selectors";
 import { ensureValidGroup } from "../redux/utils/groupUtils";
 import { ROUTES } from "../routes";
 import { ENTITIES } from "./entities";
-import { GROUP_QUERY_PREFIX } from "./groups";
+import { GROUP_QUERY_ID_PREFIX } from "./groups";
 
 const { API_URL } = ENV;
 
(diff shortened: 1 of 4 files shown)


src/frontend/src/common/queries/groups.ts
```diff
@@ -77,6 +77,12 @@ export const mapGroupMemberData = (obj: RawGroupMemberRequest): GroupMember => {
   };
 };
 
+/**
+ * starting all group-related queries with "group" allows us to invalidate
+ * all of the queries in a single call to queryClient.invalidateQueries(["group"])
+ */
+export const GROUP_QUERY_PREFIX = "group";
+
```
ehoops-cz, line 85: Alternatively we could call multiple times.  We await each one though, so this is a little faster.

src/frontend/src/views/AgreeTerms/style.tsx
```diff
@@ -20,3 +20,13 @@ export const Details = styled.p`
 export const SpacedBold = styled.b`
   padding: 0 0.3em;
 `;
+
+export const StyledIcon = styled(Icon)`
+  ${(props) => {
+    const colors = getColors(props);
+
+    return `
+      fill: ${colors?.gray[500]};
```
ehoops-cz: @Janeece I'm currently using gray[500] for the spinner, should it be lighter?

Janeece: SDS spec says it should be gray[400] because its a disabled state button with an icon. We should go with that to be consistent.

Janeece:
I love the addition of the loading state! Great idea.

src/frontend/src/common/queries/groups.ts
```diff
@@ -77,6 +77,12 @@ export const mapGroupMemberData = (obj: RawGroupMemberRequest): GroupMember => {
   };
 };
 
+/**
+ * starting all group-related queries with "group" allows us to invalidate
+ * all of the queries in a single call to queryClient.invalidateQueries(["group"])
+ */
+export const GROUP_QUERY_PREFIX = "group";
```
vincent-czi: Is this not case-sensitive? All the group queries I see start with `GROUP` for the value, eg
```
  GROUP_INVITATION_INFO = "GROUP_INVITATION_INFO",
```
... oh, wait, I see that `id` is `groupInvitationInfo`. It's the `id` that matters for invalidation, not `entities` then? Maybe call it `GROUP_QUERY_ID_PREFIX`? I dunno, maybe that's silly and would be more obvious if I knew react-query better.

vincent-czi: Regardless, I think if we're depending on `GROUP_QUERY_PREFIX` matching other values, it's better if we can force that in code rather than by convention. So we could change stuff to
```
id: `${GROUP_QUERY_PREFIX}InvitationInfo`,
```

src/frontend/src/common/queries/auth.ts
```diff
@@ -75,12 +71,10 @@ export function useUpdateUserInfo(): UseMutationResult<
...
-      await queryClient.invalidateQueries([
-        USE_USER_INFO,
-        USE_GROUP_INFO,
-        USE_GROUP_INVITATION_INFO,
-        USE_GROUP_MEMBER_INFO,
-      ]);
+      // invalidateQueries only works for multiple queries if they start with the same key
+      // https://github.com/TanStack/query/discussions/2057#discussioncomment-549333
+      await queryClient.invalidateQueries([USE_USER_INFO]);
+      await queryClient.invalidateQueries([GROUP_QUERY_PREFIX]);
```
vincent-czi: I like this, but it seems wacky to me that react-query invalidates all queries that match based on prefix. I was assuming the invalidation was based on string `===` equality, but it's... prefix matching? Okay, yeah, I see it in the docs now. What a wacky API. Seems like this is the way to go, then!

vincent-czi (approved):
Nice!

src/frontend/src/common/queries/auth.ts
ehoops-cz: ya, string equality seems much more straightforward, but here we are 🤷‍♀️

src/frontend/src/common/queries/groups.ts
ehoops-cz: renaming to be as clear as possible seems good.  I doubt we'll ever have "react-query experts" 😅

PR #1313, opened 2022-09-01
Prevent broken tree builds in the tree list UI

jgadling:
### Summary:
- **What:** Don't create new phylo run rows for swipe retries
- **Ticket:** [sc212405](https://app.shortcut.com/genepi/story/212405)
- **Env:** https://fixswipe-frontend.dev.czgenepi.org/

### Notes:
This PR does a few things:
- Reduces the amount of memory allocated to most batch jobs. Smaller instances are typically less likely to be interrupted.
- Our scheduled-job-runner script now creates phylo run rows in the DB, and then kicks off swipe jobs the same way we do for **on-demand** nextstrain runs.
- Clean up old dead code now that there's only one way to kick off nextstrain jobs.

I've manually tested that on-demand pangolin runs still work (I uploaded 2 samples and they had populated lineages ~3min later:)
Screen Shot 2022-09-01 at 9 31 27 AM

And I've also checked that the nextstrain autorun script is able to launch tree build jobs without any obvious problems -- this job failed due to missing samples in the dev environment.
Screen Shot 2022-09-01 at 9 31 54 AM

I'll kick off an auto-run job once this change is in staging to validate that nightly tree builds are still happy.

### Checklist:
- [x] I added labels to my PR


src/backend/aspen/api/settings.py
```diff
@@ -270,3 +226,58 @@ def customise_sources(
...
+
+    @property
+    def AUTH0_SERVER_METADATA_URL(self) -> str:
+        if self.SECRET_AUTH0_SERVER_METADATA_URL:
+            return self.SECRET_AUTH0_SERVER_METADATA_URL
+        else:
+            return f"https://{self.AUTH0_DOMAIN}/.well-known/openid-configuration"
+
+
+class CLISettings(Settings):
```
jgadling, line 282: The API service requires several environment variables to be set, that we currently don't (and don't have any reason to) send to batch jobs that run CLI commands. The refactoring of Settings / APISettings / CLISettings is just to get us to the point where API and CLI settings work **similarly** but with a few differences.

src/backend/aspen/api/views/phylo_runs.py
```diff
@@ -151,43 +148,8 @@ async def kick_off_phylo_run(
     await db.commit()
 
     # Step 5 - Kick off the phylo run job.
-    aws_region = os.environ.get("AWS_REGION")
```
jgadling, line 154: This stuff got moved to swipe.py

src/backend/aspen/api/views/samples.py
```diff
@@ -238,48 +235,11 @@ async def validate_ids(
     return ValidateIDsResponse(missing_sample_ids=missing_sample_ids)
 
 
-# TODO this should be a general swipe calling library instead of lame copy-pasta.
-def _kick_off_pangolin(group_prefix: str, sample_ids: Sequence[str], settings):
```
jgadling, line 242: This stuff got moved to swipe.py

src/backend/aspen/cli/db.py
```diff
@@ -130,197 +106,3 @@ def interact(ctx, connect):
 
     shell = InteractiveShellEmbed()
     shell()
-
-
-@db.command("add-can-see")
```
jgadling, line 135: We weren't using this, cleaning up dead code.

src/backend/aspen/cli/db.py
```diff
@@ -130,197 +106,3 @@ def interact(ctx, connect):
...
-                    CanSee.__table__.c.owner_group_id.name: owner_group.id,
-                    CanSee.__table__.c.viewer_group_id.name: viewer_group.id,
-                    CanSee.__table__.c.data_type.name: datatype,
-                }
-            )
-            .on_conflict_do_nothing()
-        )
-
-
-@db.command("create-phylo-run")
```
jgadling, line 184: This functionality has been moved to our nextstrain-autorun script

src/backend/aspen/cli/db.py
```diff
@@ -130,197 +106,3 @@ def interact(ctx, connect):
...
-        if user:
-            workflow.user = session.query(User).filter(User.email == user).one()
-
-        session.add(workflow)
-        session.flush()
-        workflow_id = workflow.workflow_id
-    print(workflow_id)
-
-
-@db.command("create-mega-fasta")
```
jgadling, line 271: We weren't using this, cleaning up dead code.

src/backend/aspen/util/swipe.py
```diff
@@ -0,0 +1,90 @@
...
+import re
+from typing import List
+
+from boto3 import Session
+
+from aspen.api.settings import Settings
+from aspen.database.models import Group, PhyloRun
+
+
+class SwipeJob:
```
jgadling, line 12: This is a start on centralizing the interface for kicking off Swipe jobs.

src/backend/aspen/workflows/nextstrain_run/autorun_scheduled.py
```diff
@@ -1,81 +1,65 @@
...
-    # It's important to make "OutputPrefix" unique because the s3 location is used by
-    # our WDL runner (miniwdl) as a storage area for files generated by the run,
-    # so the prefix must be unique per run to prevent collisions.
-    sfn_params[
-        "OutputPrefix"
-    ] = f'{sfn_params["OutputPrefix"]}/{group.name}/{start_datetime}'
-
-    execution_name = f"{group.prefix}-scheduled-nextstrain-{str(start_datetime)}"
-    execution_name = re.sub(r"[^0-9a-zA-Z-]", r"-", execution_name)
+def create_phylo_run(
```
jgadling, line 31: We're now creating PhyloRun rows before starting swipe jobs

src/backend/aspen/util/swipe.py
```diff
@@ -0,0 +1,90 @@
...
+        output_suffix = f"/{str(now)}"
+        execution_name = f"{group.prefix}-ondemand-pangolin-{str(now)}"
+        return self._start(execution_name, output_suffix, extra_params)
+
+
+class NextstrainJob(SwipeJob):
+    job_type = "ondemand"
+
+    def get_sfn_config(self):
+        return self.settings.AWS_NEXTSTRAIN_SFN_PARAMETERS
```
danrlu, line 75: Where does AWS_NEXTSTRAIN_SFN_PARAMETERS come from?

jgadling: Great question!!

Working backwards from this line, our Settings class is responsible for making that data available. Settings is a subclass of [Pydantic's BaseSettings](https://pydantic-docs.helpmanual.io/usage/settings/) class, which has a *lot* of magic built into it that lets it look for matching parameters in environment variables, settings files, or ${our custom config sources}.

Our settings class has been customized so it can also pull in configuration from AWS SSM (systems manager / parameter store), and Secrets Manager:
https://github.com/chanzuckerberg/czgenepi/blob/trunk/src/backend/aspen/api/settings.py#L256-L272

the AWS_NEXTSTRAIN_SFN_PARAMETERS setting is part of a map of parameters that comes from systems manager:
https://github.com/chanzuckerberg/czgenepi/blob/trunk/src/backend/aspen/api/settings.py#L251-L254

And ultimately gets loaded here:
https://github.com/chanzuckerberg/czgenepi/blob/trunk/src/backend/aspen/api/settings.py#L45-L69

danrlu:
This approach makes a lot of sense to me!! I'm still trying to figure out the wiring. With this set up, scheduled tree run will use what's defined in `swipe.py`, but ondemand runs will maintain its old set up, right? And `swipe.py` is helping to plug scheduled tree run into how ondemand tree run is initiated but I couldn't tell how/where it plugs into ondemand tree runs.

danrlu:
Per your answer above, `swipe.py` will plug into https://github.com/chanzuckerberg/czgenepi/blob/c510f2203d5e586708f6f168b7dd6704b90b5934/src/backend/aspen/api/settings.py#L252
and then go down the ondemand tree build route

jgadling:
> and then go down the ondemand tree build route

Yes, exactly -- scheduled tree builds and on-demand tree builds now work the same way.

danrlu:
Updated diagram 
image

danrlu (approved):
I cannot check the code, but the overall pic sounds good. Just approving here to enable testing in staging

lvreynoso (approved):
Amazing work! This looks really great, organizing our swipe jobs into something manageable. It appears that in scheduled runs the `PhyloRun` was created during the individual batch jobs and thus had the bug, while our ondemand jobs were not bugged because they are created beforehand, is that right?

All around lgtm.

PR #1315, opened 2022-09-01
[SC-190919] Metadata upload - loading spinner.

ehoops-cz:
### Summary:
- **What:** `Add a loading state to the metadata upload button`
- **Ticket:** [sc190919](https://app.shortcut.com/genepi/story/190919/add-loading-state-to-import-metadata-button-when-file-is-importing)
- **Env:** `none`

### Demos:
loading

### Notes:
The spinner doesn't spin because the browser is busy processing the file.  Does it look more broken this way?  Alternatively we could just have the "Loading" text.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/components/FilePicker/index.tsx
+++ b/src/frontend/src/components/FilePicker/index.tsx
@@ -1,7 +1,7 @@
 import { Button } from "czifui";
 import { useRef, useState } from "react";
 import ConfirmDialog from "../ConfirmDialog";
-import { HiddenInput } from "./style";
+import { HiddenInput, StyledIcon } from "./style";
 
 interface Props {
   handleFiles: (files: FileList | null) => void;
@@ -12,7 +12,7 @@ interface Props {
   shouldConfirm?: boolean;
   confirmTitle?: string;
   confirmContent?: string;
-  isDisabled?: boolean;
+  isLoading?: boolean;
 }
 
 export default function FilePicker({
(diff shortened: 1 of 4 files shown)


Janeece:
Looks like we already discussed the feedback for this on slack to remove spinner and just use "Loading…" text :)

src/frontend/src/components/FilePicker/index.tsx
```diff
@@ -61,9 +61,9 @@ export default function FilePicker({
         sdsType="primary"
         sdsStyle="square"
         onClick={shouldConfirm ? openDialog : handleClick}
-        disabled={isDisabled}
+        disabled={isLoading}
       >
-        {text}
+        {isLoading ? "Loading..." : text}
```
ehoops-cz, line 66: We previously had this in one of the parent components but not the rest.  Now if this shared component receives the optional "isLoading" flag, the text will change.  If isLoading is false or undefined the original text will show.

src/frontend/src/components/FilePicker/index.tsx
```diff
@@ -61,9 +61,9 @@ export default function FilePicker({
         sdsType="primary"
         sdsStyle="square"
         onClick={shouldConfirm ? openDialog : handleClick}
-        disabled={isDisabled}
+        disabled={isLoading}
```
ehoops-cz, line 64: Currently the loading state is the only time the button is disabled.  If we want to separate this functionality in the future, we can add a separate "isDisabled" prop.

vincent-czi (approved):

PR #1317, opened 2022-09-02
Write pathogen info to phylo runs on creation

jgadling:
### Summary:
- **What:** Update phylo runs create endpoint to add pathogen info to runs when they're created.
- **Ticket:** [sc211485](https://app.shortcut.com/genepi/story/211485)

### Notes:
This PR does a few things:
- Adds pathogens to the DB at the start of our test runs so each test doesn't need to add them
- Add a new `get_pathogen` dependency so endpoints don't have to do their own validation for the pathogen slug
- Updated a few queries to *start* filtering by pathogen, but there are some TODO's that will need fixing once these fields are marked as `NOT NULL`
- Fixed a problem with the data in the `pathogens` table. Inserting rows with an explicit identifier prevents the sequence from getting updated, which makes it harder to add more rows later. I bumped the sequence so the next insertion should start at `id=2`
- Add some additional samples/runs/trees to our local dev environment with pathogen data attached.

### Checklist:
- [x] I added labels to my PR


src/backend/aspen/api/conftest.py
```diff
@@ -92,7 +92,12 @@ async def async_sqlalchemy_interface(
     # Insert any core level data we're going to need for tests.
     session = test_db_interface.make_session()
     roles = [Role(name="admin"), Role(name="viewer"), Role(name="member")]
-    session.add_all(roles)
+    pathogens = [
+        Pathogen(slug="SC2", name="SARS-CoV-2"),
+        Pathogen(slug="MPX", name="Monkeypox"),
+        Pathogen(slug="TB", name="Tuberculosis"),
+    ]
+    session.add_all(roles + pathogens)
```
jgadling, line 100: This section adds pathogens to the db *really* early in our test setup so they're available to all tests.

src/backend/aspen/api/deps.py
```diff
@@ -55,3 +58,12 @@ def get_pathogen_slug(pathogen_slug=None):
...
     return pathogen_slug
+
+
+async def get_pathogen(
+    slug: str = Depends(get_pathogen_slug), db: AsyncSession = Depends(get_db)
+) -> Pathogen:
+    try:
+        return await Pathogen.get_by_slug(db, slug)
+    except NoResultFound:
+        raise ex.BadRequestException("Invalid pathogen slug")
```
jgadling, line 69: A new dependency that checks that the pathogen slug in our URL is valid and returns an error if not.

src/backend/aspen/api/utils/sample.py
```diff
@@ -32,10 +35,17 @@ async def samples_by_identifiers(
...
+                or_(
+                    # TODO - DECOVIDIFY - remove the None check!
+                    Sample.pathogen == pathogen,  # noqa: E711
+                    Sample.pathogen_id == None,  # noqa: E711
+                ),
+                or_(
+                    public_samples_query.c.id != None,  # noqa: E711
+                    private_samples_query.c.id != None,  # noqa: E711
+                ),
+            ),
```
jgadling, line 65: This adds a filter to select samples for a given type of pathogen (what we want once all samples have pathogens) OR no pathogen (legacy behavior). We need to leave this `== None` thing in place until all newly created samples have pathogen information attached, *and* we migrate all old samples to have pathogens.

src/backend/aspen/api/views/phylo_runs.py
```diff
@@ -155,14 +160,18 @@ async def kick_off_phylo_run(
...
     query = await az.authorized_query(privilege, PhyloRun)
     query = query.options(
         joinedload(PhyloRun.outputs.of_type(PhyloTree)),  # type: ignore
         joinedload(PhyloRun.user),  # For Pydantic serialization
         joinedload(PhyloRun.group),  # For Pydantic serialization
     )
+    # TODO - DECOVIDIFY - remove the None check!
+    query = query.filter(
+        or_(PhyloRun.pathogen == pathogen, PhyloRun.pathogen_id == None)  # noqa: E711
+    )
```
jgadling, line 174: Again with the lame `or` clause, this will need to get updated once our migration is complete.

src/backend/aspen/api/views/tests/test_create_phylo_runs.py
```diff
@@ -51,6 +53,49 @@ async def test_create_phylo_run(
     assert response["group"]["name"] == group.name
     assert response["user"]["name"] == user.name
     assert response["user"]["id"] == user.id
+    assert response["pathogen"]["slug"] == pathogen.slug
+    assert "id" in response
+
+
+async def test_create_phylo_run_mpx(
```
jgadling, line 60: Make sure we can create a phylo run with a non-default pathogen and everything works.

src/backend/aspen/database/models/pathogens.py
```diff
@@ -31,6 +33,11 @@ class Pathogen(idbase):  # type: ignore
     phylo_runs = relationship("PhyloRun", back_populates="pathogen")  # type: ignore
     phylo_trees = relationship("PhyloTree", back_populates="pathogen")  # type: ignore
 
+    @classmethod
+    async def get_by_slug(cls, db: AsyncSession, slug: str) -> Pathogen:
+        resp = await db.execute(sa.select(Pathogen).where(Pathogen.slug == slug))  # type: ignore
+        return resp.scalars().one()
+
```
jgadling, line 40: Helper method for getting pathogens by slug.

src/backend/database_migrations/versions/20220902_001655_bump_pathogen_sequence.py
```diff
@@ -0,0 +1,21 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "20220902_001655"
+down_revision = "20220831_150533"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute("select nextval('aspen.pathogens_id_seq')")
```
jgadling, line 17: Fix our pathogens sequence since the migration that added SC2 used an explicit ID

src/backend/scripts/setup_localdata.py
```diff
@@ -94,9 +95,28 @@ def create_location(session, region, country, division, location):
...
+            session.add(pathogen)
+        pathogens.append(pathogen)
+    return pathogens
+
+
+def create_sample(
+    session, group, pathogen, uploaded_by_user, location, suffix, is_failed=False
+):
+    private_id = f"{group.prefix}-private_identifier_{suffix}_{pathogen.slug}"
+    public_id = f"{group.prefix}-public_identifier_{suffix}_{pathogen.slug}"
```
jgadling, line 119: Changes to this script update the local dev environment to add samples/trees/runs to the db for both MPX *and* SC2

src/backend/aspen/database/models/pathogens.py
phoenixAja: why a class method and not a regular method for this?

src/backend/database_migrations/versions/20220902_001655_bump_pathogen_sequence.py
phoenixAja: oh no what did i do? i'm confused 😢

phoenixAja (approved):

src/backend/database_migrations/versions/20220902_001655_bump_pathogen_sequence.py
jgadling: the migration that does the insert, inserts a specific ID for the SC2 pathogen, instead of leaving that field out of the insert and letting the sequence fill in the blank. So we wound up with a sequence that's ready to insert row ID=1 and a table that already had row ID=1 in it. An example below:

Screen Shot 2022-09-02 at 10 25 43 AM

src/backend/aspen/database/models/pathogens.py
jgadling: Because we're trying to *get* a pathogen instance from the db based on a slug. We don't *have* an instance yet.

src/backend/database_migrations/versions/20220902_001655_bump_pathogen_sequence.py
phoenixAja: ohhhh, thanks for explaining i have learned!

src/backend/aspen/database/models/pathogens.py
phoenixAja: ah ok got it

PR #1322, opened 2022-09-06
Gisaid/Genbank download tweaks

ehoops-cz:
### Summary:
- **What:** `Update endpoint usage`
- **Ticket:** none
- **Env:** https://template-download-frontend.dev.czgenepi.org/

To get fasta file, use the `ORG_API.SAMPLES_FASTA_DOWNLOAD` endpoint and put the public repository in the request body:
```
{
        "sample_ids": [sample.public_identifier],
        "public_repository_name": "GISAID",
}
```
To get submission template, use the new `/v2/orgs/{group.id}/samples/submission_template` endpoint and pass:
```
   {
        "sample_ids": [sample.public_identifier for sample in samples],
        "date": today.strftime("%Y-%m-%d"),
        "page": "0",
        "public_repository_name": "GenBank",
    }
```
Use the "content-disposition" header to get the filename.

### Notes:
I added `resp.headers["Access-Control-Expose-Headers"] = "Content-Disposition"` to the submission_template endpoint for the filename to work.  Extra eyes on this would be great since I was just being a sheep and following what i thought should be there...

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
         exclude: ^src/backend/third-party/
         types: [file, python]
   - repo: https://github.com/psf/black
-    rev: 21.9b0
+    rev: 22.3.0 
     hooks:
       - id: black
         files: ^src/backend/
--- a/src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
@@ -1,6 +1,5 @@
 import { useTreatments } from "@splitsoftware/splitio-react";
 import { Alert, Icon, Link, Tooltip } from "czifui";
-import { isEqual } from "lodash";
 import { useCallback, useEffect, useState } from "react";
 import { CSVLink } from "react-csv";
 import {
--- a/src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
@@ -301,7 +301,8 @@ const DownloadModal = ({
                   <DownloadType>
                     {failedSampleIds.length}{" "}
                     {pluralize("sample", failedSampleIds.length)} will not be
-                    included in your Consensus Genome download
+                    included in your Consensus Genome or Submission Template
+                    downloads
                   </DownloadType>
                   <DownloadTypeInfo>
                     because they failed genome recovery. Failed samples will
--- a/src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
@@ -77,9 +77,6 @@ const DownloadModal = ({
   const flag = useTreatments([FEATURE_FLAGS.prep_files]);
   const isPrepFilesFlagOn = isFlagOn(flag, FEATURE_FLAGS.prep_files);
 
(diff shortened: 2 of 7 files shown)


.pre-commit-config.yaml
```diff
@@ -9,7 +9,7 @@ repos:
         exclude: ^src/backend/third-party/
         types: [file, python]
   - repo: https://github.com/psf/black
-    rev: 21.9b0
+    rev: 22.3.0
```
ehoops-cz, line 12: I was getting an error running the pre-commit hook.  [This fixed it](https://stackoverflow.com/questions/71673404/importerror-cannot-import-name-unicodefun-from-click).  I'm not sure how else I should be testing this though.

src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
```diff
@@ -358,24 +360,43 @@ const DownloadModal = ({
...
           sampleIds: checkedSampleIds,
-          endpoint: ORG_API.SAMPLES_GISAID_DOWNLOAD,
         });
+        const nPages = Math.ceil(nCompletedSampleIds / N_SAMPLES_PER_PAGE);
+        for (let page = 0; page < nPages; page++) {
+          downloadMutation.mutate({
+            endpoint: ORG_API.SAMPLES_TEMPLATE_DOWNLOAD,
+            page,
+            publicRepositoryName: PUBLIC_REPOSITORY_NAME.GISAID,
+            sampleIds: checkedSampleIds,
```
phoenixAja: checkedSamples needs to be subsetted to go along with the page, so for page0 we need checkedSamples[0-999]

phoenixAja: unless i'm missing something

lvreynoso:
👍

phoenixAja (changes requested):
i think we need to update the checkedSamples that are sent in to the request, besides that it looks good!

src/frontend/src/common/components/library/data_subview/components/DownloadModal/index.tsx
lvreynoso: 0-998, right? For 999 samples

phoenixAja: was it 999 or 1000? I'm having a hard time finding the supporting documentation for it. @danrlu can you confirm please?

danrlu: the limit is 1000, so per our last discussion ppl think we should play safe and just do 999 based on my memory, or I failed to communicate clearly XD Let me quickly double check and get back to you.

phoenixAja: thanks dan! let's keep it at 999 then just in case!

ehoops-cz: This other ticket also uses 999: https://app.shortcut.com/genepi/story/194965/as-a-user-if-i-select-more-than-999-samples-i-want-to-be-notified-that-my-template-files-will-be-chopped-up

ehoops-cz:
Removed chunking from the frontend including the multiple file callout.
Renamed GISAID Submitter ID to GISAID User Name.
Filtered to only send completedSampleIds and not failed ones.
Updated Failed sample callout copy.

phoenixAja (approved):

PR #1327, opened 2022-09-07
turn off set -x when dealing with setting/reading secrets config

phoenixAja:
### Summary:
- **What:** stop printing out genepi_config to std out
- **Ticket:** [sc212583](https://app.shortcut.com/genepi/story/212583>)
- **Env:** `<rdev link>`

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/.happy/terraform/modules/sfn_config/gisaid.wdl
+++ b/.happy/terraform/modules/sfn_config/gisaid.wdl
@@ -175,8 +175,12 @@ task TransformGISAID {
     aspen_creation_rev=$COMMIT_SHA
 
     # fetch aspen config
+    set +x  # don't echo secrets
+    echo "* set \$genepi_config"
     genepi_config="$(aws secretsmanager get-secret-value --secret-id ~{genepi_config_secret_name} --query SecretString --output text)"
+    echo "* set \$aspen_s3_db_bucket"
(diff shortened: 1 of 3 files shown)


.happy/terraform/modules/sfn_config/gisaid.wdl
```diff
@@ -175,8 +175,12 @@ task TransformGISAID {
     aspen_creation_rev=$COMMIT_SHA
 
     # fetch aspen config
+    set +x  # don't echo secrets
+    echo "* set \$genepi_config"
```
vincent-czi: I don't understand the `*` purpose? Is it there to indicate that something has been redacted? I think a single asterisk isn't very clear for that. I'd either use `*******` or `XXXXXXX`, more symbols. But I think it's probably clearer to just echo "Setting \$genepi_config" or maybe if you want to be really explicit about why we're doing the approach while printing out the value "Setting \$genepi_config var (not printing value because contains secrets)" or something.

vincent-czi (approved):

.happy/terraform/modules/sfn_config/gisaid.wdl
phoenixAja: ok i'll update the wording

PR #1328, opened 2022-09-07
Support pathogens in nextstrain runs

jgadling:
### Summary:
- **What:** Update tree builds to support pathogens
- **Ticket:** [sc211475](https://app.shortcut.com/genepi/story/211475)

### Notes:
This PR accomplishes two things:
- Writes `pathogen` info to phylo trees at the end of a run
- Refactors our YAML builder code into a plugin model, with plugins for tree type, pathogen, etc. This might change again when we learn more from CompBio about what nextstrain config changes we need to make per-pathogen, but we'll figure it out as we go.

### Checklist:
- [x] I added labels to my PR


src/backend/aspen/workflows/nextstrain_run/save.py
```diff
@@ -100,6 +101,7 @@ def cli(
             name=phylo_run.name,
             group=phylo_run.group,
             tree_type=phylo_run.tree_type,
+            pathogen=phylo_run.pathogen,
```
jgadling, line 104: This is what writes pathogen info to trees

src/backend/aspen/workflows/nextstrain_run/build_config.py
```diff
@@ -1,182 +1,76 @@
...
+)
+from aspen.workflows.nextstrain_run.build_plugins.type_plugins import (
+    NonContextualizedPlugin,
+    OverviewPlugin,
+    TargetedPlugin,
+    TreeType,
+)
+
+
+class TemplateBuilder:
```
jgadling, line 16: Instead of using a factory that produces a Contextual/Overview/Whatever tree, we'll use a wrapper class that manages a set of config-modification plugins.

phoenixAja (approved):
👍

PR #1329, opened 2022-09-07
use pathogen slug for create and list samples

phoenixAja:
### Summary:
- **What:** use the pathogen slug to return data according to pathogen slug, or create new samples associated with correct pathogen
- **Ticket:** [sc211484](https://app.shortcut.com/genepi/story/211484)
- **Env:** `<rdev link>`

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/views/samples.py
+++ b/src/backend/aspen/api/views/samples.py
@@ -73,12 +73,9 @@ async def list_samples(
         selectinload(Sample.uploaded_by),
         selectinload(Sample.collection_location),
         selectinload(Sample.accessions),
-        selectinload(Sample.pathogen)
-    ).where(
-        Sample.pathogen.slug == pathogen
-    )
+        selectinload(Sample.pathogen),
+    ).filter(Sample.pathogen_id == pathogen.id)
     user_visible_samples_result = await db.execute(user_visible_samples_query)
-    import pdb; pdb.set_trace()
     user_visible_samples: List[Sample] = (
         user_visible_samples_result.unique().scalars().all()
     )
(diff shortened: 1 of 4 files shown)


src/backend/aspen/api/views/samples.py
```diff
@@ -73,7 +73,8 @@ async def list_samples(
         selectinload(Sample.uploaded_by),
         selectinload(Sample.collection_location),
         selectinload(Sample.accessions),
-    )
+        selectinload(Sample.pathogen),
+    ).filter(Sample.pathogen_id == pathogen.id)
```
jgadling: This PR looks good! Only question is whether this `list` endpoint behaves reasonably well when `sample.pathogen_id` is null

phoenixAja: oh that's right we will have some samples without pathogen in the interim.. I think the best case here would be to return an exception?

jgadling: ha, we can't return an exception since we want our endpoints to keep working properly until we update all samples!

Returning 'pathogen: null` in the response is fine for the interim, but I want to make sure that samples without pathogens are still listed.

phoenixAja: so actually if we do have a null pathogen_slug the `get_pathogen_slug` will return `SC2` as a default, and in `get_pathogen` we `raise ex.BadRequestException("Invalid pathogen slug")` if no pathogen was found, so i think we're set on this and won't get any wierd edge cases. I'll add another test to verify though

lvreynoso (approved):
lgtm

jgadling (changes requested):
This isn't going to work properly for samples without pathogens.

phoenixAja:
@jgadling if no pathogen is provided we default to SC2, and if no pathogen is found for a given slug we raise an exception, i added tests for those two cases

jgadling (approved):

PR #1331, opened 2022-09-07
Add pathogens to old phylo trees/runs

jgadling:
### Summary:
- **What:** Now that all phylo runs and phylo trees get created with a pathogen attached, update any old trees/runs that don't have pathogens.
- **Ticket:** [sc211476](https://app.shortcut.com/genepi/story/211476)

### Checklist:
- [x] I added labels to my PR


lvreynoso (approved):

src/backend/database_migrations/versions/20220907_225235_add_pathogens_to_all_trees.py
```diff
@@ -0,0 +1,54 @@
...
+Create Date: 2022-09-07 22:52:41.089769
+
+"""
+import enumtables  # noqa: F401
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "20220907_225235"
+down_revision = "20220902_001655"
```
lvreynoso: This is now `20220907_220840`, from your synchronize alembic migration.

jgadling:
When adding pathogens to our unit tests, I'd originally added the pathogens to the db so every test didn't have to add pathogens to the db. I'm pretty sure that was the wrong call at this point, so I removed the automatic-pathogen-inserts and instead updated ~every test that needs pathogen info in the db to work properly. Lots of little changes, most of them fairly insubstantial.

src/backend/aspen/api/conftest.py
```diff
@@ -92,12 +92,7 @@ async def async_sqlalchemy_interface(
     # Insert any core level data we're going to need for tests.
     session = test_db_interface.make_session()
     roles = [Role(name="admin"), Role(name="viewer"), Role(name="member")]
-    pathogens = [
```
jgadling, line 95: This is where we *were* adding pathogens to the db. Tests are now responsible for handling their own pathogen data

src/backend/aspen/test_infra/models/pathogen.py
```diff
@@ -1,9 +1,30 @@
+from typing import Optional
+
 import sqlalchemy as sa
 
 from aspen.database.models import Pathogen
+from aspen.database.models.usergroup import generate_random_id
+
+
+# TODO this should be the functionality we use everywhere.
+def random_pathogen_factory(
```
jgadling, line 10: Once we update our tests to include pathogen slugs as part of request URL's, we're going to replace `pathogen_factory` with this method, and fix any reference to this method's old name. We need to make sure our app can support arbitrary pathogen names so this random generator should make that pretty easy.

lvreynoso (approved):
Looks sane to me? Kind of crazy in its own way, but consistent with our internal insanity

PR #1332, opened 2022-09-08
Handle missing location information for GISAID prepfile d/l

lvreynoso:
### Summary:
- **What:** Fixes a bug in submission template downloads. where a `None` field in the sample's collection location would raise an exception. Cleans up code that pulls information from sample, group, and user.
- **Env:** [https://prepfile-frontend.dev.czgenepi.org/](https://prepfile-frontend.dev.czgenepi.org/)

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/utils/sample.py
+++ b/src/backend/aspen/api/utils/sample.py
@@ -21,6 +21,7 @@ from aspen.database.models import (
     Accession,
     AccessionType,
     Group,
+    Location,
     Pathogen,
     PathogenRepoConfig,
     PublicRepository,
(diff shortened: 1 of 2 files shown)


src/backend/aspen/api/utils/sample.py
```diff
@@ -273,9 +270,13 @@ def sample_info_to_gisaid_rows(
 ) -> List[Dict[str, str]]:
     gisaid_metadata_rows = []
     for sample_info in submission_information:
-        gisaid_location = " / ".join(
-            [sample_info[key] for key in ["region", "country", "division", "location"]]
-        )
+        # Concat location information to a single string in Region / Country / Division / Location format
+        location_strings = []
+        for key in ["region", "country", "division", "location"]:
+            if sample_info[key]:
```
phoenixAja, line 276: will this raise a keyerror?

phoenixAja (approved):
I think there might be a keyerror, but besides that this looks great! Thanks for the quick fix!

src/backend/aspen/api/utils/sample.py
lvreynoso: We construct this dictionary as part of the route (in the above utility function `collect_submission_information()`) so these keys will always be present.

lvreynoso: The `if` test is just to check for a `None` value, since `str.join()` doesn't handle those nicely.

PR #1340, opened 2022-09-12
update sequence test url paths [EZPZ]

phoenixAja:
### Summary:
- **What:** update sequences test to use pathogen slug parameter
- **Env:** `<rdev link>`

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/views/tests/test_sequences.py
+++ b/src/backend/aspen/api/views/tests/test_sequences.py
@@ -5,7 +5,7 @@ from httpx import AsyncClient
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from aspen.api.views.sequences import get_fasta_filename
-from aspen.database.models import Sample
+from aspen.database.models import Pathogen, Sample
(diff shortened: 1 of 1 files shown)


jgadling (changes requested):
Can these tests be updated to generate random pathogens and use that pathogen slug as part of the URL instead of hardcoding SC2 ?

src/backend/aspen/api/views/tests/test_sequences.py
```diff
@@ -62,8 +62,11 @@ async def test_prepare_sequences_download_gisaid(
         "sample_ids": [sample.public_identifier],
         "public_repository_name": "GISAID",
     }
+    pathogen = await Pathogen.get_by_slug(async_session, "SC2")
```
jgadling: Can we switch these to using the random pathogen factory instead:
```
from aspen.test_infra.models.pathogen import random_pathogen_factory
```

jgadling (approved):

PR #1342, opened 2022-09-12
Maya/211446 local split

mmmmmmaya:
remove two old feature flags that are no longer in use

### DIFF
--- a/src/frontend/src/components/NavBar/components/AppNavBar/index.tsx
+++ b/src/frontend/src/components/NavBar/components/AppNavBar/index.tsx
@@ -1,12 +1,10 @@
-import { useTreatments } from "@splitsoftware/splitio-react";
 import { Icon } from "czifui";
 import Link from "next/link";
-import { Fragment, MouseEventHandler, useState } from "react";
+import { MouseEventHandler, useState } from "react";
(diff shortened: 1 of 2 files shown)


src/frontend/src/components/NavBar/components/AppNavBar/index.tsx
```diff
@@ -36,35 +34,13 @@ const AppNavBar = (): JSX.Element => {
...
   const name = group?.name;
-  const orgElements = (
-    <Fragment>
-      <Separator />
-      <NavOrg>{name}</NavOrg>
-    </Fragment>
-  );
-
-  function hasOrg(): JSX.Element | null {
-    if (group === undefined) {
```
jgadling, line 60: if it matters here, it is (or, at least it will be in the future) possible for a user to not be a member of any groups.

ehoops-cz (approved):
Looks good.  Looks like this is just for removal of old flags 👍

src/frontend/src/components/NavBar/components/AppNavBar/index.tsx
mmmmmmaya: That will be relevant in the future, though not for this component. That will probably be a change closer to redux, since that's what handles all of the current group state

PR #1343, opened 2022-09-12
Make it clearer what size instances to launch.

jgadling:
### Notes:
I did some investigation into what's going on with our batch job sizing and learned a few things:
- Using *both* vcpu *and* memory parameters to specify what size instance to launch is confusing for humans, it's easy to update one forget to update the other so it's not always clear what instance class is getting launched.
- It's more than just confusing for humans. It looks like swipe is in some cases basically ignoring the vcpu parameter. I have some time set aside next sprint to look into this, but in the meantime we can work around it by just using memory values.
- I did some testing in staging and for more ambiguous reasons, when we ask for 384000 mib of memory, we get an r5d.16xlarge instance, but when we ask for 382000 mib of memory, we get the r5d.12xlarge instance size we were aiming for. I'm not sure why this happens, but I added a "- 2000" to our terraform code to work around this.

Basically launching instances that are larger than we expected / planned for allowed AWS to schedule more jobs to the same host, which caused disk space contention. This PR *hopefully* means that all of our instances will be right-sized to each run on their own host for the moment.

There are a lot of improvements/changes we plan to take on in the future to make all this better, but I'm hoping this keeps jobs stable for now.

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/.happy/terraform/modules/sfn_config/main.tf
+++ b/.happy/terraform/modules/sfn_config/main.tf
@@ -8,8 +8,8 @@ locals {
       )
     }
     OutputPrefix = "s3://${var.swipe_comms_bucket}/swipe${var.remote_dev_prefix}/${var.app_name}/results",
-    RunSPOTMemory = var.memory,
-    RunEC2Memory = var.memory,
+    RunSPOTMemory = var.memory - 2000,
+    RunEC2Memory = var.memory - 2000,
     RunSPOTVcpu = var.vcpus,
     RunEC2Vcpu = var.vcpus,
     RUN_WDL_URI = "s3://${aws_s3_bucket_object.wdl.bucket}${aws_s3_bucket_object.wdl.key}",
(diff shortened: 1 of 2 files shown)


.happy/terraform/modules/sfn_config/main.tf
```diff
@@ -8,8 +8,8 @@ locals {
       )
     }
     OutputPrefix = "s3://${var.swipe_comms_bucket}/swipe${var.remote_dev_prefix}/${var.app_name}/results",
-    RunSPOTMemory = var.memory,
-    RunEC2Memory = var.memory,
+    RunSPOTMemory = var.memory - 2000,
```
vincent-czi, line 11: The explanation in the PR description is great to explain this, but the `- 2000` here sure feels like a magic number. Could we add a comment here or somewhere? I almost never go looking for PRs and reading PR convos on GitHub so I don't think I would remember why this is here in a month.

vincent-czi (approved):
Thanks for the clear write up!

PR #1345, opened 2022-09-13
make gisaid metadata truly csv

danrlu:
### Summary:
- **What:** 
- 1. Make GISAID metadata a CSV format. 
- 2. Update GenBank `isolate` column to be identical to the Sequence_ID column w `SARS-CoV-2/human` prefix.
- 3. Lower case `Homo sapiens`

### Demos:
(I don't know how to set up remote dev, so here is from my local dev)
image
image

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/api/utils/__init__.py
+++ b/src/backend/aspen/api/utils/__init__.py
@@ -26,6 +26,6 @@ from aspen.api.utils.sample import (  # noqa: F401
 from aspen.api.utils.tsv_streamer import (  # noqa: F401
     FieldSeparatedStreamer,
     GenBankSubmissionFormTSVStreamer,
-    GisaidSubmissionFormTSVStreamer,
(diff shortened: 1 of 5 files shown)


src/backend/aspen/api/utils/__init__.py
```diff
@@ -26,6 +26,6 @@
 from aspen.api.utils.tsv_streamer import (  # noqa: F401
     FieldSeparatedStreamer,
     GenBankSubmissionFormTSVStreamer,
-    GisaidSubmissionFormTSVStreamer,
+    GisaidSubmissionFormCSVStreamer,
```
danrlu, line 29: Not sure whether is better to change the name. Can change it back if that's more desirable

lvreynoso (approved):

PR #1346, opened 2022-09-13
[SC-214165] modal to confirm navigation to Galago

ehoops-cz:
### Summary:
- **What:** `Modal to confirm navigation to Galago`
- **Ticket:** [sc214165](https://app.shortcut.com/genepi/story/214165/modal-for-users-to-confirm-sending-out-to-galago-from-cz-ge)
- **Env:** `none`

### Demos:
Screen Shot 2022-09-13 at 8 30 50 AM

### Notes:
The modal is not behind the feature flag because the tree actions button is already behind it.

### Checklist:
- [x] I added labels to my PR


src/frontend/src/views/Data/components/GalagoConfirmationModal/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+  onClose,
+  treeId,
+}: Props): JSX.Element => {
+  const content = (
+    <>
+      By clicking “Continue” you agree to send a copy of your tree JSON to
+      Galago (Beta), a separate, but related service from CZ GEN EPI. Galago is
+      a serverless web application which runs entirely in the browser. Galago
+      does not store or share your data; however, you may choose to share the
+      URL with others.{" "}
```
vincent-czi: What's the final forced space `{" "}` getting us here? I'm used to it when there's multiple text components that get spaces stripped between them, but since this is a free-floating paragraph, I don't see what the space is preventing adjacency on.

src/frontend/src/views/Data/components/GalagoConfirmationModal/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+// currently this is just using the nextstrain url as a placeholder
+import { ConfirmButton } from "src/common/utils/TreeModal/ConfirmButton";
+
+interface Props {
+  open: boolean;
+  onClose: () => void;
+  treeId: number;
+}
+
+const NextstrainConfirmationModal = ({
```
vincent-czi: Can we change this name?

src/frontend/src/views/Data/components/GalagoConfirmationModal/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+      img={galagoLogo as unknown as string}
+      isOpen={open}
+      onClose={onClose}
+      onConfirm={onClose}
+      logoWidth={180}
+    />
+  );
+};
+
+export default NextstrainConfirmationModal;
```
vincent-czi: Same name change thing. I'm personally not a fan of `export default` because the naming changes more implicitly during import process and I like being able to "find all" across a codebase for a name. We don't have a really strong culture within the project of only using one or the other for exporting components from a file, but I think we've been moving more towards the explicit, named export type. I think @mmmmmmaya prefers it as well? I don't know if we have a preference between directly exporting where it's defined:
```
export function WarningMissingData(blah){...ReactStuffGoesHere...}
```

Or defining it without an export, then doing a named export at the file end
```
const FailedSampleAlert = ...ReactStuffGoesHere...
...

export { FailedSampleAlert };
```

I think we tend to use the latter more, especially for bigger components, but I don't think we've actually talked about that aspect before.

vincent-czi (approved):
Looks good overall. Only real request for change is swapping the internal name for component in the `GalagoConfirmationModal` file. I'd definitely like to see that, but approving because I don't need to review that change, can just merge once it's in. Thanks!

src/frontend/src/views/Data/components/GalagoConfirmationModal/index.tsx
mmmmmmaya: I prefer the named export.

src/frontend/src/views/Data/components/GalagoConfirmationModal/index.tsx
ehoops-cz: oops! yes

src/frontend/src/views/Data/components/GalagoConfirmationModal/index.tsx
ehoops-cz: I'm pretty neutral on this.  I do think it makes sense to be as consistent as possible though.  Maybe we can bring this up in a team sync and get aligned.

ehoops-cz: I prefer exporting the arrow function component.  Just from an easier to read perspective though.
```
export const GalagoConfirmationModal = ({
  open,
  onClose,
  treeId,
}: Props): JSX.Element => {...}
```

src/frontend/src/views/Data/components/GalagoConfirmationModal/index.tsx
ehoops-cz: This was just some copy pasta since the NextStrain modal has a Learn More link there.

src/frontend/src/views/Data/components/GalagoConfirmationModal/index.tsx
ehoops-cz: @mmmmmmaya @vincent-czi do you have a preference between the various ways to get named exports?

vincent-czi: I'm neutral between the two ways: directly exporting a `function`/arrowFunction where it's defined or batching up at the bottom are equally fine by me. I just don't like default exports because I've been bitten by minor typos where the import happens making it hard to realize you didn't notice some usage somewhere.

vincent-czi: I guess my one thing about named exports is I don't actually like a _batch_ of named exports at the bottom. For the non-primary things a file exports, I usually want the export on the line its definition starts. Having a default at the bottom feels like a way of signaling to other devs, "Hey, this is the primary purpose of this file. It builds up to this thing being available in other places." So having a bunch of exports in the bottom muddies the water IMO

ehoops-cz: I agree with not liking a bunch of named exports at the bottom.  It can sometimes lead to someone thinking "I'll just clean these up by making an object of exports" without realizing that the pattern is bad for tree-shaking.

ehoops-cz: One note about putting the export on the line where it's defined - I often find myself looking at those and if I don't see an export, I assume it's only used in that file (usually this is in a styled component file).  Having the information inline would help with this.  I also assume the main component of the file should match the file name/path.

PR #1349, opened 2022-09-13
Test/sc 210103/upload sample

luisczi:
### Summary:
- **What:** `UI automation task, that is focus on handling the upload module by testing the happy path scenarios`
- **Ticket:** <TEST: Upload Sample Flow>(https://app.shortcut.com/genepi/story/210103/test-upload-sample-flow/<210103>)
- **Env:** `[<rdev link>](https://staging.czgenepi.org/data/samples)`

### Checklist:
- [ ] I added labels to my PR


src/frontend/e2e/pages/upload.ts
```diff
@@ -0,0 +1,86 @@
...
+      await page.locator("//div[@role='tooltip']/descendant::input").type(uploadData.samples[i].collectionLocation,{delay:150});
+      await page.keyboard.press("ArrowDown");
+      await page.keyboard.press("Enter");
+      await page.locator(`//div[contains(@class,'MuiTableBody-root')]/div[@role='row'][${i+1}]/descendant::input[@name='sequencingDate']`).type(uploadData.samples[i].sequencingDate);
+      await page.locator(`//div[contains(@class,'MuiTableBody-root')]/div[@role='row'][${i+1}]/descendant::input[@name='keepPrivate']`).click();
+    }
+    
+  }
+
+  public static async selectSample(
```
charles-testco: I think we should rename this to getSampleDataFile

src/frontend/e2e/pages/upload.ts
```diff
@@ -0,0 +1,86 @@
...
+  public static async uploadFiles(
+    page: Page,
+    uploadData: UploadData,
+  ): Promise<any> {
+    //click upload button
+    await page.locator("a[href$='/upload/step1']").click();
+    await UploadSample.selectSample(page,uploadData.dataFile);
+    // complete form
+    for(let i = 0; i < uploadData.samples.length; i++){
+      await page.locator(`//div[contains(@class,'MuiTableBody-root')]/div[@role='row'][${i+1}]/descendant::input[@name='publicId']`).type(uploadData.samples[i].publicId);
```
charles-testco: Pls put a one-line comment for each field so we know which field is being filled/populated
Also put a TODO note to replace selectors with IDs

charles-testco (changes requested):
@luisczi Pls:
1. Complete the PR details: What it is, the linked tickets
2. Tick checkbox to indicate it has the latest trunk
3. Add Labels E2E and Test

charles-testco (approved):

src/frontend/e2e/pages/upload.ts
```diff
@@ -0,0 +1,96 @@
+import { Page } from "@playwright/test";
+import { faker } from "@faker-js/faker";
+import { SampleData } from "../utils/schemas/sampleData";
+import { UploadData } from "../utils/schemas/uploadData";
+import * as path from 'path';
+
+
+export class UploadSample {
+  public static async uploadFiles(
```
phoenixAja: can you rename this to `uploadSequencingFiles`?

src/frontend/e2e/pages/upload.ts
```diff
@@ -0,0 +1,96 @@
...
+      await page.locator(`//div[contains(@class,'MuiTableBody-root')]/div[@role='row'][${i+1}]/descendant::button[@label='Search For Location']`).click();
+      await page.locator("//div[@role='tooltip']/descendant::input").type(uploadData.samples[i].collectionLocation,{delay:150});
+      await page.keyboard.press("ArrowDown");
+      await page.keyboard.press("Enter");
+      // fill sequencing date input
+      // TODO modify selector
+      await page.locator(`//div[contains(@class,'MuiTableBody-root')]/div[@role='row'][${i+1}]/descendant::input[@name='sequencingDate']`).type(uploadData.samples[i].sequencingDate);
+       // toggle keep private switch 
+       // TODO modify selector
+      await page.locator(`//div[contains(@class,'MuiTableBody-root')]/div[@role='row'][${i+1}]/descendant::input[@name='keepPrivate']`).click();
```
phoenixAja: how much work would it take to remove the xpath selectors? i'm getting worried about having too much of them floating around.

src/frontend/e2e/pages/upload.ts
```diff
@@ -0,0 +1,96 @@
...
+    // click continue button
+    await page.locator("a[href$='/upload/step2']").click();
+    await page.waitForSelector("form div[role='table'] > .MuiTableBody-root");
+  }
+
+  // prefix each file name with path. set default if not file provided
+  public static getFullSampleFilePaths(
+    fileNames?: Array<string>
+  ): Array<string> {
+    const files = fileNames !== undefined ? fileNames : ["sampleData.fasta"];
```
phoenixAja: when would the fileName be undefined?

src/frontend/e2e/pages/upload.ts
```diff
@@ -0,0 +1,96 @@
...
+    const fullPathFiles: Array<string> = [];
+    files.forEach((file) => {
+      fullPathFiles.push(`../fixtures/${file}`);
+    });
+    return fullPathFiles;
+  }
+
+  //to be replaced by mock data in a separate PR
```
phoenixAja: i think it's fine to leave this like this since this is supposed to mimic what's in a tsv file to be uploaded, not data from an endpoint

src/frontend/e2e/pages/upload.ts
```diff
@@ -0,0 +1,96 @@
...
+  public static getSampleData(): Array<SampleData> {
+    return [
+      {
+        collectionDate: "2022-07-01",
```
phoenixAja: why are you using faker for `sequencingDate` but not `collectionDate`?

src/frontend/e2e/setup/global-setup.ts
```diff
@@ -1,14 +1,16 @@
...
+import path from "path";
+import dotenv from 'dotenv';
 
+dotenv.config({path: path.resolve(`.env.${process.env.NODE_ENV}`),});
 
-const username = process.env.USERNAME ?? "";
-const password = process.env.PASSWORD ?? "";
+
+const username = process.env.USERNAME as string;
+const password = process.env.PASSWORD as string;
```
phoenixAja: can you do this instead? 
```suggestion
const username: string = process.env.USERNAME;
const password: string = process.env.PASSWORD;

```

src/frontend/e2e/tests/home.spec.ts
```diff
@@ -1,29 +1,29 @@
...
-//todo: populate the remaining elements
 const footer: Record<string, string> = {
   Github: "https://github.com/chanzuckerberg/czgenepi/",
   Careers:
     "https://chanzuckerberg.com/careers/career-opportunities/?initiative=science",
   "Learning Center":
     "https://help.czgenepi.org/hc/en-us/categories/6217716150804-Genomic-Epidemiology-Learning-Center",
+  //"Privacy":  "/privacy",
+  //"Terms":  "/terms",
+  //"Contact Us":  "mailto:hello@czgenepi.org"
```
phoenixAja: why are these commented out?

src/frontend/e2e/tests/home.spec.ts
```diff
@@ -1,29 +1,29 @@
...
 };
 
 test.describe("Home page tests", () => {
-  test("Should verify home page", async ({ page }, workerInfo) => {
-    const { baseURL } = workerInfo.config.projects[0].use;
-    await page.goto(`${baseURL}` as string);
-    await expect(page.locator(getByTestID("navbar-landing"))).toBeVisible();
+  test("Should verify home page", async ({ page }) => {
+    await page.goto("https://staging.czgenepi.org");
```
phoenixAja: you should be pulling this from the env, not hardcoding

src/frontend/e2e/tests/sample.spec.ts
```diff
@@ -30,12 +30,21 @@ test.describe("Samples page tests", () => {
   });
 
   test("Should verify sample data", async ({ page }: { page: Page }) => {
+    await page.goto("https://staging.czgenepi.org/data/samples");
```
phoenixAja: same here

src/frontend/package.json
```diff
@@ -16,9 +16,10 @@
     "type-check": "tsc --noEmit",
     "lint": "concurrently \"node_modules/.bin/eslint **/*.{ts,tsx} --quiet --fix\" \"node_modules/.bin/stylelint --fix '**/*.{js,ts,tsx,css}'\"  \"npm run type-check\"",
     "lint-ci": "concurrently \"node_modules/.bin/eslint **/*.{ts,tsx}\" \"node_modules/.bin/stylelint '**/*.{js,ts,tsx,css}'\"  \"npm run type-check\"",
+    "lint-e2e": "concurrently \"node_modules/.bin/eslint --fix e2e/**/*.ts\" \"node_modules/.bin/stylelint --fix 'e2e/**/*.ts'\"  \"npm run type-check\"",
     "pl:staging": "NODE_ENV=staging npx playwright test --headed -c ./e2e/setup/staging.config.ts",
     "pl:local": "NODE_ENV=local npx playwright test --headed -c ./e2e/setup/staging.config.ts",
-    "pl:staging:headless": "npx playwright test --headless -c ./e2e/setup/staging.config.ts",
+    "pl:staging:headless": "npx playwright test --headed -c ./e2e/setup/staging.config.ts",
```
phoenixAja: did you mean to change this?

phoenixAja (changes requested):

phoenixAja:
please remove hardcoded references to staging and pull from env

src/frontend/e2e/pages/upload.ts
luisczi: yup, its fixed now

src/frontend/e2e/pages/upload.ts
luisczi: i believe somebody needs to dig into the app and the specific attributes valus into the markup so we can link to that info

src/frontend/e2e/pages/upload.ts
luisczi: fixed

src/frontend/e2e/pages/upload.ts
luisczi: fixed

src/frontend/e2e/pages/upload.ts
luisczi: fixed

src/frontend/e2e/setup/global-setup.ts
luisczi: fixed

src/frontend/e2e/tests/home.spec.ts
luisczi: fixed

src/frontend/e2e/tests/home.spec.ts
luisczi: fixed

src/frontend/e2e/tests/sample.spec.ts
luisczi: fixed

src/frontend/package.json
luisczi: its headed now yup

src/frontend/e2e/pages/upload.ts
phoenixAja: yeah, can we do that for some of this stuff potentially?

luisczi: i dont know how , and now that im looking at those, they are also dynamic so im wodering how should we address those

jgadling: @charles-testco should be able to provide some guidance here?

src/frontend/e2e/pages/upload.ts
```diff
@@ -0,0 +1,84 @@
...
+
+
+export class UploadSample {
+  public static async uploadSequencingFiles(
+    page: Page,
+    uploadData: UploadData,
+  ): Promise<any> {
+    //click upload button
+    await page.locator("a[href$='/upload/step1']").click();
+    await UploadSample.getSampleDataFile(page,uploadData.dataFile);
```
phoenixAja:
```suggestion
    await UploadSample.getSampleDataFile(page, uploadData.dataFile);
```

src/frontend/e2e/pages/upload.ts
```diff
@@ -0,0 +1,84 @@
...
+      await page.locator(`//div[contains(@class,'MuiTableBody-root')]/div[@role='row'][${i+1}]/descendant::input[@name='publicId']`).type(uploadData.samples[i].publicId);
+       // fill collection date input
+       // TODO modify selector
+      await page.locator(`//div[contains(@class,'MuiTableBody-root')]/div[@role='row'][${i+1}]/descendant::input[@name='collectionDate']`).type(uploadData.samples[i].collectionDate);
+       // fill search for location input from emergent widget
+       // TODO modify selector
+      await page.locator(`//div[contains(@class,'MuiTableBody-root')]/div[@role='row'][${i+1}]/descendant::button[@label='Search For Location']`).click();
+      await page.locator("//div[@role='tooltip']/descendant::input").type(uploadData.samples[i].collectionLocation,{delay:150});
+      await page.keyboard.press("ArrowDown");
+      await page.keyboard.press("Enter");
```
phoenixAja, line 29: can you add this logic to the other apply all fields (sequencing_date, collection_date)?

phoenixAja, line 48: are you testing the apply to all functionality here? That might make the test slow if you're applying all on a loop?

src/frontend/e2e/pages/upload.ts
phoenixAja: can you remove the comment then 😄 ?

src/frontend/e2e/tests/sample.spec.ts
```diff
@@ -30,12 +34,21 @@ test.describe("Samples page tests", () => {
   });
 
   test("Should verify sample data", async ({ page }: { page: Page }) => {
+    await page.goto(process.env.BASEURL+"data/samples" as string);
```
phoenixAja, line 37: does this have to be cast as string?

src/frontend/e2e/pages/upload.ts
jgadling: * best practices for adding tags is specifically part of the guidance we've been looking for with this project

src/frontend/e2e/pages/upload.ts
charles-testco: I wanted it to be generic so use can call without specifying the file names, in which it defaults sampleData.fasta. We can obviously pass the default as a parameter, which will be more cleaner

src/frontend/e2e/setup/global-setup.ts
charles-testco: The casting was to make TypeScript and IDE happy really

src/frontend/e2e/tests/home.spec.ts
charles-testco: This is left over from PoC; these elements were failing inconsistently (failing to find) and I think during our first meeting we agreed is priority 3 so we never went back to investigate and fix. Will remove for now

src/frontend/package.json
charles-testco: @luisczi this should not change really. Headed is already defined in line 20.

src/frontend/e2e/pages/upload.ts
charles-testco: @luisczi I wasn't aware you removed the original logic to handle this. Check commit f501e3635cd1ba4423147ba6eb30cec04987d208 and see how you can make use of it to address this feedback

src/frontend/e2e/tests/sample.spec.ts
charles-testco: No it does have to be; just making TS happy

src/frontend/e2e/pages/upload.ts
charles-testco: @phoenixAja , @jgadling , Yes, using test id is crucial for stable tests. I have already started adding test Id but I do have backlog and need to do this in a separate PR to avoid being a bottleneck and also get this test running. There are couple of things:
1. Some elements are complex and difficult to add ID to. For example, on sample page I could not figure out where to inject Id for the collection date field. This is typically true when using third party plugins, so we may have few elements where we will not compromise.
2. I have faced challenges with test-ids resolving very late, causing tests to fail. For example, on sample page, data-test-id="row-publicId" is showing as data-test-id="loading" initially. I need to figure out which event to hook to to wait until the Ids are resolved. Unfortunately Playwrights built-in wait for page load is not enough.
3. We also found difference in ID based on environment, for example id="username" on login page is different on local than staging.

All these factors have contributed to the use of these selectors as a temporary workaround, until I tackle those challenges.

src/frontend/e2e/pages/upload.ts
luisczi: @charles-testco the original logic was broken, thats the reason why i changed it, @phoenixAja this is needed since the element is an emergent component that needs focus on specific element prior selecting thats the reason i hit the keyboard, but this doesnt slow it down

phoenixAja: can you add a comment explaining that please? 😄

src/frontend/e2e/pages/upload.ts
phoenixAja: we do have a [follow up ticket for this ](https://app.shortcut.com/genepi/story/215051/replace-css-selectors-with-test-id) i changed its priority to `highest` so we'll pass on this for now so we can merge this in, but we need to address this soonish

src/frontend/e2e/setup/global-setup.ts
```diff
@@ -1,14 +1,16 @@
...
+import path from "path";
+import dotenv from 'dotenv';
 
+dotenv.config({path: path.resolve(`.env.${process.env.NODE_ENV}`),});
 
-const username = process.env.USERNAME ?? "";
-const password = process.env.PASSWORD ?? "";
+
+const username: string = process.env.USERNAME as string;
+const password: string = process.env.PASSWORD as string;
```
phoenixAja, line 10: somehow the string casting came back, do you need this to pass linting? If so can you add a todo to address this later or fix the downstream component to accept a different type?
```suggestion
const username: string = process.env.USERNAME;
const password: string = process.env.PASSWORD;
```

src/frontend/e2e/tests/sample.spec.ts
phoenixAja: what type is `process.env.BASEURL+"data/samples"` if it's not already a string? 😕

phoenixAja: ok i'll ignore for now, would be nice moving forwards to not make type assertions like this if possible 😄

src/frontend/package.json
```diff
@@ -16,10 +16,11 @@
     "type-check": "tsc --noEmit",
     "lint": "concurrently \"node_modules/.bin/eslint **/*.{ts,tsx} --quiet --fix\" \"node_modules/.bin/stylelint --fix '**/*.{js,ts,tsx,css}'\"  \"npm run type-check\"",
     "lint-ci": "concurrently \"node_modules/.bin/eslint **/*.{ts,tsx}\" \"node_modules/.bin/stylelint '**/*.{js,ts,tsx,css}'\"  \"npm run type-check\"",
+    "lint-e2e": "concurrently \"node_modules/.bin/eslint --fix e2e/**/*.ts\" \"node_modules/.bin/stylelint --fix 'e2e/**/*.ts'\"  \"npm run type-check\"",
     "pl:staging": "NODE_ENV=staging npx playwright test --headed -c ./e2e/setup/staging.config.ts",
-    "pl:local": "NODE_ENV=local npx playwright test --headed -c ./e2e/setup/staging.config.ts",
-    "pl:staging:headless": "npx playwright test --headless -c ./e2e/setup/staging.config.ts",
-    "pl:staging:debug": "npx playwright test --debug -c ./e2e/setup/staging.config.ts",
+    "pl:local": "NODE_ENV=local npx playwright test --headed -c ./e2e/setup/local.config.ts",
```
phoenixAja, line 21: thanks for fixing this!

src/frontend/package.json
```diff
@@ -16,10 +16,11 @@
...
+    "pl:staging:headless": "NODE_ENV=local NODE_ENV=staging npx playwright test --headless -c ./e2e/setup/staging.config.ts",
```
phoenixAja: you have 2 `NODE_ENV` s here

phoenixAja:
```suggestion
    "pl:staging:headless": "NODE_ENV=staging npx playwright test --headless -c ./e2e/setup/staging.config.ts",
```

src/frontend/e2e/setup/global-setup.ts
phoenixAja: i'll ignore for now, but let's try not to make type assertions like this in the future

phoenixAja (approved):
please remove double `NODE_ENV ` in pl:staging script before merging in.

src/frontend/e2e/pages/upload.ts
luisczi: sure, the script cannot insert the text directly to input because is hidden at component initial state, so after you click on it an emergent widget comes in and that reveals the input into the DOM actual state, i can now proceed to locate the input that i need to insert the text since it works with autocomplete it creates a list of elements on the fly so in order to **avoid to create** extra locator i rather press on those keys to select my entry :)

https://user-images.githubusercontent.com/109995236/190459882-89c4c383-9c7d-4938-b59c-cc94c38afa45.mov

src/frontend/package.json
luisczi: yup

src/frontend/package.json
luisczi: these were actually charles modifications but i have removed them now

luisczi: fixed

luisczi:
> 

node env has been removed already

src/frontend/e2e/setup/global-setup.ts
mmmmmmaya: Let's remove the assertions before we merge

src/frontend/e2e/tests/home.spec.ts
```diff
@@ -1,7 +1,10 @@
 import { expect, test } from "@playwright/test";
 import { getByTestID } from "../utils/selectors";
+import path from "path";
+import dotenv from 'dotenv';
+
+dotenv.config({path: path.resolve(`.env.${process.env.NODE_ENV}`),});
```
mmmmmmaya, line 6: is this still needed if you set it in the global setup file?

src/frontend/e2e/tests/sample.spec.ts
mmmmmmaya: Let's fix this one too before we merge. 
We can tell TS what types are in process.env, so we don't need the assertions
https://stackoverflow.com/questions/45194598/using-process-env-in-typescript

src/frontend/e2e/tests/sample.spec.ts
```diff
@@ -30,12 +34,21 @@ test.describe("Samples page tests", () => {
...
+
     //wait until data is displayed
     await page.waitForSelector('[data-test-id="table-row"]');
-
-    // assert table is populated with at least one record
-    expect(
-      await page.locator(getByTestID("table-row")).count()
-    ).toBeGreaterThan(0);
+    await expect(page.locator('text="20SCPH11281"')).not.toBeEmpty();
```
mmmmmmaya, line 41: what are all these magic values? 
how will a developer know if a new test failure is valid or not/when would these change?
won't these pass/fail depending on the environment (for example if we add a bunch of samples to staging)?

src/frontend/e2e/pages/upload.ts
```diff
@@ -0,0 +1,112 @@
...
+    fileName: string
+  ): Promise<void> {
+    await page.setInputFiles(
+      "input[type='file']",
+      path.resolve("e2e/fixtures/sampleData" + fileName)
+    );
+    await page.waitForTimeout(3000);
+    // click continue button
+    await page.locator("a[href$='/upload/step2']").click();
+    await page.waitForSelector("form div[role='table'] > .MuiTableBody-root");
```
mmmmmmaya, line 81: should this be a data-test-id? I'm not clear on whether it makes sense to add them to dynamically generated components, such as every row in a table

src/frontend/e2e/pages/upload.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  public static getSampleData(): Array<SampleData> {
```
mmmmmmaya, line 84: why did you choose to use faker data here, rather than using a fixture like you did for the fasta file?

src/frontend/e2e/pages/upload.ts
```diff
@@ -0,0 +1,112 @@
+import { Page } from "@playwright/test";
```
mmmmmmaya, line 1: I don't feel like I'm clear on when to create a class like this one for use in tests. Will you include that in the documentation?

src/frontend/e2e/pages/upload.ts
phoenixAja: oh i meant a comment in the code 😄 but thank you!

src/frontend/e2e/pages/upload.ts
phoenixAja: we have some follow up tickets to go through and add better test-id selectors. @charles-testco do you have any guidance on maya's question? Should we open a spike ticket on how to add better test-ids to dynamically generated components?

src/frontend/e2e/pages/upload.ts
phoenixAja: i'll make a follow up documentation ticket for this

phoenixAja: https://app.shortcut.com/genepi/story/215675/documentation-when-to-use-classes-for-test-page-objects

charles-testco:
@jgadling @mmmmmmaya I value your feedbacks and the follow up tickets I am working on will address these concerns. Will it be OK if I merge this PR?

phoenixAja:
@charles-testco yeah let's go ahead and merge

PR #1350, opened 2022-09-14
CORS handling to enable Galago usage

vincent-czi:
### Summary:
- **What:** Handle CORS in relation to Galago; bump link expiry time to 48 hours
- **Ticket:** [sc<214181>](https://app.shortcut.com/genepi/story/<214181>)
- **Env:** http://vince-galago-cors-frontend.dev.czgenepi.org/

### Demos:

You can see the CORS change in place for localdev here from Chrome Network tab.

image

### Notes:

- I still need to verify this works in rdev as a final check before ready to merge, but I have no reason to expect it won't work.
- I spent a bit of time looking over the amount of changes that would be necessary to move away from the "auspice" name for the magic link process. It shows up in a lot of places, so I made a note in docs that "auspice" is mostly a historical name but it's really for general tree viewing by external services. I'd like to come back and completely change the name over -- still not sure exactly what to name it though -- but I think doing that and not making any mistakes will take me a couple hours, so I'm going to kick the can on it and hopefully do it at the end of the sprint if there's time then.
  - If anyone has strong opinions that this really should be done now, I'm open to hearing that, I just know I'm not going to get the chance to code much for rest of week with upcoming trainings.
  
### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/backend/aspen/api/main.py
```diff
@@ -32,19 +32,40 @@
...
+    """Allowed origins for cross origin requests. See `CORSMiddleware`.
+
+    Gets list of all allowed origins for cross origin requests. This is pretty
+    simple right now: we only serve the FE of our app and Galago with content
+    from our BE, so that's all this needs to cover. For ease, the Galago URLs
+    are just hardcoded. If we ever need to start handling a lot more CORS
+    requests or the Galago URLs become more dynamic, we should reevaluate
+    current process.
+    """
+    GALAGO_URLS = (
```
jgadling: You can make this a literal list with [] and then not have to use list() below.

src/backend/aspen/api/main.py
```diff
@@ -32,19 +32,40 @@
...
+def get_allowed_origin_regex() -> Optional[str]:
+    """Allowed origin regex for cross origin requests. See `CORSMiddleware`.
+
+    This is here to enable working on Galago in localdev. Since we don't
+    have a standard port to run Galago on in localdev (everybody wants 3000),
+    this is just a catchall for all ports to enable CORS in localdev.
+    """
+    allowed_origin_regex = None  # default case, do not use regex for CORS
+    deployment = os.getenv("DEPLOYMENT_STAGE")
+    if deployment not in ["gestaging", "geprod", "staging", "prod"]:  # =>local
```
jgadling: I think this "deployment" check is kinda silly -- can we just check for the equivalent of "if deployment == 'localdev'" ?

jgadling (approved):

src/backend/aspen/api/main.py
vincent-czi: Yeah, that's a good point. I put it in the tuple because I was worried about the Galago URLs somehow getting mutated, but it's within a function, so I'm being silly.

src/backend/aspen/api/main.py
vincent-czi: Sure! I was just using what was there, I don't really know that aspect very well so I didn't touch the prior code. In the docker-compose for backend, I see `DEPLOYMENT_STAGE=local`, so this should `local` for the check, right? That's the only thing I have to care about here?

jgadling: Yeah, I think that's it.

PR #1360, opened 2022-09-19
remove hardcoded prefix references, pull from DB instead

phoenixAja:
### Summary:
- **What:** remove the hardcoded prefixes from tree identifier logic, pull from the DB instead using the preferred public database with the pathogen
- **Env:** `<rdev link>`

### Checklist:
- [x] I added labels to my PR


src/backend/aspen/api/conftest.py
```diff
@@ -143,7 +144,7 @@ async def auth0_oauth() -> MagicMock:
 
 @pytest.fixture()
 async def split_client() -> MagicMock:
-    return create_autospec(SplitClient)
+    return SplitClient(APISettings())
```
phoenixAja: this is needed to use the values from split.yml in tests

jgadling: Why do we want to use split's localhost functionality instead of mocking split's responses in our tests?

phoenixAja: so that we can pick up the values from `split.yml`, i felt like that was the most 'correct' way to go about this, do you have any potential concerns with this approach?

jgadling: Basically I want to make sure that we can test pathogen-specific functionality when we *truly* have features that only work for certain pathogens, but most of our tests should be able to do the right thing regardless of which pathogen we're working with. That probably means that we need to add repositories/prefixes to the test db when we're creating test pathogens

phoenixAja: ok i can change it back then

src/backend/aspen/test_infra/models/pathogen_repo_config.py
```diff
@@ -1,15 +1,16 @@
...
+from aspen.database.models.usergroup import generate_random_id
 
 
-def setup_gisaid_and_genbank_repo_configs(async_session) -> Pathogen:
+def setup_gisaid_and_genbank_repo_configs(async_session, pathogen=None) -> Pathogen:
     repo_info = {
-        "GISAID": {"prefix": "hCoV-19"},
-        "GenBank": {"prefix": "SARS-CoV-2/human"},
+        "GISAID": {"prefix": generate_random_id(5)},
+        "GenBank": {"prefix": generate_random_id(5)},
```
lvreynoso: I think I can see why you want these prefixes to be random, though I'm not sure I agree. But if you do this, would you mind also fixing the tests that check if the GISAID/GenBank prefix is being dealt with properly? I  see the ones in `test_sequences` and `test_download_tree_sample_ids` are broken.

src/backend/aspen/api/views/tests/test_update_phylo_run_and_tree.py
```diff
@@ -54,7 +61,7 @@ async def make_shared_test_data(
 
     await async_session.commit()
 
-    return user, group, samples, phylo_run, phylo_tree
+    return user, group, samples, phylo_run, phylo_tree, pathogen
```
lvreynoso, line 64: It looks like there are a lot of calls to this function scattered around our tests that need updating to handle this new signature.

src/backend/aspen/api/views/auth.py
```diff
@@ -146,7 +146,9 @@ async def login(
     )
 
 
-async def create_user_if_not_exists(db, userinfo) -> Tuple[User, Optional[Group]]:
+async def create_user_if_not_exists(
+    db, auth0_mgmt, userinfo
+) -> Tuple[User, Optional[Group]]:
```
lvreynoso: I think you need to merge in the latest from `trunk`.

lvreynoso (changes requested):
I'm sure you're aware of the issues I raised, but just in case

src/backend/aspen/api/views/auth.py
lvreynoso: Or eliminate all the changes to users / auth, not sure how they snuck into this PR.

phoenixAja:
@lvreynoso yep I’m changing around a bunch of stuff and am fixing the stuff you mentioned, I’ll convert this or to a draft to make that clear for now, sorry for the confusion

src/backend/aspen/api/utils/pathogens.py
```diff
@@ -0,0 +1,50 @@
...
+        .options(  # type: ignore
+            joinedload(PathogenRepoConfig.public_repository),
+            joinedload(PathogenRepoConfig.pathogen),
+        )
+        .where(
+            and_(
+                PathogenRepoConfig.pathogen == pathogen,
+                PathogenRepoConfig.public_repository.has(
+                    PublicRepository.name == preferred_public_database
+                ),
```
jgadling: I think this can be simplified to `PublicRepository.name == preferred_public_database`

src/backend/aspen/api/utils/pathogens.py
```diff
@@ -0,0 +1,50 @@
...
+            )
+        )
+    )
+    res = await db.execute(q)
+    pathogen_repo_config = res.scalars().one_or_none()
+
+    if pathogen_repo_config:
+        return pathogen_repo_config
```
jgadling: I think this `if` doesn't do anything -- pathogen_repo_config is already an object or None

src/backend/aspen/api/utils/phylo.py
```diff
@@ -276,6 +276,7 @@ async def process_phylo_tree(
     az: AuthZSession,
     phylo_tree_id: int,
     pathogen: Pathogen,
+    pathogen_repo_config,
```
jgadling: Type definition?

src/backend/aspen/api/views/phylo_trees.py
```diff
@@ -28,9 +36,28 @@ async def get_single_phylo_tree(
...
+        "PATHOGEN_public_repository", pathogen
+    )
+    # get the pathogen_repo_config  for given public_repository and pathogen
+    pathogen_repo_config = await get_pathogen_repo_config_for_pathogen(
+        pathogen, preferred_public_db, db
+    )
+    if pathogen_repo_config is None:
+        raise ex.ServerException(
+            "no public repository found for given pathogen public repository"
+        )
```
jgadling: I think you can put this code in a function and use it as a fastapi dependency instead of duplicating code!

src/backend/aspen/test_infra/models/pathogen_repo_config.py
```diff
@@ -1,15 +1,22 @@
...
+    pathogen=None,
+    gisaid_prefix=generate_random_id(5),
+    genbank_prefix=generate_random_id(5),
+) -> Pathogen:
+
     repo_info = {
-        "GISAID": {"prefix": "hCoV-19"},
-        "GenBank": {"prefix": "SARS-CoV-2/human"},
+        "GISAID": {"prefix": gisaid_prefix},
+        "GenBank": {"prefix": genbank_prefix},
```
jgadling, line 15: 😍

jgadling (approved):
A few nits that would be good if you can clean up if/when you get a chance, but nothing blocking, looks good!

src/backend/aspen/api/utils/pathogens.py
phoenixAja: true, i'll take it out

src/backend/aspen/api/utils/pathogens.py
phoenixAja: i had to do this because of the joinedload, i changed it up to be more clear hopefully

src/backend/aspen/api/views/phylo_trees.py
phoenixAja: that's a good idea

phoenixAja:
@lvreynoso do you mind approving so i can merge? If you're ok with this now

lvreynoso (approved):
Ahh thought I already approved this! Sorry about that!

PR #1361, opened 2022-09-19
[SC-215622] Use mocked response for testing sample filtering

luisczi:
### Summary:
- **What:** `test filters api using mock data for samples module`
- **Ticket:** [sc-215622](https://app.shortcut.com/genepi/story/215622/add-mocked-sample-data-to-sample-filtering-e2e-test)
- **Env:** `<[rdev link](https://staging.czgenepi.org/data/samples)>`

### Checklist:
- [ ] I added labels to my PR


src/frontend/e2e/tests/mock-api-filters.spec.ts
```diff
@@ -0,0 +1,104 @@
+import { expect, test } from "@playwright/test";
+import { SampleUtil } from "../utils/sample";
+import { FilterSample } from "../pages/filter";
+
+
+test.describe("Mock sample API data tests", () => {
+    const url = `${process.env.BASEURL}/data/samples/`;
+    const api = `${process.env.BASEAPI}/v2/orgs/${process.env.GROUPID}/pathogens/SC2/samples/`;
+
+  test.only("Should filter by complete genome recovery", async ({ page, context }) => {
```
charles-testco: Why only?

src/frontend/e2e/tests/mock-api-filters.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+      expect(response.ok()).toContainEqual("Complete");
+      route.fulfill({
+        response,
+        body: JSON.stringify(mockData),
+      });
+    });
+    await page.goto(url, { waitUntil: "networkidle" }); //wait until all responses have been received
+  });
+
+  test("Should filter by failed genome recovery", async ({ page, context }) => {
```
charles-testco: I cannot see the actual test. I would though after mocking response we would verify the data on UI, just as we did when hitting real API.

src/frontend/e2e/tests/mock-api-filters.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+      await page.goto(url, { waitUntil: "networkidle" });
+  });
+
+  test("Should filter by  samples", async ({ page, context }) => {
+    const mockData = {
+      sample: SampleUtil.getSampleResponseData(),
+    };
+
+   const daysForFilter = FilterSample.getDaysFromDateRange(mockData.sample.collection_date);
+   const today = new Date();
```
charles-testco: Can this be moved to a local function?

src/frontend/e2e/pages/filter.ts
```diff
@@ -99,6 +99,15 @@ export abstract class FilterSample {
     }
     return filterDate;
   }
+
+  //calculate days between two different dates, takings today as initial date to measure
+  public static getDaysFromDateRange(filteredDate: string){
```
charles-testco: Why do we need this?

src/frontend/e2e/utils/sample.ts
```diff
@@ -52,6 +52,7 @@ export class SampleUtil {
           defaults?.sample?.public_identifier,
           GeneralUtil.generatePublicSampleId()
         ) as string,
+        status: ""
```
charles-testco: Any reason we want the status to be blank?

charles-testco (changes requested):
@luisczi I would like to see actually testing that the response is rendered on correctly on UI

src/frontend/e2e/tests/mock-api-filters.spec.ts
luisczi: forgot to remove it after debuggin

src/frontend/e2e/tests/mock-api-filters.spec.ts
luisczi: no its only a request the one is being made, there no use of using the GUI anymore for this since those where the test were working before already

src/frontend/e2e/tests/mock-api-filters.spec.ts
luisczi: its moved now

src/frontend/e2e/pages/filter.ts
luisczi: i need to create a random past date and used in the request

src/frontend/e2e/utils/sample.ts
luisczi: since i changed to complete/failed i dont need any specific content on it, but i have set it as complete now

src/frontend/e2e/tests/mock-api-filters.spec.ts
ckmoga: @luisczi lets catch up when start work today. Our understanding of the requirement does not align and we can loop in @phoenixAja. If we are not testing frontend, then I see little value in this test

src/frontend/e2e/pages/filter.ts
ckmoga: > 

We already have function for this, have a look utils/common.ts

src/frontend/e2e/utils/sample.ts
ckmoga: hmm. I am not sure why we are setting status to a function that generates for upload data. status is part of response, not upload so I am curious about this change. Either remove it or clarify why this is happening

src/frontend/e2e/pages/filter.ts
ckmoga: > i need to create a random past date and used in the request

Having reviewed this in more detail, I don't agree with this approach.
1. SampleUtil.getSampleResponseData() already gives you date in the past. If it doesn't work, then that is a bug we need to fix without creating another function
2. SampleUtil.getSampleResponseData() optionally takes defaults and if we want to return specific value for any of the attributes we need to feed as input instead of generating and then overwriting.
3. If we really have go that way, then it make more sense to refactor convertDaysToDate function (rename, change parameters and add logic to handle string date) without creating another function that does almost the same job

src/frontend/e2e/tests/mock-api-filters.spec.ts
luisczi: ok understood

luisczi: this has been modified

src/frontend/e2e/pages/filter.ts
luisczi: i have removed this function

src/frontend/e2e/utils/sample.ts
luisczi: it is removed by now

ckmoga:
To be replaced by a new PR

PR #1365, opened 2022-09-21
Maya/214169 galago error

mmmmmmaya:
### Summary
- **What:** Centralize notifications in the frontend by moving them into redux
- **Why:** Notifications from different features won't overlap anymore
- **Ticket:** [[sc-214169]](https://app.shortcut.com/genepi/story/214169)
- **Env:** https://maya-notifs-frontend.dev.czgenepi.org/

### Testing
1. Test flows that generate notifs
1. Test multiple notifications from different features at once (group member invite, NS tree create)

### Notes
View PR with `?w=1` to eliminate whitespace only changes (over 100 lines)

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
```diff
@@ -128,7 +121,12 @@ export const CreateNSTreeModal = ({
 
   const mutation = useCreateTree({
     componentOnError: () => {
-      setShouldShowErrorNotification(true);
+      dispatch(addNotification({
+        reduxId: Date.now(),
```
vincent-czi: It looks like `reduxId` is always intended to be the result of `Date.now()`. I see you're using it as a `key` for displaying the notifications, so I see calling it an id kinda name, but I think something along the lines of `timestamp` or even more verbose `unixEpochTimestamp` or maybe just `dateNowId` would be clearer what's going on. While the dev could override and put anything they want, I think you want this to be a pretty strong convention, right?

src/frontend/src/components/NotificationManager/components/Notification/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { ReduxNotification } from "src/common/redux/types";
+import { CreateNSTreeFailureNotif } from "../CreateNSTreeFailureNotif";
+import { CreateNSTreeSuccessNotif } from "../CreateNSTreeSuccessNotif";
+import { DownloadFilesFailureNotif } from "../DownloadFilesFailureNotif";
+import { SendInviteFailureNotif } from "../SendInviteFailureNotif";
+import { SendInviteSuccessNotif } from "../SendInviteSuccessNotif";
+import { UsherPlacementSuccessNotif } from "../UsherPlacementSuccessNotif";
+
+/**
+ * We should only store things that are directly serializable (strings, numbers, etc) in redux.
```
vincent-czi, line 14: I agree with you so, so much. I have seen components and functions go into redux and it just feels so wrong

src/frontend/src/common/components/library/data_subview/components/DeleteSamplesConfirmationModal/index.tsx
```diff
@@ -34,10 +33,22 @@ const DeleteSamplesConfirmationModal = ({
 
   const deleteSampleMutation = useDeleteSamples({
     componentOnError: () => {
-      setShouldShowErrorNotification(true);
+      dispatch(addNotification({
+        autoDismiss: true,
+        reduxId: Date.now(),
+        intent: "error",
+        shouldShowCloseButton: true,
+        text: "We were unable to delete the selected samples. Please try again later.",
```
vincent-czi, line 40: Is there a clear line in the sand between when it should get a NotificationComponents `componentKey` vs just be a `text` attribute? I skimmed ahead in the PR and didn't see a clear reason why it goes one way or the other yet, might be nice to document if there's a clear litmus test for being one or the other.

src/frontend/src/common/redux/actions/index.ts
```diff
@@ -9,3 +9,18 @@ export const setPathogen: ActionType<Pathogen> = (pathogen) => ({
   type: CZGEReduxActions.SET_PATHOGEN_ACTION_TYPE,
   payload: pathogen,
 });
+
+export const addNotification: ActionType<Partial<ReduxNotification>> = (notification) => ({
```
vincent-czi: Won't `Partial` here mean that all the keys in `ReduxNotification` are optional? Looking at that type, some of them are explicitly marked as optional, so it doesn't seem right to then implicitly mark all of them as optional when making the action. Right now wouldn't it happily accept not getting a `reduxId` which is required downstream by reducer and displayed components?

src/frontend/src/common/redux/index.tsx
```diff
@@ -64,6 +66,16 @@ const reduxReducer = (state = getInitialState(), action: AnyAction) => {
           pathogen: payload,
         },
       };
+    case CZGEReduxActions.ADD_NOTIFICATION_ACTION_TYPE:
```
vincent-czi, line 69: I keep thinking about the `reduxId` aspect, and since every payload has to run its own `Date.now()` call, I think it would be better to remove the id from the payload entirely and just have it be generated here in the reducer. Less boilerplate where the notifications gets dispatched and seems like a bit more separation of concerns.

_Devil's Advocate_: That said, if we ever have a non-notif component that needs to kick off a `DELETE_NOTIFICATION_ACTION_TYPE` to delete, having the id get made in the reducer prevents it from being exposed. If we have a reduxId that gets set in some component (say in the tree creation modal), the modal could theoretically have react state to track the reduxId, then fire off a delete based on specific logic. But I don't think we have any use cases like that right now? Also we could have an optional uuid field and the delete action could be expanded to target reduxId OR uuid matches. ... oh! Hmmm, maybe the best of both worlds would be to just have a default `reduxId: Date.now()` in the payload portion for `addNotification` function? Expected default there, but allows for special case overriding.

src/frontend/src/common/redux/types.ts
```diff
@@ -11,10 +14,20 @@ export enum Pathogen {
...
 }
 
 // persisted names are for use in localstorage
 export enum ReduxPersistenceTokens {
   GROUP = "currentGroup",
   PATHOGEN = "currentPathogen",
 }
+
+export type ReduxNotification = ExposedNotificationProps & {
+  reduxId: number;
```
vincent-czi: I'm not sure if this is the best place to document it, but it feels like somewhere should mention that the typical use case is `Date.now()` because we want these to be unique. But could also put that comment into `addNotification` instead, especially if it gets the boiler plate `Date.now()` call instead of being in the various dispatches in the components.

src/frontend/src/components/NotificationManager/components/Notification/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const dispatch = useDispatch();
+  const { reduxId, componentKey, componentProps, shouldShowCloseButton, text, ...rest } = notification;
+
+  const onDismiss = () => {
+    dispatch(deleteNotification(reduxId));
+  };
+
+  let children: ReactNode = text;
+
+  if (!children && componentKey) {
```
vincent-czi, line 55: It would be a developer error to pass both `text` and `componentKey`, right? You only use one or the other, never both, right? Feels like that should be documented somewhere or enforced by TS. Looks like you can enforce something like that with [some ugly TS](https://stackoverflow.com/a/60489977), but I'd also be perfectly happy with just a bit of documentation on `addNotification` func.

src/frontend/src/components/NotificationManager/style.ts
```diff
@@ -1,10 +1,6 @@
 import styled from "@emotion/styled";
 import { CommonThemeProps, getSpaces } from "czifui";
 
-// TODO (mlila): when we add FE state management, we should have one piece of
```
vincent-czi, line 4: 🎉

src/frontend/src/views/GroupMembersPage/components/MembersTab/components/InviteModal/index.tsx
```diff
@@ -57,12 +53,30 @@ const InviteModal = ({
...
+          }
+        }));
+      }
 
       // show success for any invites we did send
       const successCount = invitations.length - failedInvites.length;
       setSentCount(successCount);
-      if (successCount > 0) setIsSuccessNotificationOpen(true);
+      if (successCount > 0) {
+        dispatch(addNotification({
```
vincent-czi, line 72: There's no `reduxId` on this one -- is that a problem, or does the `autoDismiss` make things okay? Would it prevent an early dismiss because it doesn't have a reduxId?

vincent-czi (approved):
Overall, looks great! Really like the approach and very easy to extend with new notifications as we build new stuff. I've got some comments that I would like to see addressed, but if you disagree strongly with any of them, reasonable people can have differences of opinion and it would be okay to merge as is.

I did _not_ do any manual testing in the rdev though, so I can't speak to verifying in that aspect. But code-wise everything looks quite good!

src/frontend/src/common/components/library/data_subview/components/DeleteSamplesConfirmationModal/index.tsx
mmmmmmaya: Basically, `text` is if your notification only requires text to be shown. if it needs anything more complicated than that, for example if you need to also display a link or a list of IDs, then you make a component for the content that will be displayed in the notification and use `component Key`

src/frontend/src/views/GroupMembersPage/components/MembersTab/components/InviteModal/index.tsx
mmmmmmaya: it would prevent the notif from being dismissed. just added

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
mmmmmmaya: That's the best convention I can think of for generating more or less unique ID (seems pretty unlikely two would be generated within the same millisecond, since notifs are always the result of user action and often involve travel over the network). I think it is important for the name of the prop to reflect what it represents. It's true that reduxId is a timestamp, but the point is that it uniquely identifies the notification.

mmmmmmaya: I can change it to notifId or something similar (unfortunately, the exposedprops that reduxnotification does not allow the use of `id` directly)

src/frontend/src/common/redux/index.tsx
mmmmmmaya: Redux reducers are supposed to be 100% pure and always return the same state given the same input. As such, I do not want to add it to the reducer. But I think it would be ok to add it to the action with the other default props. That would let it still be overwritable.

src/frontend/src/common/redux/types.ts
```diff
@@ -1,11 +1,13 @@
 import { ExposedNotificationProps } from "czifui";
 import { NotificationComponents } from "src/components/NotificationManager/components/Notification";
 
-export type ActionType<T> = (payload?: T) => {
-  type: CZGEReduxActions;
-  payload?: T;
+export type ComplexActionType<T, U> = (payload: T) => {
```
vincent-czi, line 4: So `ComplexActionType` is for one that mutates the incoming payload from "raw" input payload `T` to "cooked" output payload `U`?

vincent-czi (approved):

src/frontend/src/common/redux/types.ts
mmmmmmaya: yep

PR #1367, opened 2022-09-22
211446 local pathogen traffic

mmmmmmaya:
### Summary
- **What:** Add configuration of pathogen traffic type feature flags to the local front end
- **Why:** so that the correct features will show up for the correct pathogen locally
- **Ticket:**
  - [[sc-211276]](https://app.shortcut.com/genepi/story/211276)
  - [[sc-211446]](https://app.shortcut.com/genepi/story/211446)
- **Env:** https://maya-split-frontend.dev.czgenepi.org/

### Notes
- These configurations only apply locally. In other envs, the on/off values for each feature will be pulled from split
- There will be another PR to create new Split client for the purpose of rendering components depending on the traffic type
- TS should give an error (thereby warning a developer) if they try to add a new traffic=pathogen feature flag without also configuring the local split values for that flag

### Checklist
- [X] I merged latest `trunk`
- [x] I added labels to my PR


src/frontend/src/common/redux/types.ts
```diff
@@ -9,7 +9,7 @@ export type ComplexActionType<T, U> = (payload: T) => {
 export type ActionType<T> = ComplexActionType<T, T>;
 
 export enum Pathogen {
-  COVID = "covid",
+  COVID = "SC2",
```
vincent-czi, line 12: Is there a reason we don't match the BE name entirely? I definitely see downsides, either way, but it feels slightly better to me to standardize the BE, FE, and Split flags all on a single name for each pathogen. Any name mismatch confusion around this anywhere in the chain seems pretty bad so that's why I'm thinking this.

src/frontend/src/components/Split/index.tsx
```diff
@@ -17,42 +17,30 @@
...
- * To use a new flag
- *   1) Create the flag in Split, add it to the `FEATURE_FLAGS` enum.
+ * To use a new flag for `user` traffic type,
+ *   1) Create the flag in Split, add it to the `USER_FEATURE_FLAGS` enum.
  *      * If it's a simple flag of just "on"/"off", local dev defaults to on
  *   2) To pull the flag in a component in app, do the following
  *        import { useTreatments } from "@splitsoftware/splitio-react";
- *        import { FEATURE_FLAGS } from <<This file right here>>;
- *        const flag = useTreatments([FEATURE_FLAGS.my_flag_name]);
+ *        import { USER_FEATURE_FLAGS } from <<This file right here>>;
```
vincent-czi:
```suggestion
 *        import { USER_FEATURE_FLAGS } from <<file in this dir>>;
```

src/frontend/src/components/Split/index.tsx
```diff
@@ -106,6 +96,7 @@ interface Props {
  * functionality of our dev server because the config won't be rebuilt.
  */
 const SplitInitializer = ({ children }: Props): JSX.Element | null => {
+  const pathogen = useSelector(selectCurrentPathogen);
```
vincent-czi, line 99: It looks like -- for localdev -- the local config is basically locked to the values based on the originally loaded pathogen? I'm pretty sure what I'm seeing below will only prep and load the localdev config based on the pathogen the app is on during load. But if the pathogen changes, it will _not_ do any kind of new config, the config is basically locked until next browser refresh.

I know you're going to do another PR later to address the pathogen aspect more specifically, will that cover having the localdev config switch as part of pathogen switching? I could also believe that's just a huge hassle the way Split has things built and not really doable, but it would certainly be nice if we could.

vincent-czi (approved):
Looks good!

src/frontend/src/common/redux/types.ts
mmmmmmaya: SC2 is the backend name, and also the split name, I believe.

src/frontend/src/components/Split/index.tsx
mmmmmmaya: I will take this into consideration in the next PR!

PR #1368, opened 2022-09-23
[SC-215051] Add tdata-test-id to elements

charles-testco:
### Summary:
- **What:** `This PR adds test Ids to elements for automation. It also includes a babel plugin to strip test Ids from production. For testing purpose, I am changing this to staging for testing purpose. Once we see the Ids are stripped of staging, then we can raise another PR to change this production.
- There are cases I could not find a way to inject test Ids. In those cases will use a combination of other attributes like name, label etc.`
- **Ticket:** [sc-215051(https://app.shortcut.com/genepi/story/sc-215051)
- **Env:** `<rdev link>`

### Checklist:
- [x] I added labels to my PR


phoenixAja (approved):
looks good to me!

phoenixAja:
we should write some documentation on this: 
when unable to inject test Ids use a combination of other attributes like name, label etc.

I'll write a ticket

charles-testco:
> we should write some documentation on this: when unable to inject test Ids use a combination of other attributes like name, label etc.
> 
> I'll write a ticket

Yes, indeed we should

PR #1381, opened 2022-10-04
[SC-190388] as a user i want to select a custom location for tree creation

ehoops-cz:
### Summary:
- **What:** `select a custom location for tree creation`
- **Ticket:** [sc190388](https://app.shortcut.com/genepi/story/190388)
- **Env:** `<rdev link>`

### Demos:
Screen Shot 2022-10-04 at 11 51 07 AM
Screen Shot 2022-10-04 at 11 51 01 AM
Screen Shot 2022-10-04 at 11 50 54 AM

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/CollectionDateFilter/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/CollectionDateFilter/index.tsx
@@ -1,6 +1,11 @@
 import { useState } from "react";
+import { noop } from "src/common/constants/empty";
 import { getDateRangeLabel } from "src/common/utils/dateUtils";
-import { MENU_OPTION_ALL_TIME } from "src/components/DateFilterMenu/constants";
+import {
+  MENU_OPTIONS_COLLECTION_DATE,
+  MENU_OPTION_ALL_TIME,
+} from "src/components/DateFilterMenu/constants";
+import { StyledFilterGroup, StyledFilterGroupName } from "../../style";
 import { StyledDateFilterMenu, StyledInputDropdown } from "./style";
 
 export type StartDateFilterType = {
@@ -13,20 +18,13 @@ export type EndDateFilterType = {
   setEndDate: (d: FormattedDateType) => void;
 };
 
-interface Props extends StartDateFilterType, EndDateFilterType {
-  fieldKeyEnd: string;
-  fieldKeyStart: string;
-  menuOptions: DateMenuOption[];
-  updateDateFilter: UpdateDateFilterType;
-}
+interface Props extends StartDateFilterType, EndDateFilterType {}
 
 const CollectionDateFilter = ({
-  updateDateFilter,
   startDate,
   endDate,
   setStartDate,
   setEndDate,
-  ...props
 }: Props): JSX.Element => {
   const [anchorEl, setAnchorEl] = useState<HTMLElement | undefined>();
   // What menu option is chosen. If none chosen, `null`.
@@ -48,7 +46,8 @@ const CollectionDateFilter = ({
   });
 
   return (
-    <>
+    <StyledFilterGroup>
+      <StyledFilterGroupName>Collection Date</StyledFilterGroupName>
       <StyledInputDropdown
         sdsStyle="square"
         sdsType="singleSelect"
@@ -57,16 +56,19 @@ const CollectionDateFilter = ({
         onClick={handleClick}
       />
       <StyledDateFilterMenu
-        {...props}
         anchorEl={anchorEl}
+        data-test-id="collection-date"
+        fieldKeyEnd="collectionDateEnd"
+        fieldKeyStart="collectionDateStart"
+        menuOptions={[...MENU_OPTIONS_COLLECTION_DATE, MENU_OPTION_ALL_TIME]}
         onClose={handleClose}
         onStartDateChange={setStartDate}
         onEndDateChange={setEndDate}
         selectedDateMenuOption={selectedDateMenuOption}
         setSelectedDateMenuOption={setSelectedDateMenuOption}
-        updateDateFilter={updateDateFilter}
+        updateDateFilter={noop}
       />
-    </>
+    </StyledFilterGroup>
   );
 };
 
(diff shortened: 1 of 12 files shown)


src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/LineageFilter/index.tsx
```diff
@@ -0,0 +1,284 @@
+import { FilterOptionsState, PopperProps } from "@mui/material";
```
ehoops-cz, line 1: I moved this to a new file, but nothing here changed.

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
```diff
@@ -79,10 +84,32 @@ export const CreateNSTreeModal = ({
...
+    );
+
+  // If the group call isn't back when this is loaded, we need to update when the
+  // call returns
+  useEffect(() => {
+    setSelectedLocation(
+      groupInfo?.location ? foldInLocationName(groupInfo?.location) : null
+    );
+  }, [groupInfo?.location]);
+
```
ehoops-cz, line 111: I'm not sure why other places in the app can use useGroup without putting it in a useEffect. 

This looks awkward, open to suggestions.

src/frontend/src/common/queries/trees.ts
```diff
@@ -73,6 +75,7 @@ async function createTree({
       filter_start_date: startDate,
       filter_end_date: endDate,
       filter_pango_lineages: lineages,
+      filter_location_id: location?.id,
```
ehoops-cz: @vincent-czi is this the correct payload?

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/CollectionDateFilter/index.tsx
```diff
@@ -51,16 +56,19 @@ const CollectionDateFilter = ({
...
+        fieldKeyEnd="collectionDateEnd"
+        fieldKeyStart="collectionDateStart"
+        menuOptions={[...MENU_OPTIONS_COLLECTION_DATE, MENU_OPTION_ALL_TIME]}
         onClose={handleClose}
         onStartDateChange={setStartDate}
         onEndDateChange={setEndDate}
         selectedDateMenuOption={selectedDateMenuOption}
         setSelectedDateMenuOption={setSelectedDateMenuOption}
-        updateDateFilter={updateDateFilter}
+        updateDateFilter={noop}
```
vincent-czi, line 69: I don't understand how this became a `noop`? I don't see a corresponding change to `DateFilterMenu`, so is it that we can now handle everything just with `onStartDateChange={setStartDate}` and equivalent endDate?

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/LocationFilter/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+interface LocationFilterProps
+  extends LocationFilterType,
+    LocationFilterStyleProps {}
+
+interface AutocompleteState {
+  getOptionLabel: (option: NamedGisaidLocation) => string;
+  inputValue: string;
+}
+
+export const LocationFilter = ({
```
vincent-czi, line 33: It looks like everything in here is a pretty close match to how the `LocationField` component in the metadata upload/edit tables works. I see there are a few differences and I wouldn't be surprised if it's hard to refactor the old `LocationField` component to generalize it, but I am worried about if we start to have a bunch of these similar-but-slightly-different versions of the location picker start to show up. Maybe just throw in some docs/comments here and in the other one pointing out that, if we keep building location pickers like this, we need to refactor and extract them into some shareable core component.

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
vincent-czi: Hmmm, I'm not sure what would be causing that. Is it that you set something off the groupInfo once it gets loaded? @mmmmmmaya might have a better idea

src/frontend/src/common/queries/trees.ts
vincent-czi: Not quite: there's no word "filter" on the location aspect, it should be **just `location_id`**. So replace `filter_location_id` here and in the `CreateTreePayload` interface with just `location_id`. (The ticket names around this work are kind of a misnomer being called tree location filter. All the trees we build have precisely one location associated with them, it's just that right now the user has no control over what that location is. So it's not really a filter so much as it's just a way to set a template arg that would otherwise get defaulted. That's the reasoning behind why I named it just `location_id`, if you're curious.)

vincent-czi (approved):
Looks good! I'm approving because I don't need to re-review, but there is one **critical change required** around renaming the payload param to `location_id`, see the comment for more.

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/CollectionDateFilter/index.tsx
ehoops-cz: `updateDateFilter` isn't used here.  This still works because it's calling `setStartDate` and `setEndDate` directly.  I didn't change the implementation here, but I did move some of the props and typescript expects `updateDateFilter` to be passed in.

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/LocationFilter/index.tsx
ehoops-cz: Sounds good.  I think refactoring should wait until the 500+ sample project is complete.  I'll comment for now.

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
ehoops-cz: I do set something off of group data once it's loaded, but I am under the impression this is also what happens in other components, which seem to work without this.  😕

src/frontend/src/common/queries/trees.ts
ehoops-cz: got it, that makes sense.  I was wondering about that.

PR #1383, opened 2022-10-05
SC-214195/reset tree creation filters

ehoops-cz:
### Summary: (https://github.com/chanzuckerberg/czgenepi/pull/1381)
- **What:** `reset tree creation filters`
- **Ticket:** [sc214195](https://app.shortcut.com/genepi/story/214195)
- **Env:** [rdev](https://filter-reset-frontend.dev.czgenepi.org/)

### Demos:
Screen Shot 2022-10-05 at 1 19 25 PM
Screen Shot 2022-10-05 at 1 19 13 PM
Screen Shot 2022-10-05 at 1 18 59 PM

### Checklist:
- [x] I added labels to my PR


src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/LocationFilter/index.tsx
```diff
@@ -83,7 +83,9 @@ export const LocationFilter = ({
         namedLocations.find(
           (location) => location.name === newSelectedOptions.name
         ) || null;
-      setSelectedLocation(newLocation);
+      if (newLocation && newLocation.id !== selectedLocation?.id) {
+        setSelectedLocation(newLocation);
+      }
```
ehoops-cz, line 88: We need to avoid calling setSelectedLocation if there is no change because if we make the extra call the parent will think that a filter is enabled and the reset button will show when it's not supposed to.

Janeece (approved):
LGTM, thanks for the copy update for Targeted trees!

mmmmmmaya:
probably not necessary to fix in this PR, but for future work with this component, we probably want to start thinking about doing a minor refactor to break out each of the components in RadioLabel/index into their own files.

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
```diff
@@ -119,6 +133,37 @@ export const CreateNSTreeModal = ({
     USER_FEATURE_FLAGS.tree_location_filter
   );
 
+  const handleFilterChange = (setFilter: () => void): void => {
```
vincent-czi: Nit: This is minor, but I think I'd prefer the name of the arg to be `onChange` or something else a bit more generic? It took me awhile to realize `setFilter` wasn't actually setting some filter it was being given, it was just a callback function and the intention was it to be a callback function that would handle the setting of filters.

vincent-czi (approved):
Looks good!

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/index.tsx
```diff
@@ -68,25 +73,37 @@ export function SampleFiltering({
...
-          <SampleFilteringTooltip />
-        ) : (
-          <StyledTooltip
-            arrow
-            leaveDelay={1000}
-            title={SAMPLE_FILTERING_TOOLTIP_TEXT}
-            placement="top"
+      <StyledTitleContainer>
+        <StyledExplainerTitle>
+          {isTreeLocationFilterFlagOn
```
mmmmmmaya: Instead of having 3 conditionals in a row that checks for `isTreeLocationFilterOn`, can we condense it into one? I think that might make it much easier to see what is displayed in each case

ehoops-cz:
Fixed the text overflow for long locations on the targeted tree:
Screen Shot 2022-10-07 at 8 20 55 AM

Going to wait on updates to the blue focus border since it's not a blocker and there are dropdown changes currently happening in sds.  Once we have the most up to date changes, I'll circle back.

PR #1386, opened 2022-10-07
[SC-214191] nextstrain tree location analytics

ehoops-cz:
### Summary:
- **What:** `Add location_id to NextStrain tree creation event`
- **Ticket:** [sc214191](https://app.shortcut.com/genepi/story/214191)
- **Env:** [rdev link](https://location-filter-analytics-frontend.dev.czgenepi.org)

### Demos:
Screen Shot 2022-10-07 at 9 50 05 AM
Screen Shot 2022-10-07 at 9 49 30 AM
Screen Shot 2022-10-07 at 9 49 01 AM

### Checklist:
- [x] I added labels to my PR


src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/index.tsx
```diff
@@ -15,6 +15,7 @@ import {
   LocationFilter,
   LocationFilterType,
 } from "./components/LocationFilter";
+
```
ehoops-cz, line 18: prettier 🙄

src/frontend/src/common/analytics/eventTypes.ts
```diff
@@ -128,6 +128,8 @@ export type AnalyticsTreeCreationNextstrain = {
   phylo_run_workflow_id: number;
   // Type of tree being created
   tree_type: string;
+  // Location of samples used for tree creation
+  location_id?: number;
```
ehoops-cz: @vincent-czi The location can be null if the groups call fails for some reason.  It sounds like this will be ok for the API call.  Just a heads up that we should leave the optional logic in there to account for any issues fetching the group.  

The analytics also has this as optional for the same reason.  Does this seem ok?

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
```diff
@@ -217,7 +217,9 @@ export const CreateNSTreeModal = ({
           phylo_run_workflow_id: respData.id,
           // Safe to assert treeType is not undefined here, can't create tree
           // otherwise, it's just that checking happens in a child component.
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
mmmmmmaya, line 220: why is this override needed now, when it was working before?

src/frontend/src/common/analytics/eventTypes.ts
ehoops-cz: Per Vincent's recommendation, updated this to explicitly pass null if there isn't a value.  Also added a comment to the analytics event page to indicate why this might be null.

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/index.tsx
ehoops-cz: There was always a warning in the list of things we generally ignore.  This just makes the warning go away and seems ok based on the comment above.

vincent-czi (approved):

PR #1387, opened 2022-10-07
migration for nextclade output

phoenixAja:
### Summary:
- **What:** migration to add nextclade output tables
- **Tickets:** 
[sc217876](https://app.shortcut.com/genepi/story/217876)
[sc217869](https://app.shortcut.com/genepi/story/217869)
[sc217871](https://app.shortcut.com/genepi/story/217871)
[sc219627](https://app.shortcut.com/genepi/story/219627)
[sc217849](https://app.shortcut.com/genepi/story/217849)
- **Env:** `<rdev link>`

### Checklist:
- [x] I added labels to my PR


src/backend/database_migrations/versions/20221007_170744_nextclade_output_tables.py
```diff
@@ -0,0 +1,112 @@
...
+        sa.PrimaryKeyConstraint("entity_id", name=op.f("pk_aligned_peptides")),
+        sa.UniqueConstraint(
+            "s3_bucket", "s3_key", name="uq_aligned_peptides_s3_bucket_key"
+        ),
+        schema="aspen",
+    )
+    op.create_table(
+        "sample_mutations",
+        sa.Column("id", sa.Integer(), autoincrement=True, nullable=False),
+        sa.Column("sample_id", sa.Integer(), nullable=True),
```
jgadling: This shouldn't be nullable I think? there's no point having a row in this table (or most of the other tables in this PR) if they're not associated with a sample

src/backend/database_migrations/versions/20221007_170744_nextclade_output_tables.py
```diff
@@ -0,0 +1,112 @@
...
+down_revision = "20221004_221741"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "aligned_peptides",
+        sa.Column("sample_id", sa.Integer(), nullable=True),
+        sa.Column("entity_id", sa.Integer(), nullable=False),
```
jgadling, line 22: Just curious about using the `entity_id` field name here. We're using a different field naming convention where `uploaded_pathogen_genomes` inherits from `pathogen_genomes`

jgadling: or do we need an entity_id at all here actually? it's not in the diagram

src/backend/database_migrations/versions/20221007_170744_nextclade_output_tables.py
phoenixAja: yep! I agree, fixed

src/backend/database_migrations/versions/20221007_170744_nextclade_output_tables.py
```diff
@@ -0,0 +1,112 @@
...
+        schema="aspen",
+        type_="unique",
+    )
+    op.drop_constraint(
+        "fk_sample_qc_metrics_qc_type_qc_types",
+        "sample_qc_metrics",
+        schema="aspen",
+        type_="foreignkey",
+    )
+    op.drop_column("sample_qc_metrics", "qc_type", schema="aspen")
```
jgadling, line 106: There's a unique index on sample_qc_metrics.qc_score that doesn't make sense I think? Should we also drop that in this migration?

jgadling (approved):
added some comments, merge when you're happy with it.

src/backend/database_migrations/versions/20221007_170744_nextclade_output_tables.py
phoenixAja: i had it inheriting from entity, i'll remove it since we're moving away from that

src/backend/database_migrations/versions/20221007_170744_nextclade_output_tables.py
phoenixAja: so this was picked up from the auto generate from alembic, i decided to leave it in since it aligns with what's in the sqlalchemy model and we're hooking up these tables to use for real.

jgadling: yeah, this index needs to go away because the qc_type field is going to go away, but there's *another* unique index on the table (not mentioned in this PR) that looks weird to me.

phoenixAja: oh yeah that's what i was talking about, the sample_qc_metrics.qc_score was picked up from the automigration (even though i didn't touch that in this pr) and i decided to leave it in, I'll take it out though since it's not pertinent to this PR. I was just keeping it in since i'm planning on using that table soon and it seemed like a tiny bit of cleanup to sync up the model code and the DB

phoenixAja: oh nevermind misread your first comment, i can drop that constraint

PR #1393, opened 2022-10-14
Pin pangolin version in install_pangolin script

lvreynoso:
### Summary:
- **What:** Pangolin has updated with a vital bugfix, but docker is still using a cached version. This pins the version of pangolin we use by cloning a specific tag of pangolin, which will bust the docker cache whenever we update it.
- **Env:** `<rdev link>`

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/src/backend/Dockerfile.pangolin
+++ b/src/backend/Dockerfile.pangolin
@@ -19,7 +19,7 @@ RUN eval "$($HOME/miniconda/bin/conda shell.bash hook)" && conda init
 
 # install pangolin
 COPY aspen/workflows/pangolin/install_pangolin.sh .
-RUN bash install_pangolin.sh && echo "uncache"
+RUN bash install_pangolin.sh
 
 # Poetry: install app
 COPY pyproject.toml poetry.lock environment.yaml ./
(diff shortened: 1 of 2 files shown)


danrlu:
Thank you!!! This will pin pangolin version, then will lineage definition still gets refreshed everyday (the version we use to check whether a sample is using the latest lineage definition)? I can try look that up but you are more familiar with it, so want to check first.

danrlu (approved):

lvreynoso:
The scheduled job will still compare samples against the latest pangolearn version. So if there's a newer version than what the sample was run with, the scheduled job will make an updated lineage call.

I am a little worried that the samples that broke our ondemand job still don't have a lineage call, so I'm wondering if there's a bug in the scheduled job and they're not being picked up.

jgadling (approved):

danrlu:
That 2nd issue I feel we've seen it before. What is the pangolearn version for those un-run samples in the database?

lvreynoso:
The `pangolin_version` column is null, but I looked at some cloudwatch logs and I think I found the issue...the way we handle sample identifiers is fragile and breaking on public identifiers with spaces in them, so `export.py` isn't running, so no samples are being run through pangolin. That will be a separate fix, I'll work on it rn

PR #1396, opened 2022-10-17
Change nextstrain workdir to /nextstrain

lvreynoso:
### Summary:
- **What:** Our nextstrain images are failing to push in CI because they are failing tests; they are failing tests because the `nextstrain` user doesn't have write access to `/usr/src/app`. `/usr` and its subdirectories should never be used as working directories anyways, so this changes the working directory to `/nextstrain`, which is `$HOME` for the nextstrain user. 
- **Env:** `<rdev link>`

### Notes:
In the longer term, we should make sure the working directories for all our dockerfiles are not in `/usr` or its subdirectories, not run things as root, and make sure everything works with those changes.

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/.happy/terraform/modules/sfn_config/nextstrain-autorun.wdl
+++ b/.happy/terraform/modules/sfn_config/nextstrain-autorun.wdl
@@ -47,8 +47,7 @@ task nextstrain_workflow {
     ncov_git_rev=""
 
     # run main workflow
-    cd /usr/src/app/aspen/workflows/nextstrain_run
-    ./run_nextstrain_autorun.sh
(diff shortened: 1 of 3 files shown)


jgadling:
Are there any steps in our pipeline that need to change now that we're using a different base directory?

lvreynoso:
I just looked, and we use absolute references everywhere except for the WDL scripts, where we change directories then run the scripts (and I updated the references to be absolute), and this [run_test.sh](https://github.com/chanzuckerberg/czgenepi/blob/trunk/src/backend/aspen/workflows/nextstrain_run/run_test.sh) file I can't find any reference to.

I am happy to deploy an rdev and run a nextstrain workflow to test

jgadling (approved):

PR #1397, opened 2022-10-17
[SC-224761] Add analytics fields.

ehoops-cz:
### Summary:
- **What:** `Include other filters and group id in ns tree creation analytics event`
- **Ticket:** [sc224761](https://app.shortcut.com/genepi/story/224761)
- **Env:** [`<rdev link>`](https://tree-analytics-frontend.dev.czgenepi.org/data/samples/groupId/17/pathogen/SC2)

### Demos:
Screen Shot 2022-10-18 at 7 57 02 AM

### Checklist:
- [x] I added labels to my PR
- [-] I added relevant unit tests
- [-] I have notified others of changes they need to make locally (migrations, jobs, package updates, etc)


src/frontend/src/common/analytics/eventTypes.ts
```diff
@@ -129,9 +129,24 @@ export type AnalyticsTreeCreationNextstrain = {
...
+  // the user's group info was not successfully fetched before they created
   // a tree. This generally shouldn't happen.
   location_id: number | null;
+  // Location of the user's group. This is provided for comparison with the
+  // location_id above. A null value indicates that the user's
+  // group was not successfully fetched before they created a tree. This
+  // generally shouldn't happen.
+  group_location_id: number | null;
+  // Lineages selected for tree creation.
+  selected_lineages: string;
```
vincent-czi: For things that are specifically JSON strings, I prefer to use `JsonString` as the type, and then also have the comment document what should be inside the JSON. I assume this is a JSON array of strings, where each string is a lineage?

src/frontend/src/common/analytics/eventTypes.ts
```diff
@@ -129,9 +129,24 @@ export type AnalyticsTreeCreationNextstrain = {
...
+  // Time range filter for tree creation. A null value indicates that
+  // the user did not change the default value and there is
+  // no start date set.
+  start_date: FormattedDateType | null;
```
vincent-czi, line 146: `FormattedDateType` is generally a YYYY-MM-DD string, right? It looks like where the `FormattedDateType` is defined, it doesn't actually say what the format of the string is, but I think it must be YYYY-MM-DD. That should probably be documented there, although I know you didn't make that type.

Also... I'm not actually sure what will happen when this lands in the DB. It's a new field, so the DB Warehouse service will autogenerate a migration to add this new column, but I don't know if it will make a string or a Date out of it in the DB. ... guess we'll find out? It's probably not a big deal either way, I assume this isn't a critical field, and it should be possible to cast the string to a date during analytics if it lands as a date.

src/frontend/src/common/analytics/eventTypes.ts
vincent-czi: - [x] Convert type to `JsonString`

vincent-czi (approved):
Overall looks good, although there's at least one comment I definitely want to see addressed. Gonna take that PR TODO bot out for a spin!

src/frontend/src/common/analytics/eventTypes.ts
ehoops-cz: The bot's not here yet, but now I'm tempted to hook it up before I get back to this one 😅

ehoops-cz: @vincent-czi this is actually a single string created from the list of lineages.  Setting the type to `selected_lineages: string[];` causes a type error 
Screen Shot 2022-10-17 at 2 57 05 PM

I actually think it comes from here:
Screen Shot 2022-10-17 at 2 58 31 PM

Does it make more sense to update EventValue to also include `string[]` rather than converting the list to a string?

src/frontend/src/common/analytics/eventTypes.ts
ehoops-cz: I assumed it would land as a string, but I don't have any basis for this assumption.  I'll check with Amy - I assume she's the only person who might care.  I can also add to the comment to describe the format.

src/frontend/src/common/analytics/eventTypes.ts
ehoops-cz: Resolved on a call - using type "JsonString" to indicate to consumers that this is json.  Still using `json.stringify` to change from list to string.

PR #1398, opened 2022-10-17
Change gisaid workdir to /nextstrain

lvreynoso:
### Summary:
- **What:** The `nextstrain/base` dockerfile now uses the `nextstrain` user by default, which does not have write access to the `/usr/src/app` directory, which we have been using as a working directory. This changes the working directory of the container to `/nextstrain`, which is the `$HOME` for the nextstrain user. This is not just required for CI tests to pass, but for our workflow to function, as we do a lot of file manipulation in the working directory in the gisaid workflow. 
- **Env:** `<rdev link>`

### Checklist:
- [x] I added labels to my PR


jgadling (approved):

lvreynoso:
I've looked through the WDL file and we should be good as far as directory paths go.

PR #1406, opened 2022-10-18
add lineage_qc dockerfile, install nextclade

phoenixAja:
### Summary:
- **What:** add a new dockerfile for nextclade and pangolin, leave the old pangolin image for now until we're ready to switch it over. 
- **Ticket:**  [sc217893](https://app.shortcut.com/genepi/story/217893)
- **Env:** `<rdev link>`

### Checklist:
- [x] I added labels to my PR


src/backend/Dockerfile.lineage_qc
```diff
@@ -0,0 +1,40 @@
+FROM python:3.9.1-slim AS base
```
jgadling: Can you use python:3.10-slim instead? I've been meaning to upgrade the other images now that nextstrain is on a newer python, but haven't had a chance. At least if we build this one on the newer version we won't have to worry about fixing it when we upgrade.

src/backend/install_nextclade.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+eval "$($HOME/miniconda/bin/conda shell.bash hook)"
```
jgadling: Any reason we're using conda to install nextclade instead of using their cli-download option?
https://docs.nextstrain.org/projects/nextclade/en/stable/user/nextclade-cli.html#download-from-command-line

.github/workflows/push-tests.yml
```diff
@@ -217,6 +217,9 @@ jobs:
           - dockerfile: src/backend/Dockerfile.pangolin
             context: ./src/backend/
             name: genepi-pangolin
+          - dockerfile: src/backend/Dockerfile.lineage_qc
+            context: ./src/backend/
+            name: genepi-lineage-qc
```
jgadling, line 222: Note - this means we'll need to add the genepi-lineage-qc ECR repository to genepi-infra *before* this pr gets merged.

jgadling (changes requested):
Looks good overall - I'd prefer to avoid using conda as much as possible, and I'm flagging this as "don't merge" until we get the new ECR created.

src/backend/Dockerfile.lineage_qc
phoenixAja: oh sure!

src/backend/install_nextclade.sh
phoenixAja: i thought the conda install was cleaner than using curl, plus we already have to install miniconda for pangolin so might as well

.github/workflows/push-tests.yml
phoenixAja: i think we do have that ecr repository created already [here](https://github.com/chanzuckerberg/genepi-infra/blob/9affbe5cd2da4db089a5527f340d1dfb18eca67e/terraform/modules/genepi/main.tf#L32)

phoenixAja:
i disagree on not using conda, to me it's a nice package manager with little overhead, and it seems cleaner to use conda to install rather than using curl

src/backend/install_nextclade.sh
jgadling: I'd really prefer to avoid using conda as much as possible. If we get into a situation where we just *have* to manage conda environments due to conflicting package requirements, we can deal with it then, but in the meantime the curl installation is pretty clean (it's how we install miniconda already 😢 )

src/backend/Dockerfile.lineage_qc
phoenixAja: hmm that seems to break the poetry install..

phoenixAja: we might have to wait on upgrading for now until we have more time to debug if that's ok, maybe it can be a follow up PR

phoenixAja:
@jgadling i updated based on your feedback. I used curl to install nextclade and also updated the base image to python:3.10-slim

jgadling (approved):

PR #1407, opened 2022-10-19
Rdev vince analytics upload flow

vincent-czi:
### Summary:
- **What:** Adds a UUID to various steps of Upload flow so we can correlate when doing analytics
- **Ticket:** [sc<210828>](https://app.shortcut.com/genepi/story/<210828>)
- **Env:** https://vince-analytics-upload-flow-frontend.dev.czgenepi.org/

### Demos:
New event showing up in Segment debugger

image

---

New field of `upload_flow_uuid` added to previously existing upload success event, correlates all the events of a single upload "flow".

image
 

### Notes:

Most of the pain for doing this feature was wrapped up in problems with double-mounts causing the first UUID to get briefly used then dumped. I've seen this problem before and [filed an issue](https://app.shortcut.com/genepi/story/204578/double-mounts-throughout-fe-app); having to fight this again here is making me think it might actually be worth spending some time to diagnose and fix the issue. That said, I'm also starting to have a sneaking suspicion that it's the way we handle our routing which causes the apparent de-mount+re-mount "jitter", and I think it might actually be a huge hassle to fix. So the result of investigating the issue might just be a more informed decision that it's not actually worth trying to fix.

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR
~I tested in multiple browsers~
~I added relevant unit tests~


src/frontend/package.json
```diff
@@ -97,6 +98,7 @@
     "@types/react-csv": "^1.1.1",
     "@types/react-virtualized-auto-sizer": "^1.0.0",
     "@types/react-window": "^1.8.2",
+    "@types/uuid": "^8.3.4",
```
vincent-czi, line 101: Yeah, latest `uuid` is `9.0.0`, but the latest types available is `8.3.4`. It's annoying, but the only thing we're using is the v4 uuid generator, so that part of the types is super stable b/c it's been around a super long time, so the version mismatch here is not a big deal.

ehoops-cz (approved):
Looks good to me.  I'd be curious to hear other feedback though.

mmmmmmaya (approved):

PR #1411, opened 2022-10-20
Maya/224006 samplesview

mmmmmmaya:
### Summary
- **What:** 
  - Set up SamplesView (fetch data)
  - Set up TreesView (fetch data & transform it)
- **Why:** simplify the data fetching process as part of sample table refactor
- **Ticket:**
  -  [[sc-224013]](https://app.shortcut.com/genepi/story/224013)
  -  [[sc-224012]](https://app.shortcut.com/genepi/story/224012)
- **Env:** https://maya-fetch-frontend.dev.czgenepi.org/

### Testing
1. Turn on the feature flag
1. See the ids listed in both the samples view and trees view

### Demos
No visual changes.

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/titles.ts
+++ b/src/frontend/src/common/titles.ts
@@ -3,6 +3,5 @@ import { ROUTES } from "./routes";
 // For use in the `<head>` title of page
 export const PAGE_TITLES: Record<string, string> = {
   [ROUTES.HOMEPAGE]: "Public Health Open-Source Tool",
-  [ROUTES.DATA_SAMPLES]: "Samples",
   [ROUTES.PHYLO_TREES]: "Phylogenetic Trees",
 };
--- a/src/frontend/src/common/titles.ts
+++ b/src/frontend/src/common/titles.ts
@@ -1,7 +1,0 @@
-import { ROUTES } from "./routes";
-
-// For use in the `<head>` title of page
-export const PAGE_TITLES: Record<string, string> = {
-  [ROUTES.HOMEPAGE]: "Public Health Open-Source Tool",
-  [ROUTES.PHYLO_TREES]: "Phylogenetic Trees",
-};
--- a/src/frontend/src/common/utils/dataTransforms.ts
+++ b/src/frontend/src/common/utils/dataTransforms.ts
@@ -64,18 +64,20 @@ export const getDownloadLinks = (phyloRun: PhyloRun): PhyloRunLinks => {
 /**
  * Converts a completely uppercase tree type to a capitalcase tree type.
  */
-export const getTreeType = (phyloRun: PhyloRun): string | undefined => {
+export const getCapitalCaseTreeType = (
+  phyloRun: PhyloRun
+): string | undefined => {
   const { treeType } = phyloRun;
 
   if (typeof treeType !== "string" || treeType.toLowerCase() == "unknown") {
     return undefined;
   }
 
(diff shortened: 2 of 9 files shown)


src/frontend/src/common/titles.ts
```diff
@@ -3,6 +3,5 @@ import { ROUTES } from "./routes";
 // For use in the `<head>` title of page
 export const PAGE_TITLES: Record<string, string> = {
   [ROUTES.HOMEPAGE]: "Public Health Open-Source Tool",
-  [ROUTES.DATA_SAMPLES]: "Samples",
```
ehoops-cz, line 6: I see that this was moved below.  I'm not sure why we originally had titles here.  I'd prefer consistency though - either all here or all in their respective index files.

src/frontend/src/common/utils/dataTransforms.ts
```diff
@@ -0,0 +1,81 @@
...
+ * Phylo run transforms
+ */
+
+/**
+ * If there is a tree associated with this run, return links to download tree data
+ */
+interface PhyloRunLinks {
+  downloadLinkIdStylePublicIdentifiers: string | undefined;
+  downloadLinkIdStylePrivateIdentifiers: string | undefined;
+  accessionsLink: string | undefined;
```
ehoops-cz, line 35: Why not use `accessionsLink?: string`?  I'm assuming this is readability preference.  Maybe we should document stuff like this somewhere.

src/frontend/src/common/utils/dataTransforms.ts
```diff
@@ -0,0 +1,81 @@
...
+ * Converts a completely uppercase tree type to a capitalcase tree type.
+ */
+export const getTreeType = (phyloRun: PhyloRun): string | undefined => {
+  const { treeType } = phyloRun;
+
+  if (typeof treeType !== "string" || treeType.toLowerCase() == "unknown") {
+    return undefined;
+  }
+
+  const name_parts = treeType.toLowerCase().split("_");
```
ehoops-cz: Can this be camelCase?

src/frontend/src/common/utils/dataTransforms.ts
```diff
@@ -0,0 +1,81 @@
...
+    ORG_API.PHYLO_TREES
+  )}${id}/sample_ids`;
+
+  return links;
+};
+
+/**
+ * Converts a completely uppercase tree type to a capitalcase tree type.
+ */
+export const getTreeType = (phyloRun: PhyloRun): string | undefined => {
```
ehoops-cz: It would be good for the name to reflect the format change as well.  Maybe something like `getCapitalCaseTreeType`

ehoops-cz (approved):

src/frontend/src/common/utils/dataTransforms.ts
mmmmmmaya: I think a key not being present in an object is parsed differently than an object with a key that corresponds to an `undefined` value. Since I explicitly set all 3 keys to undefined in the beginning of the function, I need to tell TS that the value of each key can be `undefined`, rather than saying the key itself is optional

src/frontend/src/common/utils/dataTransforms.ts
mmmmmmaya: sure, just copy pasted it but happy to update

PR #1412, opened 2022-10-20
Surface validation errors in static upload metadata table

lvreynoso:
### Summary:
- **What:** Displays validation errors in table in the appropriate column. Shows a dismissible infobox at the top of the metadata page when loading the static metadata table.
- **Ticket:** [sc214142](https://app.shortcut.com/genepi/story/214142), [sc214140](https://app.shortcut.com/genepi/story/214140), [sc216284](https://app.shortcut.com/genepi/story/216284)
- **Env:** [https://static-metadata-frontend.dev.czgenepi.org](https://static-metadata-frontend.dev.czgenepi.org)

### Demos:
Example metadata with issues:
Screen Shot 2022-10-20 at 12 32 57

### Checklist:
- [x] I added labels to my PR


Janeece (approved):
Looks great on design side, thanks Lucia!

src/frontend/src/views/Upload/components/Metadata/components/StaticTable/components/Row/index.tsx
```diff
@@ -26,21 +34,53 @@ export default memo(function Row({ id, metadata }: Props): JSX.Element {
     publicId,
   } = metadata;
 
+  const validatedCellData: Record<
+    string,
+    React.ReactElement | string | undefined
+  > = {
+    privateId: privateId || "--",
```
vincent-czi: Nit: We have UNDEFINED_TEXT (`frontend/src/common/components/library/data_table/index.tsx`) and NO_CONTENT_FALLBACK (`frontend/src/common/components/library/Table/components/Cell/index.tsx`) doing same. Wonder if it's worth using one of those, although it already looks a bit scattered in our code base for standardizing on what we display when there's nothing to display.

mmmmmmaya: `--` should probably be a constant (in fact, I think it might already be one somewhere in the codebase) that can be reused

src/frontend/src/views/Upload/components/Metadata/components/StaticTable/components/Row/style.ts
```diff
@@ -78,3 +83,31 @@ export const PrivateContent = styled.div`
...
+export const StyledAlertText = styled.p`
+  vertical-align: middle;
+  ${fontBodyXxxs}
+
+  ${(props: CommonThemeProps) => {
+    const colors = getColors(props);
+
+    return `
+      color: ${colors?.error[500]};
+      padding: 0 0 0 0;
```
mmmmmmaya: when using padding, giving a single integer applies that value to all 4 top bottom left right.
```suggestion
      padding: 0;
```

src/frontend/src/views/Upload/components/Metadata/components/StaticTable/components/Row/style.ts
```diff
@@ -78,3 +83,31 @@ export const PrivateContent = styled.div`
...
+export const StyledExclamationMark = styled(Icon)`
+  vertical-align: middle;
+
+  ${(props: CommonThemeProps) => {
+    const spaces = getSpaces(props);
+    const colors = getColors(props);
+
+    return `
+      margin: 0 ${spaces?.xxs}px ${spaces?.xxs}px ${spaces?.xxs}px;
+      fill: ${colors?.error[500]}
```
mmmmmmaya:
```suggestion
      fill: ${colors?.error[500]};
```

src/frontend/src/views/Upload/components/Metadata/components/StaticTable/index.tsx
```diff
@@ -50,78 +62,132 @@ const validationSchema = object({
 export default function StaticTable({
   metadata,
   setIsValid,
+  hasImportedMetadataFile,
 }: Props): JSX.Element {
+  const [validationErrors, setValidationErrors] =
+    useState<Record<string, ValidationErrorRecord | null>>(EMPTY_OBJECT);
```
mmmmmmaya: it seems like the type is used in multiple places. Does it make sense to define this as a named type?

src/frontend/src/views/Upload/components/Metadata/components/StaticTable/index.tsx
```diff
@@ -50,78 +62,132 @@ const validationSchema = object({
...
-    validateMetadata(metadata);
-  }, [metadata]);
+    if (hasImportedMetadataFile) {
+      validateMetadata(metadata);
+    }
+  }, [metadata, hasImportedMetadataFile]);
+
+  // Sort entries by error status, then by sampleId
+  let errorSortedMetadata: [string, Metadata][] = [];
+  if (metadata != null) {
```
mmmmmmaya: Are you intending to do a non-strict comparison here? If not, consider using `!==` (that's what we use in most of the rest of the code base)
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Equality

src/frontend/src/views/Upload/components/Metadata/components/StaticTable/index.tsx
```diff
@@ -50,78 +62,132 @@ const validationSchema = object({
...
+  // If one field has multiple error messages we just overwrite the message in
+  // the ValidationErrorRecord, since we only want to display one error at a time.
   const validateMetadata = useCallback(
     async (metadata: SampleIdToMetadata | null) => {
       if (metadata == null) {
         setIsValid(false);
         return;
       }
-      const rowValidation: Record<string, boolean> = {};
+      const validationErrors: Record<string, ValidationErrorRecord | null> =
```
vincent-czi: I'm a bit torn on this process. Looking over it, everything looks good, and it does seem nice to be able to use Yup for the validation since that matches how the dynamic table works. That said, because it's a static table, we can rely on the user passing data through the metadata `parseFile` function, which already does validation. But on the other hand, I know moving those warnings around is a pain and might introduce extra complexity if it's doing extra checks for `>100` samples, so what's here might be the best way. But if you've already looked into that and this seemed best, then I'm happy.

vincent-czi (commented):
Looks good!

src/frontend/src/views/Upload/components/Metadata/components/StaticTable/index.tsx
```diff
@@ -50,78 +62,132 @@ const validationSchema = object({
...
+      validateMetadata(metadata);
+    }
+  }, [metadata, hasImportedMetadataFile]);
+
+  // Sort entries by error status, then by sampleId
+  let errorSortedMetadata: [string, Metadata][] = [];
+  if (metadata != null) {
+    errorSortedMetadata = Object.entries(metadata)
+      .map(([sampleId, sampleMetadata]) => {
+        return [sampleId, sampleMetadata] as [string, Metadata];
```
mmmmmmaya: is this type assertion required?

src/frontend/src/views/Upload/components/Metadata/components/StaticTable/index.tsx
```diff
@@ -50,78 +62,132 @@ const validationSchema = object({
...
+      })
+      .sort((a, b) => {
+        const a_error_sort = validationErrors[a[0]] == null ? 1 : 0;
```
mmmmmmaya: camelcase the var names plz

src/frontend/src/views/Upload/components/Metadata/components/StaticTable/components/Row/index.tsx
```diff
@@ -26,21 +34,53 @@ export default memo(function Row({ id, metadata }: Props): JSX.Element {
...
+  if (validationError != null) {
+    Object.entries(validationError).forEach(([key, message]) => {
+      // The validation error for a location is for an id,
+      // which is internal and not something a user would know about.
+      // The only case in which the field does not validate is if the
+      // value is missing entirely.
+      if (key == "collectionLocation") {
+        message = "Required";
+      }
+      validatedCellData[key] = (
```
mmmmmmaya, line 59: this is probably ok the way it is right now, but how would you feel about moving it into its own component file if it gets any larger?

src/frontend/src/views/Upload/components/Metadata/components/StaticTable/index.tsx
```diff
@@ -50,78 +62,132 @@ const validationSchema = object({
...
-        rowValidation[sampleId] = isRowValid;
+        try {
+          await validationSchema.validate(sampleMetadata, {
+            abortEarly: false,
+          });
+        } catch (error) {
+          if (error instanceof ValidationError) {
+            const errorRecord: ValidationErrorRecord = {};
+            error.inner.forEach((validationError) => {
+              if (validationError.path != undefined) {
```
mmmmmmaya: same note about equality operator here

src/frontend/src/views/Upload/components/Metadata/components/StaticTable/index.tsx
```diff
@@ -50,78 +62,132 @@ const validationSchema = object({
...
+                const rootMetadataKey = validationError.path.split(".")[0];
+                errorRecord[rootMetadataKey as keyof ValidationErrorRecord] =
```
mmmmmmaya: is this type assertion required?

src/frontend/src/views/Upload/components/Metadata/components/StaticTable/index.tsx
lvreynoso: The user does pass data through the metadata parseFile function. It probably would have been better overall to validate there and surface any errors, but it also would have been much more complex. I went for the simplest route.

src/frontend/src/views/Upload/components/Metadata/components/StaticTable/index.tsx
lvreynoso: Yeah, the linter was not happy without it for some reason. It seems to think that the type is `[string | Metadata]` without it.

src/frontend/src/views/Upload/components/Metadata/components/StaticTable/components/Row/index.tsx
lvreynoso: I don't mind

src/frontend/src/views/Upload/components/Metadata/components/StaticTable/index.tsx
lvreynoso: Yeah, I only put the type assertions in when TS was making noises at me.

lvreynoso: Actually, on a second thought I just changed the type definition to make this assertion go away.

src/frontend/src/views/Upload/components/Metadata/components/StaticTable/index.tsx
mmmmmmaya: I'm also slightly confused why this map is needed at all. it looks like the map returns exactly what the args are, so it's probably extraneous.
 
Screenshot 2022-11-07 at 11 14 00 AM

Also, removing the `map` removes the type error so you can take out the type assertion.

src/frontend/src/views/Upload/components/Metadata/components/StaticTable/index.tsx
```diff
@@ -50,78 +59,130 @@ const validationSchema = object({
...
+    }
+  }, [metadata, hasImportedMetadataFile]);
+
+  // Sort entries by error status, then by sampleId
+  let errorSortedMetadata: [string, Metadata][] = [];
+  if (metadata !== null) {
+    errorSortedMetadata = Object.entries(metadata)
+      .map(([sampleId, sampleMetadata]) => {
+        return [sampleId, sampleMetadata] as [string, Metadata];
+      })
```
mmmmmmaya:
```suggestion
```

src/frontend/src/views/Upload/components/Metadata/components/StaticTable/index.tsx
mmmmmmaya: Yea, typically that is the way to go. A type assertion shouldn't be used as a panacea for type compiler complaints. Usually when the type checker fails it's because you have two types somewhere that don't line up properly.

mmmmmmaya:
good to go after map is removed

PR #1415, opened 2022-10-21
Lineage filter bugfix

ehoops-cz:
### Summary:
- **What:** `Lineage filter bug fix` I did not end up finding the root cause, but we can remove the "All" option from the dropdown for now - we have the reset button, but will likely come back to this to rework the "All" option.  [Related slack thread.](https://czi-sci.slack.com/archives/C01HYJYA50W/p1666374420991309)
- **Ticket:** none
- **Env:** [rdev link](https://lineage-filter-bug-fix-frontend.dev.czgenepi.org/)

### Demos:
Screen Shot 2022-10-21 at 1 09 54 PM

Screen Shot 2022-10-21 at 1 18 45 PM

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/LineageFilter/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/LineageFilter/index.tsx
@@ -2,7 +2,6 @@ import { FilterOptionsState, PopperProps } from "@mui/material";
 import { createFilterOptions } from "@mui/material/Autocomplete";
 import { DefaultMenuSelectOption } from "czifui";
 import { isEqual } from "lodash";
-import { useEffect, useState } from "react";
 import { useSelector } from "react-redux";
 import { selectCurrentPathogen } from "src/common/redux/selectors";
 import { SplitPathogenWrapper } from "src/components/Split/SplitPathogenWrapper";
--- a/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/LineageFilter/index.tsx
+++ b/src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/LineageFilter/index.tsx
@@ -1,6 +1,7 @@
 import { FilterOptionsState, PopperProps } from "@mui/material";
 import { createFilterOptions } from "@mui/material/Autocomplete";
 import { DefaultMenuSelectOption } from "czifui";
+import { isEqual } from "lodash";
 import { useEffect, useState } from "react";
 import { useSelector } from "react-redux";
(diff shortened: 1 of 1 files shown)


src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/LineageFilter/index.tsx
```diff
@@ -19,31 +19,12 @@ export type LineageFilterType = {
...
-const ALL_LINEAGES_KEYWORD = "All";
-
 function makeDropdownOption(name: string): DefaultMenuSelectOption {
   return { name: name };
 }
-// Generate only once because we need to reference same object throughout.
-const ALL_LINEAGES_CHOICE = makeDropdownOption(ALL_LINEAGES_KEYWORD);
 
 /**
- * `Dropdown` defaults to checking if option is selected (`value`) by equality.
```
vincent-czi, line 33: I just want to double-check here: we are no longer dynamically generating the the options, right? It's just an array we ingest once and then never have to regenerate and would never accidentally get regenerated as we prep what we display to user? I've definitely been surprised by being bitten by the object-equality aspect instead of value-equality with our dropdown options, so wanted to make sure you're quite confident on this one.

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/LineageFilter/index.tsx
```diff
@@ -219,44 +160,18 @@ export function LineageFilter({
     const newSelectedLineages = newSelectedOptions.map(
       (option: DefaultMenuSelectOption) => option.name
     );
-    // What we actually emit as selection does not always match user selected.
-    let emittedSelection: string[];
 
     // When beginning selection process, had nothing selected / "All" selected
```
vincent-czi: Nit: should probably remove the `/ "All" selected` clause of comment here since we don't have "All" as a real option and it's just what we display when 0 length of options chosen.

vincent-czi (approved):
Nice! Very happy to see all that brittle "All" code go bye-bye! So much simpler and stable seeming now!

src/frontend/src/common/components/library/data_subview/components/CreateNSTreeModal/components/SampleFiltering/components/LineageFilter/index.tsx
ehoops-cz: SDS updated their equality check to also check the name field rather than object equality.  As of recent, this prop was renamed, so the one we're passing here isn't being used.  It shouldn't matter though since it no longer relies on object equality.

vincent-czi: Good to know, thanks!

PR #1417, opened 2022-10-21
Generic pathogen repository tables

jgadling:
### Summary:
- **What:** Create new tables to support workflows for different repositories/pathogens.

### Notes:
We currently have a variety of tables in our db named something_**genbank**_something. We need to expand the scope for all of these tables to support multiple pathogens and multiple pathogen repositories. While we're at it, we may as well update the table names to reflect their true purpose as well. 

Most of these changes are adding new tables, so shouldn't affect existing functionality, however there is one thing we'll need to keep an eye on while migrations run:
I updated the naming convention for unique constraints to include the names of *all* fields in the index, instead of just the name of the *first* field in the index. We sometimes have multiple constraints on the same table, and the auto-generated migrations were creating constraints with name collisions. I don't think there's any harm in going back and renaming existing constraints as well.

### Checklist:
- [x] I added labels to my PR


src/backend/aspen/database/models/base.py
```diff
@@ -7,7 +7,7 @@
     schema="aspen",
     naming_convention={
         "ix": "ix_%(column_0_label)s",
-        "uq": "uq_%(table_name)s_%(column_0_name)s",
+        "uq": "uq_%(table_name)s_%(column_0_N_name)s",
```
jgadling, line 10: This changes the name of unique constraints to include *all* column names in the constraint instead of just the first one. It reduces the number of conflicting index names when alembic auto-generates migrations

src/backend/aspen/database/models/public_repository_metadata.py
```diff
@@ -0,0 +1,40 @@
...
+    pathogen = relationship(Pathogen, back_populates="public_repository_metadata")  # type: ignore
+    public_repository_id = Column(
+        Integer,
+        ForeignKey(PublicRepository.id),
+        nullable=False,
+    )
+    public_repository = relationship(PublicRepository, back_populates="public_repository_metadata")  # type: ignore
+    strain = Column(String, primary_key=True)
+    lineage = Column(String, nullable=True)
+    clade = Column(String, nullable=True)
```
phoenixAja: i believe lineage and clade will be used for lineage calls (we're using pangolin for sars-cov-2 lineage, clade output from nextclade for all other lineages) cc @sidneymbell

jgadling: We populate this table with samples downloaded from gisaid/genbank, and we use the data in it to validate strain names (for tree building). We may run nextclade on these samples once we support HCV, but for now they'll likely be left empty.

src/backend/aspen/database/models/repository_workflows.py
```diff
@@ -0,0 +1,111 @@
...
+    public_repository_id = Column(
+        Integer, ForeignKey(PublicRepository.id), nullable=False
+    )
+    public_repository = relationship(PublicRepository, back_populates="raw_repository_data")  # type: ignore
+    download_date = Column(DateTime, nullable=False)
+    s3_bucket = Column(String, nullable=False)
+    s3_key = Column(String, nullable=False)
+
+
+class ProcessedRepositoryData(Entity):
```
phoenixAja, line 30: i thought we were migrating away from entities?

jgadling: Eventually yes, but I'm not currently ready to replace this part of our infrastructure. This is essentially just a table that will replace the existing ProcessedGisaidDump table

src/backend/database_migrations/versions/20221021_205409_fix_constraint_names.py
```diff
@@ -0,0 +1,68 @@
...
+depends_on = None
+
+
+def upgrade():
+    op.drop_constraint(
+        "uq_phylo_trees_s3_bucket",
+        "phylo_trees",
+        schema="aspen",
+        type_="unique",
+    )
```
phoenixAja, line 22: did you mean to drop this?

src/backend/aspen/database/models/public_repository_metadata.py
sidneymbell: FWIW, `lineage` and `clade` mean the same thing. I haven't dug into the context of this line, so grain of salt, but might be worth considering whether we need both fields.

src/backend/aspen/database/models/workflow.py
```diff
@@ -15,6 +15,8 @@ class WorkflowType(enum.Enum):
     PROCESS_GISAID_DUMP = "PROCESS_GISAID_DUMP"
     ALIGN_GISAID_DUMP = "ALIGN_GISAID_DUMP"
     PHYLO_RUN = "PHYLO_RUN"
+    DOWNLOAD_PUBLIC_REPOSITORY = "DOWNLOAD_PUBLIC_REPOSITORY"
```
phoenixAja: this is a nit, but i feel like DOWNLOAD_PUBLIC_REPOSITORY_DATA makes more sense? When i see DOWNLOAD_PUBLIC_REPOSITORY i immediately think you mean gisaid or genbank

src/backend/aspen/database/models/base.py
phoenixAja, line 10: i'm a little unclear on what this is for.

src/backend/aspen/database/models/entity.py
```diff
@@ -35,6 +35,9 @@ class EntityType(enum.Enum):
     PROCESSED_GISAID_DUMP = "PROCESSED_GISAID_DUMP"
     ALIGNED_GISAID_DUMP = "ALIGNED_GISAID_DUMP"
     PHYLO_TREE = "PHYLO_TREE"
+    RAW_PUBLIC_REPOSITORY_DATA = "RAW_PUBLIC_REPOSITORY_DATA"
+    PROCESSED_PUBLIC_REPOSITORY_DATA = "PROCESSED_PUBLIC_REPOSITORY_DATA"
+    ALIGNED_PUBLIC_REPOSITORY_DATA = "ALIGNED_PUBLIC_REPOSITORY_DATA"
```
phoenixAja, line 40: how is processed different from aligned? Can you give an example?

src/backend/aspen/database/models/public_repository_metadata.py
phoenixAja: so i feel like we can remove the clade field?

src/backend/aspen/database/models/public_repository_metadata.py
```diff
@@ -0,0 +1,40 @@
+from sqlalchemy import Column, DateTime, ForeignKey, Integer, String, UniqueConstraint
+from sqlalchemy.orm import relationship
+
+from aspen.database.models.base import idbase
+from aspen.database.models.mixins import DictMixin
+from aspen.database.models.pathogens import Pathogen
+from aspen.database.models.public_repositories import PublicRepository
+
+
+class PublicRepositoryMetadata(idbase, DictMixin):  # type: ignore
```
phoenixAja: in general i'm a little confused on this table... so for a particular pathogen and public_repository we are expecting these fields? (strain, clade, lineage, isl, etc)

src/backend/aspen/database/models/base.py
phoenixAja: nevermind just saw your comment above

src/backend/aspen/database/models/public_repository_metadata.py
phoenixAja: we'll continue this discussion in above thread

src/backend/aspen/database/models/public_repository_metadata.py
phoenixAja: i'm still a little confused how we're using this table, so for example we download some samples from genbank, does genbank give us some metadata on the collection of samples we just downloaded? and that metadata is the same for all the samples as long as it's for the same pathogen?

phoenixAja (approved):

PR #1421, opened 2022-10-26
Write to `pathogen_lineages` alongside `pango_lineages`

jgadling:
### Summary:
- **What:** Update our pango-lineage import script to write to both old & new lineage tables

### Notes:
I duplicated the existing "write data to `pango_lineages`" function instead of doing something more clever to make it easier to delete it later (since that table is going away) but open to other suggestions if we think there's a better way to manage table migrations and duplicated code.

Since the `pathogen_lineages` table supports multiple pathogens, our old "delete data from destination table and insert data from temp table" library had to be modified to support a `where` clause to **selectively** delete rows before copying in updated data.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/workflows/import_pango_lineages/load_lineages.py
+++ b/src/backend/aspen/workflows/import_pango_lineages/load_lineages.py
@@ -91,6 +91,7 @@ def get_lineages(lineage_notes_file: io.TextIOBase) -> list[str]:
     ]
 
 
+# TODO this function is only here for reverse-compatibility. We'll remove it soon
 def load_old_lineages_data(lineages: list[str]) -> None:
     """Loads all the lineages into DB.
 
@@ -220,7 +221,7 @@ def cli(
         return  # End here to avoid importing to DB
 
     print("Loading Pango lineages to DB...")
-    # load_old_lineages_data(lineages)
+    load_old_lineages_data(lineages)
     load_lineages_data(lineages)
     print("Loading Pango lineages complete!")
 
--- a/src/backend/aspen/workflows/import_pango_lineages/load_lineages.py
+++ b/src/backend/aspen/workflows/import_pango_lineages/load_lineages.py
@@ -2,6 +2,7 @@ import io
 from typing import Optional
 
 import click
+import sqlalchemy as sa
 
 from aspen.config.config import Config
 from aspen.database.connection import (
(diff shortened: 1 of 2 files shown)


src/backend/aspen/workflows/import_pango_lineages/load_lineages.py
```diff
@@ -90,7 +91,8 @@ def get_lineages(lineage_notes_file: io.TextIOBase) -> list[str]:
     ]
 
 
-def load_lineages_data(lineages: list[str]) -> None:
+# TODO this function is only here for reverse-compatibility. We'll remove it soon
+def load_old_lineages_data(lineages: list[str]) -> None:
```
vincent-czi, line 95: Nit: I think `legacy_load_lineages_data` or `deprecated_load_lineages_data` might be a slightly clearer name, to me the `old` part of this makes it seem like we now have two categories of data, but doesn't make me necessarily think we're going to be dropping support for the old flavor.

src/backend/aspen/workflows/shared_utils/database.py
```diff
@@ -34,15 +36,23 @@ def create_temp_table(session: Session, source_table: Table) -> Table:
...
 
 
-def mv_table_contents(session: Session, source_table: Table, dest_table: Table) -> None:
+def mv_table_contents(
+    session: Session,
+    source_table: Table,
+    dest_table: Table,
+    filters: Optional[Iterable[ClauseElement]] = None,
+) -> None:
     """Deletes contents of dest, copies in contents from source to dest.
```
vincent-czi, line 45: Nit: I'd like to see the docstring updated to address that the use case is not always a drop all --> move all operation anymore.

vincent-czi (approved):
Looks good! I don't know that much about databases, but might be worth it to add an index to `pathogen_lineages` DB table on the `pathogen_id` value since whenever we interact with it, we tend to always do stuff that reads/drops all the rows for a given `pathogen_id` at once.

src/backend/aspen/workflows/import_pango_lineages/load_lineages.py
jgadling: > Looks good! I don't know that much about databases, but might be worth it to add an index to `pathogen_lineages` DB table on the `pathogen_id` value since whenever we interact with it, we tend to always do stuff that reads/drops all the rows for a given `pathogen_id` at once.

Good call out! The primary key on this table is on [pathogen_id, strain] so it should help us with queries on just pathogen_id. (more info on multi-column indexes here: https://www.postgresql.org/docs/current/indexes-multicolumn.html)

I'll rename the functions in the next pr (which depends on this one)

PR #1425, opened 2022-10-28
very basic samples table set up

mmmmmmaya:
### Summary
- **What:** Very basic (incomplete) setup for new samples table
- **Why:** start playing around with react table to make sure the diagram I made can actually be implemented
- **Ticket:** [[sc-223983]](https://app.shortcut.com/genepi/story/223983)
- **Env:** https://maya-[url]-frontend.dev.czgenepi.org/

### Demos
Screenshot 2022-10-28 at 3 31 50 PM

### Notes
- obviously there will need to be styling work
- the configs for the columns will move to their own .js file in the table component dir after I completely iron them out. The current inline config is incomplete
- this is only a partially complete table because I wanted to get something working to see where things do/don't overlap with trees table

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/views/Data/components/SamplesView/components/SamplesTable/index.tsx
+++ b/src/frontend/src/views/Data/components/SamplesView/components/SamplesTable/index.tsx
@@ -1,0 +1,108 @@
+import { CellBasic, CellHeader, Table, TableHeader, TableRow } from "czifui";
+import {
+  createColumnHelper,
+  flexRender,
+  getCoreRowModel,
+  useReactTable,
+} from "@tanstack/react-table";
+import { IdMap } from "src/common/utils/dataTransforms";
+import { map } from "lodash";
+import { useEffect, useState } from "react";
+
+// TODO-TR (mlila): types
+interface Props {
+  data: IdMap<Sample> | undefined;
+  isLoading: boolean;
+}
+
+const columnHelper = createColumnHelper<Sample>();
+
+const columns = [
+  columnHelper.accessor("privateId", {
+    header: "Private ID",
+  }),
+  columnHelper.accessor("publicId", {
+    header: "Public ID",
+  }),
+  columnHelper.accessor("collectionDate", {
+    header: "Collection Date",
+  }),
+  columnHelper.accessor((obj) => JSON.stringify(obj.lineage), {
+    id: "lineage",
+    header: "Lineage",
+  }),
+  columnHelper.accessor("uploadDate", {
+    header: "Upload Date",
+  }),
+  columnHelper.accessor((obj) => JSON.stringify(obj.collectionLocation), {
+    id: "collectionLocation",
+    header: "Collection Location",
+  }),
+  columnHelper.accessor("sequencingDate", {
+    header: "Sequencing Date",
+  }),
+  columnHelper.accessor((obj) => JSON.stringify(obj.gisaid), {
+    id: "gisaid",
+    header: "GISAID",
+  }),
+];
+
+const SamplesTable = ({ data, isLoading }: Props): JSX.Element => {
+  const [samples, setSamples] = useState<Sample[]>([]);
+
+  useEffect(() => {
+    if (!data) return;
+
+    const newSamples = map(data, (v) => v);
+    setSamples(newSamples);
+  }, [data]);
+
+  const table = useReactTable({
+    data: samples,
+    columns,
+    getCoreRowModel: getCoreRowModel(),
+  });
+
+  if (isLoading) {
+    return <div>Loading ...</div>;
+  }
+
+  return (
+    <Table>
+      <TableHeader>
+        {table.getHeaderGroups().map((headerGroup) => (
+          <>
+            {headerGroup.headers.map((header) => (
+              <CellHeader key={header.id} horizontalAlign="left">
+                {header.isPlaceholder
+                  ? ""
+                  : (flexRender(
+                      header.column.columnDef.header,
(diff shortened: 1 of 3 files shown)


src/frontend/src/views/Data/components/SamplesView/components/SamplesTable/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { map } from "lodash";
+import { useEffect, useState } from "react";
+
+// TODO-TR (mlila): types
+interface Props {
+  data: IdMap<Sample> | undefined;
+  isLoading: boolean;
+}
+
+const columnHelper = createColumnHelper<Sample>();
```
vincent-czi, line 18: From what I'm reading in the docs for react-table, they say `createColumnHelper` is intended to be used with the row shape. But looking at the docs closely, I'm not really sure that matters, is it just a type convenience sorta thing? Is it an easy way to have type safety where you're guaranteeing that the row/data object being discussed per-column has all the attributes given? Basically, is the provided type more about the data structure or the row structure? If it's the case that it's about data more, then I think `createColumnHelper<Sample>` is totally fine and reasonable. But if it actually ties in more deeply to the row shape than the data shape, it feels like we need another `SampleRow` type or something.

src/frontend/src/views/Data/components/SamplesView/components/SamplesTable/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+  columnHelper.accessor("privateId", {
+    header: "Private ID",
+  }),
+  columnHelper.accessor("publicId", {
+    header: "Public ID",
+  }),
+  columnHelper.accessor("collectionDate", {
+    header: "Collection Date",
+  }),
+  columnHelper.accessor((obj) => JSON.stringify(obj.lineage), {
```
vincent-czi, line 30: I assume the `(obj) => JSON.stringify(obj.foo)` stuff here and below is mostly about proof of concept to see the data is available and being passed to the table? Any plans yet on passing the more complicated data chunks downstream to the cells (eg, passing down the full `lineage` to those cells so they can do the full data in their tooltip)? This was a pain for me in the past, but I did with a combination of custom accessors and then custom cell components that were specialized to those outputs. I'm looking over the current react-table docs, that looks doable, although how you call the specified `cell` component seems confusing to me at the moment.

src/frontend/src/views/Data/components/SamplesView/components/SamplesTable/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+];
+
+const SamplesTable = ({ data, isLoading }: Props): JSX.Element => {
+  const [samples, setSamples] = useState<Sample[]>([]);
+
+  useEffect(() => {
+    if (!data) return;
+
+    const newSamples = map(data, (v) => v);
+    setSamples(newSamples);
```
vincent-czi, line 57: I'm not sure I understand the purpose of making a copy of the upstream data here in the SamplesTable. Is it mostly here to convert the `data` from being a lookup dict into being an array so it can become rows of a table?

src/frontend/src/views/Data/components/SamplesView/components/SamplesTable/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+            ))}
+          </>
+        ))}
+      </TableHeader>
+      <tbody>
+        {table.getRowModel().rows.map((row) => (
+          <TableRow key={row.id}>
+            {row.getVisibleCells().map((cell) => {
+              return (
+                <CellBasic
```
vincent-czi, line 93: Overall, I think it makes more sense for the column config to specify the kind of cell component that gets rendered than for it to be written in to the table itself. I expect some columns will be have pretty bespoke cells and specialty data objects to build those cells. To me, it makes the most sense for that to get configured per-column. (Of course, that might already be the plan and this is version is a stepping stone. Also, I'm not sure how easy the approach I just sketched out will be with the new react-table API; it felt a lot more natural with the old version I used to work with, maybe the new version really should be handled in a different way.)

vincent-czi (approved):
Approach looks good to me. The big open question for me is in the various interactions to create the complicated cells: config <-> data <-> cell component. But I expect that's the next major focus of work. Nice work!

src/frontend/src/views/Data/components/SamplesView/components/SamplesTable/index.tsx
mmmmmmaya: yes

src/frontend/src/views/Data/components/SamplesView/components/SamplesTable/index.tsx
mmmmmmaya: I agree with you and that is part of the future plan.

src/frontend/src/views/Data/components/SamplesView/components/SamplesTable/index.tsx
mmmmmmaya: yes exactly. this is the most basic table just to prove we can render the thing

src/frontend/src/views/Data/components/SamplesView/components/SamplesTable/index.tsx
mmmmmmaya: https://app.shortcut.com/genepi/story/227054

PR #1427, opened 2022-11-01
Capture original tree build args

vincent-czi:
### Summary:
- **What:** Capture original (user-provided) tree build args; save resolved/interpreted args once tree done
- **Ticket:** [sc<220375>](https://app.shortcut.com/genepi/story/<220375>)
- **Env:** None -- Since rdevs can't run tree builds right now, just verified it locally (although piecemeal, running each step alone, never all in one go). Planning to run it in Staging once merged and verify working as expected.

### Demos:

Capture of original, non-expanded version of build args in `phylo_runs.template_args`

image

---
Resulting `phylo_trees.resolved_template_args` from tree build process

image

### Notes:

- Overarching idea here is to have two different versions of the build args that go in to kicking off a tree run. The `phylo_runs.template_args`, a JSON column that is (mostly) a raw copy of what params were given in the request to kick off the tree build, and `phylo_trees.resolved_template_args`, a new JSON column that is the "resolved" version of the args: pango lineages expanded from wildcards and synonym names, location expanded into all its attributes, etc.
  - Migration for the new column done in a previous PR -- #1420 
- Overarching idea was necessary because spec for "Details" in tree list requires us to show the pango lineages as originally specified by the user (eg, in wildcard form, _not_ resolved). But overall this wound up being less valuable than I had originally thought it was going to be. Two reasons for not being so helpful:
  - When I started working on this, I had assumed the `resolved_template_args` would be useful for populating the "Details" boxes in Trees list. They might be in some cases, but I overlooked that the "Details" were present on failed and in-progress trees. I was just thinking about completed trees. Since I added `resolved_template_args` to the completed phylo tree object instead of as part of the run itself, this extra info isn't actually guaranteed to be available for all trees. That was an oversight in my design.
  - Furthermore, after looking at the current Trees list response longer, I realized we do actually have all the info we need for the "Details" from the existing `template_args`. Some aspects are a bit more of a pain, it requires more FE effort to make sense of it (eg, need to interpolate the location_id against the locations list pulled from another API call), but there's nothing particularly necessary to change about the current response beyond tracking the original, wildcard-version of filter_pango_lineages, which this PR does handle.
- But at this point it doesn't really feel worth it to go back and try to change the approach. Everything does work as intended -- we capture the original args so we can display them to the user in "Details" for a tree, and we keep a record of what those args became during build process in case we ever need an authoritative record for troubleshooting, etc -- so I think it's best to just move on.

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/backend/aspen/database/models/phylo_tree.py
```diff
@@ -102,6 +110,13 @@ class PhyloRun(Workflow):
     pathogen: Pathogen = relationship(Pathogen, back_populates="phylo_runs")  # type: ignore
 
     template_file_path = Column(String, nullable=True)
+    """[Tony Tung, Spring2021] This is the path, relative to aspen root, for
+    the builds template file.  For all new builds, this should be set.
+    However, for runs imported from covidhub, this would not be available.
+
+    [Vincent, Fall2022] Above appears not to be true anymore, and seems to have
+    stopped being true at end of 2021. TODO Potentially remove this column?
```
jgadling, line 118: I suspect we're not using this column at all anymore. It should be safe to do an audit of our code and remove it.

src/backend/aspen/workflows/nextstrain_run/export.py
```diff
@@ -236,19 +249,41 @@ def get_phylo_run(session, phylo_run_id):
...
+    provided `filter_pango_lineages` arg and converting it. If arg was not
+    present in original template_args, returns None instead.
+    """
+    lineage_list = template_args.get("filter_pango_lineages")
+    if lineage_list is None:  # short-circuit if template arg was not present
+        return None
+    # Utility that does expansion depends on having set of all lineages.
+    all_lineages = {
+        query_result[0]  # query_result is 1-length tuple of just lineage
+        for query_result in session.query(PangoLineage.lineage).all()
```
jgadling: nit: Just a reminder that we should try to use SA 2.0 style queries (`sa.select()`) as much as possible

vincent-czi: Yeah, I waffled on that while coding this, but opted in favor of matching the format used by the rest of the file currently. If you think it's better to do 2.0 style even if it will be the odd duck out in the file, I'm happy to rewrite it.

jgadling (approved):

src/backend/aspen/workflows/nextstrain_run/export.py
jgadling: Your call. One less query to translate later would be nice, but it's not a reason to block this feature.

vincent-czi: Converted it over to v2 style

PR #1429, opened 2022-11-01
Use new tables

jgadling:
### Summary:
- **What:** This is the next step in migrating to new public repo tables and should be deployed after #1423

### Notes:
This PR basically sets us up to use our new repository-based workflow and entity tables that we began to populate in #1423:
- Always read from our new `repository`-workflows tables instead of `gisaid`-workflow tables (for example, AlignedRepositoryData instead of AlignedGisaidDump)
  - since these tables have a `pathogen_id` and `repository_id` field in them, we need to include those values in queries.
  - We read the default pathogen repository data from a split feature flag
- Also migrate all of our *existing* `gisaid`-workflow table data to `repository`-workflow tables

### Checklist:
- [x] I added labels to my PR


src/backend/aspen/api/views/phylo_runs.py
```diff
@@ -102,16 +111,18 @@ async def kick_off_phylo_run(
...
-    # 4B - AlignedGisaidDump
-    aligned_gisaid_dump_query = (  # type: ignore
-        sa.select(AlignedGisaidDump)  # type: ignore
-        .join(AlignedGisaidDump.producing_workflow)  # type: ignore
+    # 4B - AlignedRepositoryData
+    aligned_repo_data_query = (  # type: ignore
+        sa.select(AlignedRepositoryData)  # type: ignore
+        .join(AlignedRepositoryData.producing_workflow)  # type: ignore
+        .where(AlignedRepositoryData.pathogen == pathogen)  # type: ignore
+        .where(AlignedRepositoryData.public_repository == public_repository)  # type: ignore
```
lvreynoso, line 119: This is totally just an aside, but I learned today that `.where()` also accepts multiple conditions as comma-separated arguments; they get treated as parts of `AND` clause (functionally the same as what you wrote here). Pretty cool.

lvreynoso (approved):
Great work!

PR #1435, opened 2022-11-03
Handle end dates for all tree titles (SC 214190)

vincent-czi:
### Summary:
- **What:** Always provide an end date to the tree title (for JSON tree, not the name we show on aspen trees list page)
- **Ticket:** [sc<214190>](https://app.shortcut.com/genepi/story/<214190>)
- **Env:** None -- testing this on an rdev would be attached to creating a tree, and that's currently broken on rdev

### Demos:
Phylo run with input template arg of `filter_start_date` but no `filter_end_date`, so falls back to using the phylo run's `start_datetime` to produce the title

image

image

image

---

Phylo run that has no `filter_start_date`, title template language changes to only mention end date. (In this case, explicitly provided by `filter_end_date`, but would default to run `start_datetime` if that was not available.)

image

image

### Notes:

There is the possibility of an off-by-one error for the end date when it's implicitly pulled from the phylo run's `start_datetime`. This happens because the user's timezone is unknown: eg, if I live in the USA and kick off a tree build on Nov 1 at 10pm, that will be early Nov 2 in UTC time. Since all the datetimes on the server are UTC, the implicit end date given by `start_datetime` might actually be different than the date the user saw when they kicked off the tree build. I discussed this with Dan, and there's enough ambiguity around date already that this potential error seems acceptable to both of us. (If our users ever encounter it and care, we could try to implement a fix, but I don't think there's a perfect solution: imagine collaborators on opposite sides of the international date line. If one of them creates a tree, they each see the tree creation as happening on a different day, so we can't write both those dates to the tree JSON. There will have to be at least some level of compromise on this.)

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR
~[ ] I added relevant unit tests~

### DIFF
--- a/src/backend/aspen/workflows/nextstrain_run/build_plugins/base_plugin.py
+++ b/src/backend/aspen/workflows/nextstrain_run/build_plugins/base_plugin.py
@@ -16,7 +16,7 @@ class BaseConfigPlugin:
         self.template_args = template_args
         self.template = None
         self.tree_build_level = "location"
-        # Set self.num_county_sequences, self.num_sequences, etc.
+        # Set self.num_sequences, self.run_start_datetime, etc.
         for k, v in kwargs.items():
             setattr(self, k, v)
 
--- a/src/backend/aspen/workflows/nextstrain_run/build_plugins/type_plugins.py
+++ b/src/backend/aspen/workflows/nextstrain_run/build_plugins/type_plugins.py
@@ -40,9 +40,10 @@ class TreeTypePlugin(BaseConfigPlugin):
         build["subsampling_scheme"] = self.subsampling_scheme
 
         # Update the tree's title with build type, location and date range.
-        if (self.template_args.get("filter_start_date") is not None) and (
-            self.template_args.get("filter_end_date") is not None
-        ):
+        # We always provide some form of end date in the title.
+        end_date = self._get_formatted_tree_end_date()
+        # We base format of title on whether we have a `filter_start_date`
+        if self.template_args.get("filter_start_date") is not None:
             title_template = "{tree_type} tree for samples collected in {location} between {start_date} and {end_date}"
             build["title"] = title_template.format(
                 tree_type=self.subsampling_scheme.title(),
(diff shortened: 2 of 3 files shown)


src/backend/aspen/workflows/nextstrain_run/build_plugins/base_plugin.py
```diff
@@ -16,7 +16,7 @@ def __init__(
         self.template_args = template_args
         self.template = None
         self.tree_build_level = "location"
-        # Set self.num_county_sequences, self.num_sequences, etc.
```
vincent-czi, line 19: `num_county_sequences` had been used as an example, but it's no longer anywhere in our code base, so I removed since I don't think it makes for a very good example any more.

src/backend/aspen/workflows/nextstrain_run/build_plugins/type_plugins.py
```diff
@@ -40,32 +40,61 @@ def _update_config_params(self, config):
...
                 location=", ".join(location_values),
+                end_date=end_date,
             )
 
         config["files"]["description"] = config["files"]["description"].format(
             tree_type=self.subsampling_scheme.lower()
         )
         config["priorities"]["crowding_penalty"] = self.crowding_penalty
 
+    def _get_formatted_tree_end_date(self):
```
danrlu, line 69: Out of curiosity, when do you guys use "_" in the beginning of the method name? I see this in nextstrain repo a lot

danrlu (approved):
Looks great to me! Thanks very much!

src/backend/aspen/workflows/nextstrain_run/build_plugins/type_plugins.py
vincent-czi: Sorry for the delay: it's basically a convention in Python to say "hey, this is really just intended for internal use, don't touch this thing unless you really know what you're doing". In other languages, there's the concept of "Public" and "Private" methods -- you can import/access "Public" stuff elsewhere, but "Private" is not directly accessible -- but Python does not have that. So it's a stylistic convention that most people in the Python space are aware of and follow (you'll also see it in other languages, I'm pretty sure I've seen it in the wild in JavaScript code, etc).

PR #1437, opened 2022-11-04
[SC-214189] I want to know which location my tree is from

ehoops-cz:
### Summary:
- **What:** `Add details to tree table`
- **Ticket:** [sc214189](https://app.shortcut.com/genepi/story/214189)
- **Env:** `https://tree-details-frontend.dev.czgenepi.org/`

### Demos:
Screenshot 2022-11-04 at 12 26 48 PM
Screenshot 2022-11-04 at 12 26 42 PM

### Checklist:
- [ ] I added labels to my PR


src/frontend/src/views/Data/components/TreeTableNameCell/style.ts
```diff
@@ -80,3 +86,19 @@ export const StyledTreeIconWrapper = styled.div`
...
+      padding: "10px !important",
+    },
+    "& .MuiTableRow-root": {
+      verticalAlign: "top",
+    },
+    "& .MuiTableCell-root": {
+      width: "25%",
+    },
+  },
+};
```
ehoops-cz, line 107: Definitely on the struggle bus regarding styling the tooltip.  I wasn't able to get the mui classes to style directly, but passing the style to the popper props works.  Unfortunately, this doesn't directly target `.MuiTooltip-tooltip` and targets `.MuiPopper-root-MuiTooltip-popper .MuiTooltip-tooltip` instead - thus the `!important`.

Any feedback/suggestions would be great.

As an aside - I learned to open the sources tab of dev tools and use the F8 key to disable javascript.  Then hover, press F8 and then you can inspect and such.

Janeece (approved):
@ehoops-zz LGTM. Is it possible to increase the space between the data rows to spaceS (8px)? It'd be nice to have a bit more separation because of the multi-line content, but not a blocker (especially if the SDS component makes that tricky).

ehoops-cz:
> @ehoops-zz LGTM. Is it possible to increase the space between the data rows to spaceS (8px)? It'd be nice to have a bit more separation because of the multi-line content, but not a blocker (especially if the SDS component makes that tricky).

Done, now it looks like this:
Screenshot 2022-11-04 at 2 43 24 PM

src/frontend/src/common/types/bioinformatics.d.ts
```diff
@@ -51,6 +51,13 @@ interface Tree {
   id: number;
 }
 
+interface TemplateArgs {
+  filterPangoLineages?: Array<string>;
+  filterStartDate?: string;
+  filterEndDate?: string;
+  locationId?: string;
```
vincent-czi: Pretty sure this should be a number. Here's what I'm seeing locally for the API response: 
image

vincent-czi: Although maybe this will cause some fun annoyances later because you'll need to use that number to do a lookup on an object, so it will be implicitly cast to a string because JavaScript, and perhaps TypeScript will not be happy about that?

src/frontend/src/views/Data/components/TreeTableNameCell/index.tsx
```diff
@@ -18,12 +31,55 @@ interface Props {
   item: PhyloRun;
 }
 
+const getDateRangeString = (item: PhyloRun): string => {
+  // NOTE: The start date default is covid-specific. We will need to update for
+  // other pathogens
+  const startDate = item.templateArgs?.filterStartDate || "2019-12-23";
+  const endDate =
+    item.templateArgs?.filterEndDate || item.startedDate.slice(0, 10);
```
vincent-czi, line 41: Nit: could you add a brief comment explainer about `item.startedDate` making sense when the filterEndDate is missing? I think it makes sense in the context of actively thinking about this like we are, but in a couple months when I forget that context, I think having the `startedDate` for the `endDate` is just going to seem like a typo rather than a purposeful choice because of it being an implicit end to when the tree had samples available for use.

src/frontend/src/views/Data/components/TreeTableNameCell/index.tsx
```diff
@@ -18,12 +31,55 @@ interface Props {
...
   const treeId = phyloTree?.id;
   const userName = user?.name;
   const isDisabled = status !== TREE_STATUS.Completed || !treeId;
+  const { data: namedLocationsData } = useNamedLocations();
+
+  const { data: groupInfo } = useGroupInfo();
+  const { namedLocationsById } = useMemo(() => {
+    if (!namedLocationsData) return {};
+    return {
+      namedLocationsById: reduceObjectArrayToLookupDict(
```
vincent-czi: Nit: Doing this reduce operation for every single tree name cell seems kind of expensive, especially if we have a lot of tree rows in the table. We just need the same lookup object for all of them, right?

This might definitely be in the category of premature optimization, so grain of salt and all that, but it would be nice if we could create this at a higher level one time and just pass down the same lookup object to all of the name cells. Counter-point: getting that to work with our current table architecture would probably be a real ugly pain. The parent to this -- if I recall correctly -- is very dynamic, so inserting it in a sane way would be tough.

Alternatively, you could create a new react-query for locations, ala the existing `useNamedLocations`. But instead if would have a more complicated `select` function to generate the lookup. I'm _pretty sure_ that would only cause the lookup generation to run once, even if all the cells are calling the new `useLocationsLookup`, but... double check me on that and verify the `select` really does only run once.

All that said, if you don't think it's worth the hassle / don't have the time, don't worry. But if you feel like you can spare the time, it seems like a reasonable nice to have: the locations response is reasonably big, like ~0.5MB and ~35k location objects in there (those numbers are from me checking Staging).

src/frontend/src/views/Data/components/TreeTableNameCell/index.tsx
```diff
@@ -18,12 +31,55 @@ interface Props {
...
+        namedLocationsData.namedLocations,
+        "id"
+      ),
+    };
+  }, [namedLocationsData]);
+
+  const getLocationName = () => {
+    const templateLocationId = item.templateArgs?.locationId;
+    // If there is no locationId in templateArgs, either this is an old
```
vincent-czi, line 58: This is such a good, clear comment.

src/frontend/src/views/Data/components/TreeTableNameCell/style.ts
vincent-czi: I wish I had something useful for you here, but I don't think I can come up with anything helpful.

Thanks for the F8 tip though! I've definitely had that annoyance in the past and never really done well with stuff disappearing when I move my mouse cursor but need to inspect it.

vincent-czi (approved):
Looks great! Most of the comments in here are just nice-to-haves, but I do think addressing the `locationId` type is probably important before merging. Everything else I'm fine with as is if you don't think it's worth it / disagree.

src/frontend/src/common/types/bioinformatics.d.ts
ehoops-cz: I thought typescript was mad about this, but apparently not.  It must have been something else.  Changing back to `number` seems to work fine 👍

src/frontend/src/views/Data/components/TreeTableNameCell/index.tsx
ehoops-cz: Good call.  Adding a `useNamedLocationsById` select function was easy enough

PR #1444, opened 2022-11-07
Maya/223991 sample modal manager

mmmmmmaya:
### Summary
- **What:** 
  - duplicate the modal files and move them to the appropriate space in the component tree
  - row selection (aka checked sample) state management in the new samples table
  - modal manager for the samples that controls modal state and rendering
- **Why:** Having the modals isolated in their own manager ensures the code for the samples view doesn't get very crowded with state that is only used in one place (ie: whether or not any of the 5 modals are shown). 
- **Ticket:** [[sc-223991]](https://app.shortcut.com/genepi/story/223991)
- **Env:** https://maya-modals-frontend.dev.czgenepi.org/

### Demos
after

### Notes
- There is one bug in this code (that I know about 😬), which is when closing the edit/delete sample modals, the checkboxes do not visually uncheck, even though in the component logic they are no longer checked.
- I am probably going to slightly modify the way the samples table and view keep track of checked sample state in the future, but tbd.
- You should be able to completely ignore the content of most of these files. They were copied as entire directories and pasted in place. Here are some interesting files you might want to check for correctness:
  - SamplesView/index
  - SamplesTable/index
  - SampleTableModalManager/index
  - SampleTableActions/index

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/frontend/src/views/Data/components/SamplesView/components/SamplesTable/index.tsx
```diff
@@ -136,6 +141,14 @@ const SamplesTable = ({ data, isLoading }: Props): JSX.Element => {
     onRowSelectionChange: setRowSelection,
   });
 
+  useEffect(() => {
+    const newCheckedSamples = table
+      .getSelectedRowModel()
+      .rows.map((r) => r.original);
```
vincent-czi, line 149: `original` seems kind of magical here in the context of whether or not a row is selected.
...
Okay, I've thought about it for awhile, and I'm pretty sure this is what's going on: the table has some kind of model for pulling out just those rows that are selected. We then want to convert all of those rows back into the original data behind the row, so we can set them as the checked samples, which will then allow the data for those checked samples to be used elsewhere. Is that right?

Either way, I'd like to see a short comment somewhere in here saying what's going on. Something like "Go through selected rows from table and pull data from those rows so can be used elsewhere", etc. As it stands, I think it's a bit unclear what's happening without context or needing to think for a minute or two.

src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+    const numberOfCheckedSamples = checkedSampleIds.length;
+    if (numberOfCheckedSamples > 0 && numberOfCheckedSamples <= 100) {
+      setCanEditSamples(true);
+    } else {
+      setCanEditSamples(false);
+    }
+  }, [checkedSampleIds]);
+
+  useEffect(() => {
+    if (shouldStartUsherFlow) setShouldStartUsherFlow(false);
```
vincent-czi, line 56: Nice-to-have: it looks like this is replicating the prior behavior, so it can be left as-is, but it seems odd that the Usher modal has this kind of "trip-wire" / signal behavior where it pulses true before flipping back to false, while all the other modals are simple on/off toggles. If you can spare the time and it's not too hard to refactor the Usher modal so it can operate the same way, I think that would be an improvement.

vincent-czi (approved):
Looks good. I only really looked over the four files you called out as being the ones with real changes, so I can't speak to everything, but those four made sense. I have a couple nit-type comments, but nothing that would block merging this as is.

src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/index.tsx
mmmmmmaya: for context, it was originally like this because it was a flow that contained 3 modals whose state depended on each other (to determine which of the 3 were open). But now it's been downsized to two and it hasn't caused us problems so far, which is why we haven't changed it yet

PR #1447, opened 2022-11-09
Bump GISAID ingest instance type to r5d.x16large

lvreynoso:
### Summary:
- **What:** GISAID ingest jobs are failing due to a lack of storage space. This bumps us to an instance size with 2.4 TB storage from our current 1.8 TB.
- **Ticket:** [sc227167](https://app.shortcut.com/genepi/story/227167)
- **Env:** `<rdev link>`

### Notes:
This is just a quick fix. Over the long term it might be worth it to ask the SWIPE team to support storage-optimized instance types - if we want to bump up the storage capacity any more, any im4gn is going to be cheaper than an r5.x24large.

See also #1343 

### Checklist:
- [x] I added labels to my PR


jgadling (approved):
I'm provisionally approving this -- I'm not sure we're specifically running an instance out of disk space vs running too many jobs on the same host, but I think this won't hurt at least.

PR #1449, opened 2022-11-09
Maya/227588 sample table styles

mmmmmmaya:
### Summary
- **What:** Some styling polish for the samples table
- **Ticket:** [[sc-227588]](https://app.shortcut.com/genepi/story/227588)
- **Design:** https://www.figma.com/file/BhgBQIacuxHYRGumpLJfDx/CZ-GE-Table-Refactor?node-id=0%3A1

#### Before: 
Screenshot 2022-11-09 at 1 24 01 PM

#### After: 
Screenshot 2022-11-09 at 1 26 04 PM

### Notes
There will need to be more styling work in the future. This is a first pass polishing.

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/views/Data/components/SamplesView/components/SamplesTable/index.tsx
+++ b/src/frontend/src/views/Data/components/SamplesView/components/SamplesTable/index.tsx
@@ -209,15 +209,6 @@ const columns: ColumnDef<Sample, any>[] = [
     cell: DefaultCell,
   },
   {
-    id: "lineage",
-    accessorKey: "lineage",
-    header: ({ header }) => (
-      <SortableHeader header={header}>Lineage</SortableHeader>
-    ),
-    cell: DefaultCell,
-    enableSorting: true,
-  },
-  {
     id: "gisaid",
     accessorKey: "gisaid",
     header: ({ header }) => (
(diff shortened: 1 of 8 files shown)


src/frontend/src/views/Data/components/SamplesView/components/SamplesTable/index.tsx
```diff
@@ -72,24 +88,48 @@ const columns: ColumnDef<Sample, any>[] = [
...
     ),
-    cell: ({ getValue }) => <CellBasic primaryText={getValue()} />,
+    cell: ({ getValue, row }) => {
+      const uploader = row?.original?.uploadedBy.name;
+      return (
+        <StyledPrivateId
+          primaryText={getValue()}
+          secondaryText={uploader}
+          shouldTextWrap
+          primaryTextWrapLineCount={1}
```
vincent-czi, line 124: What exactly does `primaryTextWrapLineCount` this mean? After 1 line of wrapped text it truncates? This is a Janeece question and maybe something you've already talked about this with her, but it seems like this could cause problems if users can't always be guaranteed to be able to see the entirety of the private id value. (That said, if we have user re-sizable columns -- I forget if that's being set up with the refactor -- it's not so big a deal since if they really do want to read their exact private id, they could widen the column as needed.)

vincent-czi (approved):
Looks good!

src/frontend/src/views/Data/components/SamplesView/components/SamplesTable/index.tsx
mmmmmmaya: It wraps up to x number of lines, and then it truncates with an ellipsis. However, the id column has a tooltip on hover that the user can use to view the entire ID even if the column is truncated

PR #1452, opened 2022-11-09
[SC-225249] add pathogen tab for monkeypox

ehoops-cz:
### Summary:
- **What:** `add pathogen tab for monkeypox`
- **Ticket:** [sc225249](https://app.shortcut.com/genepi/story/225249)
- **Env:** `none`

### Demos:

https://user-images.githubusercontent.com/109251328/200962345-a231bf2a-9da7-4895-998d-d274ab2f2b98.mov

https://user-images.githubusercontent.com/109251328/200962364-abff206a-088f-44ef-8467-c81d50e74ab1.mov

### Notes:
* The rest of the nav bar updates are in a separate ticket.
* The copy updates are in a separate ticket.

### Checklist:
- [x] I added labels to my PR


src/frontend/src/components/NavBar/components/AppNavBar/components/PathogenTabs/style.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+export const StyledTabs = styled(Tabs)`
+  align-items: center;
+
+  ${(props: CommonThemeProps) => {
+    const colors = getColors(props);
+    const palette = getPalette(props);
+
+    return `
+      // color: ${colors?.gray[400]};
```
mmmmmmaya: should all this stay?

src/frontend/src/components/NavBar/components/AppNavBar/components/PathogenTabs/style.tsx
```diff
@@ -0,0 +1,73 @@
...
+    const colors = getColors(props);
+    const palette = getPalette(props);
+    const spaces = getSpaces(props);
+    const fontWeights = getFontWeights(props);
+
+    return `
+      margin-right: ${spaces?.xl}px;
+      font-weight: ${fontWeights?.semibold};
+      color: ${colors?.gray[400]};
+      &:hover, :focus {
```
mmmmmmaya: could the next 3 blocks be all combined into one since they all have the same styling?

src/frontend/src/common/queries/phyloRuns.ts
```diff
@@ -43,14 +44,18 @@ export const USE_PHYLO_RUN_INFO = {
  * custom hook to automatically expire tree info when needed
  * such as when trees are deleted
  */
-export function usePhyloRunInfo(): UseQueryResult<PhyloRunResponse, unknown> {
-  return useQuery([USE_PHYLO_RUN_INFO], fetchPhyloRuns, {
+export function usePhyloRunInfo(
```
mmmmmmaya: instead of having every single consumer pass in the pathogen, you can read it from redux directly inside this function. See an example of how to read from the redux store without a hook at `api/index.tsx:28-34`

src/frontend/src/common/queries/samples.ts
```diff
@@ -278,14 +279,18 @@ export const USE_SAMPLE_INFO = {
   id: "sampleInfo",
 };
 
-export function useSampleInfo(): UseQueryResult<SampleResponse, unknown> {
-  return useQuery([USE_SAMPLE_INFO], () => fetchSamples(), {
+export function useSampleInfo(
+  pathogen: Pathogen
+): UseQueryResult<SampleResponse, unknown> {
+  return useQuery([USE_SAMPLE_INFO, pathogen], () => fetchSamples(), {
```
mmmmmmaya, line 286: same redux comment

src/frontend/src/common/queries/phyloRuns.ts
ehoops-cz: Nice, thank you!

ehoops-cz: @mmmmmmaya Making this change made switching between pathogens noticeably slower.  Any idea why?

ehoops-cz: Seems like it's because `useSampleInfo` isn't called again when the new tab loads.  The components are no longer watching for changes to the current pathogen.  I know putting the hook inside of useEffect is not good, but I essentially want something that will watch for when the current pathogen changes and then get new data when it does.

mmmmmmaya: I think you should be able to use redux middleware to run additional code (like expire an fe cache) whenever you dispatch a setCurrentPathogen redux action. You can see examples of existing middleware in middleware.ts

src/frontend/src/common/redux/middleware/index.ts
```diff
@@ -56,11 +61,25 @@ export const setGroupMiddleware: Middleware =
     return next(action);
   };
 
+const expirePathogenCaches = async (): Promise<void> => {
```
ehoops-cz, line 64: Is this the right file to put this in?  The groups one is in `src/common/queries/groups.ts`.  This one has query keys from multiple queries files, so I put it here.  Maybe we would want a separate queries file for it?

Janeece (approved):
LGTM! thanks

src/frontend/src/common/redux/middleware/index.ts
vincent-czi: I wasn't part of the original conversation around this, but it seems like a reasonable place to me, at least until we have to start wrangling more of these. This is the only file `expirePathogenCaches` gets called right now, so having it next to where it gets called seems as good as anything else.

src/frontend/src/common/redux/middleware/index.ts
```diff
@@ -56,11 +61,25 @@ export const setGroupMiddleware: Middleware =
...
+    await queryClient.fetchQuery([q]);
+  });
+};
+
 export const setPathogenMiddleware: Middleware =
   () => (next) => (action: AnyAction) => {
     const { type, payload } = action;
     if (type === CZGEReduxActions.SET_PATHOGEN_ACTION_TYPE) {
       setLocalStorage(ReduxPersistenceTokens.PATHOGEN, payload);
+      expirePathogenCaches();
```
vincent-czi, line 82: If you've been seeing this work fine, I have no reason to think this instance will give us trouble.

That said, I think this approach could open us up to some race conditions in certain edge cases. If the code that handled `SET_PATHOGEN_ACTION_TYPE` in the reducer happened to be very slow -- say it does a lot of computation in the reducer from the payload and current state before coming to a result that then gets placed in state -- we could theoretically have a situation where the query cache gets expired and the query refetches data, but while the underlying state is still stale (b/c reducer still running b/c slow), so the new query is doing a fetch for stale data. The code feels sequential: expire cache in middleware, set new value in reducer, query fetch based on new value. But I don't think we have a real guarantee that the query fetch has to happen after the new value is set.

All that said, for this specific instance (and probably almost every case in our app as currently build), I'd be pretty shocked if it didn't run in that sequence. I expect the speed of setting a new value in redux vs doing the whole cache expiration, run query fetch is so huge we can ignore the potential race condition.

vincent-czi (approved):
Looks good to me!

src/frontend/src/common/redux/middleware/index.ts
ehoops-cz: Hmm... I do see the potential race condition.  It seems ok for now, but if there's weirdness in the future we can use thunk middleware to address.

mmmmmmaya: I don't think that this creates a race condition, per se. But almost worse, it guarantees that the caches are updated before the pathogen is set. That is, it will make the API call using what `pathogen` is in the store before your update. Redux middleware executes before the action is applied to the state. One way you could get around this is to expire the caches, but to tell them which pathogen you want to fetch with when you refresh them.

https://redux.js.org/tutorials/fundamentals/part-4-store#middleware

PR #1459, opened 2022-11-15
Stop using the GisaidMetadata table

jgadling:
### Summary:
- **What:** This is the last step in migrating away from `gisaid`-specific tables
- **Env:** `<rdev link>`

### Notes:
This removes all of our dependencies on the `GisaidMetadata` and `PangoLineage` tables. This PR looks slightly larger than just that those changes because the replacement tables have additional fields that we need to add filters for: `pathogen_id` and/or `public_repository_id`

### Checklist:
- [x] I added labels to my PR


src/backend/aspen/api/deps.py
```diff
@@ -74,6 +74,22 @@ async def get_pathogen(
         raise ex.BadRequestException("Invalid pathogen slug")
 
 
+async def get_public_repository(
```
jgadling, line 77: A fastapi dependency that gets the default public repo for a given pathogen

src/backend/aspen/api/utils/repo_metadata.py
```diff
@@ -0,0 +1,86 @@
+from typing import Iterable, Mapping, Set, Tuple
```
jgadling, line 1: This is mostly renaming the gisaid.py file and updating the code to support extra columns

lvreynoso (approved):
lgtm

PR #1461, opened 2022-11-15
Prevent logging of sample ids in Pangolin jobs

lvreynoso:
### Summary:
- **What:** Currently we expand the WDL input containing a list of sample ids, and they got logged to CloudWatch. This turns off bash command tracing when we do that.
- **Env:** `<rdev link>`

### Checklist:
- [x] I added labels to my PR


vincent-czi (approved):
Code looks good, but may I ask why we're trying to avoid logging the sample ids? I didn't think we considered it an issue to capture the sample ids we were processing for debug purposes since they don't contain any PII in and of themselves. Is there another reason we want to avoid logging the sample ids, or is this in the PII realm?

lvreynoso:
I was concerned it could be a PII issue, on top of clogging up logs.

PR #1462, opened 2022-11-15
Maya/223992 tree modals

mmmmmmaya:
### Summary
- **What:** Add modal mgmt for trees table
- **Ticket:** [[sc-223992]](https://app.shortcut.com/genepi/story/223992)
- **Env:** https://maya-modalmgmt-frontend.dev.czgenepi.org/

### Testing
1. Test everything in the more actions menu

### Demos
after

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR


src/frontend/src/views/Data/components/TreesView/components/TreeCreateHelpLink/styles.ts
```diff
@@ -6,7 +6,7 @@ import {
   getFontWeights,
   getSpaces,
 } from "czifui";
-import { NewTabLink } from "../../../NewTabLink";
+import { NewTabLink } from "../../../../../../common/components/library/NewTabLink";
```
ehoops-cz: can this be an absolute path instead?  Seems far enough away.

src/frontend/src/views/Data/components/TreesView/components/TreesTable/components/TreeActionMenu/components/MoreActionsMenu/index.tsx
```diff
@@ -51,10 +68,40 @@ const MoreActionsMenu = ({
...
   const open = Boolean(anchorEl);
 
   return (
     <>
+      {usesTableRefactor && (
+        <>
+          <EditTreeConfirmationModal
+            open={isEditTreeModalOpen}
+            phyloRun={item}
+            onClose={() => setIsEditTreeModalOpen(false)}
```
ehoops-cz, line 96: I think moving this from an inline arrow function to a not-inline function makes sense for these.  We have a row for every tree and afaik we aren't paginating.  Seems like an easy (if small) performance win.

src/frontend/src/views/Data/components/TreesView/components/TreesTable/components/TreeActionMenu/components/MoreActionsMenu/index.tsx
```diff
@@ -51,10 +68,40 @@ const MoreActionsMenu = ({
...
+          />
+          <DeleteTreeConfirmationModal
+            open={isDeleteTreeModalOpen}
+            phyloRun={item}
+            onClose={() => setIsDeleteTreeModalOpen(false)}
```
ehoops-cz, line 101: - [x] ~~same^~~

src/frontend/src/views/Data/components/TreesView/components/TreesTable/components/TreeActionMenu/components/MoreActionsMenu/index.tsx
```diff
@@ -63,8 +110,7 @@ const MoreActionsMenu = ({
       >
         <span>
           <ButtonIcon
-            // TODO: it would be helpful for this to indicate which tree it's for
-            aria-label="more actions"
+            aria-label={`more actions for ${name}`}
```
ehoops-cz, line 113: 🎉

src/frontend/src/views/Data/components/TreesView/components/TreesTable/components/TreeActionMenu/components/OpenInNextstrainButton/index.tsx
```diff
@@ -1,7 +1,7 @@
 import { ButtonIcon, Icon, Tooltip } from "czifui";
 import { useState } from "react";
 import { TREE_STATUS } from "src/common/constants/types";
-import NextstrainConfirmationModal from "../../../NextstrainConfirmationModal";
+import NextstrainConfirmationModal from "../../../../../../../NextstrainConfirmationModal";
```
ehoops-cz: Absolute path?

src/frontend/src/views/Data/components/TreesView/components/TreesTable/components/TreeActionMenu/components/MoreActionsMenu/index.tsx
ehoops-cz: - [x] ~~move to not inline function~~

ehoops-cz (approved):
Looks good.  A couple minor comments.

src/frontend/src/views/Data/components/TreesView/components/TreesTable/components/TreeActionMenu/components/MoreActionsMenu/index.tsx
mmmmmmaya: @ehoops-cz Is this actually a performance win? I'm not clear on why. Even if I define the handlers as variables, wouldn't they still be redefined on every re-render unless I memoize?

ehoops-cz: I am under the impression that defining them as variables means they would be recreated on every re-mount of the component rather than recreating them on every re-render.  I might be incorrect though.

mmmmmmaya: I'm fairly certain it must be the case that variables are redefined on every re-render, at least for functional components (almost all of ours are). If not, then variables like `const shouldShowTooltip = Boolean(tooltipProps);` wouldn't hold different values on any render after the first one, even if the props changed. 

The reason I chose to inline these is because I feel it's more readable when the callback is just one line, rather than defining variables up top and then inlining them. Though I do prefer the vars if the callback is any more than just one line.

mmmmmmaya: I imagine this is a big part of why `useState` is a thing. If variables were not reset every render, then you could use a regular variable to hold component state. See [this](https://beta.reactjs.org/learn/state-a-components-memory#when-a-regular-variable-isnt-enough) `handleClick` example.

ehoops-cz: I'm convinced.  I also find them more readable.

vincent-czi: I agree with where this conversation landed. I don't think we need to worry about the function being created in-line every render.

That said, if it was a different situation and we really did need a way to avoid re-creating the function on every re-render, I think `useCallback` is what you'd want to reach for: https://reactjs.org/docs/hooks-reference.html#usecallback

My understanding is:
- `useMemo` -- You have a function that is expensive to run, but its output stays (mostly) the same between re-renders, so you want to only run the function when its inputs change, not on every single render. `useMemo` lets you capture that value and only re-run if an input changes, regardless of renders happening.
- `useCallback` -- You have a function that is expensive to create -- say, you have a function factory that outputs the function, but that factory is expensive to run -- or when keeping reference equality for the function between renders is helpful for child components. `useCallback` lets you capture that function and hold on to it so it doesn't need re-creation every re-render. It's basically a `useMemo`, but for functions (which the React docs explicitly call out).

PR #1464, opened 2022-11-17
[SC-225225 generate the general viral template upload/edit

ehoops-cz:
### Summary:
- **What:** `Refactor downloadable upload/edit template constants to work with multiple pathogens`
- **Env:** `<rdev link>`

**Would like some feedback on where to get the pathogen information.**  I think helper functions should take this as an argument rather than looking it up in redux directly.  However, passing the pathogen param through multiple layers of helper functions didn't seem great either.  I am currently in an inconsistent state with components getting the pathogen where they can, but some helpers also look in redux.  **Thoughts?**

### Demos:
Screenshot 2022-11-17 at 9 28 51 AM
Screenshot 2022-11-17 at 9 28 27 AM
Screenshot 2022-11-17 at 9 27 08 AM
Screenshot 2022-11-17 at 9 26 48 AM

### Notes:
* This includes updating the webform table headers
* This does not include updating copy on the edit/upload flows

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/components/DownloadMetadataTemplate/common/constants.ts
+++ b/src/frontend/src/components/DownloadMetadataTemplate/common/constants.ts
@@ -59,3 +59,160 @@ export const SAMPLE_EDIT_METADATA_KEYS_TO_HEADERS: PathogenConfigType<
   },
   [Pathogen.MONKEY_POX]: GENERAL_VIRAL_SAMPLE_EDIT_METADATA_KEYS_TO_HEADERS,
 };
+
+const DATE_FORMAT = "YYYY-MM-DD";
+const BOOLEAN_FORMAT = "Yes/No";
+
+export const EXAMPLE_SAMPLE_IDS = [
+  "Example Sample A",
+  "Example Sample B",
+  "Example Sample C",
+];
+
+export const EXAMPLE_CURRENT_PRIVATE_IDENTIFIERS = [
+  "example private ID A",
+  "example private ID B",
+  "example private ID C",
+];
+
+const GENERAL_VIRAL_UPLOAD_EXAMPLE_ROWS = [
+  // Very first example row helps explain usage, but not fully valid.
+  [
+    EXAMPLE_SAMPLE_IDS[0], // sampleId
+    "Private sample name", // privateId
+    "(if available) GenBank Accession", // publicId -- here as explainer
+    DATE_FORMAT, // collectionDate -- not valid, here as explainer in template
+    "North America/USA/California/Los Angeles County", // collectionLocation
+    DATE_FORMAT, // sequencingDate -- not valid, here as explainer in template
+    BOOLEAN_FORMAT, // keepPrivate -- not valid, here as explainer in template
+  ],
+  // Subsequent example rows are mostly valid, honest-to-goodness examples...
+  // ... except for the dates. This is to avoid Excel auto "correct".
+  [
+    EXAMPLE_SAMPLE_IDS[1], // sampleId
+    "id101", // privateId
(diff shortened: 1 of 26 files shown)


src/frontend/src/components/DownloadMetadataTemplate/common/constants.ts
```diff
@@ -14,22 +16,203 @@ export const BASE_METADATA_HEADERS = {
...
+        exampleCollectionLocation, //collectionLocation
+        "Delete", // sequencingDate
+        BOOLEAN_FORMAT, // keepPrivate
+      ],
+    ],
+    [Pathogen.MONKEY_POX]: [
+      [
+        EXAMPLE_CURRENT_PRIVATE_IDENTIFIERS[0], // currentPrivateID
+        "X3421876", // newPrivateID
+        "hMpxV/USA/demo-17806/2021", // publicId
```
vincent-czi, line 190: I'm not particularly concerned right now, but if we start to have to do this for more pathogens, we'll want to change this into a function that generates the array. Looks like the only thing I see changing is the pathogen prefix in the publicId (`hCoV-19` vs `hMpxV`).

vincent-czi (approved):
Overall, this looks fine. I agree that the way it's organized to actually get the pathogen info is kind of a pain, but I'm not sure I really have anything much better to suggest. I'd be happy to chat on a call if you want and we can try spitballing approaches, but I certainly don't have a silver bullet.

Two other notes:
- I'm a little apprehensive about what happens if the value selected from state for `pathogen` ever goes wonky. If we somehow pulled `null`, `undefined`, or `"WARBARGL"`, a bunch of lookups would immediately error out when the first look for the pathogen returned `undefined` instead of a proper object to then do the next lookup (`.sampleId`, etc). I spent a bit of time poking through the redux code for how those values get init-ed, and I'm pretty confident we don't have to worry about that ever happening, but if we ever get into a fall-through case I think stuff will just explode. But like I said, looking into a bit it seemed safe to make the assumption that it would always be a legit pathogen value?
- In some cases, maybe it would feel a little nicer to have a helper that returns a lookup for that pathogen, rather than looking up via pathogen. Easier to explain in code:
```
const pathogenSampleUploadMetadataKeysToHeaders = getMetadataKeysToHeaders(pathogen);
... later ...
pathogenSampleUploadMetadataKeysToHeaders.sampleId, etc

=== rather than current approach of ===
SAMPLE_UPLOAD_METADATA_KEYS_TO_HEADERS[pathogen].sampleId, etc
```
I don't really strongly believe in that second point, but if you think it looks nicer or would make some things simpler, maybe a good idea? Then again, it is kinda nice to have very obvious lookups for this kind of thing, that way we don't have some clever function that obfuscates we completely missed handling a certain pathogen's strings properly... Like I said, I really don't have any silver bullets, everything in the PR seems pretty reasonable to me.

src/frontend/src/components/DownloadMetadataTemplate/common/constants.ts
ehoops-cz: I agree.  The Upload example data has the GISAID -> GenBank change and the Edit example data has the publicId change.  All of the other data is the same.  I don't know off the top of my head an eloquent way to mix and match custom vs non-custom fields.  Using the generator idea from above might simplify this a bit.  Given the tight timeline of getting this project done and the uncertain outlook going forward, I'm leaning toward leaving this as is.  Maybe we will have some more time to revisit when we implement hepC.

vincent-czi: I agree, seems like the best move!

PR #1465, opened 2022-11-17
[SC-225226] warn user if user clicks from upload flow

ehoops-cz:
### Summary:
- **What:** `warn user if user clicks from upload flow`
- **Ticket:** [sc225226](https://app.shortcut.com/genepi/story/225226)
- **Env:** `none`

### Demos:
Screenshot 2022-11-17 at 11 21 29 AM

### Notes:
The current change makes switching between pathogens go to the samples page for the new pathogen.  With this change, nothing else is needed for the pop-up to work in the upload flow.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/components/NavBar/components/AppNavBar/components/PathogenTabs/index.tsx
+++ b/src/frontend/src/components/NavBar/components/AppNavBar/components/PathogenTabs/index.tsx
@@ -1,7 +1,12 @@
 import { useRouter } from "next/router";
 import { useSelector } from "react-redux";
+import {
+  GROUP_URL_INDICATOR,
+  PATHOGEN_URL_INDICATOR,
+} from "src/common/appRouting";
 import { selectCurrentPathogen } from "src/common/redux/selectors";
 import { Pathogen } from "src/common/redux/types";
+import { ROUTES } from "src/common/routes";
 import { StyledTab, StyledTabs } from "./style";
 
 type PathogenTabEventHandler = (
@@ -11,15 +16,18 @@ type PathogenTabEventHandler = (
 
 export const PathogenTabs = (): JSX.Element => {
   const router = useRouter();
-
   const currentPathogen = useSelector(selectCurrentPathogen);
 
   const handleTabClick: PathogenTabEventHandler = (_, value) => {
     const currentRoute = router.asPath;
-    const newRoute = currentRoute.replace(
-      /pathogen\/.*?(\/|$)/,
-      `pathogen/${value}/`
-    );
+
+    const groupRegex = new RegExp(`${GROUP_URL_INDICATOR}\\/.*?(\\/|$)`);
+    const groupUrlMatch = currentRoute.match(groupRegex);
+    const groupUrl = groupUrlMatch ? groupUrlMatch[0] : "";
+
+    // When the user changes between pathogens, we always start them on the samples page
+    const newRoute = `${ROUTES.DATA_SAMPLES}/${groupUrl}${PATHOGEN_URL_INDICATOR}/${value}`;
+
     router.push(newRoute);
   };
 
--- a/src/frontend/src/components/NavBar/components/AppNavBar/components/PathogenTabs/index.tsx
+++ b/src/frontend/src/components/NavBar/components/AppNavBar/components/PathogenTabs/index.tsx
@@ -1,9 +1,6 @@
 import { useRouter } from "next/router";
-import { useSelector } from "react-redux";
-import {
-  GROUP_URL_INDICATOR,
-  PATHOGEN_URL_INDICATOR,
-} from "src/common/appRouting";
+import { useDispatch, useSelector } from "react-redux";
+import { setPathogen } from "src/common/redux/actions";
 import { selectCurrentPathogen } from "src/common/redux/selectors";
 import { Pathogen } from "src/common/redux/types";
 import { ROUTES } from "src/common/routes";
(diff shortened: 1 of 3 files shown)


src/frontend/src/components/NavBar/components/AppNavBar/components/PathogenTabs/index.tsx
```diff
@@ -11,15 +16,18 @@ type PathogenTabEventHandler = (
...
-      /pathogen\/.*?(\/|$)/,
-      `pathogen/${value}/`
-    );
+
+    const groupRegex = new RegExp(`${GROUP_URL_INDICATOR}\\/.*?(\\/|$)`);
+    const groupUrlMatch = currentRoute.match(groupRegex);
+    const groupUrl = groupUrlMatch ? groupUrlMatch[0] : "";
+
+    // When the user changes between pathogens, we always start them on the samples page
+    const newRoute = `${ROUTES.DATA_SAMPLES}/${groupUrl}${PATHOGEN_URL_INDICATOR}/${value}`;
```
mmmmmmaya, line 29: I don't think this should be necessary. I think if you set the current pathogen in redux, the path should automatically update to reflect that.

```suggestion
  const currentPathogen = useSelector(selectCurrentPathogen);
  const dispatch = useDispatch();

  const handleTabClick: PathogenTabEventHandler = (_, value) => {
    dispatch(setPathogen(value));
  };
```

src/frontend/src/components/NavBar/components/AppNavBar/components/PathogenTabs/index.tsx
```diff
@@ -11,15 +16,18 @@ type PathogenTabEventHandler = (
...
+
     router.push(newRoute);
```
mmmmmmaya: I don't think this should be necessary. I think if you set the current pathogen in redux, the path should automatically update to reflect that.

```suggestion
  const currentPathogen = useSelector(selectCurrentPathogen);
  const dispatch = useDispatch();

  const handleTabClick: PathogenTabEventHandler = (_, value) => {
    dispatch(setPathogen(value));
  };
```

ehoops-cz: Thanks! I ended up needing to add `router.push(ROUTES.DATA_SAMPLES)` to make sure we end up on the samples page for the new pathogen.

vincent-czi: I could be wrong, but I _think_ the `router.push(ROUTES.DATA_SAMPLES)` is actually necessary to cause the path to get updated. Messing around locally with manual redux dispatches and looking through the code, I think the path only gets checked/updated when the user navigates to one of the `workspacePaths` (DATA, DATA_SAMPLES, PHYLO_TREES, and UPLOAD). But if the redux pathogen value gets updated and no navigation takes place, the route stays stale. So my theory is that it's a combination of the redux pathogen changing and the router.push event (to a workspacePath) that causes the route to update.

If you don't include the `router.push(ROUTES.DATA_SAMPLES)` line at all, does the `pathogen/FOO` part of the route still update?

vincent-czi (approved):
Looks good! I left a comment about the router push part: I think that might be pretty important to how the pathogen part of the browser route gets updated, not sure the pathogen change on its own does it. I don't know if it's worth chasing any of that down though if you see it working consistently.

src/frontend/src/components/NavBar/components/AppNavBar/components/PathogenTabs/index.tsx
ehoops-cz: No it doesn't update without the `router.push` portion

ehoops-cz: Doing some testing by:
1. clicking SC2 tab
2. clicking on the Trees tab
3. clicking on the MPX tab
and vice versa

I'm seeing weird behavior with both versions.  In the explicit path version sometimes the tree table switches to the new pathogen (causing a flash of the mpx tree tab) and then switches to the mpx sample tab.  Usually this starts happening if I click through a few times.  

With the redux dispatch/router push version sometimes there is a long delay until all of the tabs align with the data on the screen.  Similarly, this happens mostly after multiple clicks through the steps above. 

Honestly both of these seem pretty jank.  I'm planning to see what I can do this afternoon.


https://user-images.githubusercontent.com/109251328/202783317-9561bd52-e1cb-46a4-a62b-c4e4619084d4.mov

ehoops-cz: ^this video is using the redux dispatch/router approach

ehoops-cz: I'm not finding a much better solution.  The issues with this seem to only happen when rapidly switching around between sample/tree tabs as well as pathogen tabs.  I don't think this is a very natural flow for real users.  I'm leaning towards leaving it as is and we can revisit in the future.  Do either of you have concerns?

mmmmmmaya (approved):

src/frontend/src/components/NavBar/components/AppNavBar/components/PathogenTabs/index.tsx
vincent-czi: I'm not super concerned. I think as long as we're aware that it's a possible issue we can continue to keep an eye on it and maybe re-factor some of how the routing works down the road once we have a better idea of what's needed/important.

PR #1472, opened 2022-11-23
MPX and HCV DB support

phoenixAja:
### Summary:
- **What:** db migration to add MPX and HCV support to the DB for pathogen and prefixes
- **Ticket:** [sc225240](https://app.shortcut.com/genepi/story/225240)
- **Env:** `<rdev link>`

### Checklist:
- [ ] I added labels to my PR


src/backend/database_migrations/versions/20221122_222651_mpx_hepc_support.py
```diff
@@ -0,0 +1,35 @@
...
+revision = "20221122_222651"
+down_revision = "20221101_201105"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute(
+        """
+        INSERT INTO aspen.pathogens (id, slug, name) VALUES ((SELECT MAX(id)+1 FROM aspen.pathogens), 'HCV', 'Hepatitis C'), ((SELECT MAX(id)+2 FROM aspen.pathogens), 'MPX', 'Monkeypox')
```
jgadling: instead of max(id)+1, please use DEFAULT so PostgreSQL uses its sequences, or you can just omit the `id` field from the insert query entirely.

jgadling: For example, either:
```
insert INTO my_table (id, name) VALUES (DEFAULT, 'bob')
```
or
```
INSERT INTO my_table (name) VALUES ('bob')
```

Basically either of the above queries uses postgresql's default field value, which is attached to a sequence that keeps track of the next available id for our tables.

phoenixAja: yeah the issue with that is that i accidentally added the ids manually when i did the last insert for the pathogens/pathogen_repo_config tables, I'm working on the solution now.

jgadling: Err, isn't that fixed now? You should be able to update the max value of the sequences if you really have to.

vincent-czi:
oh, uh, I have pretty much the same migration in my work, but just without HCV

src/backend/database_migrations/versions/20221122_222651_mpx_hepc_support.py
vincent-czi: Yeah, the default thing works as far as I can tell. In the migration I made, it does it via default (omit setting the id in the SQL) and the result is as expected.

image

https://github.com/chanzuckerberg/czgenepi/blob/rdev-vince-workflow-nextclade/src/backend/database_migrations/versions/20221122_231047_add_mpx_pathogen_and_nextclade_datasets.py

phoenixAja: sorry i meant i still needed to add it for the pathogen_repo_configs

jgadling (approved):
Please coordinate with @vincent-czi on resolving migration versions.

phoenixAja:
@vincent-czi can i merge this since you are going back to the drawing board with your migrations?

vincent-czi:
@phoenixAja I'd rather you didn't, I've actually got what I'm doing figured and just need a little more time to wrap up this afternoon.

PR #1476, opened 2022-11-23
On-demand Lineage/QC/Mutations workflow (Nextclade)

vincent-czi:
### Summary:
- **What:** Adds new pipeline/workflow for doing on-demand calling of Lineage/QC/Mutations calling (via Nextclade)
- **Ticket:** [sc<227162>](https://app.shortcut.com/genepi/story/<227162>)
- **Env:** vince-nextclade-test-frontend.dev.czgenepi.org (no changes in FE, but step func / workflow is live there)

### Demos:

I uploaded a total of 12 samples: 10 of them were real covid sequences, 1 of them was just a total garbage string, and 1 of them was a real monkeypox sequence. In that order, those were PK sample_ids `17814 <-> 17823`, `17824`, and `17825`. Below is the DB result in the rdev for the `sample_qc_metrics` table afterwards. Everything worked as expected. (There is a `raw_qc_output` available in there too that looks as expected, but it's sooooo long it blows up the ability to take a screenshot and point at it for demo purposes.)

Screen Shot 2022-11-22 at 8 43 46 PM

Here is one example of the `raw_qc_output`: it parses correctly as JSON and seems to have everything I would expect.

image

### Notes:

- This PR has a number of DB migrations. I'm going to create a separate PR in a little while that has just those migrations on their own, all cherry-picked out. We can merge that one ahead of this one to prevent SQLAlchemy from briefly freaking out before the migrations land.
- In talking to Sidney, it turns out we need to capture certain bits of info about the Nextclade dataset we're using in the calling process to be able to unambiguously know the provenance of the call / know when a call is stale compared to current info and it's time for a refresh. Those pieces of info are the datasets `name`, `accession`, and `tag`, along with the software and software version (Mutations are a bit of an exception, they can skip name and tag, but still need accession). For now, I've just tacked on new columns as needed to the various tables and each row entry saves that.
  - It's arguable that this is a bad approach: it's not very normalized, we're duplicating the info for a given run in a bunch of rows. I hear that and agree, but I also think it wouldn't be very hard to later go and improve things with a centralized, normalized table that tracks tool calls. Since we only ever care about the latest run, we wouldn't have much difficulty in migrating over to that improved way. Plus we already store a ton of info in the raw output columns, so I think the current approach doesn't really change things much either way and we can wait on improving it if we decide it's worth it.
- Working on this, I discovered that some samples can come out as fully error-ed: they show up with data in the `nextclade.errors.csv` and also have corresponding error-type columns in the usual `nextclade.csv`. I worked through this with Sidney, and she came to the decision that any such sample should have its run info discarded. We also discussed this with Christine and Janeece, and we all agreed to introduce an extra `qc_status` of "invalid" in that case (in addition to the three we get from Nextclade of good, mediocre, and bad).

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR
- [x] I ~added~ updated relevant unit tests


.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -244,6 +243,48 @@ module pangolin_ondemand_sfn_config {
...
+  memory   = 16000
+  wdl_path = "workflows/lineage-qc-autorun.wdl"
+  custom_stack_name     = local.custom_stack_name
+  deployment_stage      = local.deployment_stage
+  remote_dev_prefix     = local.remote_dev_prefix
+  stack_resource_prefix = local.stack_resource_prefix
+  swipe_comms_bucket    = local.swipe_comms_bucket
+  swipe_wdl_bucket      = local.swipe_wdl_bucket
+  sfn_arn               = local.swipe_sfn_arn
+  # TODO set up a schedule once we decide on one.
```
phoenixAja, line 259: we do have a schedule decided in the design doc for the cadence this should be run for the different pathogens

src/backend/aspen/database/models/lineages.py
```diff
@@ -106,13 +108,61 @@ class SampleLineage(idbase):  # type: ignore
     lineage_probability = Column(Float, nullable=True)
     raw_lineage_output = Column(JSONB, nullable=True)
 
+    # For Nextclade, we need to track the underlying reference data bundle
+    # that was involved to know how the lineage call was made.
+    # TODO FIXME: Lots of table/row duplication here. _Probably_ better to
+    # normalize it as a single entry somewhere and then reference that.
+    # Evaluate later -- because of how we use this, not hard to change later.
+    # Search this to find all instances: §NextcladeTagDuplicationTODO
+    reference_dataset_name = Column(String, nullable=True)
```
phoenixAja, line 117: is this the nextclade dataset name? We have that field in the pathogens table.

phoenixAja: sidney said the dataset name is fairly stable and should rarely change.

vincent-czi: You're right, this should be duplicating the info 99.9% of the time, but it gives us an unambiguous way to have all the call info together. It's related to pulling the info out of the `tag.json` file that comes with the nextclade dataset. In case things ever get muddled, I wanted to consider the `tag.json` the real source of truth.

Alternatively, we could raise an error in the workflow if the `tag.json` name and the pathogens.nextclade_dataset_name were different.

src/backend/aspen/workflows/nextclade/save.py
```diff
@@ -13,117 +14,235 @@
...
+            # We always record QC info for any sample run, even if invalid.
+            qc_score = row["qc.overallScore"]
+            qc_status = row["qc.overallStatus"]
+            if not is_result_valid:
+                # HACK Setting `qc_score` to empty string is a bit of an abuse.
+                # DB has that field as a non-nullable string, but Nextclade's
+                # qc.overallScore is actually a number. We store CSV text of
+                # that number, so this is a compromise to indicate no result.
+                qc_score = ""
+                qc_status = INVALID_RESULT_STATUS
```
phoenixAja: we could just change the qc_score to be nullable?

src/backend/aspen/database/models/lineages.py
vincent-czi: Oh, I forgot, the other benefit is that if we ever changed the `pathogens.nextclade_dataset_name` for a given pathogen in the future, that would cause it to auto-trigger a new run on all the impacted samples. That should be a very rare event, but it means that we definitely wouldn't need to go manually trigger the run for those samples. (But that might also be silly, if we change the name, the underlying accession and/or tag is likely to change... but now that I write that out, I'm not so sure. Some of the datasets are just specialized versions of others, so they might have more overlap than I think.)

src/backend/aspen/workflows/nextclade/save.py
vincent-czi: Yeah, if you think that's the way to go, I'm happy to go make it nullable. I think you and Jess were both out when I wrote that, and I wasn't sure if there was a strong need for `qc_score` to not be nullable, so I just did that in the meantime and forgot to ask when you were back.

phoenixAja (approved):
this all looks pretty good to me, though i would love @jgadling or @lvreynoso to take a look too if possible before merging. Nice work!

.happy/terraform/modules/ecs-stack/main.tf
vincent-czi: You're right. I do want to talk that over with you though: the schedule has it running on different times for different pathogens, but I think we can just have it run weekly for all of them, but the ones that update less frequently will just not actually execute anything because they won't need updating.

But whatever we do, we'll need to keep the schedule part commented out until that workflow gets written.

vincent-czi:
From Phoenix's notes, it sounds like there might be the need for some more DB migration tweaking, so I'm going to hold off on making that PR with the cherry-picked migration commits until that's figured.

src/backend/aspen/workflows/nextclade/prep_samples.py
```diff
@@ -0,0 +1,119 @@
...
+def cli(
+    sample_ids_fh: io.TextIOBase,
+    sequences_fh: io.TextIOBase,
+    pathogen_info_fh: IO[str],
+):
+    """
+    Writes out a FASTA for the specified samples.
+
+    - sample_ids_fh: A file of sample ID primary keys, one ID per line.
+        Each sample must be for the same pathogen (all SARS-CoV-2, etc)
```
jgadling, line 42: I think (?) our API prevents users from uploading samples for more than one pathogen per batch, but making this script more flexible (for example, implementing a `--pathogen mpx` flag) might make it easier to implement the scheduled job.

Basically converting this from an assumption to an explicit input might both simplify the code and make this script more reusable.

src/backend/aspen/workflows/nextclade/prep_samples.py
```diff
@@ -0,0 +1,119 @@
...
+                    f"{pathogen_of_all_samples.slug} in given samples. There "
+                    f"may also be others, this is just first difference found."
+                )
+                print(err_msg)
+                raise RuntimeError("Samples given are not all same pathogen")
+
+            uploaded_pathogen_genome = sample.uploaded_pathogen_genome
+            # Samples _should_ always have uploaded_pathogen_genome with
+            # sequence data on them in theory, but if it's missing, blow up.
+            if uploaded_pathogen_genome is None:
```
jgadling: if you add an `innerjoin=True` parameter to the joinedload above, you won't have to check for errors here.

src/backend/aspen/workflows/nextclade/save.py
jgadling: if it needs to be nullable, let's make it nullable 🤷‍♀️

src/backend/aspen/workflows/nextclade/prep_samples.py
vincent-czi: Each batch of samples would have to be for a single pathogen at a time, though, right? I had actually originally planned to do it with some kind of explicit call, but I realized that the provided batch of samples implicitly says what the pathogen being operated on is, because the samples in the batch point at their pathogen.

We could make it explicit, but I don't think it makes it more flexible for the current approach. I think it could arguably make it safer: if we think we're calling on MPX, but the provided samples were something else, we should probably fail so we notice the mismatch sooner.

For the scheduled job, I'm planning on having the runner batch up the sample lists by pathogen and then run the on-demand script on each of those in turn.

jgadling: yeah, but why bother having a separate batch-creation script? This script can handle that functionality with a few minor adjustments.

src/backend/aspen/database/models/sequences.py
```diff
@@ -100,6 +100,31 @@ class UploadedPathogenGenome(PathogenGenome):
     sequencing_depth = Column(Float)
     upload_date = Column(DateTime, nullable=False, server_default=func.now())
 
+    def get_stripped_sequence(self) -> str:
```
jgadling, line 103: Are there standard fasta-parsing libraries we should be using instead of rolling our own?

jgadling (approved):
Approved with nits.

src/backend/aspen/workflows/nextclade/prep_samples.py
vincent-czi: Wouldn't it need to do a mixed batch of pathogens then? The batch-creation script is intended to handle all pathogens at once on a scheduled basis. I figured a good dividing line was for the on-demand to handle a single pathogen at a time.

jgadling: The scheduled job can just kick off one sub-job per pathogen, it doesn't have to create the batches itself.

vincent-czi: Talked on Zoom about this: I'm sold! We'll use the same process for the sub-job and it will use a lack of sample_ids (or maybe another arg?) as a way to indicate a general run for the pathgoen. Then also adding a little additional logic for checking over the entire set of samples for a given pathogen and updating the stales in that case.

For now, I'll leave this PR as-is around this aspect, but I'll plan to use that approach when I add in the scheduled aspect down the road.

jgadling: I think the sub-job (and tbh, even this on-demand job) would benefit from getting an explicit pathogen argument.

vincent-czi: Sure, but I still think we should make use of the implicit pathogen in the listed samples, even if it's just to raise an error when they differ.

jgadling: fair enough.

PR #1477, opened 2022-11-28
Migrations: Lineage QC

vincent-czi:
### Summary:
- **What:** Migrations in support of PR #1476
- **Ticket:** None
- **Env:** None (although PR above has a rdev I've been testing on)

### Notes:
This breaks out all the commits from #1476 that are for the migrations. Doing so we can merge and deploy this before the other PR to prevent any outages while SQLAlchemy model mismatches DB.

### Checklist:
- [x] I merged latest `trunk`
- [x] I manually verified the change (both locally and on rdev)
- [x] I added labels to my PR

### DIFF
--- a/src/backend/database_migrations/versions/20221122_231047_add_mpx_pathogen_and_nextclade_datasets.py
+++ b/src/backend/database_migrations/versions/20221122_231047_add_mpx_pathogen_and_nextclade_datasets.py
@@ -1,0 +1,40 @@
+"""Add MPX pathogen and nextclade datasets
+
+Create Date: 2022-11-22 23:10:48.896041
+
+Adds in new pathogen of MPX (Monkeypox).
+Also adds in appropriate nextclade_dataset_name for all pathogens so far
+(the two, SC2 and MPX). This is the value used to fetch the reference
+dataset for running Nextclade tool. If you need to add or update
+nextclade_dataset_name values in the future, check with Comp Bio
+about what the appropriate value to use is for each pathogen: Nextclade
+usually has multiple, different possible datasets per pathogen that focus
+on different aspects. Don't just pick the first you see!
+"""
+import enumtables  # noqa: F401
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "20221122_231047"
+down_revision = "20221122_064351"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # Appropriate `nextclade_dataset_name` gotten from Comp Bio team member
+    op.execute(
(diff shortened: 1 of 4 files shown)


jgadling (approved):

vincent-czi:
I had missed some commits that weren't the migration themselves, but one important adjustment and one linting thing. Added those in.

src/backend/database_migrations/versions/20221122_231047_add_mpx_pathogen_and_nextclade_datasets.py
```diff
@@ -0,0 +1,40 @@
...
+    # Appropriate `nextclade_dataset_name` gotten from Comp Bio team member
+    op.execute(
+        """UPDATE aspen.pathogens
+        SET nextclade_dataset_name = 'sars-cov-2'
+        WHERE slug = 'SC2'
+        """
+    )
+    op.execute(
+        """INSERT INTO aspen.pathogens (slug, name, nextclade_dataset_name)
+        VALUES ('MPX', 'Monkeypox', 'hMPXV')
```
phoenixAja, line 34: i think locally we already have an MPX entry, does this migration still successfully run locally?

vincent-czi: Nope! I'm going to tell folks to start a new DB from scratch locally. 😅

It's not great, but I think it's better that we have consistency. The way the local is getting the MPX entry is totally different than how we get it on our deployed DBs. I chatted about this a little with @jgadling and it seemed like the better thing to do was to bring them in to alignment, even if it does cause a bit of local dev pain.

vincent-czi: To be clear, SC2 is part of the migrations, but locally we have MPX being brought in via the `setup_localdata.py`. I think it's better to standardize on migrations, since then it's the same thing locally as deployed, and we had already been using migrations as our way of brining in static pathogen stuff.

phoenixAja: so we'll need to update the `setup_localdata` script to go along with this then right?

vincent-czi: It's actually fine as is. We could potentially deprecate that part of it, but as currently written it just becomes a no-op with the migration in place.

PR #1478, opened 2022-11-29
Write Sample Lineages

jgadling:
### Summary:
- **What:** Have the Pangolin job write pangolin lineage data to `sample_lineages`
- **Ticket:** [sc226127](https://app.shortcut.com/genepi/story/226127)

### Notes:
There's not a lot to see here, this PR does what it says on the tin. The major changes are:
- Update the nightly pangolin job to only fetch SC2 samples
- Create-or-update SampleLineage rows for each sample we do a lineage call on, as well as updating the `pathogen_genome.pango*` columns, for reverse compatibility
- Update the tests to generate samples for multiple pathogens, to make sure filtering/etc works. The tests rely much less on hardcoded values now, but I couldn't find a reasonable way to avoid hardcoded ID's in the `lineage_report.csv` file, so maybe we can take another stab at that later.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/workflows/pangolin/find_samples.py
+++ b/src/backend/aspen/workflows/pangolin/find_samples.py
@@ -12,7 +12,7 @@ from aspen.database.connection import (
     session_scope,
     SqlAlchemyInterface,
 )
-from aspen.database.models import Sample
+from aspen.database.models import Pathogen, Sample
 
 
 def check_latest_pangolin_version() -> str:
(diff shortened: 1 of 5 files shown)


src/backend/aspen/workflows/pangolin/find_samples.py
```diff
@@ -25,15 +25,18 @@ def check_latest_pangolin_version() -> str:
...
     with session_scope(interface) as session:
         # filter for sequences that were run with an older version of pangolin
 
-        all_samples: Collection[Sample] = session.query(Sample).options(
-            joinedload(Sample.uploaded_pathogen_genome)
+        all_samples: Collection[Sample] = (
+            session.query(Sample)
+            .options(joinedload(Sample.uploaded_pathogen_genome))
+            .join(Sample.pathogen)
+            .where(Pathogen.slug == pathogen)
```
jgadling, line 39: Filter for SC2 by default.

src/backend/aspen/workflows/pangolin/save.py
```diff
@@ -71,7 +101,30 @@ def cli(pangolin_fh: io.TextIOBase, pangolin_last_updated: datetime):
...
+            else:
+                lineage = SampleLineage(
+                    sample=pathogen_genome.sample,
+                    lineage_type=lineage_type,
+                    lineage=pango_info["lineage"],  # type: ignore
+                    lineage_software_version=pango_info["version"],  # type: ignore
+                    lineage_probability=pango_info["probability"],  # type: ignore
+                    raw_lineage_output=pango_info["full_output"],  # type: ignore
+                )
+                pathogen_genome.sample.lineages.append(lineage)
```
jgadling, line 122: Add or update a SampleLineage row depending on whether one already exists

src/backend/aspen/workflows/pangolin/save.py
```diff
@@ -71,7 +101,30 @@ def cli(pangolin_fh: io.TextIOBase, pangolin_last_updated: datetime):
...
+
+            if current_chunk_size > commit_chunk_size:
+                session.commit()
+                current_chunk_size = 0
```
jgadling, line 126: Commit every N rows instead of for every row. Should speed things up for large #'s of samples.

phoenixAja (approved):

PR #1479, opened 2022-11-29
update samples list response 

phoenixAja:
### Summary:
- **What:** update samples list response to include SampleQCMetrics and SampleLineages, also includes FE changes to accept the new response format. Sorry this PR got a little beefy, just figured it would be best to do the fullstack implementation at once. 
- **Ticket:** [sc226127](https://app.shortcut.com/genepi/story/226127)
- **Env:** [`<rdev link>`](https://update-samples-response-frontend.dev.czgenepi.org)

### Checklist:
- [x] I added labels to my PR


src/backend/aspen/api/schemas/samples.py
```diff
@@ -22,14 +24,20 @@ class SampleGisaidResponse(BaseResponse):
     status: str
 
 
+# TODO: add field to capture pangolin_last_updated (last_updated) once we add support for that
 class SampleLineageResponse(BaseResponse):
-    last_updated: Optional[datetime.datetime]
```
phoenixAja, line 26: we didn't move this over in the last migration to populate this table, are we dropping this field from the DB?

src/backend/scripts/setup_localdata.py
```diff
@@ -127,6 +132,79 @@ def create_pathogens(session):
     return pathogens
 
 
+def create_sample_lineage(session, sample):
```
phoenixAja, line 135: adding some test data to localdev

phoenixAja:
realizing the filtering FE logic broke with this, converting back to a draft until i have a fix!

phoenixAja:
actually nevermind it is working 😅

phoenixAja:
creating rdev now

src/backend/aspen/api/schemas/samples.py
phoenixAja: i'm adding this field back in a future migration, but for now let's leave it off since that field currently doesn't exist on the SampleLineages table

src/backend/aspen/api/schemas/samples.py
```diff
@@ -60,7 +81,7 @@ class SampleGetterDict(GetterDict):
             if obj.uploaded_pathogen_genome
             else None
         ),
-        "lineage": format_sample_lineage,
+        "lineages": format_sample_lineage,
```
jgadling, line 84: Do we still need to use `format_sample_lineage` or can we rely on pydantic's built-in serialization now?

phoenixAja: it's a little nice for adding the `qc_score` to the lineages output  (to pull from the associated `sample.qc_metrics`) and also we still need to pull the scorpio calls from the `raw_lineage_output`, depending on sample.pathogen. So i think it's still worth having `format_sample_lineages`

jgadling: I think there might be a more pydantic-style way of serializing this object. Can you look into whether using `@validator()` methods can help us pull out the scorpio/other data we need? I think the `PhyloRunResponse` object has some interesting logic in it that might be a useful reference

src/backend/aspen/api/schemas/samples.py
```diff
@@ -84,7 +105,6 @@ class Config:
     collection_location: LocationResponse
     czb_failed_genome_recovery: bool
     gisaid: Optional[SampleGisaidResponse]
-    lineage: Optional[SampleLineageResponse]
```
jgadling, line 87: Keep in mind that removing this field from the backend in the same deployment that updates the frontend to stop using it could make the UI behave strangely while we update our service to the latest version. I don't think we need to stress it for every change, but I want to make sure we're making conscious decisions to accept these tradeoffs.

jgadling (changes requested):

src/frontend/e2e/utils/schemas/sampleResponseData.ts
```diff
@@ -35,4 +26,30 @@ export type SampleResponseData = {
...
+      scorpio_call?: string;
+      scorpio_support?: string;
+      qc_status?: string;
+    }
+  ];
+  qc_metrics: [
+    {
+      qc_score?: string;
+      qc_software_version?: string;
+      qc_status?: string;
```
vincent-czi: Aren't some of these -- like `qc_status` -- guaranteed if we get a qc_metric showing up? I know a lot of the fields here in lineages and qc_metrics are optional / null-able in BE / DB, but some of them are guaranteed to come through right? Is there a reason the FE type needs the optional `?` there?

src/frontend/src/common/types/bioinformatics.d.ts
```diff
@@ -9,12 +9,26 @@ interface GISAID {
...
+  reference_dataset_tag?: string;
+  scorpio_call?: string;
+  scorpio_support?: string;
+  qc_status?: string;
+}
+
+interface QCMetrics {
+  qc_score?: string;
+  qc_software_version?: string;
+  qc_status?: string;
```
vincent-czi: Same comment about `?` optional-ness

src/backend/aspen/api/schemas/samples.py
phoenixAja: yeah in the interest of time i was trying to do all this at once, i personally think it would be an ok tradeoff this one time? I was thinking that the update would be fairly fast, so the amount of time there would potentially be funkiness would be an ok tradeoff? I do agree that we should minimize this in the future.

jgadling: yeah, I think it's ok.

vincent-czi (approved):
I only looked over the FE aspects, but it generally looks good, certainly no blockers.

That said -- and this might be a controversial idea -- I think it might be better for the BE to _lie_ about the multiple lineages aspect for now. We're going through a lot of `d.lineages[0]` type stuff in the FE to filter down to only the first lineage/QC entry, but we don't have any designs in place for handling multiple entries. So we'd need to go back eventually and change all that hardcoding to using the first entry (and hopefully not miss any somewhere when we do that). We have a clear assumption that all our samples only ever have one lineage/QC/mutation entry at a time. We've built it so it can eventually support multiple, but we assume in our usage at the moment that there really is only one. We might as well compartmentalize that to the BE and let the FE stay in the dark about the list aspect for now.

That said, the work here is already done and it's not bad or anything, so grain of salt. I think it just makes a cleaner break to compartmentalize the list aspect in the BE and eventually move to supporting multiple entries in one go once we actually have the FE designs to support them.

src/frontend/e2e/utils/schemas/sampleResponseData.ts
phoenixAja: for the brief period of time when a user uploads new samples and the nextclade job is runnign the samples may not have the associated lineage/qc_metrics right?

phoenixAja:
@vincent-czi that's a good idea about the BE faking only one entry, if i have some time to go back and add that in a follow up PR

src/frontend/e2e/utils/schemas/sampleResponseData.ts
vincent-czi: Yeah, but they wouldn't have anything in here then, right? It's an array of objects, but if the object exists in the array that key should have a guaranteed value. I don't think it's possible for the lineage/qc_metric to exist but that value to be missing from it. Either it's an empty list or it'll be there.

src/backend/aspen/api/schemas/samples.py
phoenixAja: in this case i think it's ok, i was doing this all together because of the time constraints to get it finished before the sprint ended. I think i would accept UI funkiness while the backend is updating for this, but i agree with trying to minimize this in the future, and i will take more time to do that next time.

phoenixAja: my original plan was to do this in steps but the time got away from me a bit and i wanted to get this in on time. I will definitely take more care in the future around this.

src/frontend/e2e/utils/schemas/sampleResponseData.ts
phoenixAja: yeah good point, i'll go back and add not optional to some of these, for some reason i was thinking they needed to be optional if the whole object could not exist but that was the wrong thinking.

jgadling: wait - the backend supports multiple lineage/qc types, and so the frontend should too, right? I also don't think the frontend should grab the first element in the list - it can use a feature flag to figure out which lineage/qc call is the default and use that.

jgadling (approved):

PR #1481, opened 2022-11-29
Add multi-pathogen support to Usher options

lvreynoso:
### Summary:
- **What:** Adds support for storing usher options by pathogen and correctly displaying the options according to the pathogen selected on the frontend.
- **Ticket:** [sc<226107>](https://app.shortcut.com/genepi/story/226107)
- **Env:** `<rdev link>`

### Demos:
Screenshot 2022-11-28 at 20 56 54

### Checklist:
- [x] I added labels to my PR


lvreynoso:
Looks like I need to update some tests.

src/backend/aspen/database/models/usher.py
```diff
@@ -14,6 +16,8 @@ class UsherOption(idbase, DictMixin):  # type: ignore
     value = Column(String, unique=True, nullable=False)
     # `priority` is order we display options to user. LOWEST number is max priority.
     priority = Column(Integer, unique=True)
```
vincent-czi: I saw you dropped the specific, `priority` only unique constraint in the migrations -- which is great, I was definitely thinking that would be a good thing to do as I read through the PR -- can you also drop the `unique=True` here in the model for that column?

The migration also adds a multi-col unique constraint, so can you also add a `__table_args__` to match that here in the model as well please? Something like this:
```
    __table_args__ = (
        UniqueConstraint(
            "sample_id",
            "mutations_caller",
        ),
    )
```

I definitely think you did the right thing in your migration, just want to keep the model in sync with the DB.

vincent-czi (commented):
Looks good! I'm going to hold off on actually putting in the "Approve" just until the tests get fixed, but it's spiritually approved. I've got one comment in there that should get addressed to keep the model in sync with the DB, but other than that minor thing, everything looks solid.

src/backend/aspen/database/models/usher.py
lvreynoso: Right, I overlooked that. Thanks for catching it!

src/backend/database_migrations/versions/20221129_012949_add_pathogen_column_to_usher_options.py
```diff
@@ -0,0 +1,68 @@
...
+        WHERE pathogen_id IS NULL
+        """
+    )
+    op.alter_column(
+        "usher_options",
+        "pathogen_id",
+        existing_type=sa.INTEGER(),
+        nullable=False,
+        schema="aspen",
+    )
```
jgadling, line 64: Should this PR include some `INSERT` statements for the options for MPX trees?

lvreynoso: added!

jgadling (approved):

PR #1482, opened 2022-11-29
Add an indicator so we can easily determine if an account is in the CZGE_internal segment in split.

ehoops-cz:
### Summary:
- **What:** `Add section to account page to indicate whether an account is in the 'internal_user' split group.`
- **Ticket:** none
- **Env:** `none`

### Demos:
Screenshot 2022-11-29 at 9 05 01 AM

### Notes:
We will use the CZGE_internal segment in split to test the new samples and trees tables.  Since we expect no change, it will be helpful to have a quick way to determine whether an account is in the segment.  This will not show for users who are not in the CZGE_internal segment.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/components/Split/types.ts
+++ b/src/frontend/src/components/Split/types.ts
@@ -18,6 +18,7 @@ export enum SPLIT_SIMPLE_FLAG {
 export enum USER_FEATURE_FLAGS {
   // my_flag_name = "my_flag_name", (<-- format example)
   galago_integration = "galago_integration",
+  internal_user = "internal_user",
   multi_pathogen = "multi_pathogen",
   table_refactor = "table_refactor",
   tree_location_filter = "tree_location_filter",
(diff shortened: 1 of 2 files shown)


vincent-czi (approved):
Looks good! Long-term, this might be a nice place to stow indicators for any flags we want to watch generally. Only `internal_user` folk will get the section displayed, but within that section we could display a bunch of flag statuses. Or maybe just have some iterator that runs through all our flags and drops the values. But I'm just spit-balling long-term ideas, this is a great start.

ehoops-cz:
> Looks good! Long-term, this might be a nice place to stow indicators for any flags we want to watch generally. Only `internal_user` folk will get the section displayed, but within that section we could display a bunch of flag statuses. Or maybe just have some iterator that runs through all our flags and drops the values. But I'm just spit-balling long-term ideas, this is a great start.

Agreed!

PR #1484, opened 2022-11-30
Select Usher db query param argument based on current pathogen

lvreynoso:
### Summary:
- **What:** Selects the proper value for the `db` query parameter in Usher links based on the user's currently selected pathogen.
- **Ticket:** [sc<226107>](https://app.shortcut.com/genepi/story/226107)
- **Env:** `<rdev link>`

### Checklist:
- [x] I added labels to my PR


src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/components/UsherTreeFlow/index.tsx
```diff
@@ -14,14 +15,20 @@ interface Props {
...
   remoteFile: string,
   treeType: string,
-  sampleCount: number
+  sampleCount: number,
+  pathogen: Pathogen
 ) => {
   const encodedFileLink = encodeURIComponent(remoteFile);
 
-  const DB_PARAM = `db=wuhCor1`;
+  const DB_PARAM = USHER_DBS[pathogen];
```
vincent-czi: I don't believe this will work as-is. The old version was a query param of `db=foobar`, but the `USHER_DBS` lookup is just pulling `foobar` by itself. I assume it still needs to be turned into a query param?

src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/components/UsherTreeFlow/index.tsx
```diff
@@ -14,14 +15,20 @@ interface Props {
   shouldStartUsherFlow: boolean;
 }
 
+const USHER_DBS: Record<Pathogen, string> = {
```
vincent-czi: This feels like something the FE shouldn't need to know. That said, it's what we were doing in the past, the FE knew the DB_PARAM, this is just generalizing it, so it's not any worse than it was previously. But if you can see a clean/easy way to make this a piece of info the BE is returning with the rest of the USHER options info, that would be a nice-to-have.

src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/components/UsherTreeFlow/index.tsx
lvreynoso: ahh I overlooked that, good point. I'll fix it

src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/components/UsherTreeFlow/index.tsx
lvreynoso: Yeah, that makes sense.

ehoops-cz: Can this type should use the `PathogenConfigType`?  This was originally just for the strings.ts files, but I've been using it for anything that is pathogen-specific. `src/components/DownloadMetadataTemplate/prepMetadataTemplate.ts` has and example on line 19.

- [x] use PathogenConfigType (or just mark complete if it's not a good idea)

ehoops-cz: Plus one to this doesn't really belong on the FE, but also plus one to doing something practical

ehoops-cz (approved):

PR #1485, opened 2022-11-30
[SC-225247] move upload button

ehoops-cz:
### Summary:
- **What:** `Move upload button from App Nav to top of samples table`
- **Ticket:** [sc225247](https://app.shortcut.com/genepi/story/225247)
- **Env:** `none`

### Demos:
Screenshot 2022-11-30 at 2 34 04 PM

### Notes:
The pathogen tabs remain behind the multi_pathogen split flag.  However, moving the upload button and moving the group menu is not behind a flag.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/components/SampleTableActions/index.tsx
+++ b/src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/components/SampleTableActions/index.tsx
@@ -1,3 +1,5 @@
+import { Link } from "czifui";
+import { ROUTES } from "src/common/routes";
 import { IconButton } from "../IconButton";
 import { MoreActionsMenu } from "./components/MoreActionMenu";
 import { TreeSelectionMenu } from "./components/TreeSelectionMenu";
@@ -5,6 +7,7 @@ import {
   Divider,
   StyledChip,
   StyledSelectedCount,
+  StyledUploadButton,
   StyledWrapper,
   TooltipDescriptionText,
   TooltipHeaderText,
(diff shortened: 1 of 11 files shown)


Janeece (approved):
LGTM!

src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/components/SampleTableActions/index.tsx
```diff
@@ -77,6 +80,17 @@ const SampleTableActions = ({
         onEditSelected={openEditSampleModal}
         data-test-id="sample-page-more-action-btn"
       />
+      <Link href={ROUTES.UPLOAD_STEP1} passHref>
```
vincent-czi, line 83: I think this is not actually an issue, but I'm confused how there are two `Link href={ROUTES.UPLOAD_STEP1}` in different files that seem to be the same link? In the code, it appears to be the whole set of menu buttons is identical here (in the `SampleTableModalManager`) and in the `data_subview`. I guess I'm just curious what's going on...

ohhhh, is this the new table vs the old table? okay, okay, I get it then.

vincent-czi (approved):
Looks good!

src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/components/SampleTableActions/index.tsx
ehoops-cz: Ya, old and new tables.  It's awkward because the previous implementation used next link, similar to this one, but I ended up removing the next link because it wasn't passing the href properly (probably because the sds component doesn't actually take that field).  It's possible this link will need to be updated to also not use next link.

PR #1486, opened 2022-12-01
Kickoff for on-demand Mutations/QC/Lineages

vincent-czi:
### Summary:
- **What:** Adds kicking off the job/workflow for calling Mutations/QC/Lineages via Nextclade upon sample upload
- **Ticket:** [sc<220816>](https://app.shortcut.com/genepi/story/<220816>)
- **Env:** yes... but no. I had an rdev going for this `vince-kickoff-ondemand-spikemutations`, but there was separately some Happy weirdness, and it wound up getting cleaned up over the course of working. But I did manage to complete the testing on it and you can see that in the demos below.

### Demos:

I uploaded a set of 12 samples total, 10 known good, actual COVID genomes, the other 2 as bad genomes (one trash, one MPX). Job ran as expected when inspecting step function and logs. Results came out as expected in various tables.

For all the demos below, these are not the entirety of the DB row, just partials. The entire row for each has a gigantic chunk of the raw output as JSON, so makes for some pretty ugly screenshots, so I just made it enough cols to use for reasonable demo screenshots.

The two invalid samples are `17826` and `17827`.

---

**QC Metrics**
You can see the 10 good samples have QC accordingly, while the bad samples have the `invalid` keyword to show a failed Nextclade run. (There's an extra 2 samples in here from an earlier test with 2 other invalid samples. You can see those at the top, 17814 and 17815.)

Screen Shot 2022-11-30 at 3 46 37 PM

---

**Mutations**

You can see the mutations result calls for the 10 good samples. The bad samples do not have a row at all, because if a sample fails the Nextclade run, we do not record its mutation info (because it's trash, there's nothing there, and Sidney told me to do so).

Screen Shot 2022-11-30 at 3 49 18 PM

---

**Lineages**

Is this a demo? This is a demo of not having done anything, I guess? Because these samples were all given as SC2 pathogens, we did not get their lineages from Nextclade. These values were all populated by the already existing Pangolin on-demand job/workflow. The new work of this PR did not interfere or accidentally change any of that. (You can also see that the two invalid samples have a trash Pangolin call because they're not SC2 genomes.)

Also note that all the reference dataset info (name, accession, tag) is left as NULL, because that's a Nextclade specific thing. Since these entries did not come Nextclade, there's nothing there for any of them. 

Screen Shot 2022-11-30 at 3 51 52 PM

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR


jgadling:
Is there anything we can do re: test coverage?

scripts/setup_dev_data.sh
```diff
@@ -154,19 +154,21 @@ ${local_aws} ssm put-parameter --name /genepi/local/localstack/nextstrain-sfn --
   "RunSPOTVcpu":10,
   "StateMachineArn":'${LOCAL_SFN_ARN}'
 }'
-${local_aws} ssm delete-parameter --name /genepi/local/localstack/pangolin-ondemand-sfn
```
vincent-czi, line 157: This commit chunk looks weird and wrong -- why is pangolin changing into lineage-qc, we need both! -- but it's really just two different pieces of work smushed together.

1. Added in a new local SSM param for doing lineage-qc
2. I noticed that we had duplicated `pangolin-ondemand-sfn` local SSM params. There were two of them, with the second simply clobbering the first, meaning the first was doing nothing in the loading script. I chatted a bit with Jess about it, it seems like the first is probably a better representation of real SSM param at this point, so I deleted this one, the second one. The second actually was what we were deving against since it was doing the clobbering, but I tested it locally and there's no issues.

vincent-czi:
> Is there anything we can do re: test coverage?

I'm not really sure. I looked a bit and didn't see any tests for kicking off the Pangolin jobs. We do have some test coverage of the Nextclade run itself in its workflow folder, though. But I don't see an easy way to test the end-to-end kick-off aspect.

jgadling (approved):

PR #1488, opened 2022-12-01
mpx hcv support - migration

phoenixAja:
### Summary:
- **What:** in my last PR i renamed my migration file and it got dropped from the PR 😵 SORRY
- **Env:** `<rdev link>`

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/src/backend/database_migrations/versions/20221130_222651_mpx_hepc_support.py
+++ b/src/backend/database_migrations/versions/20221130_222651_mpx_hepc_support.py
@@ -1,4 +1,5 @@
-"""
+"""add MPX and HCV DB support for gisaid/genbank template
+
 
 Create Date: 2022-11-30 22:26:58.178142
 
--- a/src/backend/database_migrations/versions/20221130_222651_mpx_hepc_support.py
+++ b/src/backend/database_migrations/versions/20221130_222651_mpx_hepc_support.py
@@ -21,8 +21,11 @@ def upgrade():
         """
     )
 
-    op.execute("select nextval('aspen.pathogen_repo_configs_id_seq')")
-    op.execute("select nextval('aspen.pathogen_repo_configs_id_seq')")
+    op.execute(
+        """
+        SELECT setval('aspen.pathogen_repo_configs_id_seq', (SELECT MAX(id) FROM aspen.pathogen_repo_configs)+1);
+        """
+    )
 
     op.execute(
         """
--- a/src/backend/database_migrations/versions/20221130_222651_mpx_hepc_support.py
+++ b/src/backend/database_migrations/versions/20221130_222651_mpx_hepc_support.py
@@ -31,7 +31,7 @@ def upgrade():
         """
         INSERT INTO aspen.pathogen_repo_configs (prefix, public_repository_id, pathogen_id) VALUES
              ( 'hcv', (SELECT id from aspen.public_repositories WHERE name='GenBank'), (SELECT id from aspen.pathogens WHERE slug='HCV') ),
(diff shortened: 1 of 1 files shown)


src/backend/database_migrations/versions/20221130_222651_mpx_hepc_support.py
```diff
@@ -0,0 +1,37 @@
...
+
+def upgrade():
+    op.execute(
+        """
+        INSERT INTO aspen.pathogens (slug, name) VALUES ('HCV', 'Hepatitis C')
+        """
+    )
+
+    op.execute("select nextval('aspen.pathogen_repo_configs_id_seq')")
+    op.execute("select nextval('aspen.pathogen_repo_configs_id_seq')")
```
jgadling:
```suggestion
    op.execute(
        """
        SELECT setval('aspen.pathogen_repo_configs_id_seq', (SELECT MAX(id) FROM aspen.pathogen_repo_configs)+1);
        """
    )
```

jgadling: It's more robust to use the current max val to update the sequence

src/backend/database_migrations/versions/20221130_222651_mpx_hepc_support.py
```diff
@@ -0,0 +1,37 @@
...
+
+    op.execute(
+        """
+        INSERT INTO aspen.pathogen_repo_configs (prefix, public_repository_id, pathogen_id) VALUES
+             ( 'hcv', (SELECT id from aspen.public_repositories WHERE name='GenBank'), (SELECT id from aspen.pathogens WHERE slug='HCV') ),
+             ( 'hMpxV', (SELECT id from aspen.public_repositories WHERE name='GenBank'), (SELECT id from aspen.pathogens WHERE slug='MPX') )
```
jgadling: I think the genbank prefix is 'MPXV' or 'MpxV' [based on these search results](https://www.ncbi.nlm.nih.gov/labs/virus/vssi/#/virus?SeqType_s=Nucleotide&VirusLineage_ss=Monkeypox%20virus%20(monkeypox),%20taxid:10244&Completeness_s=complete&HostLineage_ss=Homo%20sapiens%20(human),%20taxid:9606) maybe @sidneymbell can clarify?

sidneymbell: Ooh good catch. You are correct, @jgadling . Weird / annoying that genbank and gisaid differ on this but /shrug

src/backend/database_migrations/versions/20221130_222651_mpx_hepc_support.py
vincent-czi: I agree that if we want contiguous ids, this is the better approach, but I'm not sure about the `+1` part. When the default insert pulls via `nextval` on the sequence, I think it might already have a +1 in the returned value from `nextval`. If @jgadling is confident this code is correct, then just go ahead, but if not, you might want to mess around on your local DB for a bit, verify there's no off-by-one-error.

src/backend/database_migrations/versions/20221130_222651_mpx_hepc_support.py
phoenixAja: do we prefer MPXV or MpxV?

src/backend/database_migrations/versions/20221130_222651_mpx_hepc_support.py
phoenixAja: awesome thanks @jgadling for the suggestion

jgadling (approved):

PR #1491, opened 2022-12-02
Use repository config prefix to auto-generate public id's

jgadling:
### Summary:
- **What:** We were hardcoding `hCoV-19` as the default public identifier prefix for samples. This uses the split flag value to identify the default repository for a pathogen and uses the appropriate prefix for generating public id's
- **Ticket:** [sc228425](https://app.shortcut.com/genepi/story/228425)
- **Env:**  https://fixprefix-frontend.dev.czgenepi.org/

### Notes:
Since I updated the API endpoints to use the split feature flag for a pathogen's default public repository, I had to update a lot of tests to include a SplitClient dependency as well.

### Checklist:
- [ ] I added labels to my PR


src/backend/aspen/api/views/phylo_runs.py
```diff
@@ -56,6 +63,7 @@ async def kick_off_phylo_run(
     group: Group = Depends(require_group_privilege("create_phylorun")),
     pathogen: Pathogen = Depends(get_pathogen),
     public_repository: PublicRepository = Depends(get_public_repository),
+    pathogen_repo_config: PathogenRepoConfig = Depends(get_pathogen_repo_config),
```
jgadling, line 66: This dependency uses split.io to determine the default repo for a pathogen.

src/backend/aspen/api/views/samples.py
```diff
@@ -200,7 +208,9 @@ async def update_samples(
         # workaround for our response serializer
         sample.show_private_identifier = True
 
-        sample.generate_public_identifier(already_exists=True)
+        sample.generate_public_identifier(
+            pathogen_repo_config.prefix, already_exists=True
```
jgadling, line 212: We're passing the prefix into the name generator here so we're not stuck with `hcov-19/` for everything

src/backend/aspen/database/models/sample.py
```diff
@@ -217,11 +217,11 @@ def generate_public_identifier(self, already_exists=False):
         if already_exists:
             id = self.id
             self.public_identifier = (
-                f"hCoV-19/{country}/{group_prefix}-{id}/{current_year}"
+                f"{prefix}/{country}/{group_prefix}-{id}/{current_year}"
```
jgadling, line 220: Use the prefix we got instead of a hardcoded identifier.

src/backend/aspen/test_infra/models/pathogen_repo_config.py
```diff
@@ -1,35 +1,65 @@
...
-def setup_gisaid_and_genbank_repo_configs(
+def pathogen_repo_config_factory(prefix, pathogen, public_repository):
+    pathogen_repo_config = PathogenRepoConfig(
+        prefix=prefix, pathogen=pathogen, public_repository=public_repository
+    )
+
+    return pathogen_repo_config
+
+
+def setup_random_repo_configs(
```
jgadling, line 18: By default, creates a pathogen, two pathogen repos, and sets a split feature flag to determine which one is the default.

src/backend/aspen/test_infra/models/pathogen_repo_config.py
```diff
@@ -1,35 +1,65 @@
...
+        split_client.get_pathogen_treatment.return_value = (
+            default_config.public_repository.name
+        )
+    return pathogen, default_config
 
 
-def pathogen_repo_config_factory(prefix, pathogen, public_repository):
-    pathogen_repo_config = PathogenRepoConfig(
-        prefix=prefix, pathogen=pathogen, public_repository=public_repository
+def setup_gisaid_and_genbank_repo_configs(
```
jgadling, line 53: For the few endpoints that have specific `gisaid` vs `genbank` switches (like the metadata download), make it easy to create those public repositories instead of random ones.

src/backend/aspen/api/views/tests/test_create_phylo_runs.py
```diff
@@ -31,53 +30,11 @@ async def test_create_phylo_run(
...
-    assert response["template_args"] == {}
-    assert response["workflow_status"] == "STARTED"
-    assert response["group"]["name"] == group.name
-    assert response["user"]["name"] == user.name
-    assert response["user"]["id"] == user.id
-    assert response["pathogen"]["slug"] == pathogen.slug
-    assert "id" in response
-
-
-async def test_create_phylo_run_mpx(
```
jgadling, line 65: We didn't need specific/separate SC2/MPX tests since our test now generates random pathogens.

lvreynoso (approved):
lgtm. I did wonder if we should be storing the default repo for a pathogen in split instead of our db, but it seems like it's a bit late to be asking that question 😅

jgadling:
> lgtm. I did wonder if we should be storing the default repo for a pathogen in split instead of our db, but it seems like it's a bit late to be asking that question 😅



Yeah, I'm (clearly) a huge fan of feature flags, but *this* one seems significant enough that maaaaybe it belongs in the DB instead. We should consider moving it when we get a chance.

PR #1499, opened 2022-12-07
fix the last_updated migration

phoenixAja:
### Summary:
- **What:** there was an issue with the last migration where[ null values were being inserted into the db on the last update](https://github.com/chanzuckerberg/czgenepi/actions/runs/3633234849/jobs/6130011309)
- **Env:** `<rdev link>`

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/src/backend/database_migrations/versions/20221201_174505_add_last_updated_lineages.py
+++ b/src/backend/database_migrations/versions/20221201_174505_add_last_updated_lineages.py
@@ -23,14 +23,14 @@ def upgrade():
 
     conn = op.get_bind()
     insert_last_updated_sql = sa.sql.text(
-    """
-    UPDATE aspen.sample_lineages
-    SET last_updated = t.pangolin_last_updated FROM (SELECT *
-    FROM aspen.uploaded_pathogen_genomes upg
-    INNER JOIN aspen.pathogen_genomes pg
-    ON pg.entity_id = upg.pathogen_genome_id
-    ) t
-    WHERE aspen.sample_lineages.sample_id = t.sample_id
+        """
+        UPDATE aspen.sample_lineages
+        SET last_updated = t.pangolin_last_updated FROM (SELECT *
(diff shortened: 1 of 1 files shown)


src/backend/database_migrations/versions/20221201_174505_add_last_updated_lineages.py
```diff
@@ -22,10 +22,16 @@ def upgrade():
...
+    # only tacking on the last_updated portion to the query.
+    # this is still valid at this time because we are still
+    # populating the pathogen_genome fields from the pangolin save script.
+
     insert_accessions_sql = sa.sql.text(
         """
-        INSERT INTO aspen.sample_lineages (last_updated)
-        SELECT pg.pangolin_last_updated
+        INSERT INTO aspen.sample_lineages (sample_id, lineage_type, lineage, lineage_probability, lineage_software_version, raw_lineage_output, last_updated)
+        SELECT s.id, 'PANGOLIN', pg.pangolin_lineage, pg.pangolin_probability, pg.pangolin_version, pg.pangolin_output, pg.pangolin_last_updated
```
phoenixAja: @jgadling is it ok that i'm modifying the previous migration that failed or should i create a new migration file? Also if you know a better way to do this let me know

jgadling: I think you need to update the existing sample lineages rows to include the new field?

phoenixAja: yes i updated the query

jgadling (approved):

src/backend/database_migrations/versions/20221201_174505_add_last_updated_lineages.py
phoenixAja: fyi @vincent-czi this is the update query

PR #1505, opened 2022-12-13
[SC-228029] Add privacy policy preview

lvreynoso:
### Summary:
- **What:** Adds a preview of the Jan 3 privacy policy, adds an announcement banner on all non-public pages, includes an anchor link that goes directly to the **Data CZ Gen Epi Collects** section of the policy.
- **Ticket:** [sc228029](https://app.shortcut.com/genepi/story/228029)
- **Env:** `<rdev link>`

### Demos:
Screenshot 2022-12-13 at 11 17 17
Screenshot 2022-12-13 at 11 18 00

### Notes:
Adds an `Announcements` component, rendered in the `AppNavBar`, for this and future announcements.

The link `/privacy-preview#privacy-data-collection` will go straight to the aforementioned data collection section of the policy.

### Checklist:
- [ ] I added labels to my PR


lvreynoso:
[sc-228029]

src/frontend/src/common/redux/utils/groupUtils.ts
```diff
@@ -14,7 +14,15 @@ export const ensureValidGroup = async (): Promise<void> => {
 
   const { dispatch, getState } = store;
   const state = getState();
-  const rawUserInfo = await fetchUserInfo();
+  let rawUserInfo: RawUserRequest | null = null;
+  try {
+    rawUserInfo = await fetchUserInfo();
+  } catch (e) {
+    return;
```
vincent-czi: Nit: I generally prefer `return null` when it's being used to say "hey, don't render this component", it feels slightly clearer to me.

src/frontend/src/common/redux/utils/groupUtils.ts
```diff
@@ -14,7 +14,15 @@ export const ensureValidGroup = async (): Promise<void> => {
...
+  }
+  if (rawUserInfo == null) {
```
vincent-czi: Should be triple `===`, not double

vincent-czi: (aside: I would have thought our linter would flip out over this? I guess not. We should probably configure it that way.)

src/frontend/src/views/PrivacyPreview/index.tsx
```diff
@@ -0,0 +1,1217 @@
...
+  TopRow,
+  UnderLineHeader,
+} from "./style";
+
+const PrivacyPolicyPreview = (): JSX.Element => {
+  const renderIntro = () => (
+    <>
+      <Title>
+        <H1>Chan Zuckerberg GEN EPI (formerly Aspen) Privacy Policy</H1>
+        <H4>Last Updated: January 1, 2023</H4>
```
vincent-czi, line 31: @Janeece  The announcements banner says "We are updating our Privacy Policy to comply with CCPA, effective January 3, 2023." but this date is set for Jan 1. Seems like they should match?

Also, it feels to me like the preview privacy policy page should indicate somewhere that it is currently only a **preview** and has not actually gone into force yet. Along those lines, maybe shouldn't say "Last Updated", but rather "will take effect on" or something? That's more of a legal question though.

Anyway, neither of those are code issues, but wanted to flag them still.

src/frontend/src/common/redux/utils/groupUtils.ts
```diff
@@ -14,7 +14,15 @@ export const ensureValidGroup = async (): Promise<void> => {
 
   const { dispatch, getState } = store;
   const state = getState();
-  const rawUserInfo = await fetchUserInfo();
+  let rawUserInfo: RawUserRequest | null = null;
+  try {
```
vincent-czi: I'm surprised about needing the try/catch here around the `fetchUserInfo`. It probably shouldn't bubble up an error even if something goes wrong. Is this the outstanding problem for non-logged-in users on the homepage where the requests get swatted down b/c no auth, so we have errors because of that, or is it something else?

I do think we should deal with that at some point, but I feel like it should happen more inside the level of the react-query `fetchUserInfo` or sub-functions doing the actual request than up here at the call site. Maybe we could just make an eng excellence ticket to deal with these kinds of expected auth "errors" when not logged in and hopefully we'll get to it in the future?

vincent-czi (approved):
Code-wise, everything looks good. I only skimmed the actual new privacy policy component since it was mostly legal document and not really code, so I can't fully speak to that aspect.

Assuming that Product/Design/Legal will review on Staging very soon, I think we're good to merge.

src/frontend/src/views/PrivacyPreview/index.tsx
Janeece: Yeah, lets set the date to match when it will go live (Jan 1) (cc @lvreynoso ). I'm not so sure about the "last updated", I agree that its more of a legal question — I'd check in with Olivia/Christine about that since they were a part of those conversations with legal.

src/frontend/src/common/redux/utils/groupUtils.ts
lvreynoso: This is a personal patch I've been using to deal with the frontend app erroring out because the userinfo request failed for whatever reason, which happens to me all the time in localdev. I didn't actually mean to add it to this PR, I'll take it out

PR #1510, opened 2022-12-14
qc status filter, qc_status tags

phoenixAja:
### Summary:
- **What:** add the qc status filter, add chips
- **Env:** `<rdev link>`

### Demos:
Screenshot 2022-12-14 at 8 33 07 AM
Screenshot 2022-12-14 at 8 32 48 AM

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/src/frontend/src/components/FilterPanel/components/QCStatusFilter/index.tsx
+++ b/src/frontend/src/components/FilterPanel/components/QCStatusFilter/index.tsx
@@ -1,4 +1,3 @@
-import { CatchingPokemonSharp } from "@mui/icons-material";
 import { DefaultMenuSelectOption } from "czifui";
 import { useEffect, useState } from "react";
 import { StyledComplexFilter } from "../../style";
@@ -34,7 +33,7 @@ const QCStatusFilter = (props: Props): JSX.Element => {
 
   useEffect(() => {
     if (value) {
-        updateQCStatusFilter(
+      updateQCStatusFilter(
         (value as DefaultMenuSelectOption[]).map((d) => d.name)
       );
     }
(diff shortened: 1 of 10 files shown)


src/frontend/src/components/FilterPanel/components/QCStatusFilter/index.tsx
```diff
@@ -0,0 +1,71 @@
+import { DefaultMenuSelectOption } from "czifui";
```
phoenixAja, line 1: this was mostly cribbed from the lineage filter

src/frontend/src/components/FilterPanel/components/QCStatusFilter/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+export type ComplexFilterValue =
+  | DefaultMenuSelectOption[]
+  | DefaultMenuSelectOption
+  | null;
+
+// ComplexFilter defaults to checking if option is selected (is a value) by equality.
+// However, because our options are objects that are dynamically generated, object
+// equality is always false because they're not the same object, even if same content.
+// Instead, we must create a custom selection checker to check underlying value.
```
ehoops-cz, line 18: I don't think this is needed anymore since an SDS update.

src/frontend/src/views/Data/cellRenderers.tsx
```diff
@@ -137,8 +157,8 @@ const SAMPLE_CUSTOM_RENDERERS: Record<string | number, CellRenderer> = {
               <StyledChip
                 data-test-id="row-sample-status"
                 size="small"
-                label={label.label}
-                status={label.status}
+                label={label()?.label}
+                status={label()?.status}
```
ehoops-cz: Can this ever be undefined/null?  It looks like `label()` would just return the default case if qcMetrics is null.  If typescript is mad, adding a return type to the function may help.

ehoops-cz (approved):

src/frontend/src/components/FilterPanel/components/QCStatusFilter/index.tsx
ehoops-cz: (not that high of priority since I'm pretty sure SDS just implemented this exact thing, there aren't any performance improvements etc.)

phoenixAja: ok cool, good to know!

PR #1513, opened 2022-12-15
Add in QC/Mutations download BE endpoint

vincent-czi:
### Summary:
- **What:** Adds in BE endpoint for downloading TSV of QC/Mutations data for specified samples
- **Ticket:** [sc<220817>](https://app.shortcut.com/genepi/story/<220817>)
- **Env:** None (was going to put one up, but then realized I wouldn't be able to hit it because rdev) -- planning to test on Staging. Changes should be purely additive and pretty much invisible to most users, so even if it's totally broken, no real harm.

### Demos:
Resulting TSV opened in Excel

image

---

Running Postman locally

image

### Notes:
This version of QC/Mutations download is very tied to our current approach where Nextclade is the only game in town and we only ever have a single `sample_qc_metrics` row for any given sample. That should eventually change when we start supporting other tools, but I think it's acceptable for now: a lot of our other stuff is very strongly making that assumption as well.

There are no tests currently. That's probably a bad thing! But I'd like to merge this in so I can verify it's working on Staging and we can ship the download feature that much sooner. I definitely want to write an automated test for it though, but not quite sure the best way to approach: I need to look over other download endpoint tests, maybe discuss with someone else who has written one of those tests. 

I also have not updated the aspencli.py to be able to download from the new endpoint. I want to do that as well, but would like to verify this is working in Staging first (just going to use Postman to start). I also need to update the cli tool to handle pathogens in the URL since that's starting to matter. But I think that's similar to the above where we could swing back in a week after the initial download version in FE is working.

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR
~[ ] I added relevant unit tests~ (would be good in near future though, see above)~


jgadling:
Re: using postman - imop it's *easier* to just add the command to the aspen cli than it is to fight with getting auth tacked into postman. The CLI is your friend!! :)

Afaik this should work just fine in local dev - is it working locally?

And writing tests is pretty straightforward - there's already an example of working with raw data endpoints in `aspen/api/views/tests/test_submission_templates.py`

vincent-czi:
> Afaik this should work just fine in local dev - is it working locally?

oh yeah! I was running Postman against my local and everything works as expected. I've got demo images above of it operating locally.

phoenixAja:
it looks good, but it would be nice to have a couple tests 🙏

phoenixAja (approved):

vincent-czi:
I'm going to merge this in now because I want to make sure it's available for having the FE download the file to avoid blocking things while I'm out the next couple days. But I'll have my next bit of work be adding some tests and CLI for it.

PR #1524, opened 2022-12-21
CLI: Download for QC/Mutations output

vincent-czi:
### Summary:
- **What:** Changes to CLI tool: adds support for pathogens in URLs and puts in command to download QC/Mutations
- **Ticket:** [sc<228746>](https://app.shortcut.com/genepi/story/<228746>)
- **Env:** None

### Demos:

image

### Notes:

Previously, it was easy to miss needing to add the org_id (via `--org`) when using commands that hit endpoints which are org-dependent. When that happened, you'd get 404-type responses but no explanation of why. I added a check that an org is present and it will now raise an exception with an error message if it's missing that tells you what you need to do.

I also added handling for pathogens in URLs so you can now also pass the `--pathogen-slug`. If you don't, it uses the other, org-only form of the URL and we let the BE do defaulting of the pathogen (to SC2).

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/cli/aspencli.py
+++ b/src/cli/aspencli.py
@@ -133,9 +133,10 @@ class TokenHandler:
 
 
 class ApiClient:
-    def __init__(self, url, token_handler, org_id):
+    def __init__(self, url, token_handler, org_id, pathogen_slug):
         self.url = url
         self.org_id = org_id
+        self.pathogen_slug = pathogen_slug
         self.token_handler = token_handler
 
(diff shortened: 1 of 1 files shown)


jgadling (approved):

jgadling:
Just a headsup that the backend shouldn't be doing default pathogen handling for much longer. We should remove those old routes.

PR #1527, opened 2022-12-24
Handle lineage `last_updated` for non-covid samples

vincent-czi:
### Summary:
- **What:** Fix for "Last Updated" missing from MPOX lineage tooltips
- **Ticket:** [sc<228768>](https://app.shortcut.com/genepi/story/<228768>)
- **Env:** https://vince-mpox-last-updated-frontend.dev.czgenepi.org

### Demos:

image

^^^ "Last Updated" is now filled in instead of blank

### Notes:

The core issue was that Pangolin/SC2 has a `last_updated` value/field that we fill in when we run Pangolin on a sample and create a `sample_lineages` row, but that Nextclade does not have quite the same thing. We rely on `last_updated` existing for the lineage tooltip in the FE, so when Nextclade provides the lineage, we did not have a date available.

Nextclade does provide a `tag` attribute for every dataset (what we call `reference_dataset_tag`, and these tags are a timestamp (eg, "2021-06-25T00:00:00Z"), so we can effectively read the `last_updated` value for a lineage from the tag. However, while everything I've seen so far in Nextclade has these as always being timestamps in that format, I have not seen any documentation _guaranteeing_ this structure. All they guarantee is that the tag is unique (within a pathogen type, I believe). Nonetheless, I think it's reasonable to assume we will continue to get timestamps, read out the dates accordingly, and if it ever turns out we get something different, fire off a Sentry alert so we go figure out how to deal with it then.

I want to mention that this is not the only way we could handle this: we could also interpret the lineage's `reference_dataset_tag` value in the FE instead. I don't think that's crazy, but I think it's generally better to abstract needing to know those specifics away from the FE so it just relies on a fairly standardized lineage coming along. Another option would be to just populate the `last_updated` value for every single lineage row, but that would basically be the same as duplicating the already existing `reference_dataset_tag` right now anyway, so this seemed like the better approach to me.

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR
[x] I modified relevant unit tests


src/backend/aspen/api/utils/sample.py
```diff
@@ -243,12 +244,48 @@ def format_sample_lineage(sample: Sample) -> List[Dict[str, Any]]:
...
+    Nextclade attaches a `tag` to every dataset to uniquely identify it. This
+    tag is consistently just a datetime (eg, "2021-06-25T00:00:00Z"), so we
+    can extract the date from it. While Nextclade seems to strongly adhere to
+    this convention for tags, there is no documentation that future tags /must/
+    keep this structure. If they ever change it we'll need to alter how we
+    handle capturing the date associated with a given dataset.
+    """
+    date = None  # default fall-through just in case tag is missing
+    if reference_dataset_tag:
+        m = re.match(r"^\d{4}-\d+-\d+(?=T)", reference_dataset_tag)
```
jgadling, line 273: personally I'd use [arrow](https://arrow.readthedocs.io/en/latest/) here instead of a regex, but this should get the job done if nextclade's data is pretty clean

jgadling (approved):

src/backend/aspen/api/utils/sample.py
vincent-czi: I thought about that, but I'm currently kinda thinking of this as a canary in a coal mine: if anything about how they structure their `tag` values changes, I want to find out quickly, because it might indicate some other stuff having changed that we need to look into.

phoenixAja (approved):
thanks for fixing this

PR #1530, opened 2023-01-04
Add space between Learn More link and tooltip text.

ehoops-cz:
### Summary:
- **What:** Add space between Learn More link and tooltip text. 🤦‍♀️ forgot to commit this with yesterday's pr.
- **Ticket:** none
- **Env:** `<rdev link>`

### Demos:
Screenshot 2023-01-03 at 4 02 17 PM

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/views/Data/table-headers/commonSampleHeaders.tsx
+++ b/src/frontend/src/views/Data/table-headers/commonSampleHeaders.tsx
@@ -117,7 +117,7 @@ export const QC_METRICS_HEADER: Header = {
       linkText: "Learn more.",
     },
     regularText:
-      "Overall QC score from Nextclade which considers genome completion and screens for potential contamination and sequencing or bioinformatics errors.",
+      "Overall QC score from Nextclade which considers genome completion and screens for potential contamination and sequencing or bioinformatics errors. ",
   },
 };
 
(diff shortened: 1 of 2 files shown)


src/frontend/src/views/Data/table-headers/commonSampleHeaders.tsx
```diff
@@ -117,7 +117,7 @@ export const QC_METRICS_HEADER: Header = {
       linkText: "Learn more.",
     },
     regularText:
-      "Overall QC score from Nextclade which considers genome completion and screens for potential contamination and sequencing or bioinformatics errors.",
+      "Overall QC score from Nextclade which considers genome completion and screens for potential contamination and sequencing or bioinformatics errors. ",
```
vincent-czi, line 120: I feel like this kind of forgetting-to-drop-a-terminating-space-char for the various strings is going to wind up being common. With the other change here, I think I would probably save the file and verify on my local browser, so I'd notice missing that weird terminating JSX space. But for creating/updating a set of strings, I think I'd be a lot more likely to not see the issue, just save my work, and make a PR. I wonder if it would be less accident-prone / clearer to have wherever the interpolation happens handle the extra space instead?

I'm **not** actually suggesting this change for this PR. It just popped into my head as a mistake we might regularly make as we expand to having more sets of strings for various pathogens, so something I wanted to call out. If we notice ourselves regularly bumping our knees on this coffee table, we should come back and see if we can tweak the interpolation to make the string look more like what you'd expect at a glance.

vincent-czi (approved):

src/frontend/src/views/Data/table-headers/commonSampleHeaders.tsx
ehoops-cz: Ya, I agree with this. It's the placement of the following link that actually causes weird spacing anyway.

PR #1533, opened 2023-01-04
nextclade analytics event

phoenixAja:
### Summary:
- **What:** add analytics event to capture that nextclade template was downloaded
- **Ticket:** [sc227789](https://app.shortcut.com/genepi/story/227789)
- **Env:** `<rdev link>`

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/docs/DEV_ENV.md
+++ b/docs/DEV_ENV.md
@@ -148,3 +148,7 @@ Notes:
 Drop out of remote container mode:
 * Run the `Remote-Containers: Reopen Folder Locally` command
 
+
+### Capturing Analytics Events:
+
+* [Confluence document on some basic notes on where to get started](https://czi.atlassian.net/wiki/spaces/SCI/pages/2637725697/Analytics+Notes)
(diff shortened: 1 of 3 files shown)


phoenixAja:
@vincent-czi do we need to do anything on the segment side to account for the change in data that's being sent? (Since i modified the `AnalyticsSamplesDownloadFile` type to include `includes_nextclade_data`)

vincent-czi:
@phoenixAja The way Segment works and data goes from Segment --> Snowflake (where the data lands ultimately), as long as it's an _additive_ change where we're just adding new stuff, everything is fine and happy. So we're good here. That said, if you were removing an attribute or changing the underlying data type of an attribute (you used to be an int, but now you're a string), things get tougher.

Adding new stuff == good
Removing stuff == probably good, but be careful
Modifying stuff that already exists == uh oh

vincent-czi (approved):
Looks good! Thanks for making the doc in Confluence!

PR #1534, opened 2023-01-04
add analytics event to capture user filtering

phoenixAja:
### Summary:
- **What:** capture filtering events for analytics
- **Ticket:** [sc227789](https://app.shortcut.com/genepi/story/227789)
- **Env:** `<rdev link>`

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/analytics/eventTypes.ts
+++ b/src/frontend/src/common/analytics/eventTypes.ts
@@ -62,6 +62,9 @@ export enum EVENT_TYPES {
 
   // User is in multiple groups and is changing which group they are acting in.
   ACTIVE_GROUP_CHANGE = "ACTIVE_GROUP_CHANGE",
+
+  // User has filtered data on the Samples page
+  SAMPLES_FILTER = "SAMPLES_FILTER",
 }
 
 /**
@@ -82,12 +85,18 @@ export enum EVENT_TYPES {
  * that kicks off the underlying Segment `track` by ensuring that all events
  * extend from that base structure.
  */
-type EventValue = string | number | boolean | null;
+type EventValue = string | number | boolean | FilterDate | null;
 export type EventData = Record<string, EventValue | undefined>;
 
 // While we only send values of EventValue, sometimes we send a JSON string.
 // For ease of readability below, we alias string for those cases.
 type JsonString = string;
+
+// For sending date types back to segment (e.g. for filtering on upload or collection dates)
+type FilterDate = {
+  start?: Date;
+  end?: Date;
+};
 
 /**
  * Structure of additionalEventData for each EVENT_TYPES type that sends it.
@@ -258,4 +267,24 @@ export type AnalyticsActiveGroupChange = {
   previous_group_id: number;
   // The group ID that the user is switching to viewing/acting as.
   new_group_id: number;
+};
+
+/** EVENT_TYPES.SAMPLES_FILTER */
+export type AnalyticsSamplesFilter = {
+  // User is filtering by uploadDate
+  filtering_by_upload_date: boolean;
+  // User is filtering by collectionDate
+  filtering_by_collection_date: boolean;
+  // User is filtering by lineage
+  filtering_by_lineage: boolean;
+  // User is filtering by qcStatus
+  filtering_by_qc_status: boolean;
+  // Upload dates that user is filtering on
+  upload_date_start_end: FilterDate;
+  // Collection dates that user is filtering on
+  collection_date_start_end: FilterDate;
+  // Lineages that user is filtering on
+  lineages: JsonString;
+  // qcStatus that user is filtering on
+  qc_statuses: JsonString;
 };
--- a/src/frontend/src/common/analytics/eventTypes.ts
+++ b/src/frontend/src/common/analytics/eventTypes.ts
@@ -85,18 +85,12 @@ export enum EVENT_TYPES {
  * that kicks off the underlying Segment `track` by ensuring that all events
  * extend from that base structure.
  */
-type EventValue = string | number | boolean | FilterDate | null;
+type EventValue = string | number | boolean | null;
 export type EventData = Record<string, EventValue | undefined>;
 
 // While we only send values of EventValue, sometimes we send a JSON string.
 // For ease of readability below, we alias string for those cases.
 type JsonString = string;
-
-// For sending date types back to segment (e.g. for filtering on upload or collection dates)
-type FilterDate = {
-  start?: Date;
-  end?: Date;
-};
 
 /**
  * Structure of additionalEventData for each EVENT_TYPES type that sends it.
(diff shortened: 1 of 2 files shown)


src/frontend/src/common/analytics/eventTypes.ts
```diff
@@ -82,13 +85,19 @@ export enum EVENT_TYPES {
  * that kicks off the underlying Segment `track` by ensuring that all events
  * extend from that base structure.
  */
-type EventValue = string | number | boolean | null;
+type EventValue = string | number | boolean | FilterDate | null;
```
vincent-czi: This should not be expanded. Because of how the data lands from Segment -> Snowflake, these are really the only kind of values that should ever be allowed. Adding a new kind of type here causes a lot of downstream pain, so it really should ONLY be string | number | boolean | null

src/frontend/src/common/analytics/eventTypes.ts
```diff
@@ -82,13 +85,19 @@ export enum EVENT_TYPES {
...
 export type EventData = Record<string, EventValue | undefined>;
 
 // While we only send values of EventValue, sometimes we send a JSON string.
 // For ease of readability below, we alias string for those cases.
 type JsonString = string;
 
+// For sending date types back to segment (e.g. for filtering on upload or collection dates)
```
vincent-czi: Following up on what I said above with EventValue, we almost always want to avoid having objects be what we send as event data. Since this ultimately lands as data in DB columns for our Data Analyst teammate, we want to try to avoid sending anything other than atomic `EventValue` data. Simple, flat objects with atomic values for everything is what we want to emit. There are rarely some cases where we break this rule, and in that case we fall back to JSONifying objects before smushing them into event data payloads. For some examples of this, look for references elsewhere in the codebase to the above `JsonString` type.

src/frontend/src/common/analytics/eventTypes.ts
```diff
@@ -259,3 +268,23 @@ export type AnalyticsActiveGroupChange = {
...
+  // User is filtering by uploadDate
+  filtering_by_upload_date: boolean;
+  // User is filtering by collectionDate
+  filtering_by_collection_date: boolean;
+  // User is filtering by lineage
+  filtering_by_lineage: boolean;
+  // User is filtering by qcStatus
+  filtering_by_qc_status: boolean;
+  // Upload dates that user is filtering on
+  upload_date_start_end: FilterDate;
```
vincent-czi: Continuing the discussion about `FilterDate` here, I think the best thing is to break out each of the filter chunks into its pieces, that way the payload stays flatter. So we'd have
```
upload_date_start: string;
upload_date_end: string;
...etc...
```

vincent-czi: The one thing I'm not sure about is _maybe_ expanding the `EventValue` type to include a plain `Date` type. There's a lot of auto-magic stuff going on in how Segment lands data over in Snowflake. Segment is basically auto-generating migrations as it stuffs data into Snowflake. A new event type or modifying an existing event type means it also creates a corresponding migration so the data can land. But it really abstracts a lot of that away from you as the developer. I _think_ a Date would just land as a string (or maybe the auto-magic recognizes ISO8601 strings?), but Segment is ultimately sending JSON payloads itself to their own API, so my guess is that a Date type would not actually help, and gives us a false sense of how things will work.

A lot of how I structured stuff with the `EventValue` and the payloads was intended to avoid running into problems with the auto-magic. I was trying to de-magic our interaction so we'd be less likely to make a reasonable-but-wrong assumption for how data would ultimately land in the analytics DB.

phoenixAja:
thanks for the comments @vincent-czi

src/frontend/src/common/analytics/eventTypes.ts
phoenixAja: makes sense, i can adjust that

src/frontend/src/common/analytics/eventTypes.ts
phoenixAja: ok removed

src/frontend/src/common/analytics/eventTypes.ts
```diff
@@ -259,3 +262,25 @@ export type AnalyticsActiveGroupChange = {
...
+  // User is filtering by qcStatus
+  filtering_by_qc_status: boolean;
+  // Upload dates that user is filtering on
+  upload_date_start: FormattedDateType | null;
+  upload_date_end: FormattedDateType | null;
+  // Collection dates that user is filtering on
+  collection_date_start: FormattedDateType | null;
+  collection_date_end: FormattedDateType | null;
+  // Lineages that user is filtering on
+  lineages: JsonString;
```
vincent-czi, line 283: For things that `JsonString` type, please also add a comment to mention what the structure of the JSON is, just to make things easier for the Data Analyst looking at the events downstream. The intention of the `eventTypes.ts` file is meant to be part schema for code, part documentation for analysts.

For example, elsewhere in the file I have this:
```
  // JSON array of all the IDs for newly created samples for this upload
  sample_ids: JsonString;
```

src/frontend/src/common/analytics/eventTypes.ts
```diff
@@ -259,3 +262,25 @@ export type AnalyticsActiveGroupChange = {
...
+  // qcStatus that user is filtering on
+  qc_statuses: JsonString;
```
vincent-czi, line 285: same additional comment bit here please

src/frontend/src/components/FilterPanel/index.tsx
```diff
@@ -148,6 +154,43 @@ const FilterPanel: FC<Props> = ({
...
+    const uploadDate = activeFilters.find((e) => e.key === "uploadDate");
+    const collectionDate = activeFilters.find(
+      (e) => e.key === "collectionDate"
+    );
+    const qcStatus = activeFilters.find((e) => e.key === "qcMetrics");
+    const lineage = activeFilters.find((e) => e.key === "lineage");
+    const qcStatuses = qcStatus?.params.multiSelected || [];
+    const lineages = lineage?.params.multiSelected || [];
+
+    const anyFiltersActive: boolean[] = [
```
vincent-czi, line 167: Nit: this eventually collapses into a single boolean with `anyFiltersActive.includes(true)` and that's the only intended use -- right? -- so I think doing something like
```
(!!uploadDate || !!collectionDate || !!qcStatus | !!lineage)
```
gets across the intent slightly better.

vincent-czi (approved):
Overall looks good! I've got two minor comments about documentation I'd like addressed before merging, but they're really small. The other comment is whatever. Thank you!

src/frontend/src/components/FilterPanel/index.tsx
phoenixAja: I like to created intermediate variables sometimes to make the logic more clear. I can take it out though if that is not really coming across.

PR #1537, opened 2023-01-07
qc filter/ nextclade download cleanup

phoenixAja:
### Summary:
- **What:** remove feature flag, update warnings to include failed qc samples, design feedback 
- **Ticket:** [sc228887](https://app.shortcut.com/genepi/story/228887)
- **Ticket:** [sc228884](https://app.shortcut.com/genepi/story/228884)
- **Ticket:** [sc228823](https://app.shortcut.com/genepi/story/228823)
- **Env:** `<rdev link>`

### Notes:
~~PR #1539 has to be merged in first (where we delete the samples that have failed genome recovery)~~

### Checklist:
- [ ] I added labels to my PR


src/frontend/e2e/utils/schemas/sampleResponseDefaults.ts
```diff
@@ -1,11 +1,11 @@
 export type SampleResponseDefaults = {
   collection_date?: string;
   collection_location?: number;
-  czb_failed_genome_recovery?: boolean;
   gisaid_id?: string | null;
   gisaid_status?: string;
   id?: number;
-  lineage?: string;
+  lineages?: [Lineage];
```
mmmmmmaya, line 7: The way these types are written indicate that `lineages` is a one-item tuple, where the only item is a `Lineage`. Is that what you meant? If you meant a Lineage array, this should be changed to `Lineage[]`

src/frontend/src/common/components/library/data_table/index.tsx
```diff
@@ -204,6 +193,12 @@ export const DataTable: FunctionComponent<Props> = ({
     }
   }, [data, checkedSampleIds, setHeaderIndeterminant, setIsHeaderChecked]);
 
+  useEffect(() => {
```
mmmmmmaya: can you merge this in with the useEffect above where all the other logic is for whether or not the header is checked?

src/frontend/src/views/Data/cellRenderers.tsx
```diff
@@ -154,13 +126,8 @@ const PrivateId = ({
...
 
-  const label = usesNextcladeDownload
-    ? qcStatusLabel.label
-    : labelCZBFailedGenomeRecovery.label;
-
-  const status = usesNextcladeDownload
-    ? qcStatusLabel?.status
-    : labelCZBFailedGenomeRecovery.status;
+  const label = qcStatusLabel.label;
+  const status = qcStatusLabel.status;
```
mmmmmmaya:
```suggestion
  const { label, status } = qcStatusLabel;
```

src/frontend/src/common/components/library/data_table/index.tsx
phoenixAja: sure!

src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/index.tsx
```diff
@@ -34,16 +35,16 @@ const SampleTableModalManager = ({
...
     const checkedIds = checkedSamples.map((s) => s.publicId);
-    const failedIds = checkedSamples
-      .filter((s) => s.CZBFailedGenomeRecovery)
-      .map((s) => s.publicId);
-    const badQCIds = checkedSamples
+    const badOrFailedQCIds = checkedSamples
       // for now there should only ever be one qcMetrics entry per sample
-      .filter((s) => s.qcMetrics[0].qc_status === "Bad")
+      .filter(
+        (s) =>
```
mmmmmmaya: I noticed you used this pattern 3 times in this PR. I think it is worth making a util function that accepts an item, and returns whether its qc status is bad or failed. What do you think?

mmmmmmaya:
imo, for changes of this size, it would be nice to have one PR per ticket to make it easier to review.

src/frontend/e2e/utils/schemas/sampleResponseDefaults.ts
phoenixAja: yeah because as of now there should only be one Lineage entry per sample. We were future proofing the backend by having the capability to have a list of associated lineages or qc_metrics in case we wanted to implement that in the future, but for right now there should only be one Lineage or qcMetric per sample

phoenixAja:
> imo, for changes of this size, it would be nice to have one PR per ticket to make it easier to review.

I totally agree, this one got a little out of hand because the changes sort of went together (removing the nextclade integration flag went in hand with removing the czb_genome_recovery filter, then i figured i would strip out the genome recovery stuff in general which lead to updating the alerts) so i figured i would group them together, usually i break things up as much as i can, this one was an exception. Usually i do try to make smaller PRs

src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/index.tsx
phoenixAja: i think it's small enough to go either way, i can make a function

mmmmmmaya (approved):

PR #1541, opened 2023-01-10
remove failed genome recovery field dependency

phoenixAja:
### Summary:
- **What:** remove the backend dependency on `czb_failed_genome_recovery` field before removing from the DB
- **Ticket:** **[sc217890](https://app.shortcut.com/genepi/story/217890)**
- **Env:** `<rdev link>`

### Notes:
#1537 needs to be merged in before this

### Checklist:
- [ ] I added labels to my PR


src/backend/scripts/setup_localdata.py
```diff
@@ -510,7 +503,7 @@ def create_samples(
     for suffix in range(num_successful):
         _ = create_sample(session, group, user, pathogen, location, suffix)
     for suffix in range(num_failures):
-        _ = create_sample(session, group, user, pathogen, location, suffix, True)
+        _ = create_sample(session, group, user, pathogen, location, suffix)
```
lvreynoso: We can probably eliminate this loop, right? And eliminate the `num_failures` parameter in the `create_samples()` function it lives in. We would just have to modify line 565 slightly.

lvreynoso (approved):
one suggestion, otherwise lgtm

src/backend/database_migrations/versions/20230110_175025_remove_czb_failed_genome_recovery_field.py
```diff
@@ -0,0 +1,20 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "20230110_175025"
+down_revision = "20230109_223207"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.drop_column("samples", "czb_failed_genome_recovery", schema="aspen")
```
jgadling: Can we move this migration to a separate PR and deploy it after we've stopped depending on this field?

src/backend/scripts/setup_localdata.py
phoenixAja: yes thank you

phoenixAja: i accidentally missed that

src/backend/database_migrations/versions/20230110_175025_remove_czb_failed_genome_recovery_field.py
phoenixAja: sure

jgadling (approved):

PR #1544, opened 2023-01-11
[SC-228214] Create tree modal - radio button copy updates.

ehoops-cz:
### Summary:
- **What:** `Create tree modal - radio button copy updates.`
- **Ticket:** [sc228214](https://app.shortcut.com/genepi/story/228214)
- **Env:** `none`

### Demos:
Screenshot 2023-01-10 at 3 55 31 PM
Screenshot 2023-01-10 at 3 55 26 PM
Screenshot 2023-01-10 at 3 55 22 PM

### Notes:
I created a constant and reused it rather than removing the strings file entirely.  This seems ok, but open to feedback.

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/src/frontend/src/components/Split/pathogenSplitConfig.ts
+++ b/src/frontend/src/components/Split/pathogenSplitConfig.ts
@@ -27,7 +27,7 @@ const pathogenTrafficFlagConfig: PathogenTrafficFlagConfig = {
   },
   [PATHOGEN_FEATURE_FLAGS.nextstrain_enabled]: {
     [Pathogen.COVID]: SPLIT_SIMPLE_FLAG.ON,
-    [Pathogen.MONKEY_POX]: SPLIT_SIMPLE_FLAG.OFF,
+    [Pathogen.MONKEY_POX]: SPLIT_SIMPLE_FLAG.ON,
   },
   [PATHOGEN_FEATURE_FLAGS.public_repository]: {
     [Pathogen.COVID]: SPLIT_SIMPLE_FLAG.ON,
--- a/src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/components/CreateNSTreeModal/components/RadioLabel/strings.ts
+++ b/src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/components/CreateNSTreeModal/components/RadioLabel/strings.ts
@@ -15,50 +15,29 @@ interface RadioLabelStrings {
   targetedGoodFor: string;
 }
 
+const generalViralRadioLabelStrings: RadioLabelStrings = {
+  nonContextualizedBestFor:
+    "Best for uncovering sampling bias in your own sampling effort.",
+  nonContextualizedDescription: "Builds tree with only samples of interest.",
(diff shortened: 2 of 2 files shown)


src/frontend/src/components/Split/pathogenSplitConfig.ts
```diff
@@ -27,7 +27,7 @@ const pathogenTrafficFlagConfig: PathogenTrafficFlagConfig = {
   },
   [PATHOGEN_FEATURE_FLAGS.nextstrain_enabled]: {
     [Pathogen.COVID]: SPLIT_SIMPLE_FLAG.ON,
-    [Pathogen.MONKEY_POX]: SPLIT_SIMPLE_FLAG.OFF,
+    [Pathogen.MONKEY_POX]: SPLIT_SIMPLE_FLAG.ON,
```
ehoops-cz, line 30: These are just for creating the local flags for development

Janeece (approved):
LGMT! thanks Erin!

phoenixAja (approved):

src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/components/CreateNSTreeModal/components/RadioLabel/strings.ts
```diff
@@ -15,50 +15,29 @@ interface RadioLabelStrings {
...
-    overviewGoodFor1: "Good for identifying possible local outbreaks.",
-    overviewGoodFor2:
-      "Good for specifying sample location, collection date, or lineages instead of the defaults used for your CZ GEN EPI weekly build.",
-    targetedBestFor: "Best for investigating an identified outbreak.",
-    targetedDescription:
-      "Builds tree with selected samples (from the sample table or the ID box below) and closely-related samples, at a ratio of roughly 1:2.",
-    targetedGoodFor:
-      "Good for identifying samples most closely related to the selected samples among all samples in GISAID and your CZ GEN EPI samples.",
-  },
+  [Pathogen.COVID]: generalViralRadioLabelStrings,
```
mmmmmmaya, line 41: I'm wondering if you think it's worth having a `default` case or something for `PathogenConfigType`, so that a pathogen's strings only need to be specified when it differs from the default?

I'm on the fence about the idea, though. May make more sense to wait until we have 3+ pathogens because then default will actually mean something other than "are they both the same?".

mmmmmmaya (approved):

src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/components/CreateNSTreeModal/components/RadioLabel/strings.ts
ehoops-cz: Definitely worth considering.  For a few different cases I created a similar generalViral string object and used that for mpox (and was planning to use for hepC as well). Either this or a default would potentially work with a spread operator and then just changing the specific strings that need it.  

I agree that we should revisit once we have 3+ pathogens.

PR #1546, opened 2023-01-11
install nginx extras

phoenixAja:
### Summary:
- **What:** install nginx extras to get the fair load balancer back
- **Env:** `<rdev link>`

### Checklist:
- [x] I added labels to my PR


src/backend/aspen/tests/test_backend_serves_requests.py
```diff
@@ -0,0 +1,9 @@
+import requests
+
+
+def test_backend_serves_requests():
+
+    backend_request = requests.get(
+        "http://backend.genepinet.localdev:3000/v2/openapi.json"
```
vincent-czi: I think it would be better to use the health check endpoint rather than the openapi stuff since the health check is more closely part of our app's working structure and the openapi thing is more something provided by FastAPI.

`/v2/health`, see `health.py` in views

vincent-czi (approved):
Overall looks good. Minor comment about improving a test. Is this something we should have an rdev for or is the way nginx is breaking only impact localdev?

src/backend/aspen/tests/test_backend_serves_requests.py
phoenixAja: yeah using the health endpoint seems better, also i think we're ok with just checking localdev, because if localdev is working that means that the backend container is able to serve requests (which should be the same case for staging, prod, etc). @jgadling what do you think?

jgadling: por qué no los dos?

src/backend/Dockerfile
```diff
@@ -31,12 +31,12 @@ RUN apt-get update && apt-get install -y vim procps wget make jq nano
 # Orchestration. Supervisor is heavyweight but afaik nothing else supports sane child reaping and signal forwarding.
 RUN pip install --no-cache-dir supervisor
 
-# Proxy/Router: Set up nginx
-RUN apt install -y nginx
+# Proxy/Router: Set up nginx (also install extras to get ngx_http_upstream_fair_module)
+RUN apt install -y nginx nginx-extras
```
phoenixAja, line 35: nginx-extras also installs nginx by default

PR #1547, opened 2023-01-11
Script to ingest Mpox lineages

jgadling:
### Summary:
- **What:** Download lineage json data from our source of truth and write them to the db.
- **Ticket:** [sc226110](https://app.shortcut.com/genepi/story/226110)
- **Env:** `<rdev link>`

### Notes:
This script isn't currently hooked up to any existing scheduled jobs. We can run it manually if necessary, but the idea is that we'll add it in to the NCBI Virus download workflow when that's ready to go later this sprint.

This script is intentionally a bit overkill since I'm hoping we can expand it to support downloading lineages for more pathogens down the road.

### Checklist:
- [x] I added labels to my PR


src/backend/aspen/util/pathogen_configs.py
```diff
@@ -0,0 +1,15 @@
+from typing import Any, Dict
+
+
+def get_lineage_urls(pathogen_slug: str) -> Dict[str, Any]:
```
vincent-czi, line 4: Since the plan is for this to eventually build up to supporting many pathogen-lineages down the road, I'd like to see a some documentation on what the intent here with the structure is. There's some assumption about how the lineages data will be structured, and while I can figure it out by cross-referencing `format_lineage_data` where the path/keys stuff gets used, I'd like to see it written out explicitly. As much to make it easier to crank them out when we add more pathogen-lineages as to help see where to extend the structure when we encounter a pathogen-lineages info structure that's totally foreign to this one.

That said, I think having that documentation over in `format_lineage_data` makes an equal amount of sense. I just want it somewhere and a comment on the other one saying "yo, go look at the docs there to understand the usage structure".

src/backend/aspen/workflows/import_lineages/load_lineages.py
```diff
@@ -0,0 +1,168 @@
...
+    drop_temp_table,
+    mv_table_contents,
+)
+
+
+def download_lineages(
+    url: str,
+    format: str,
+    print_response: bool,
+) -> Any[str, List[str]]:
```
vincent-czi: What does this mean? I've never seen this construct before with `Any`, I didn't think it could take ... args? types? type-args? I have no idea what to generically call the X in `List[X]`, but that thing. I thought `Any` stood alone? I could totally be wrong and learning something new here though!

src/backend/aspen/workflows/import_lineages/load_lineages.py
```diff
@@ -0,0 +1,168 @@
...
+            for lineage in lineages
+        ]
+        session.execute(temp_table.insert(), lineage_objects)
+
+        # Replace previous data with new data from temp_table
+        mv_table_contents(
+            session,
+            temp_table,
+            dest_table,
+            [(PathogenLineage.pathogen_id == pathogen.id)],
```
vincent-czi, line 112: Unrelated to this work, but do we have an index on the `pathogens_lineages.pathogen_id` column yet? We should remember to slap one on eventually if not.

src/backend/aspen/workflows/import_lineages/load_lineages.py
```diff
@@ -0,0 +1,168 @@
...
+@click.command()
+@click.argument(
+    "pathogen_slug",
+    required=True,
+)
+@click.option(
+    "--print-source-file",
+    type=bool,
+    is_flag=True,
+    help="Print the list of lineages we got from the remote url",
```
vincent-czi: Nit: This prints the raw response, not necessarily the parsed out list of lineages, right? Can we tweak this to call that out: "Print raw text of lineage data we got from..." or something?

vincent-czi (approved):
Looks good! I've got some minor comments in there, but definitely nothing serious.

src/backend/aspen/workflows/import_lineages/load_lineages.py
lvreynoso: I think the type hint here is just `Any`. That's the hint you wrote for `format_lineage_data()`, which is the consumer of this function's output, and I think it works fine. Especially since JSON objects are a nightmare to type in Python.

If you are certain that the two expected types are `str` and `List[str]` you can write `Union[str, List[str]]` instead. In that case you'll have to also use is as the type hint for the `source_data` parameter.

The import test worked with either type on my machine.

lvreynoso (approved):
Just a small change needed.

PR #1550, opened 2023-01-13
remove pangolin fields from workflows

phoenixAja:
### Summary:
- **What:** remove uploaded pathogen fields support from workflows
- **Env:** `<rdev link>`

### Checklist:
- [ ] I added labels to my PR


src/backend/aspen/workflows/pangolin/find_samples.py
```diff
@@ -25,6 +25,12 @@ def check_latest_pangolin_version() -> str:
     return latest_version
 
 
+def should_sample_be_updated(sample: Sample, most_recent_pango_version: str) -> bool:
```
jgadling, line 28: Just a head's up that @vincent-czi has worked on another way to do this check via SQL in his nextclade workflow. Up to you whether to pull in his method now or later.

jgadling (approved):

src/backend/aspen/workflows/pangolin/find_samples.py
vincent-czi: I think it's fine to pull it in later. I don't think this approach ever blocked anything or really seriously slowed it down, so I'd say it's better to merge sooner just to move on to the next bit of work.

PR #1553, opened 2023-01-13
Maya/228059 demo feedback

mmmmmmaya:
### Summary
- **What:** Address feedback from [table feedback doc](https://docs.google.com/document/d/1kz2TigEtzbHmkZglQ4QTozqncpfby4_EnNaFHs8OARE/edit)
- **Why:** Going to try to get more feedback and release next week!
- **Ticket:** [[sc-228059]](https://app.shortcut.com/genepi/story/228059)
- **Design:** https://www.figma.com/file/BhgBQIacuxHYRGumpLJfDx/CZ-GE-Table-Refactor?node-id=0%3A1&t=HqvE2JvT4cOqUg5e-0

#### Before: 
Screenshot 2023-01-12 at 11 23 26 PM

#### After: 
Screenshot 2023-01-12 at 11 21 26 PM

### Notes
- I'm going to upgrade the new table to use the QC tags that were implemented while I was gone in a follow up PR since it's a lot of lines (#1554)
- Doesn't include performance improvements, but addresses the majority of the design feedback.

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/components/DateFilterMenu/index.tsx
+++ b/src/frontend/src/components/DateFilterMenu/index.tsx
@@ -102,7 +102,9 @@ export const DateFilterMenu: FC<Props> = ({
     setFieldValue(fieldKeyEnd, undefined);
 
     let start = undefined;
-    if (dateOption.numDaysStartOffset) {
+    // need to ccompare to undefined because `Today` offset is 0
+    // (which evaluates to false if used to check a conditional)
+    if (dateOption.numDaysStartOffset !== undefined) {
       start = new Date();
       start.setDate(start.getDate() - dateOption.numDaysStartOffset);
       start.setHours(0, 0, 0);
--- a/src/frontend/src/common/queries/queryClient.ts
+++ b/src/frontend/src/common/queries/queryClient.ts
@@ -15,7 +15,7 @@ import { QueryClient } from "@tanstack/react-query";
  * avoid direct use: for example, if you need queryClient data in a redux
  * reducer, you could have a dispatch from a React component include the
  * queryClient data you need as well as whatever other info is required for
- * the payload. But in a few cases, you it really, truly is better to directly
+ * the payload. But in a few cases, it really, truly is better to directly
  * import and use the queryClient.
  *
  * Finally, the reason this is separated out and not just in the same file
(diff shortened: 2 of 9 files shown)


src/frontend/src/components/DateFilterMenu/index.tsx
```diff
@@ -102,7 +102,9 @@ export const DateFilterMenu: FC<Props> = ({
     setFieldValue(fieldKeyEnd, undefined);
 
     let start = undefined;
-    if (dateOption.numDaysStartOffset) {
+    // need to ccompare to undefined because `Today` offset is 0
```
vincent-czi, line 105: Very helpful comment for understanding!

vincent-czi (approved):
Looks good!

happyimadesignr:
Hey @mmmmmmaya, sorry my comment didn't come through!! Here it is:

Hey Maya! Here are some questions and comments from me and Janeece!
- Are the QC tooltips included in this?
- What does the table row hover look like?
- Do the tooltips still linger too long?
- Please decrease the space between the checkbox and the private ID
- We suggest decreasing the width of the Private ID column and truncating the name when needed. Looks like Janeece had it at 350px width.
- We wonder if we should increase the max page width to larger than our standard 1300.
- From the sceenshot, it looks like the columns are a bit close together, making the content look like it’s running together a bit and a bit hard to read, it looks like there more space and separation in the designs.

mmmmmmaya:
> Are the QC tooltips included in this?

Yes, you can see them in the rdev by hovering over the tags.

> What does the table row hover look like?

It is primary 100. It doesn't work in rdev, but it does work in staging and production.

> Do the tooltips still linger too long?

I adjusted the tooltip timers, and also prevented multiple tooltips from propagating too quickly. You can test this by quickly moving the cursor across all columns in the header row.

> Please decrease the space between the checkbox and the private ID

By how much?

> We suggest decreasing the width of the Private ID column and truncating the name when needed. Looks like Janeece had it at 350px width.

The width of the private ID column is currently set to 350px. Please let me know if you would like me to reduce it.

> We wonder if we should increase the max page width to larger than our standard 1300.

That is up to you. But fwiw, on my laptop screen there is plenty of extra space to increase the table size (16" macbook pro). 
Screenshot 2023-01-19 at 2 36 04 PM

> From the sceenshot, it looks like the columns are a bit close together, making the content look like it’s running together a bit and a bit hard to read, it looks like there more space and separation in the designs.

The columns are each 50px, which is the defined minimum. Because we have so many columns and such a narrow table, we are basically guaranteed to have all the columns be at the minimum width, and this causes everything to look squished because all the values and headers are truncated. Please let me know if you would like me to either increase the minimum column width or increase cell padding, and what the adjustment values would be.

happyimadesignr:
Thanks @mmmmmmaya, thanks for your responses!
- QC tooltip - thank you, is it supposed to have the link to help docs as in the design?
- Checkbox spacing - in the design it’s space-s on both sides of the check box. I don’t see Janeece calling it out specifically, but I measured: https://www.figma.com/file/BhgBQIacuxHYRGumpLJfDx/CZ-GE-Table-Refactor?node-id=8%3A23567&t=Zhp2Hsd7ltyUgGon-1
- Private ID column - when I inspect it, it is 460px, I couldn’t find where that’s defined. It seems to be fluid rather than the 350px.
- Page max width - I think it should increase. To around 1600px, but want to check standards and it would help to see the Private ID column and checkbox update to judge this.

mmmmmmaya:
Re: QC tooltips, the links are already implemented. Processing tags don't have a link in the design, but the others do. So if you sort the column and hover one of the other tag types you will see the Learn more link. I did notice it was not added to the header when Phoenix implemented it, so I will add it.

PR #1556, opened 2023-01-17
Stop using python 3.9

jgadling:
### Summary:
- **What:** Fix pangolin workflow

### Notes:
We used to have multiple versions of python in some docker containers since we had a system-managed python and a miniconda-managed python. We've stopped using conda entirely so we only have a single version of python3, and that version is now 3.10. Just use the system managed `python3` alias to select the right python version automatically.

### Checklist:
- [x] I added labels to my PR


vincent-czi (approved):
It pleases me to see everything going to being able to just call `python3`!

PR #1564, opened 2023-01-19
Fix the Augur CLI

jgadling:
### Summary:
- **What:** This is a quickfix (hack, frankly) to update one of Augur's dependencies to a version it's happy with. We can/should investigate this further and come up with a cleaner fix, but should get trees working again in the meantime.
- **why:** We recently updated poetry, and it requires `jsonschema` version 4+, and `augur` requires `jsonschema` 3.x. It was error'ing with the following message:

Screenshot 2023-01-18 at 4 02 16 PM

I've added a test to our CI job that ensures augur doesn't fail due to import problems.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/Dockerfile.nextstrain
+++ b/src/backend/Dockerfile.nextstrain
@@ -65,4 +65,7 @@ COPY . .
 # Install the aspen package
 RUN poetry install
 
+# TODO - temporary workaround
+RUN pip install jsonschema==3.*
+
 WORKDIR /nextstrain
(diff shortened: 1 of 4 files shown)


vincent-czi (approved):

src/backend/Dockerfile.nextstrain
```diff
@@ -65,4 +65,7 @@ COPY . .
 # Install the aspen package
 RUN poetry install
 
+# TODO - temporary workaround
```
vincent-czi, line 68: Would be nice to have an additional comment line saying why the workaround is necessary:
"Mismatch between poetry and augur deps forces us to manually install jsonschema v3 here" or similar

lvreynoso (approved):

PR #1573, opened 2023-01-23
Maya/trees perf

mmmmmmaya:
### Summary
- **What:** Apply performance updates to the trees table
- **Why:** Parity in functionality with samples table

### Demos
No visual changes.

### Notes
Depends on #1567. 

### Checklist
- [x] I merged latest `trunk`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/utils/memo.ts
+++ b/src/frontend/src/common/utils/memo.ts
@@ -1,0 +1,19 @@
+import React, { PropsWithChildren, SFC } from "react";
+
+export const memo = <P extends Record<string, unknown>>(
+  Component: SFC<P>,
+  propsAreEqual?: (
+    prevProps: Readonly<PropsWithChildren<P>>,
+    nextProps: Readonly<PropsWithChildren<P>>
+  ) => boolean
+) => {
+  const memoized = React.memo(Component, propsAreEqual);
+
+  Object.defineProperty(memoized, "displayName", {
+    set(name) {
+      Component.displayName = name;
+    },
+  });
+
+  return memoized;
+};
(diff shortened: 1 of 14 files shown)


phoenixAja (approved):

src/frontend/src/common/utils/memo.ts
```diff
@@ -0,0 +1,19 @@
...
+  const memoized = React.memo(Component, propsAreEqual);
+
+  Object.defineProperty(memoized, "displayName", {
+    set(name) {
+      Component.displayName = name;
+    },
+  });
+
+  return memoized;
+};
```
ehoops-cz, line 19: What does React.memo do that is different from this?  It seems like it isn't actually doing its job 🤯

src/frontend/src/views/Data/components/SamplesView/components/SamplesTable/index.tsx
```diff
@@ -382,6 +389,20 @@ const SamplesTable = ({
...
 
+  // adds virtualization to the table
+  const tableContainerRef = useRef<HTMLDivElement>(null);
+
+  const { rows } = table.getRowModel();
+  const rowVirtualizer = useVirtual({
+    parentRef: tableContainerRef,
+    size: rows.length,
+    // increasing the `overscan` number will enable smoother scrolling, but will also add more nodes
+    // to the DOM, and may impact performance of other things in view, such as filters and modals
```
ehoops-cz, line 400: 👍

src/frontend/src/common/utils/memo.ts
mmmmmmaya: If you try to replace instances of the new `memo` with `React.memo` you will likely see the problem in your editor. For FC that are stored in variables, they are technically anonymous functions. For this reason, React does not assign them a `displayName`, which causes the linter to fail. 

`displayName` helps with a few things in debugging. Among them, providing a name to React devtools when displaying the component tree and giving a component a human-readable name in error messages. Otherwise it would just say `anonymous`. This isn't really React's fault. It's just an unfortunate side effect of FC and HOC sometimes.

https://reactjs.org/docs/react-component.html#displayname

PR #1576, opened 2023-01-25
[SC-229039] monkeypox analytics events

ehoops-cz:
### Summary:
- **What:** `Add pathogen to existing analytics events`
- **Ticket:** [sc229039](https://app.shortcut.com/genepi/story/229039)
- **Env:** `none`

### Notes:
will test in staging 

Adding pathogen
TREE_VIEW_NEXTSTRAIN
TREE_CREATION_NEXTSTRAIN
TREE_DOWNLOAD_TREE_FILE
TREE_DOWNLOAD_SELECTED_SAMPLES_TEMPLATE
SAMPLES_UPLOAD_PAGE_CHANGE
SAMPLES_UPLOAD_SUCCESS
SAMPLES_DOWNLOAD_FILE
SAMPLES_FILTER

Not adding pathogen
TREE_ACTIONS_CLICK_GALAGO
TREE_VIEW_GALAGO
TREE_DETAILS_VIEW
ACTIVE_GROUP_CHANGE

### Checklist:
- [ ] I added labels to my PR


vincent-czi (approved):
Looks good!

vincent-czi:
This isn't really a positive or negative comment, but I just wanted to mention that for a lot of these events it is _possible_ to get the pathogen from already available information. For example, a TREE_VIEW_NEXTSTRAIN has the tree_id, which then can be cross-referenced against the trees to find the inputs, which then allows digging down to the samples, where the pathogen of the samples can be looked up (and implicitly the same pathogen should be present on all the samples or something went terribly, terribly wrong). But that's a really convoluted way to find out the pathogen, so it makes a lot of sense to just chuck it into the events so analytics are easier to run. Still, if we had a different data model, it might make more sense to just look up the pathogen from the associated object (such as tree_id) in the DB.

ehoops-cz:
> This isn't really a positive or negative comment, but I just wanted to mention that for a lot of these events it is _possible_ to get the pathogen from already available information. For example, a TREE_VIEW_NEXTSTRAIN has the tree_id, which then can be cross-referenced against the trees to find the inputs, which then allows digging down to the samples, where the pathogen of the samples can be looked up (and implicitly the same pathogen should be present on all the samples or something went terribly, terribly wrong). But that's a really convoluted way to find out the pathogen, so it makes a lot of sense to just chuck it into the events so analytics are easier to run. Still, if we had a different data model, it might make more sense to just look up the pathogen from the associated object (such as tree_id) in the DB.

This makes sense.  It seems like the trade off is a slightly larger payload for each event.  This seems ok for now, but I could see this getting bloated over time, so it might be something to watch out for in the future.

PR #1582, opened 2023-01-26
Remove old table types and remediate exposed type errors 

mmmmmmaya:
### Summary
- **What:**
  - Remove types associated with the old table
  - Remediate type conflicts that were exposed as a result
  - Remove CZB genome recovery from the metadata export
  - Refactor a function for metadata export that was challenging to read
- **Why:** table refactor clean up
- **Ticket:**
  - [[sc-224267]](https://app.shortcut.com/genepi/story/224267)
  - [[sc-224266]](https://app.shortcut.com/genepi/story/224266)
- **Env:** https://maya-types-frontend.dev.czgenepi.org/

### Testing
Primarily need verification for metadata export.

### Demos
No visual changes.

### Notes
- The old table type was fairly permissive, as it had to support both samples and trees in a variety of shapes and contexts. When those old types were removed, it exposed a lot of type errors that should never have existed. As a result, there is a lot of type shuffling going on in this PR, but since most of the changes are just for TypeScript they should not impact the functionality of the site. 
- I had to do a few small refactors to get some of the types to play nicely together.
- I refactored the function that expanded subheaders for the metadata tsv export. I tagged @phoenixAja to take a look at that and make sure it looks good, since errors here could impact the user experience.

### Checklist
- [x] I added labels to my PR


src/frontend/src/views/Data/tableHeaders/commonSampleHeaders.tsx
```diff
@@ -9,9 +10,8 @@ export const PRIVATE_ID_HEADER: Header = {
   },
 };
 
-export const PUBLIC_ID_HEADER: Header = {
+export const PUBLIC_ID_HEADER: TableHeader<Sample> = {
```
ehoops-cz: Are these header definitions used for the new table?

src/frontend/src/views/Data/tableHeaders/types.ts
```diff
@@ -0,0 +1,21 @@
+export interface SubHeader {
+  key: string;
+  text: string;
+}
+
+export type TooltipTextType = {
+  boldText: string;
+  regularText: string;
+  link?: {
```
ehoops-cz, line 9: 😭 I know we do this all over, but we really shouldn't have links in tooltips since they aren't reachable with the keyboard.

ehoops-cz (approved):
Looks good.  One question about the headers, but jw if this is doing some unnecessary work.

src/frontend/src/views/Data/tableHeaders/commonSampleHeaders.tsx
mmmmmmaya: Good point. This is primarily for exported metadata. I should probably rename the type (we will still need it for the export script), but we can remove the tooltips from the type since it lives in the cell def now.

src/frontend/src/views/Data/tableHeaders/types.ts
mmmmmmaya: Make a ticket in the eng backlog and maybe we can prioritize changing things!

phoenixAja (approved):

PR #1585, opened 2023-01-27
Scheduled Nextclade running for lineage-qc job

vincent-czi:
### Summary:
- **What:** Process to kick off scheduled Nextclade runs and refresh calls against any samples that are now stale.
- **Ticket:** [sc<228314>](https://app.shortcut.com/genepi/story/<228314>)
- **Env:** http://vince-scheduled-nextclade-frontend.dev.czgenepi.org/ (may still be deploying when you read this)

### Notes:

- Over time, samples get "stale" in their QC/Lineage/Mutations call. Nextclade releases new versions of the reference datasets behind those QC/Lineage/Mutations calls, so if a new dataset becomes available, any previous calls are now stale and all those samples should get re-called with the latest dataset.
- Core idea in the code here is taking the main process from before -- for on-demand calling -- and expanding it so it's possible to capture which samples, if any, need "refreshing", then just doing the same process on those samples.
- Introduces multiple run types to that main process. Previously, we just did `specified-ids-only` where a list of samples would be provided to run Nextclade against. Now we also have `refresh-stale` which takes info about the most recent Nextclade dataset and compares that to what's in the DB for the various samples. Any samples on a stale dataset get a Nextclade run.
- If there are no stale samples, gracefully exits and turns into a no-op.
- The scheduled autorun aspect just kicks off the "on-demand" workflow for each pathogen in our DB under one of those new `refresh-stale` type runs.

### Checklist:
- [x] I merged latest `trunk`
- [x] I added labels to my PR
~[ ] I added relevant unit tests~
~[ ] I have notified others of changes they need to make locally (migrations, jobs, package updates, etc)~


src/backend/aspen/workflows/nextclade/run_nextclade.sh
```diff
@@ -1,37 +1,49 @@
...
-  --name "${nextclade_dataset_name}" \
-  --output-dir "${NEXTCLADE_DATASET_DIR}"
-# Inside bundle, this file has info about the bundle we want to capture.
-DATASET_TAG_FILE=tag.json
+# In some cases, we discover nothing needs to be done, can exit early.
+should_exit_because_no_samples=$(jq --raw-output ".should_exit_because_no_samples" "${JOB_INFO_FILE}")
+if [ "${should_exit_because_no_samples}" = true ] ; then
+    echo "No samples to run Nextclade against. Exiting workflow."
+    exit 0
+fi
```
jgadling, line 48: Do we still need to run `nextclade dataset get` here?

vincent-czi: No, not any more. I moved the `dataset get` call into the Python prep_samples script since we use the tag info as part of how that script operates.

src/backend/aspen/workflows/nextclade/prep_samples.py
```diff
@@ -18,44 +35,134 @@
...
+                nextclade_dataset_info["tag"],
+            )
+        elif run_type == RunType.FORCE_ALL:
+            sample_ids = get_all_sample_ids_for_pathogen(
+                session,
+                target_pathogen,
+            )
+
+        # We should abandon this workflow run if there are no samples to run.
+        if sample_ids == []:
```
jgadling:
```suggestion
        if not sample_ids:
```

src/backend/aspen/workflows/nextclade/prep_samples.py
```diff
@@ -18,44 +35,134 @@
...
+            if run_type == RunType.REFRESH_STALE:
```
jgadling, line 173:
```suggestion
            if run_type != RunType.SPECIFIED_IDS_ONLY:
```
It's possible we won't have any samples in our system for a given pathogen, and we probably don't want to spam alerts if that's the case

jgadling (approved):

src/backend/aspen/workflows/nextclade/prep_samples.py
vincent-czi: I agree generally, but right now the only other run type that would kick off an alert would be a `force-all` -- which is currently just for manual firing by an engineer in case something weird happens -- so I think it's reasonable for the alert to happen then. If we tried to trigger a `force-all` run for a pathogen with no samples at all, that seems like a usage mistake we would want to be notified about, not something quietly logged.

jgadling: An engineer running this manually will see the "no samples found" message and be able to decide what to do, while raising alerts when we have no samples is going to make it harder to automate this force-all job.

vincent-czi: I guess that I was imagining the force-all job to never be automated, manual only, so it wouldn't get too spammy. If an engineer is watching the logs, yeah, I agree, the alert is annoying. But I was thinking fail loud was better in case the eng forgets to monitor after kicking off a force-all.

PR #1594, opened 2023-01-31
Make sure aligned sequence dumps have an associated workflow.

jgadling:
### Summary:
- **What:** Make sure we have enough data in the DB to support MPX tree builds.

### Notes:
We were having trouble kicking off MPX tree builds in staging even though we'd already executed the ingestion workflow. This is because our "start a tree build" job looks for the *most recent* alignment data, which we can only infer by looking at the workflow run that produced the alignment data. Our script wasn't creating an alignment workflow object at all, so tree builds couldn't be started.

I made a few other minor changes here too:
1. Renamed the workflow script to reflect that it's for working with *aligned* data rather than *raw* data
2. Updated our ingestion test script so that it doesn't download all the MPX data from nextstrain on every run. This will make it easier to write tests (Coming Soon(TM)!) against a successful import.

### Checklist:
- [x] I added labels to my PR


phoenixAja (approved):
Thanks!

PR #1597, opened 2023-02-01
Maya/229189 sample table pathogen config

mmmmmmaya:
### Summary
- **What:** Use pathogen-specific column defs for the new samples table
- **Why:** mpox uses different columns than sc2
- **Env:** https://maya-columns-frontend.dev.czgenepi.org/

#### Before: 
before

#### After: 
after

### Checklist
- [x] I added labels to my PR


src/frontend/src/views/Data/components/SamplesView/components/SamplesTable/columnDefinitions/publicId.tsx
```diff
@@ -0,0 +1,24 @@
...
+  id: "publicId",
+  accessorKey: "publicId",
+  header: ({ header, column }) => (
+    <SortableHeader
+      header={header}
+      style={generateWidthStyles(column)}
+      tooltipStrings={{
+        boldText: "Public ID",
+        regularText:
+          "This is your GISAID ID or public ID generated by CZ Gen Epi.",
```
ehoops-cz: This text is different for mpox:
"This is your GenBank Accession or public ID generated by CZ GEN EPI.

src/frontend/src/views/Data/components/SamplesView/components/SamplesTable/index.tsx
```diff
@@ -293,6 +17,12 @@ const SamplesTable = ({
   isLoading,
   setCheckedSamples,
 }: Props): JSX.Element => {
+  const pathogen = useSelector(selectCurrentPathogen);
+  const columns: ColumnDef<Sample, any>[] = useMemo(
```
ehoops-cz, line 21: 👍 , I wonder if I should have been memoizing other config getters

src/frontend/src/views/Data/components/SamplesView/components/SamplesTable/columnDefinitions/publicId.tsx
ehoops-cz: Everything else in this column is the same, I'm not sure if we should add a strings file for this or create a separate column file...

ehoops-cz (changes requested):
Just one copy update for mpox.  Otherwise, this looks great!  Thanks for doing this!

src/frontend/src/views/Data/components/SamplesView/components/SamplesTable/index.tsx
mmmmmmaya: if you define the columns outside the component, you don't have to. I only have done it here since I need to put it inside the component in order to pass the pathogen in

ehoops-cz (approved):
thanks 🎉

PR #1600, opened 2023-02-01
Tree location default - use max depth

ehoops-cz:
### Summary:
- **What:** `For mpox we are only using division rather than location.  We need to apply this to the create nextstrain tree modal default location.`
- **Ticket:** none
- **Env:** `none`

### Demos:
mpox:
Screenshot 2023-02-01 at 12 39 51 PM
sc2:
Screenshot 2023-02-01 at 12 46 10 PM

Mpox upload flow vs tree flow:
Screenshot 2023-02-02 at 3 10 07 PM
Screenshot 2023-02-02 at 3 10 45 PM

SC2 tree flow still has location-level data
Screenshot 2023-02-02 at 3 12 24 PM

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/queries/locations.ts
+++ b/src/frontend/src/common/queries/locations.ts
@@ -12,7 +12,7 @@ export interface LocationsResponse {
   locations: GisaidLocation[];
 }
 
-const locationDepthPathogenConfig: PathogenConfigType<
+export const locationDepthPathogenConfig: PathogenConfigType<
   keyof GisaidLocation | null
 > = {
   [Pathogen.COVID]: null,
--- a/src/frontend/src/common/queries/locations.ts
+++ b/src/frontend/src/common/queries/locations.ts
@@ -19,7 +19,30 @@ export const locationDepthPathogenConfig: PathogenConfigType<
   [Pathogen.MONKEY_POX]: "division",
 };
 
-const fetchLocations = (): Promise<LocationsResponse> => {
+const fetchAllLocations = (): Promise<LocationsResponse> => {
+  return getBackendApiJson(API.LOCATIONS);
+};
+
+export const USE_LOCATIONS_INFO_QUERY_KEY = {
+  entities: [ENTITIES.LOCATION_INFO],
+  id: "locationInfo",
+};
+
+const ONE_HOUR = 1000 * 60 * 60; // for react-query; milliseconds
+
+export function useLocations(): UseQueryResult<LocationsResponse, unknown> {
+  return useQuery([USE_LOCATIONS_INFO_QUERY_KEY], fetchAllLocations, {
+    retry: false,
+    // Because locations are very stable and mildly heavy (~2Meg), we mark
+    // stale and re-fetch after an hour rather than default insta-stale.
+    // Note, this does not prevent garbage cleanup due to `cacheTime` if user
+    // does not have data "actively" used. In that case, re-fetch would happen
+    // next time a component using this data mounts and uses this func.
+    staleTime: ONE_HOUR,
+  });
+}
+
+const fetchPathogenDepthLocations = (): Promise<LocationsResponse> => {
   const state = store.getState();
   const pathogen = selectCurrentPathogen(state);
 
@@ -31,23 +54,31 @@ const fetchLocations = (): Promise<LocationsResponse> => {
   return getBackendApiJson(route);
 };
 
-export const USE_LOCATIONS_INFO_QUERY_KEY = {
+export const USE_PATHOGEN_DEPTH_LOCATIONS_INFO_QUERY_KEY = {
   entities: [ENTITIES.LOCATION_INFO],
-  id: "locationInfo",
+  id: "pathogenDepthocationInfo",
 };
 
-const ONE_HOUR = 1000 * 60 * 60; // for react-query; milliseconds
-
-export function useLocations(): UseQueryResult<LocationsResponse, unknown> {
-  return useQuery([USE_LOCATIONS_INFO_QUERY_KEY], fetchLocations, {
-    retry: false,
-    // Because locations are very stable and mildly heavy (~2Meg), we mark
-    // stale and re-fetch after an hour rather than default insta-stale.
-    // Note, this does not prevent garbage cleanup due to `cacheTime` if user
-    // does not have data "actively" used. In that case, re-fetch would happen
-    // next time a component using this data mounts and uses this func.
-    staleTime: ONE_HOUR,
-  });
+export function useNamedPathogenDepthLocations(): UseQueryResult<
+  NamedLocationsResponse,
+  unknown
+> {
+  return useQuery(
+    [USE_PATHOGEN_DEPTH_LOCATIONS_INFO_QUERY_KEY],
+    fetchPathogenDepthLocations,
+    {
+      retry: false,
+      // Using `select` allows it to share cache with other USE_LOCATIONS_INFO_QUERY_KEY,
+      // but give a different view on the same data after processed by `select` func.
+      select: foldInNamesToLocations,
+      // Because locations are very stable and mildly heavy (~2Meg), we mark
+      // stale and re-fetch after an hour rather than default insta-stale.
+      // Note, this does not prevent garbage cleanup due to `cacheTime` if user
+      // does not have data "actively" used. In that case, re-fetch would happen
+      // next time a component using this data mounts and uses this func.
+      staleTime: ONE_HOUR,
+    }
+  );
 }
 
 interface NamedLocationsResponse {
@@ -77,7 +108,7 @@ export function useNamedLocations(): UseQueryResult<
   NamedLocationsResponse,
   unknown
 > {
-  return useQuery([USE_LOCATIONS_INFO_QUERY_KEY], fetchLocations, {
+  return useQuery([USE_LOCATIONS_INFO_QUERY_KEY], fetchAllLocations, {
     retry: false,
     // Using `select` allows it to share cache with other USE_LOCATIONS_INFO_QUERY_KEY,
     // but give a different view on the same data after processed by `select` func.
@@ -105,7 +136,7 @@ export function useNamedLocationsById(): UseQueryResult<
   NamedLocationsByIdResponse,
   unknown
 > {
-  return useQuery([USE_LOCATIONS_INFO_QUERY_KEY], fetchLocations, {
+  return useQuery([USE_LOCATIONS_INFO_QUERY_KEY], fetchAllLocations, {
     retry: false,
     // Using `select` allows it to share cache with other USE_LOCATIONS_INFO_QUERY_KEY,
     // but give a different view on the same data after processed by `select` func.
--- a/src/frontend/src/common/queries/locations.ts
+++ b/src/frontend/src/common/queries/locations.ts
@@ -63,22 +63,27 @@ export function useNamedPathogenDepthLocations(): UseQueryResult<
   NamedLocationsResponse,
   unknown
 > {
-  return useQuery(
-    [USE_PATHOGEN_DEPTH_LOCATIONS_INFO_QUERY_KEY],
-    fetchPathogenDepthLocations,
-    {
-      retry: false,
-      // Using `select` allows it to share cache with other USE_LOCATIONS_INFO_QUERY_KEY,
-      // but give a different view on the same data after processed by `select` func.
-      select: foldInNamesToLocations,
-      // Because locations are very stable and mildly heavy (~2Meg), we mark
-      // stale and re-fetch after an hour rather than default insta-stale.
-      // Note, this does not prevent garbage cleanup due to `cacheTime` if user
-      // does not have data "actively" used. In that case, re-fetch would happen
-      // next time a component using this data mounts and uses this func.
-      staleTime: ONE_HOUR,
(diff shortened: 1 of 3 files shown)


src/frontend/src/common/utils/locationUtils.ts
```diff
@@ -151,3 +151,70 @@ export const getIdFromCollectionLocation = (
...
+  const finalIndex = locationHierarchy.findIndex(
+    (element) => element === maxDepth
+  );
+
+  const foundLocations = locationHierarchy.reduce((acc, tier, index) => {
+    // Until we get to the index of the maxDepth, search for the searchLocation value
+    // After that, search for null
+    const filterValue = index <= finalIndex ? searchLocation[tier] : null;
+    return acc.filter((value) => value[tier] === filterValue);
+  }, locations);
```
ehoops-cz, line 191: There is an optimization here where we don't need to check further tiers if the previous tier is null - i.e. if division is null, don't check location.  It's not coming to me though... also not sure how relevant this is since we're currently using max depth of "location" for sc2 and "division" for mpox.

src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/components/CreateNSTreeModal/index.tsx
```diff
@@ -100,18 +104,44 @@ export const CreateNSTreeModal = ({
...
     useState<NamedGisaidLocation | null>(getLocationFromGroup(groupInfo));
 
+  const locationMaxDepthFinder = useMemo(() => {
+    return createMaxDepthLocationFinder(namedLocations);
+  }, [namedLocations]);
+
+  const setLocationToGroupDefault = () => {
+    const locationDepth = locationDepthPathogenConfig[pathogen];
+
+    const defaultTreeLocation =
```
mmmmmmaya: i think this might be more readable as actual if/else-if statements, but I hold that lightly as it's personal preference.

src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/components/CreateNSTreeModal/index.tsx
```diff
@@ -100,18 +104,44 @@ export const CreateNSTreeModal = ({
...
+      locationDepth === null
+        ? // If locationDepth is not specified, use the group's location
+          getLocationFromGroup(groupInfo)
+        : // if the locationDepth is specified and location is defined, then search
+        groupInfo?.location
+        ? // Search for the location id that matches the max depth of the group's location
+          // For example, for mpox the max depth is "division", we want to find the id of
+          // the location that has the same division as the group's location, but location is null
+          locationMaxDepthFinder(groupInfo?.location, locationDepth)
```
mmmmmmaya: if you're in this part of the ternary, you already know groupInfo.location is defined, so you can drop the `?`

src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/components/CreateNSTreeModal/index.tsx
```diff
@@ -100,18 +104,44 @@ export const CreateNSTreeModal = ({
   // Filter based on location
   const { data: groupInfo } = useGroupInfo();
   const { data: namedLocationsData } = useNamedLocations();
-  const namedLocations: NamedGisaidLocation[] =
-    namedLocationsData?.namedLocations ?? [];
+  const namedLocations: NamedGisaidLocation[] = useMemo(() => {
```
mmmmmmaya, line 140: I'm a little confused on this use of useMemo. Usually useMemo would be used when a function is computationally expensive. What is the reason for using it here?

src/frontend/src/common/utils/locationUtils.ts
```diff
@@ -151,3 +151,70 @@ export const getIdFromCollectionLocation = (
     return collectionLocation.id;
   }
 };
+
+function findMaxDepthLocation(
+  searchLocation: GisaidLocation | NamedGisaidLocation,
+  maxDepth: keyof GisaidLocation,
```
mmmmmmaya, line 157: 👍  love the types

mmmmmmaya:
Would love for Vince to look this over, too!

src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/components/CreateNSTreeModal/index.tsx
vincent-czi: Yeah, I also find ternary-in-ternary hard to read

src/frontend/src/common/utils/locationUtils.ts
```diff
@@ -151,3 +151,70 @@ export const getIdFromCollectionLocation = (
...
+  if (locations.length === 0) {
+    return undefined;
+  }
+
+  // Only do cache interactions if one was provided
+  if (
+    locationFinderCache !== null &&
+    locationFinderCache[`${searchLocation.id}${maxDepth}`]
+  ) {
+    return locationFinderCache[`${searchLocation.id}${maxDepth}`];
```
vincent-czi, line 173: I'm confused about the usage/intent of the `locationFinderCache` here. Isn't the `searchLocation.id` the PK for the location? It feels like if we already have the PK to identify the location, there's not much to look up anymore?

Maybe I'm misunderstanding the intent. Is it more that we want a way of boiling more-specific locations off into less-specific ones? Like we know id 47 is a `location` level Location, but we're regularly trying to pull the `country` level Location that id 47 is inside, so we have `47country` to look it up frequently?

To give some background, I was the one who originally set up the `locationFinderCache` stuff for the `findLocationFromString`, and the intent was to speed up how metadata TSV import worked in the Upload flow. Since we can have 100s of rows being imported at once, searching the whole set of Locations for every row would be a lot of waste, since most rows are identical location search terms. So it was primarily there as a way to avoid duplicate searches and speed up runtime. But with that framework as where I'm coming from, I don't really see what the intent of the cache is here? Because, if starting with a string search term, we'd still have to search the whole Locations set before we'd have the id to do a lookup with, right?

src/frontend/src/common/utils/locationUtils.ts
vincent-czi: This is kind of orthogonal, but it feels like the easiest approach is to just ensure that whatever provides the locations upstream pre-filters the Locations list down to  already dealing with the `maxDepth` aspect, so nothing of greater depth is provided in the Locations list already. The big challenge there is how do we deal with the group's location, but maybe we could just put in a special look-up somehow? Although isn't that already happening via the BE by passing the query param?

I guess my above thought could screw things up for TSV import. If a user puts in "Los Angeles" for their location in TSV when doing Mpox, should "California" get populated (b/c Mpox max depth is division), or should just nothing get populated at all? How I'd approach this depends on how we need to handle metadata import regarding locations, but I just don't know what the specs are there.

vincent-czi (approved):
I feel like this all works, but I don't fully understand the intent of some of what's going on. That could just be on me though -- I was trying to read it while in a meeting -- so if you're confident it's working and know it's reasonably robust (make sure to check with uploading a metadata TSV that has a few 100 rows), I'm fine with it as is.

src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/components/CreateNSTreeModal/index.tsx
ehoops-cz: I like them, but I'll be overruled here.

src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/components/CreateNSTreeModal/index.tsx
ehoops-cz: real reason: The browser ate it without the memo and the editor suggested it.
I'm not entirely sure why this kept getting recomputed.

src/frontend/src/common/utils/locationUtils.ts
ehoops-cz: This may not need to be cached.  It should be one lookup value per group.  My reasoning was if things were recomputed we wouldn't have to search again. At that point the cache may be gone from re-rendering though.

src/frontend/src/common/queries/locations.ts
```diff
@@ -50,6 +42,45 @@ export function useLocations(): UseQueryResult<LocationsResponse, unknown> {
   });
 }
 
+const fetchPathogenDepthLocations = (): Promise<LocationsResponse> => {
```
ehoops-cz, line 50: Created a new query because mpox needs the full locations list for things like sample upload but needs the list at the division level for the nextstrain tree creation modal.

src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/components/CreateNSTreeModal/index.tsx
```diff
@@ -99,19 +103,50 @@ export const CreateNSTreeModal = ({
 
   // Filter based on location
   const { data: groupInfo } = useGroupInfo();
-  const { data: namedLocationsData } = useNamedLocations();
-  const namedLocations: NamedGisaidLocation[] =
-    namedLocationsData?.namedLocations ?? [];
+  const { data: namedLocationsData } = useNamedPathogenDepthLocations();
```
ehoops-cz: Now this refetches locations even for covid - I'm not sure where to put the conditional since a hook is not supposed to be inside of one.
@mmmmmmaya do you know off the top of your head?

ehoops-cz:
Update: Vince uncovered a missed requirement where locations of sample uploads are still supposed to use the location-level, while the mpox trees are supposed to use the division-level.  I attempted to make this change - since we're on a time crunch, the easiest way seemed to be to add a second react-query.  However, I am currently using the second react query in all cases on the NS Tree Modal.  This isn't optimal since users should not need to refetch locations if they are working on covid which has trees on the location level.  Any suggestions would be greatly appreciated 🙏

src/frontend/src/common/queries/locations.ts
```diff
@@ -31,16 +54,29 @@ const fetchLocations = (): Promise<LocationsResponse> => {
   return getBackendApiJson(route);
 };
 
-export const USE_LOCATIONS_INFO_QUERY_KEY = {
+export const USE_PATHOGEN_DEPTH_LOCATIONS_INFO_QUERY_KEY = {
   entities: [ENTITIES.LOCATION_INFO],
-  id: "locationInfo",
+  id: "pathogenDepthocationInfo",
```
vincent-czi: nit: pathogenDepthocationInfo -> `pathogenDepthLocationInfo`

src/frontend/src/common/queries/locations.ts
```diff
@@ -31,16 +54,29 @@ const fetchLocations = (): Promise<LocationsResponse> => {
...
+> {
+  const state = store.getState();
+  const pathogen = selectCurrentPathogen(state);
+  // Choosing the query key based on the pathogen ensures that pathogens
+  // that use all of the locations for their trees don' have to re-fetch
+  // everything
+  const queryKey =
+    locationDepthPathogenConfig[pathogen] === null
+      ? USE_LOCATIONS_INFO_QUERY_KEY
+      : USE_PATHOGEN_DEPTH_LOCATIONS_INFO_QUERY_KEY;
```
vincent-czi: I don't know if it's possible to trigger this given how things are currently set up -- it's quite possible it never happens right now -- but I think this wouldn't actually fire a new fetch if we changed from pathogen to pathogen while the query was active somewhere in the component tree. Since `USE_PATHOGEN_DEPTH_LOCATIONS_INFO_QUERY_KEY` is a constant, both covid and mpox would see the same query key when firing `useNamedPathogenDepthLocations`. I think we'd need the query key to be lightly dynamic, and somehow also incorporate the pathogen slug or something so there's a difference in keys when we change pathogens.

src/frontend/src/views/Data/components/SamplesView/components/SampleTableModalManager/components/CreateNSTreeModal/index.tsx
```diff
@@ -103,7 +103,7 @@ export const CreateNSTreeModal = ({
 
   // Filter based on location
   const { data: groupInfo } = useGroupInfo();
-  const { data: namedLocationsData } = useNamedLocations();
+  const { data: namedLocationsData } = useNamedPathogenDepthLocations();
```
vincent-czi: I'd like a tiny comment here just explaining that the NSTree creation process restricts the locations you can choose from based on the type of pathogen. Seems like a useful thing to explain here.

vincent-czi (approved):
Overall looks good, although I do think there might be the seed of a bug based on how the query key is set up right now. I don't think it will actually cause a bug because of how our app is using react-query at the moment, but something to be aware of.

src/frontend/src/common/queries/locations.ts
ehoops-cz: This bug really did occur.  To fix it, I moved setting the queryKey out to where this is called in the NSTreeModal.  That component seems to do a lot, but it does work.

PR #1601, opened 2023-02-01
Overwrite include.txt before copying to s3

jgadling:
### Notes:
We copy include.txt to S3 to aid in debugging of tree failures. Let's make sure we're 100% *done* processing that file before we copy it to s3 so our logging makes more sense.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/workflows/nextstrain_run/run_nextstrain_sc2.sh
+++ b/src/backend/aspen/workflows/nextstrain_run/run_nextstrain_sc2.sh
@@ -57,6 +57,12 @@ aligned_gisaid_location=$(
 )
 
 
+# Tree builds break if include.txt is empty. If we don't have any selected samples,
+# then copy in nextstrain's defaults.
+if [ ! -s /ncov/data/include.txt ]; then
+    cp /ncov/defaults/include.txt /ncov/data/include.txt
+fi
+
 # Persist the build config we generated.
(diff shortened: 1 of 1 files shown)


danrlu (approved):
Makes sense!

src/backend/aspen/workflows/nextstrain_run/run_nextstrain_sc2.sh
```diff
@@ -57,6 +57,12 @@ aligned_gisaid_location=$(
...
 
+# Tree builds break if include.txt is empty. If we don't have any selected samples,
+# then copy in nextstrain's defaults.
+if [ ! -s /ncov/data/include.txt ]; then
+    cp /ncov/defaults/include.txt /ncov/data/include.txt
+fi
+
 # Persist the build config we generated.
 $aws s3 cp /ncov/my_profiles/aspen/builds.yaml "${s3_prefix}/builds.yaml"
 $aws s3 cp /ncov/data/include.txt "${s3_prefix}/include.txt"
```
danrlu, line 68: how come this `cp` command hasn't failed so far when this `data/include.txt` doesn't exist (like for all automatic trees)?

jgadling: the file always exists, it's just empty sometimes!

PR #1604, opened 2023-02-02
Maya/229189 close modal bug

mmmmmmaya:
### Summary
- **What:** Resolve a bug where closing a modal (such as download modal) would clear checked sample state
- **Why:** Design parity with old table
- **Env:** https://maya-checked-bug-frontend.dev.czgenepi.org/

### Testing
1. Check some rows
1. Open the download modal
2. Close the download modal without downloading anything
3. Rows should no longer be checked

#### Before: 
before

#### After: 
after

### Checklist
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/components/Table/index.tsx
+++ b/src/frontend/src/components/Table/index.tsx
@@ -9,7 +9,7 @@ import {
   useReactTable,
 } from "@tanstack/react-table";
 import { Table as SDSTable, TableHeader, TableRow } from "czifui";
-import { map } from "lodash";
+import { isEqual, map } from "lodash";
 import { useEffect, useRef, useState } from "react";
 import { useVirtual, VirtualItem } from "react-virtual";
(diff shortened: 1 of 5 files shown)


src/frontend/src/components/Table/index.tsx
```diff
@@ -23,7 +23,14 @@ interface Props<T> {
   tableData: IdMap<T> | undefined;
   initialSortKey?: string;
   isLoading?: boolean;
+  uniqueIdentifier: keyof T;
+  // the following three props should either
+  // 1) always appear together
+  // 2) none appear at all
+  // it's hard to programmatically enforce this, so leaving it as a convention for now
```
ehoops-cz, line 30: 👍

ehoops-cz (approved):
This looks good.  I assume someone from product/design wanted this functionality?  If they were neutral, I think it would be cleaner to not have to do this.  But in any case, I don't see any issues.

PR #1605, opened 2023-02-02
analytics events for 500 sample limit performance updates

phoenixAja:
### Summary:
- **What:** add analytics events for sample upload (now that we can handle more samples in upload process)
- **Ticket:** [sc228799](https://app.shortcut.com/genepi/story/228799)
- **Env:** `<rdev link>`

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/analytics/eventTypes.ts
+++ b/src/frontend/src/common/analytics/eventTypes.ts
@@ -272,10 +272,8 @@ export type AnalyticsSamplesUploadSuccess = {
 
 /** EVENT_TYPES.SAMPLES_UPLOAD_FAILED */
 export type AnalyticsSamplesUploadFailed = {
-  // How many samples the user just uploaded
-  sample_count: number;
-  // JSON array of all the IDs for newly created samples for this upload
-  sample_ids: JsonString;
+  // What was the error message for the failed upload
+  failed_message: string;
   // See above docs on `AnalyticsSamplesUploadPageChange.upload_flow_uuid`.
   // For an Upload "flow" that ends in successful or failed upload, this will match up.
   upload_flow_uuid: string;
--- a/src/frontend/src/common/analytics/eventTypes.ts
+++ b/src/frontend/src/common/analytics/eventTypes.ts
@@ -246,7 +246,7 @@ export type AnalyticsUploadMetadataType = {
   // The type of metadata the user is uploading
   metadata_entry_type: "MANUAL" | "TSV" | "BOTH";
   // See above docs on `AnalyticsSamplesUploadPageChange.upload_flow_uuid`.
-  // For an Upload "flow" that ends in successful upload, this will match up.
+  // For an Upload "flow" that ends in successful or failed upload, this will match up.
   upload_flow_uuid: string;
   // number of samples in the upload
   sample_count: number;
@@ -259,7 +259,7 @@ export type AnalyticsSamplesUploadSuccess = {
   // JSON array of all the IDs for newly created samples for this upload
   sample_ids: JsonString;
   // See above docs on `AnalyticsSamplesUploadPageChange.upload_flow_uuid`.
-  // For an Upload "flow" that ends in successful upload, this will match up.
+  // For an Upload "flow" that ends in successful or failed upload, this will match up.
   upload_flow_uuid: string;
   // The current pathogen. For example, "SC2" or "MPX".
   pathogen: string;
@@ -272,7 +272,7 @@ export type AnalyticsSamplesUploadFailed = {
   // JSON array of all the IDs for newly created samples for this upload
   sample_ids: JsonString;
   // See above docs on `AnalyticsSamplesUploadPageChange.upload_flow_uuid`.
-  // For an Upload "flow" that ends in successful upload, this will match up.
+  // For an Upload "flow" that ends in successful or failed upload, this will match up.
   upload_flow_uuid: string;
   // The current pathogen. For example, "SC2" or "MPX".
   pathogen: string;
--- a/src/frontend/src/common/analytics/eventTypes.ts
+++ b/src/frontend/src/common/analytics/eventTypes.ts
@@ -1,3 +1,5 @@
+import { Pathogen } from "../redux/types";
+
 /**
  * Canonical list of all the various event types we track with analytics.
  *
@@ -98,6 +100,9 @@ export type EventData = Record<string, EventValue | undefined>;
 // For ease of readability below, we alias string for those cases.
 type JsonString = string;
 
+// capture how the user is uploading or entering metadata
+export type UploadFormMetadataType = "BOTH" | "MANUAL" | "TSV";
+
 /**
  * Structure of additionalEventData for each EVENT_TYPES type that sends it.
  *
@@ -141,7 +146,7 @@ export type AnalyticsTreeViewNextstrain = {
   // Tree that user is being sent to view
   tree_id: number;
   // The current pathogen. For example, "SC2" or "MPX".
-  pathogen: string;
+  pathogen: Pathogen;
 };
 
 /** EVENT_TYPES.TREE_CREATION_NEXTSTRAIN */
@@ -161,7 +166,7 @@ export type AnalyticsTreeCreationNextstrain = {
   // generally shouldn't happen.
   group_location_id: number | null;
   // The current pathogen. For example, "SC2" or "MPX".
-  pathogen: string;
+  pathogen: Pathogen;
   // Lineages selected for tree creation. json stringified list of strings.
   // "[]" indicates "All lineages"
   selected_lineages: JsonString;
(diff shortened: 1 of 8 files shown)


src/frontend/src/common/analytics/eventTypes.ts
```diff
@@ -233,14 +239,40 @@ export type AnalyticsSamplesUploadPageChange = {
   upload_flow_uuid: string;
 };
 
+/** EVENT_TYPES.UPLOAD_METADATA_TYPE */
+export type AnalyticsUploadMetadataType = {
+  // The current pathogen. For example, "SC2" or "MPX".
+  pathogen: string;
```
mmmmmmaya: this should be `Pathogen`, to restrict to the pathogens we have available.

src/frontend/src/common/analytics/eventTypes.ts
```diff
@@ -233,14 +239,40 @@ export type AnalyticsSamplesUploadPageChange = {
...
   // How many samples the user just uploaded
   sample_count: number;
   // JSON array of all the IDs for newly created samples for this upload
   sample_ids: JsonString;
   // See above docs on `AnalyticsSamplesUploadPageChange.upload_flow_uuid`.
-  // For an Upload "flow" that ends in successful upload, this will match up.
+  // For an Upload "flow" that ends in successful or failed upload, this will match up.
+  upload_flow_uuid: string;
+  // The current pathogen. For example, "SC2" or "MPX".
+  pathogen: string;
```
mmmmmmaya: same type comment here

src/frontend/src/views/Upload/components/Metadata/index.tsx
```diff
@@ -33,24 +33,57 @@ import {
...
   const [autocorrectWarnings, setAutocorrectWarnings] =
     useState<SampleIdToWarningMessages>(EMPTY_OBJECT);
+  // this is used to track whether the user has manually edited the metadata for analytics
+  const [hasManuallyEditedMetadata, setHasManuallyEditedMetadata] =
+    useState<boolean>(false);
 
   let numberOfDetectedSamples = 0;
   if (samples != null) {
-    numberOfDetectedSamples = Object.keys(samples).length;
+    numberOfDetectedSamples = (Object.keys(samples).length);
```
mmmmmmaya: what's up with these parens?

src/frontend/src/views/Upload/components/Metadata/index.tsx
```diff
@@ -33,24 +33,57 @@ import {
...
   }
 
+  useEffect(() => {
+    const hasMetadataBeenEdited =
+      hasImportedMetadataFile || hasManuallyEditedMetadata;
+    if (!hasMetadataBeenEdited) return;
+
+    type MetadataType = "BOTH" | "MANUAL" | "TSV";
```
mmmmmmaya: this should be outside the component so it is only defined once, like an `interface`

src/frontend/src/views/Upload/components/Metadata/index.tsx
```diff
@@ -33,24 +33,57 @@ import {
...
+    let metadataType: MetadataType = "MANUAL";
+    if (hasImportedMetadataFile) metadataType = "TSV";
+    if (hasImportedMetadataFile && hasManuallyEditedMetadata)
+      metadataType = "BOTH";
+    analyticsTrackEvent<AnalyticsUploadMetadataType>(
+      EVENT_TYPES.UPLOAD_METADATA_TYPE,
```
mmmmmmaya: should it be the case that if someone imports a metadata file and then manually changes metadata that two events should be fired?

mmmmmmaya:
mostly lgtm, just the one question and a couple type comments

src/frontend/src/views/Upload/components/Metadata/index.tsx
phoenixAja: oh woops that was an artifact from something that i was trying earlier, thanks for catching

src/frontend/src/views/Upload/components/Metadata/index.tsx
phoenixAja: hmm good point, i was thinking that it would be ok since this event is linked to the upload flow id, so you could filter out double events easily, but i'll try moving this to the review step to only fire once

src/frontend/src/views/Upload/components/Review/index.tsx
```diff
@@ -31,19 +31,53 @@ import {
...
   const [isConsentChecked, setIsConsentChecked] = useState(false);
 
+  let numberOfDetectedSamples = 0;
+  if (samples != null) {
+    numberOfDetectedSamples = Object.keys(samples).length;
+  }
+
+  // firing off analytics event in review since metadata editing should be finalized at this point.
+  useEffect(() => {
+    const hasMetadataBeenEdited =
```
mmmmmmaya, line 62: If I'm understanding it correctly, this conditional should not be needed. I don't think a person can get to the review step without either uploading a file or manually entering metadata.

src/frontend/src/views/Upload/index.tsx
```diff
@@ -35,6 +35,11 @@ export default function Upload(): JSX.Element | null {
   const pathogen = useSelector(selectCurrentPathogen);
   const [samples, setSamples] = useState<ISamples | null>(null);
   const [metadata, setMetadata] = useState<SampleIdToMetadata | null>(null);
+  const [hasManuallyEditedMetadata, setHasManuallyEditedMetadata] =
```
mmmmmmaya, line 38: Do all three of the upload components really need to know this state? It seems to me it's likely only needed on the metadata step.

phoenixAja: no but i didn't see another way to let the review page know how the metadata had been edited from the metadata page, do you know of a better way?

src/frontend/src/views/Upload/components/Review/index.tsx
vincent-czi: I wish, but you can actually refresh the page (or directly navigate via URL) while you're in the middle of the Upload flow and you get to that page of the Upload flow but with no data. It's arguable whether the defensive coding approach is necessary to cover this corner case, but I think it doesn't hurt to have it.

src/frontend/src/views/Upload/index.tsx
vincent-czi: I think for the Review step to have access to the information of how the metadata was set so the analytics event can include it, this is probably the most reasonable way.

src/frontend/src/views/Upload/components/Review/index.tsx
phoenixAja: yeah :/

src/frontend/src/views/Upload/index.tsx
phoenixAja: yep

phoenixAja:
i changed up the failed events target, re-requesting review

src/frontend/src/views/Upload/index.tsx
mmmmmmaya: I don't think sending this on the review step is necessarily the right call. What are we trying to track specifically? Any time a person does something metadata related? Their first upload/edit? The last? 

IMO, if it's about tracking actions on the metadata page, then the state should live in that component only, the events should fire from that component, and the Review component should not be involved.

If necessary, let's acquire a clearer understanding of what/when we need to track.

phoenixAja: so we want to track how the user is entering metadata, whether it's via manual entry or tsv upload. the issue with doing this on the metadata page is that it's hard to tell when the user is done with editing metadata. We don't want to trigger events multiple times if they enter data via tsv and then use the webform. I added this event to the review step because by that time the user would have been done entering metadata and we can trigger only one analytics event.

phoenixAja: @mmmmmmaya is that explanation enough? Can we merge this?

mmmmmmaya: Ok, I thought about this a bit and I think we can merge it as-is for now. However, I would like you to make a ticket for the eng backlog and post it on this thread and in a slack channel cc Christine that we should refactor the upload components. I think a couple of things here have code smell to me. I don't like the idea that all 3 need to have the same props with the way things are set up, especially if some of those props will not be used, and I think the state should probably be managed in `Upload` instead of the metadata page if it's used by multiple children.

- [x] make shortcut ticket to refactor upload flow
- [x] link to new ticket in this thread
- [x] post ticket in slack cc Christine

mmmmmmaya (approved):

src/frontend/src/views/Upload/index.tsx
phoenixAja: https://app.shortcut.com/genepi/story/229448/refactor-upload-flow-shared-props

PR #1606, opened 2023-02-02
[SC-225223] Add blank state for samples.

ehoops-cz:
### Summary:
- **What:** `Add blank state when the use does not have any samples for a given pathogen`
- **Ticket:** [sc225223](https://app.shortcut.com/genepi/story/225223)
- **Env:** `none`

### Demos:
Screenshot 2023-02-02 at 1 28 28 PM

### Notes:
This is only implemented on the new table and won't show until we swap over.

### Checklist:
- [x] I added labels to my PR


mmmmmmaya:
You forgot to include the code for the new component

src/frontend/src/views/Data/components/SamplesView/index.tsx
```diff
@@ -27,6 +28,15 @@ const SamplesView = (): JSX.Element => {
 
   // load sample data from server
   const { data: samples, isLoading } = useSampleInfo();
+  const [showBlankState, setShowBlankState] = useState<boolean>(false);
+
+  useEffect(() => {
```
mmmmmmaya: I don't think the useEffect is needed here. Couldn't think be a regular if statement that returns a blank state if no samples are available?

ehoops-cz: Was thinking that we would want to update this if someone deleted their last sample.  Would the component re-render in that case?

ehoops-cz:
> You forgot to include the code for the new component

Thanks! Added now

Janeece (approved):
I've been waiting for this moment for like 1.5 yrs 🤣 — looks great!

src/frontend/src/views/Data/components/SamplesView/index.tsx
mmmmmmaya: I think so, because `useSampleInfo` should update since it will expire the cache whenever you delete a sample. 

https://user-images.githubusercontent.com/7562933/216653976-54be7f88-e166-425c-ba9a-bcb97676f0d7.mov

src/frontend/src/views/Data/components/SamplesView/components/BlankState/style.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export const Container = styled.div`
+  display: flex;
+  flex-direction: row;
+  align-self: center;
+  ${(props: CommonThemeProps) => {
+    const spaces = getSpaces(props);
+    return `
+    margin-top: ${spaces?.xl}px;
+  `;
```
mmmmmmaya:
```suggestion
      margin-top: ${spaces?.xl}px;
    `;
```

src/frontend/src/views/Data/components/SamplesView/components/BlankState/style.ts
```diff
@@ -0,0 +1,67 @@
...
+`;
+
+export const StyledLinkText = styled.p`
+  margin: 0;
+  ${fontHeaderM}
+  ${(props: CommonThemeProps) => {
+    const spaces = getSpaces(props);
+    return `
+    margin-right: ${spaces?.xs}px;
+  `;
```
mmmmmmaya:
```suggestion
      margin-right: ${spaces?.xs}px;
    `;
```

mmmmmmaya (approved):

PR #1607, opened 2023-02-02
Maya/229189 trees perf

mmmmmmaya:
### Summary
- **What:** Improvements for the trees table
  - add ids and memoize components to improve perf on trees table
  - move trees table column defs to their own files
  - make a huge perf improvement for tree name cell location code -- see inline comment
  - minor style & import updates
  - remove old/unused samples hook
- **Why:** Last bits of work for table refactor 🙌 🚀 
- **Env:** https://maya-tree-table-perf-frontend.dev.czgenepi.org/

#### Before: 
before

#### After: 
after

### Checklist
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/queries/locations.ts
+++ b/src/frontend/src/common/queries/locations.ts
@@ -86,31 +86,3 @@ export function useNamedLocations(): UseQueryResult<
     staleTime: ONE_HOUR,
   });
 }
-
-interface NamedLocationsByIdResponse {
-  namedLocationsById: IdMap<NamedGisaidLocation>;
-}
-
-const getNamedLocationsById = (
-  data: LocationsResponse
-): NamedLocationsByIdResponse => {
-  const { locations } = data;
-  const namedLocations = locations.map(foldInLocationName);
-  return {
-    namedLocationsById: reduceObjectArrayToLookupDict(namedLocations, "id"),
-  };
-};
-
-export function useNamedLocationsById(): UseQueryResult<
-  NamedLocationsByIdResponse,
-  unknown
-> {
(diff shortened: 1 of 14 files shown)


src/frontend/src/common/queries/locations.ts
```diff
@@ -86,31 +85,3 @@ export function useNamedLocations(): UseQueryResult<
...
-
-export function useNamedLocationsById(): UseQueryResult<
-  NamedLocationsByIdResponse,
-  unknown
-> {
-  return useQuery([USE_LOCATIONS_INFO_QUERY_KEY], fetchLocations, {
-    retry: false,
-    // Using `select` allows it to share cache with other USE_LOCATIONS_INFO_QUERY_KEY,
-    // but give a different view on the same data after processed by `select` func.
-    select: getNamedLocationsById,
```
mmmmmmaya, line 112: Using select does not actually update what is stored in the cache, or create a secondary cache with the newly selected information. As a result, using the hook has little benefit when used in contexts where many components will call the hook. In this instance, the hook was only used in rendering the tree table. The select was run once for every table row that got created.  Since there are 20k+ locations, the execution time added up quickly when there were a large number of trees in the table (order of hundreds). Because this select executed a map on a 20k object hundreds of times at once, it created a noticeable lag in rendering the table, especially when scrolling with a virtualized list since rows are continually created and destroyed on scroll and are not reused.

As a result, I decided to move this to a helper function, which I could call once at the table level, and then pass in that memoized list to the tree table name cell to find the name as needed. This way, we only execute this code each time the table re-renders (or the locations cache expires), which is far, far more infrequent than whenever we render a new row in the trees table.

ehoops-cz (approved):

PR #1611, opened 2023-02-03
Maya/api camelize

mmmmmmaya:
### Summary
- **What:**
  - Remove maps/submaps for object keys to convert api responses from snake_case to camelCase.
  - I also fixed a couple of bugs I found in this branch
- **Why:** We can do this programmatically -- 99% of the time we just want a straight camelCase conversion.
- **Env:** https://maya-maps-frontend.dev.czgenepi.org/
- **Discussion:** https://czi-sci.slack.com/archives/C01HYJYA50W/p1675457987504069

### Testing
1. I tested pretty much everything in the app, including upload, download, link outs, static pages, table functionality like filter and sort, etc

### Demos
No visual changes.

### Checklist
- [x] I merged latest `maya-remove-old-tables`
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/src/common/queries/samples.ts
+++ b/src/frontend/src/common/queries/samples.ts
@@ -18,7 +18,9 @@ import {
   SampleResponse,
 } from "../api";
 import { API_URL } from "../constants/ENV";
-import { IdMap, reduceObjectArrayToLookupDict } from "../utils/dataTransforms";
+import { store } from "../redux";
+import { selectCurrentPathogen } from "../redux/selectors";
+import { IdMap, reduceObjectArrayToLookupDict, replaceKeyName } from "../utils/dataTransforms";
 import { ENTITIES } from "./entities";
 import { MutationCallbacks } from "./types";
 
@@ -271,7 +273,14 @@ export function useCreateSamples({
  * sample cache
  */
 const mapSampleData = (data: SampleResponse) => {
-  return reduceObjectArrayToLookupDict<Sample>(data.samples, "publicId");
+  const { samples } = data;
+
+  samples.forEach((s) => {
+    replaceKeyName(s, "privateIdentifier", "privateId");
+    replaceKeyName(s, "publicIdentifier", "publicId");
+  });
+
+  return reduceObjectArrayToLookupDict<Sample>(samples, "publicId");
 };
 
 export const USE_SAMPLE_INFO = {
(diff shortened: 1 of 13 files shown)


src/frontend/src/common/queries/samples.ts
```diff
@@ -271,7 +277,14 @@ export function useCreateSamples({
  * sample cache
  */
 const mapSampleData = (data: SampleResponse) => {
-  return reduceObjectArrayToLookupDict<Sample>(data.samples, "publicId");
+  const { samples } = data;
```
phoenixAja, line 280: do you need to call `camelize` on `samples` here?

mmmmmmaya: No, because samples are fetched via `apiResponse`, which already calls `camelize`

src/frontend/src/common/utils/dataTransforms.ts
```diff
@@ -83,3 +84,37 @@ export const getCapitalCaseTreeType = (
...
+  obj: any,
+  oldKey: string,
+  newKey: string
+): any => {
+  if (oldKey === newKey) return;
+
+  Object.defineProperty(
+    obj,
+    newKey,
+    Object.getOwnPropertyDescriptor(obj, oldKey) ?? {}
```
phoenixAja, line 116: when would this be `{}`?

src/frontend/src/common/utils/samples.ts
```diff
@@ -1,2 +1,6 @@
 export const getLineageFromSampleLineages = (lineages: Lineage[]): Lineage =>
   lineages?.[0];
```
phoenixAja, line 2: do we still need this?

phoenixAja (approved):
looks good, just a few qs

src/frontend/src/common/utils/dataTransforms.ts
mmmmmmaya: If you provide an `oldKey` that does not have an associated value, `getOwnPropertyDescriptor` returns `undefined` instead of a set of options. However, you cannot call `Object.defineProperty` without a set of options. It requires you to pass in a set of options when you define the property, such as whether or not the field is read-only. In the case where the `oldKey` is not defined on the object, we just give an empty set of options.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty

src/frontend/src/common/utils/samples.ts
mmmmmmaya: Yes, it is used in the lineages column definition in the samples tables.

src/frontend/src/common/utils/dataTransforms.ts
phoenixAja: i was more wondering when that would happen in our code? Like when would the oldKey not have an associated value? If that key had already been replaced?

mmmmmmaya: Like if the field is optional on the backend, but the backend sends the key without a value. For example, if a sample doesn't have qc metrics associated with it, the backend might send something like `{ qc_metrics: undefined, ...other fields }`. Even though there isn't a value associated with the key, the backend still sends it as part of the api response.

phoenixAja: ok cool thanks!

PR #1627, opened 2023-02-16
Expand time window for failed tree run

danrlu:
Ethiopia only has samples older than 1 year ago

### DIFF
--- a/src/backend/aspen/workflows/nextstrain_run/create_phyloruns.py
+++ b/src/backend/aspen/workflows/nextstrain_run/create_phyloruns.py
@@ -150,7 +150,7 @@ def retry_template_args_for_focal_group(
     for level in location_hierarchy:
         if getattr(location, level):
             current_location_level.append(level)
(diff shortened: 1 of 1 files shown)


src/backend/aspen/workflows/nextstrain_run/create_phyloruns.py
```diff
@@ -150,7 +150,7 @@ def retry_template_args_for_focal_group(
     for level in location_hierarchy:
         if getattr(location, level):
             current_location_level.append(level)
-    start_date_attempts = ["6 months ago", "12 months ago"]
+    start_date_attempts = ["6 months ago", "24 months ago"]
```
jgadling: should we try 6mo then 12mo then 24mo? or go straight from 6mo to 2yrs?

danrlu: good point~~

jgadling (approved):

PR #1630, opened 2023-02-27
Make it possible for version strings to match.

jgadling:
### Summary:
- **What:** Reduce the number of samples we update on each pangolin run.
- **Ticket:** [sc229721](https://app.shortcut.com/genepi/story/229721)

### Notes:
This fix isn't perfect, but it should be a step towards making the scheduled pango job behave as expected. Previously, the `find_samples.py` script was looking for all samples whose version (as outputted by pangolin in the CSV file and written to our db) didn't match the current version of pangolin (as fetched from the pangoLEARN release version). The latest PangoLEARN version looks something like `2022-07-09` while versions outputted by pangolin look something like `PUSHER-v1.18.1.1` or `SCORPIO_v0.1.10`.

This PR:
1. Updates the code that gets the current pango version to use the output of `pangolin -pv` (which is currently `pangolin-data 1.18.1.1`) instead of pangolearn releases
2. We then compare the last sequence of digits and `.` characters at the end of the strings output by `pangolin -pv` and pangolin's CSV output to determine which sequences are out of date.

I tested this script in local dev and prod, and we went from updating many 000's of sample lineage calls to dozens on every run.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/aspen/workflows/pangolin/find_samples.py
+++ b/src/backend/aspen/workflows/pangolin/find_samples.py
@@ -1,5 +1,6 @@
 import io
-import urllib.request
+import re
+import subprocess
 from typing import Collection
 
 import click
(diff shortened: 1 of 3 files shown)


src/backend/aspen/workflows/pangolin/find_samples.py
```diff
@@ -16,19 +17,20 @@
...
+    installed_version = re.sub(".*?([0-9\\.]+)$", "\\1", installed_version)
+    return installed_version
 
 
 def should_sample_be_updated(sample: Sample, most_recent_pango_version: str) -> bool:
     if not sample.lineages:
         return True
-    return sample.lineages[0].lineage_software_version != most_recent_pango_version
+    sample_lineage_version = re.sub(
+        ".*?([0-9\\.]+)$", "\\1", sample.lineages[0].lineage_software_version
```
vincent-czi: Nit: I think the strings here would be good to use raw strings instead for readability
```
"\\1" ---> r"\1"
```

src/backend/aspen/workflows/pangolin/find_samples.py
```diff
@@ -16,19 +17,20 @@
...
-    return latest_version
+    installed_version = re.sub(".*?([0-9\\.]+)$", "\\1", installed_version)
+    return installed_version
 
 
 def should_sample_be_updated(sample: Sample, most_recent_pango_version: str) -> bool:
     if not sample.lineages:
         return True
-    return sample.lineages[0].lineage_software_version != most_recent_pango_version
+    sample_lineage_version = re.sub(
```
vincent-czi, line 33: I'd like either a comment on this var declaration or a docstring for the `should_sample_be_updated` function: in the context of this PR with the description, it makes sense what you're trying to do -- although I'll be honest, it took me quite a bit of looking at that regex to make sense of what it was doing, but I'm not the best at regex -- but I think without the PR description as context, why this is necessary would be pretty mystifying.

vincent-czi (approved):
Approach makes sense, seems good!

PR #1635, opened 2023-03-21
Ingest genbank data instead of gisaid

jgadling:
### Summary:
- **Env:** `<rdev link>`

### Checklist:
- [ ] I added labels to my PR


.happy/terraform/modules/sfn_config/gisaid.wdl
```diff
@@ -109,16 +113,14 @@ task IngestGISAID {
...
 
-    # fetch the gisaid dataset and transform it.
-    wget "~{gisaid_ndjson_url}" --user "${gisaid_username}" --password "${gisaid_password}" --continue --tries=2 -O gisaid_dump.fasta.bz2
-    bunzip2 gisaid_dump.fasta.bz2 
-    zstdmt gisaid_dump.fasta -o sequences.fasta.zst
-    rm gisaid_dump.fasta
-
-    ${aws} s3 cp sequences.fasta.zst "s3://${aspen_s3_db_bucket}/${sequences_key}"
+    # fetch the genbank dataset and transform it.
+    # ${aws} s3 cp "~{upstream_ndjson_url}" genbank.ndjson.zst
```
vincent-czi, line 120: Were these commented code lines left intentionally? I'm not sure what they're serving if so.

.happy/terraform/modules/sfn_config/gisaid.wdl
```diff
@@ -109,16 +113,14 @@ task IngestGISAID {
...
+    # ${aws} s3 cp genbank.ndjson.zst "s3://${aspen_s3_db_bucket}/${sequences_key}"
+    wget -nv -O genbank.ndjson.zst "~{upstream_ndjson_url}"
```
vincent-czi: I don't know how chunky this file is and I expect this connection is more reliable, but I think it's probably still a good idea to keep the `--continue --tries=2` flags around if it's real big. If it's on the smaller side (<1GB, I dunno), just the `tries` in case there's a brief outage during download still seems worth it.

.happy/terraform/modules/sfn_config/gisaid.wdl
```diff
@@ -181,21 +183,23 @@ task TransformGISAID {
     mv /ncov-ingest/source-data/czgenepi_locations.tsv \
        /ncov-ingest/source-data/gisaid_geoLocationRules.tsv
 
-    # decompress the gisaid dataset and transform it.
-    ${aws} s3 cp --no-progress "s3://${raw_gisaid_s3_bucket}/${raw_gisaid_s3_key}" - | zstdmt -d > gisaid.ndjson
-    /ncov-ingest/bin/transform-gisaid     \
-        gisaid.ndjson                     \
+    # decompress the raw dataset and transform it.
+    ${aws} s3 cp --no-progress "s3://${raw_data_s3_bucket}/${raw_data_s3_key}" - | zstdmt -d > genbank.ndjson
```
vincent-czi, line 187: Nit (that I might even retract if I had more context): everything else has moved pretty carefully to really generalized language of "raw dataset" etc instead of saying where the data came from, but this switches back to saying genbank explicitly. That said, I expect this `transform-genbank` call is inherently built to working against genbank, not generalized, so trying to name things generically here is probably just silly. But maybe a comment saying we're specifically built with genbank in mind for the time being would make sense somewhere? I dunno.

.happy/terraform/modules/sfn_config/gisaid.wdl
```diff
@@ -181,21 +183,23 @@ task TransformGISAID {
...
+    /ncov-ingest/bin/transform-genbank    \
+        genbank.ndjson                    \
         --output-metadata metadata.tsv    \
         --output-fasta sequences.fasta    \
+        --biosample=""                    \
         --output-unix-newline
 
     zstdmt sequences.fasta
+    xz metadata.tsv
```
vincent-czi, line 196: Just curious: why do we use `zstdmt` for compression on the first, but then `xz` for the other?

.happy/terraform/modules/sfn_config/gisaid.wdl
```diff
@@ -257,48 +261,52 @@ task AlignGISAID {
...
-    # fetch the gisaid dataset
-    ${aws} s3 cp --no-progress "s3://${processed_gisaid_s3_bucket}/${processed_gisaid_sequences_s3_key}" - | zstdmt -d > /ncov/data/sequences.fasta
-    ${aws} s3 cp --no-progress "s3://${processed_gisaid_s3_bucket}/${processed_gisaid_metadata_s3_key}" /ncov/data/metadata.tsv
-    mkdir /ncov/my_profiles/align_gisaid/
-    cp /usr/src/app/aspen/workflows/align_gisaid/{builds.yaml,config.yaml} /ncov/my_profiles/align_gisaid/
-    # run snakemake, if run fails export the logs from snakemake and ncov to s3 
-    (cd /ncov && snakemake --printshellcmds results/aligned_gisaid.fasta.xz results/sanitized_metadata_gisaid.tsv.xz --profile my_profiles/align_gisaid) || { ${aws} s3 cp /ncov/.snakemake/log/ "s3://${aspen_s3_db_bucket}/aligned_gisaid_dump/${build_id}/logs/snakemake/" --recursive ; ${aws} s3 cp /ncov/logs/ "s3://${aspen_s3_db_bucket}/aligned_gisaid_dump/${build_id}/logs/ncov/" --recursive ; }
+    # fetch the upstream repo dataset
+    ${aws} s3 cp --no-progress "s3://${processed_genbank_s3_bucket}/${transformed_metadata_s3_key}" /ncov/data/metadata.tsv.xz
+    # ${aws} s3 cp --no-progress "~{upstream_aligned_url}" aligned.fasta.xz
```
vincent-czi: Leaving commented code intentional?

.happy/terraform/modules/sfn_config/gisaid.wdl
```diff
@@ -257,48 +261,52 @@ task AlignGISAID {
...
+    wget -nv -O aligned.fasta.xz "~{upstream_aligned_url}"
```
vincent-czi: Same comment about maybe wanting `--continue --tries=2` depending on size, etc.

.happy/terraform/modules/sfn_config/gisaid.wdl
```diff
@@ -257,48 +261,52 @@ task AlignGISAID {
...
+    # create db objects
     python3 /usr/src/app/aspen/workflows/align_gisaid/save.py                       \
             --aspen-workflow-rev "${aspen_workflow_rev}"                            \
             --aspen-creation-rev "${aspen_creation_rev}"                            \
             --ncov-rev "${ncov_git_rev}"                                            \
             --aspen-docker-image-version "externally_managed"                       \
             --start-time "${start_time}"                                            \
             --end-time "${end_time}"                                                \
-            --processed-gisaid-object-id "~{processed_gisaid_object_id}"            \
+            --processed-gisaid-object-id "~{transformed_metadata_entity_id}"        \
```
vincent-czi, line 300: Should this flag stay as `processed-gisaid-object-id`?

vincent-czi (approved):
Probably should still have Dan look it over since I really don't have great knowledge about the genbank aspect, but I definitely don't see any issues from what I read here!

Not sure if it would be worth it to convert the workflow name to something more generic -- eg, the `gisaid` part of `src/backend/aspen/workflows/align_gisaid/config.yaml`, `src/backend/pipeline_tests/test_gisaid_pipeline.py`, etc -- but that would be mildly nice. Still, if it takes any real effort, I don't think it's worth it given the larger context.

.happy/terraform/modules/sfn_config/gisaid.wdl
jgadling: Thanks, I'll add the switch. These files are in the ~1gb range but fewer silly failures is is best.

.happy/terraform/modules/sfn_config/gisaid.wdl
jgadling: I'm just trying to keep to the conventions of the previous script, even though switching between 3 different compression formats is super confusing 🤯

.happy/terraform/modules/sfn_config/gisaid.wdl
jgadling: Yep, I'm trying to make it easy(ish) to revert back to gisaid at some point in the future

.happy/terraform/modules/sfn_config/gisaid.wdl
jgadling: I'd like to switch to downloading these files via S3 instead of HTTPS -- nexstrain makes them available via a public s3 bucket, and it's cheaper/better for everyone to fetch them directly from the bucket. I was having problems with the batch job's IAM configuration so I fell back to HTTPS for now. I'm leaving the comments in as a reminder to debug this further in the future

PR #1637, opened 2023-03-28
chore: replacing biohub logo in footer

sethfeingold:
simple update to biohub logo in footer


src/frontend/src/views/LandingPageV2/components/Footer/index.tsx
```diff
@@ -1,6 +1,6 @@
 import Image from "next/image";
 import { useEffect } from "react";
-import BiohubLogoImg from "src/common/images/cz-biohub-logo.png";
+import BiohubLogoImg from "src/common/images/biohub-sf-logo.svg";
```
mmmmmmaya: It seems like this is the only use of the old logo in the repo. What do you think about removing it?

src/frontend/src/views/LandingPageV2/components/Footer/index.tsx
```diff
@@ -1,6 +1,6 @@
 import Image from "next/image";
 import { useEffect } from "react";
-import BiohubLogoImg from "src/common/images/cz-biohub-logo.png";
```
mmmmmmaya, line 3: It seems like you replaced a png with an svg, which actually does not work as a direct swap out. With this code, the logo is not visible.
Screenshot 2023-03-28 at 9 46 37 AM

src/frontend/src/views/LandingPageV2/components/Footer/index.tsx
sethfeingold: Definitely okay with removing the asset from the images dir 👍🏻

src/frontend/src/views/LandingPageV2/components/Footer/index.tsx
sethfeingold: Replaced with png version 😄  thanks for the heads up!

mmmmmmaya (approved):

sethfeingold:
Thanks so much for the review and help @mmmmmmaya 😄

PR #1639, opened 2023-04-05
unpin Pangolin data

danrlu:
What do you think if we unpin the underlying data for Pango lineages? So we can have the newest lineage assignments as new lineages become available... which also means we need to reinstate auto deployment, but that is a separate conversation. 

I can only test the changes don't break things once they're on staging fyi.

### DIFF
--- a/src/backend/Dockerfile.pangolin
+++ b/src/backend/Dockerfile.pangolin
@@ -41,10 +41,10 @@ RUN apt-get update && apt-get install -y  \
 # Install Pangolin and PangoLEARN + deps (for ncov)
 # The cov-lineages projects aren't available on PyPI, so install via git URLs.
 RUN pip3 install snakemake==6.8.0
-RUN pip3 install git+https://github.com/cov-lineages/pangolin.git@v4.1.3
-RUN pip3 install git+https://github.com/cov-lineages/pangolin-data.git@v1.16
-RUN pip3 install git+https://github.com/cov-lineages/scorpio.git@v0.3.17
-RUN pip3 install git+https://github.com/cov-lineages/constellations.git@v0.1.10
+RUN pip3 install git+https://github.com/cov-lineages/pangolin.git@v4.2
+RUN pip3 install git+https://github.com/cov-lineages/pangolin-data.git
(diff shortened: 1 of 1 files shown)


src/backend/Dockerfile.pangolin
```diff
@@ -41,10 +41,10 @@ RUN apt-get update && apt-get install -y  \
 # Install Pangolin and PangoLEARN + deps (for ncov)
 # The cov-lineages projects aren't available on PyPI, so install via git URLs.
 RUN pip3 install snakemake==6.8.0
-RUN pip3 install git+https://github.com/cov-lineages/pangolin.git@v4.1.3
-RUN pip3 install git+https://github.com/cov-lineages/pangolin-data.git@v1.16
-RUN pip3 install git+https://github.com/cov-lineages/scorpio.git@v0.3.17
-RUN pip3 install git+https://github.com/cov-lineages/constellations.git@v0.1.10
+RUN pip3 install git+https://github.com/cov-lineages/pangolin.git@v4.2
```
danrlu, line 44: this is the tool, the 3 repos below are where the lineage data is actually stored so I removed their pins.

danrlu: Pangolin removed requirement on Snakemake version https://github.com/cov-lineages/pangolin/commit/3ecf8e9d845a0cecc1753054db42423aeba1555b. So if we want to remove this pin on pangolin.git, we may consider remove snakemake pin together.

jgadling:
I think we'll get best results by pinning to specific versions for a couple reasons:
- We won't have our deploys fail unexpectedly due to changes unrelated to our code 
- Somewhat more importantly: the result each of those "RUN" commands gets cached in our CI pipeline. So even if the underlying version of pangolin / pangolin-data change, our CI builds won't *actually* re-run those commands and install the latest version unless something breaks the cache. Changing a version number will break the cache.

danrlu:
Good points!! Is there an easy way to keep our Pango lineages up to date then? Lineage definitions evolve as new lineages become available. For example SPHERE call today mentioned XBB.1.16 that's of interest but we don't have the definition for that lineage in our system, so Chicago will not know even if an XBB.1.16 show up in their samples

PR #1640, opened 2023-04-05
update pins for pango data

danrlu:
Thank you!!


danrlu:
Also no rush to bring the changes in. We can wait for the next deploy whenever that is.

jgadling (approved):

PR #1642, opened 2023-04-07
Fix metadata and locations import scripts

jgadling:
### Summary:
We've been having two major problems with location imports:
- Sometimes we get `''` empty string values in division columns when we expect `NULL` values, so we wind up with (almost!) duplicate rows in the db for locations with empty string vs null values.
- Some MPX data only has region metadata. We need to store locations no broader than country level.

### Notes:
I rewrote parts of the locations import script to put more work on the database and shuffle less data through python (🤞 that the DB doesn't mind), and added some tests to make sure the end result is correct.

### Checklist:
- [x] I added labels to my PR


src/backend/aspen/workflows/import_locations/tests/test_import.py
```diff
@@ -0,0 +1,125 @@
...
+    assert expected == found
+
+
+# Make sure the build config is working properly, and our location/group info is populated.
+def test_mean_data(mocker, session, split_client, postgres_database):
+    mock_remote_db_uri(mocker, postgres_database.as_uri())
+
+    input_metadata = [
+        # Region level
+        ["Asia", "Bangladesh", None, None],
```
vincent-czi:
```suggestion
        ["Asia", None, None, None],
```

vincent-czi (approved):
Looks great! I have one minor tweak to a test as a code suggestion, but everything looks great otherwise.

To be clear, this doesn't get rid of any of our region-only level rows that are currently in `locations`, right? We need to either drop those separately or filter them out in the FE from selection? (If we decide to go with dropping them,I was thinking that might cause problems for historical trees/samples that had been given locations of region-only. Not sure how serious the impact would be or if there's actually much out there that would be impacted, but we should remember to check.)

jgadling:
> To be clear, this doesn't get rid of any of our region-only level rows that are currently in `locations`, right? We need to either drop those separately or filter them out in the FE from selection? (If we decide to go with dropping them,I was thinking that might cause problems for historical trees/samples that had been given locations of region-only. Not sure how serious the impact would be or if there's actually much out there that would be impacted, but we should remember to check.)

Exactly right -- we'll need to clean up the old duplicate locations data once this PR is in prod. I doubt it will have much impact, but I'll try the naive approach first and if that runs into trouble (the important stuff like sample and group location associations have FK constraints that will make themselves known) I'll write a script to do it.

PR #1648, opened 2023-04-14
Update copy per GenBank change

danrlu:
### Summary:
tl;dr make texts throughout to mention GenBank and point to the appropriate ID format.

After switching to GenBank, it actually doesn't necessarily impact what the users do. They can still use their GISAID ID as public ID to upload, so we're keeping the mentioning of GISAID IDs throughout. 

Keeping the SARS-CoV-2 sample upload template/table header as `GISAID ID (Public ID)` otherwise it will break current user's upload if they use the old template. Only edited the example rows to mention GenBank.

Side note:
The sample's GISAID ID would look like `hCoV-19/USA/CA-CZB-0000/2020`
The sample's GenBank Isolate Name would look like `SARS-CoV-2/human/USA/CA-CZB-0000/2020`
All we need in their public ID is `USA/CA-CZB-0000/2020`
Btw the sample's GenBank Accession would look like `AF12345`

### Notes:
I'll check again on staging since I can't set up rdev.

### Checklist:
- [x] I added labels to my PR


src/frontend/src/components/DownloadMetadataTemplate/common/constants.ts
```diff
@@ -80,7 +80,7 @@ const GENERAL_VIRAL_UPLOAD_EXAMPLE_ROWS = [
   [
     EXAMPLE_SAMPLE_IDS[0], // sampleId
     "Private sample name", // privateId
-    "(if available) GenBank Accession", // publicId -- here as explainer
```
danrlu, line 83: @vincent-czi if we update the example row, current users won't have to update their upload template, right? I remember you saying the example rows are ignored during upload

vincent-czi: Yes. I just double-checked, that's right. All the data in an example row (that is, a row which starts with one of the EXAMPLE_SAMPLE_IDS) will be ignored during upload. So it's safe to change that description to whatever you want.

vincent-czi (approved):

danrlu:
Hold until we decide on mpox changes. I may have changed them the wrong way.

PR #1660, opened 2023-06-13
Disallow spaces in privateId

ehoops-cz:
### Summary:
- **What:** Disallow spaces because nextstrain tree builds don't use them in the way we expect.
- **Ticket:** no ticket, on-call

### Demos: Fix for both SC2 and MPOX.  With spaces we no longer allow the upload. Without spaces we allow the upload.
Screenshot 2023-06-13 at 9 54 05 AM
Screenshot 2023-06-13 at 9 53 42 AM
Screenshot 2023-06-13 at 9 52 14 AM
Screenshot 2023-06-13 at 9 51 49 AM

### Checklist:
- [x] I added labels to my PR


lvreynoso (approved):
lgtm

src/frontend/src/views/Upload/components/common/constants.ts
```diff
@@ -70,7 +70,7 @@ export const MAX_NAME_LENGTH = 120;
 // All the characters that we do NOT allow for sample name.
 // Effectively just a whitelist of characters, then negated to produce what's forbidden.
 // Allowed characters: all latin alphabet, all digits, ` ` (space), `.`, `_`, `/`, `-`
-export const FORBIDDEN_NAME_CHARACTERS_REGEX = /[^a-zA-Z0-9 ._/-]/;
+export const FORBIDDEN_NAME_CHARACTERS_REGEX = /[^a-zA-Z0-9._/-]/;
```
phoenixAja, line 73: i'm confused, this regex seems to be capturing all the charachters we do want, should this be renamed to ALLOWED_NAME_CHARACTERS_REGEX lol

phoenixAja (approved):

src/backend/aspen/database/models/sample.py
```diff
@@ -213,17 +215,20 @@ def generate_public_identifier(self, prefix, already_exists=False):
...
             )
+            self.public_identifier = sub(FORBIDDEN_NAME_CHARACTERS_REGEX, "", raw_public_identifier)
         else:
-            self.public_identifier = func.concat(
+            raw_public_identifier = func.concat(
                 f"{prefix}/{country}/{group_prefix}-",
                 text("currval('aspen.samples_id_seq')"),
                 f"/{current_year}",
             )
+            self.public_identifier = sub(FORBIDDEN_NAME_CHARACTERS_REGEX, "", raw_public_identifier)
```
lvreynoso: I think the regex string should leave out the leading and trailing forward slashes for Python. Also I think we can dedupe the sub call.

lvreynoso (changes requested):

src/frontend/src/views/Upload/components/common/constants.ts
ehoops-cz: I believe the `^` makes it so this selects for any character not in the defined characters

ehoops-cz: I updated the comment to remove space from the allowed characters.

lvreynoso (approved):
looks great!

PR #1674, opened 2023-11-09
Request a larger instance for pangolin

danrlu:
This is the size used in testing run [here](https://us-west-2.console.aws.amazon.com/states/home?region=us-west-2#/v2/executions/details/arn:aws:states:us-west-2:829407189049:execution:genepi-geprod-swipe-default-wdl:pango-32core-rerun) which should get an instance with 32 threads (note somewhere between Terraform and WDL the requested memory lose 2000 across all jobs).
This is the size of instance we use for tree runs, so not a big ask.

### DIFF
--- a/.happy/terraform/modules/ecs-stack/main.tf
+++ b/.happy/terraform/modules/ecs-stack/main.tf
@@ -227,7 +227,7 @@ module pangolin_sfn_config {
   source   = "../sfn_config"
   app_name = "pangolin-sfn"
   image    = local.pangolin_image
-  memory   = 16000
+  memory   = 120000
   wdl_path = "workflows/pangolin.wdl"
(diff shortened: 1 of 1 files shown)


danrlu:
Note that currently we ask pango to use 16 threads [here](https://github.com/chanzuckerberg/czgenepi/blob/c2ec44613c4934fe373aed521c63561415a99324/src/backend/aspen/workflows/pangolin/run_pangolin.sh#L20), which finishes the job in 8hr (with 32 thread CPU), or we can give pangolin 32 threads which finishes the job in 6hr. I'm not super confident about making this change b/c pango command is shared between on demand run and scheduled runs. 
I'll hand it back to eng if we want to do further optimization, but I'm OK with it as is 🙇‍♀️

jgadling (approved):

PR #1677, opened 2024-04-24
[pango fix] strip whitespace from first line for safety check

phoenixAja:
### Summary:
pango jobs were failling because we were expecting the first line in the lineages file to be  "Lineage\tDescription\n" but it was  "Lineage\tDescription \n" instead. This PR removes whitespaces before doing the safety check

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/src/backend/aspen/workflows/import_pango_lineages/load_pango_lineages.py
+++ b/src/backend/aspen/workflows/import_pango_lineages/load_pango_lineages.py
@@ -85,7 +85,7 @@ def get_lineages(lineage_notes_file: io.TextIOBase) -> list[str]:
     """
     # First line is header, no lineage data, but we use as a safety check.
     first_line = next(lineage_notes_file)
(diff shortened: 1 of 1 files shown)


vincent-czi (approved):
Looks good. I do wonder why they changed the structure by that tiny bit. Did you look over the new lineages file beyond that? I assume they're otherwise identical, but would be good to sanity check before we merge and move on.

phoenixAja:
> Looks good. I do wonder why they changed the structure by that tiny bit. Did you look over the new lineages file beyond that? I assume they're otherwise identical, but would be good to sanity check before we merge and move on.

i think it was just someone editing the file and leaving in a space? The comment above said that it's human edited so that seems likely what happened to me, the file is otherwise identical

danrlu:
The nightly jobs are still failing. I think maybe the prod deploy didn't succeed?

PR #1678, opened 2024-05-06
Security Updates

ehoops-cz:
### Summary:
- **What:** update all bable packages to latest - these are all minor version updates.  Remove faker-js - this was only used to create a fake name for one test.  imo it's better to avoid importing a whole package for this (plus it had a vulnerability).
- **Ticket:** https://czi-tech.atlassian.net/browse/CZID-9570
- **Env:** none

### Notes:
This seems to work after stopping and restarting containers.  I remember there being something weird about updating npm dependencies, but I don't remember the details and I can't find anything in the docs.  I'm planning to monitor and make sure everything looks good when this gets to staging.

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/frontend/e2e/utils/nexstraintree.ts
+++ b/src/frontend/e2e/utils/nexstraintree.ts
@@ -1,4 +1,3 @@
-import { faker } from "@faker-js/faker";
 import { sample } from "lodash";
 import { getADateInThePast, getValueOrDefault } from "./common";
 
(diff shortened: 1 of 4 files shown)


vincent-czi (approved):
PR looks good, but I think our deploy process is currently broken from something that came up a couple weeks ago? We can merge this, but I'm pretty sure we'll need to fix deployments before anything happens. (Security-wise though, we're fine, we both agreed this wasn't pressing because of the attack vector.)

PR #1693, opened 2024-07-11
Use group prefix for phylo tree s3 key

lvreynoso:
### Summary:
- **What:** Uses the unique, required group prefix instead of the nullable, optional group location column for s3 keys for phylo trees.
- **Env:** `<rdev link>`

### Checklist:
- [ ] I added labels to my PR

### DIFF
--- a/src/backend/aspen/util/swipe.py
+++ b/src/backend/aspen/util/swipe.py
@@ -80,7 +80,7 @@ class NextstrainJob(SwipeJob):
         output_suffix = f"/{group.name}/{str(now)}"
         execution_name = f"{group.prefix}-{self.job_type}-nextstrain-{str(now)}"
         extra_params = {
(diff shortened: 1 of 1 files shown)


vincent-czi (approved):
This seems like a good idea to me! The only thing I'm a little worried about is that we have somewhere in code that assumes the structure of our S3 keys instead of directly reading it out of a DB column. I know CZID has a lot of places where the spot in code that grabs the file from S3 assumes it knows how the S3 key is structured (from some combo of project and sample id, etc); I don't think we did that kind of thing in CZGE, I think the keys were always written to the DB and we never assume we know their structure when we read from S3. But that's my main fear. My only ask would be that you double-check that is the case before merging it in. Thanks!

lvreynoso:
We do specifically store the s3 keys for phylo trees in a column on the phylo trees table. I'll double check to make sure there isn't anything hardcoded anywhere.

PR #1703, opened 2024-08-21
Update lineage file to color samples on the mpox tree

danrlu:
### Summary:
Bring in the latest lineage info to color samples on the tree. See discussion [here](https://discussion.nextstrain.org/t/mpox-lineage-in-metadata-is-different-from-what-shows-in-the-tree/1714/2)

### Checklist:
- [x] I added labels to my PR

### DIFF
--- a/src/backend/Dockerfile.nextstrain
+++ b/src/backend/Dockerfile.nextstrain
@@ -58,7 +58,7 @@ RUN mkdir /mpox && \
     git remote add origin https://github.com/chanzuckerberg/monkeypox.git && \
     git fetch origin subsampling && \
     git reset --hard fd74f4b5f219035c9cbb7909b6f84f8a06fda76d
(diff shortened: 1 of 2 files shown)


vincent-czi (approved):
This seems like a reasonable extension of what's already happening!

Personally, I'm a bit confused how it's operating: I thought that the Dockerfile.nextstrain produced the docker image that the `run_nextstrain_mpx.sh` script ran inside of. So I'm surprised that we can run the `chown` during image build on files that we haven't downloaded yet? But we were already doing it for `exclude_accessions_mpxv.txt` and I assume that was fine then, so this seems like it would be fine as well? Perhaps the base nextstrain docker image already has those files present, so we can `chown` them during build, but then we download and swap-in-place so the ownership is now for the same file, but totally different content in the file? Dunno.

danrlu:
Oh? I thought the files are downloaded into the image when the Docker image was built? Let me push it to staging and we can see whether it works as expected.

PR #1707, opened 2024-09-05
Remove serial international samples for targeted trees

danrlu:
The evolutionary rates for different mpox clades vary greatly. We currently don't tree builds that focus on a clade, so targeted trees are the closest alternatives. This PR removes the randomly selected international samples from targeted trees so we don't add samples from random clades to the tree.


phoenixAja (approved):
yay!

PR #1711, opened 2024-09-10
Bump up number of closest sequences for targeted trees

danrlu:
The changes in https://github.com/chanzuckerberg/czgenepi/pull/1709 didn't do much. This is the place we need to update.


vincent-czi (approved):
I assume this makes sense comp bio wise, but I wanted to call out that this will change SC2 in addition to MPX. I expect you're already aware of that, but just wanted to make sure that the change is intended for both pathogens, not just mpox.

